PR #369, opened 2013-07-23
A servlet filter that adds request identifiers to slf4j's mapped diagnos...

ajordens:
...tic context (MDC), it allows logging output at multiple levels of your service to be easily grouped by the originating request.

- Updated RequestIdLoggingFilter to add support for checking clientId's against a regex for validity.
- Updated LoggingExceptionMapper to use the Service-Request-Id if it's available in the MDC.


JonMR:
The project mentioned above is a little more real now.  Please check it out @ https://github.com/service-enabled/dropwizard-request-tracker.  0.2.0 should be up in maven central.  We'd like add support for customizing the ID generation strategy and welcome PRs :)

carlo-rtr:
Sound like the right path is to enhance and upgrade dropwizard-request-tracker given the age of the last commit, the fact that people want to see a more general version, and the fact that someone has developed a module for this.

PR #370, opened 2013-07-24
experimenting with a configurable pattern for access requests using logback-access

eliast:
- Implemented our own DropwizardRequestLog to avoid any file-based configuration from RequestLogImpl.
- Added a new interface AccessAppendFactory to build access vs classic appenders

TODO: I'm still working on rolling file appender support.

### DIFF
--- a/dropwizard-example/example.yml
+++ b/dropwizard-example/example.yml
@@ -31,6 +31,14 @@ database:
 server:
 #  softNofileLimit: 1000
 #  hardNofileLimit: 1000
+  requestLog:
+    pattern: "%h %l %u %t \"%r\" %s %b \"%i{Referer}\" \"%i{User-Agent}\""
+    appenders:
+      - type: file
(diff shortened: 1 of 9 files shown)


carlo-rtr:
I think this is a good idea, unfortunately the code has gone a bit stale.

If you are still interested in seeing this happen, a PR would be welcome. I think you would have to make ```RequestLogFactory``` behave similar to the hierarchy of ```ServerFactory```, ```DefaultServerFactory``` and ```SimpleServerFactory```. This would allow you to provide your own impl of RequestLogFactory by specifying the type in the configuration and including the class at runtime. The framework would have to find it via the java provider mechanism ("META-INF.service") like this:

https://github.com/dropwizard/dropwizard/tree/master/dropwizard-core/src/main/resources/META-INF/services

https://github.com/dropwizard/dropwizard/issues/972

PR #377, opened 2013-08-03
Add JDBI SQL Object integration test support.

thody:
This changeset adds a simple base class for implementing JDBI integration tests. 

Extending `JdbiIntegrationTest` allows you to quickly get instances of JDBI SQL Objects for testing. When run, the tests will first execute your migrations to ensure the state of your database is clean and consistent – this also ensures that the queries in your SQL Object are tested against the same schema used to create your production database.

Note: this implementation does assume that users are using migrations in their service.


sw00:
Will this ever get merged? Seems to be essential for testing.

arteam:
For anyone interested, I've recently started a project that simplifies testing JDBI code in Dropwizard. Checkout https://github.com/arteam/jdit

It uses an alternative approach with `JUnitRunner` and annotations rather than JUnit rules.
Features: 
* Support for defining DB schema. 
* Test-level DDL scripts 
* DB is started for all tests and data is swept after every test 
* Can be used outside Dropwizard as well.

Known limitations:
* Hard dependency on HSQLDB

Feedback is welcome.

carlo-rtr:
@arteam does your project completely cover this case? If so, should we close this and make sure you have an entry https://github.com/dropwizard/modules.dropwizard.io/blob/master/_data/modules/thirdparty.yaml

arteam:
I've added a link to JDIT to the list of thirdparty modules.

This project covers the needed functionality (integration testig of JDBI SQL Objects against the schema), has additional features and well maintaned, so I encourage people to use it.

Thanks, @thody, for the pull request, though. It was a source of inspiration for JDIT.

PR #383, opened 2013-08-27
Extract HandlerFactory from ServerFactory

nicktelford:
Decouples settings specific to a context (e.g. "admin", "application") from the server. The motivation for this was to allow per-context configuration of the RequestLog.

  * `requestLog`, `gzip`, `minThreads`, `maxThreads`, `maxQueuedRequests` and `idleThreadTimeout` now all configurable per context (handler).
  * For `DefaultServerFactory`, each context accepts a list of `connectors`.
  * For `SimpleServerFactory`, the server is configured with a single `connector`, and each context accepts a context `path`.

See [Example Configuration](https://github.com/nicktelford/dropwizard/blob/feature/extract-servlet-factory/dropwizard-example/example.yml) for an example of the new configuration.


carlo-rtr:
Given this isn't mergeable, if your still interested in pursuing this, please open up an issue and we can discuss.

PR #488, opened 2014-03-07
Generalize DropwizardAppRule for other test frameworks

csabapalfi:
Hi,

I was looking at starting/stopping a Dropwizard app in Cucumber tests and DropwizardAppRule has all the functionality I need but obviously it doesn't expose startIfRequired and stop methods.

I'd happy to extract a DropWizardAppTestSupport class from DropwizardAppRule and make the DropwizardAppRule depend on that and contribute the code if there's any interest.

What do you think?

Thanks,
Csaba

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/ConfigOverride.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/ConfigOverride.java
@@ -1,4 +1,4 @@
-package io.dropwizard.testing.junit;
+package io.dropwizard.testing;
 
 public class ConfigOverride {
 
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
@@ -1,0 +1,192 @@
+package io.dropwizard.testing;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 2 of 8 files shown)


joschi:
If someone steps up to update the PR to apply cleanly to current master, we'll happily merge it. (@ddunwoody, @hslater-okta)

ddunwoody:
I'll take a look this week.  It's my last week on the project that uses this, so it would be a nice feeling to get it merged in.

csabapalfi:
Sorry about the delay. 

Re-done the change on the latest master (as it diverged a lot).

2 tests were failing locally when run with all others (but passing in isolation). That still needs fixing.

csabapalfi:
The patch keeps everything backwards compatible. Is that required? If not then the code can be a bit simpler in some cases.

csabapalfi:
Found the cause of the test failures. One of the test was not cleaning up after itself properly and left an extra system property confusing the tests run afterwards.

csabapalfi:
Just spotted that the same test problem is already on master. Introduced by https://github.com/dropwizard/dropwizard/commit/9848d9b5275dff6255b1aa255b8075638a1cdb20

 If you merge my PR and that'll fix the tests on master, too.

joschi:
Great! Thanks for the PR.

ddunwoody:
Thanks everyone.  I was saving this for work in the morning, but looks like you beat me to it!

PR #539, opened 2014-04-14
GZip's compressedMimeTypes is not working

rkapsi:
Dropwizard is setting the MIME types before GzipFilter#init() is being called. The latter will overwrite our custom list if FilterConfig#getInitParameter("mimeTypes") is null (see GzipFilter line 199 and the first if-block).


jplock:
Is this still an issue in Dropwizard 0.8?

carlo-rtr:
Given no response since @jplock commented, I'm going to assume no

PR #567, opened 2014-05-09
Allow normal Logback facilities to be used to configure logging

justrudd:
As the title states, this PR allows the normal Logback facilities to be used to configure logging. 

Basically what it does is extract LoggingFactory into an interface like ServerFactory. It is a discoverable interface and defaults to DefaultLoggingFactory (which is the current code renamed).

You have to do three things in order to use the normal Logback facilities...

1. Add a constructor to your Application derived class that passes false to the super class. This tells Dropwizard to not bootstrap logging. Logback is lazy and doesn't initialize until it has to (which given the current code is when ServerCommand is created in Application.run).
2. In your YML file, make the type of logging noop.

    ```
logging:
  -type: noop
```

3. Add a logback.xml, logback.groovy, or other configuration file to your classpath.

### CODE REVIEW

**Comment by DeanPoulin** [comment]
Based on comments in this thread about keeping this functionality able to be natively configured via yaml for a given dropwizard service, I took a stab at achieving that.

We also had the desire to log certain messages to separate log files. Here's the approach I took. This seems to play very nicely with current logging functionality with the additional capability to log to separate log files via configuration. We are currently using dropwizard 0.6.2 in production, so I did this off that tag but should be trivial to apply to 0.7.x & 0.8.x as well.

Here's the general idea:

Current 0.6.x configuration for file logging:

```yaml
  file:
    enabled: true
    timeZone: EST5EDT
    threshold: ALL
    currentLogFilename: /var/log/my-service/my-service.log
    archivedLogFilenamePattern: /var/log/my-service/my-service-%d.log.gz
```

New configuration with optional additional appenders:

```yaml
file:
    enabled: true
    timeZone: EST5EDT
    threshold: ALL
    currentLogFilename: /var/log/sui/sui.log
    archivedLogFilenamePattern: /var/log/sui/sui-%d.log.gz
    
    #optional additional appenders
    appenders:
      - enabled: true
        timeZone: EST5EDT
        threshold: ALL
        currentLogFilename: /var/log/my-service/activity.log
        archivedLogFilenamePattern: /var/log/my-service/activity-%d.log.gz

        # the class/namespace you want to have logged to another file
        logger: com.mycompany.myservice.webservice.logging.LoggingFilter

        # should the logs also show up in the main file (true = yes | default = yes)?
        additive: false

      - enabled: true
        timeZone: EST5EDT
        threshold: ALL
        currentLogFilename: /var/log/my-service/foo.log
        archivedLogFilenamePattern: /var/log/my-service/foo-%d.log.gz
        logger: com.mycompany.myservice.webservice.Foo
```

With the above configuration, logs for `com.mycompany.myservice.webservice.logging.LoggingFilter` will be logged to an additional file `/var/log/my-service/activity.log` and will not show up in the main log file for the service. Logs for `com.mycompany.myservice.webservice.Foo` will show up in the main log file and will also be logged to a separate file `/var/log/my-service/foo.log`.

Here's the changes I made to facilitate this functionality: https://github.com/DeanPoulin/dropwizard/commit/b9df4187dba758305ad2a9ecbdf6ff126dbf8651

justrudd:
So this has been open 9 months. It is out of date (I didn't keep it up to date with 0.8 code), and it appears that it isn't going to be applied. It looks like a pretty even split between those that do want it and those that don't want it.  

Regardless, I have moved on from DW to other tools (primarily because I'm not using Java day-to-day) so I have no interest in this anymore. Luckily, it would appear that @DeanPoulin did what most of you want - reimplemented logback configuration in YAML.

kennycason:
I know this is closed, but I literally wasted days figuring out how to get logback working (I just reset the context after DropWizard was up and running). All this was because I wanted to control logging level for different loggers. I think if the documentation was updated to show an example, the number of complains about logging would drop significantly. Thanks!
```yaml
logging:
  level: INFO
  loggers:
    org.hibernate: WARN <--- this was the golden nugget i was looking for! 
  appenders:
    - type: console
      threshold: ALL
      timeZone: UTC
      target: stdout
    - type: file
      threshold: ALL
      currentLogFilename: /srv/kraken/log/kraken.log
      archivedLogFilenamePattern: kraken-%d.log.gz
      archivedFileCount: 5
      timeZone: UTC
```

Issue #202, opened 2012-11-01
Poor URL hygiene


jglick:
`ConfigurationFactoryTest` has e.g.

```
private final File validFile = new File(Resources.getResource("factory-test-valid.yml")
                                                 .getFile());
```

which does not work when run from a checkout with a space in its name:

```
java.io.FileNotFoundException: /…/Dropwizard%20Core/dropwizard-core/target/test-classes/factory-test-valid.yml (No such file or directory)
```

`new File(url.getFile())` is wrong; use `new File(url.toURI())` instead.

Better still is to make the factory method take a `URL` to begin with, which would let this test run from a JAR &c.

PR #580, opened 2014-05-16
Replacing dropwizard request logging with logback-access

reines:
WIP: This isn't ready to merge!

This is an attempt at replacing the Slf4jRequestLog with logback-access, and exposing the log pattern in configuration.

The primary driver behind this is the ability to use more complex log patterns, specifically the %replace(){} function from logback to redact sensitive information from request logs. It also adds the ability to make use of MDC in request logs.

- [x] This relies on [#201](https://github.com/qos-ch/logback/pull/201) and [#202](https://github.com/qos-ch/logback/pull/202) being merged in to logback first.
- [ ] It specifies logback-access 1.1.3-SNAPSHOT in the dropwizard-logging/pom.xml because I built logback locally with these changes, obviously this would actually be ${logback.version}.
- [x] This removes the timeZone configuration field as the timezone is now specified as part of the logFormat configuration field instead.
- [x] This needs tests written. I know, I'm a bad person :(

I'm primarily looking for feedback on this. Assuming the logback PRs get accepted, would this be considered?

There is already #370 which has similar goals, but is badly outdated now, and works by adding a new build method to each appender factory, rather than genericising them.


carlo-rtr:
* Looks like the logback deps got merged in, so it's no longer blocking
* Flexible request logging is definitely something we want: https://github.com/dropwizard/dropwizard/issues/972

@reines You mind getting this to a point where it's mergeable, I'll make sure it's reviewed.

carlo-rtr:
I'm going to close until there is an update.

PR #610, opened 2014-06-08
First attempt at an Eclipse formatter settings file

saadmufti:
First attempt at an Eclipse formatter settings file to format contributions according to Dropwizard's preferred style.

Note that this file started out as the Google Java Code Style settings at https://code.google.com/p/google-styleguide/source/browse/trunk/eclipse-java-google-style.xml and I tweaked them a little. I am mostly looking for feedback at this point and refinements as this settings file has many moving parts. The best way to play with it is to use it in Eclipse and modify it using the Eclipse GUI for project specific Java Code Style->Formatter settings. The individual elements in the XML file itself can be a bit opaque and are not very well documented.

The intent is to include this in Dropwizard once we agree on the settings so that contributors can have an easier time conforming to the project's code formatting style.

Cheers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,10 +20,16 @@ they will be accepted, we ask that you follow these simple guidelines:
 
 Code style
 ---
-When submitting code, please make every effort to follow existing conventions and
-style in order to keep the code as readable as possible. We realize that the style
-used in Dropwizard might be different that what is used in your projects, but in the end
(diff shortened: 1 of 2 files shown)


carlo-rtr:
The committers discussed this.

I think we should follow http://google-styleguide.googlecode.com/svn/trunk/javaguide.html

With that said, we shouldn't reformat all the code or make it fail the build. We'll follow this on a best effort basis.

PR #631, opened 2014-06-24
Added dropwizard-embedded module, to embed a service in a daemon

AlexTalks:
This new module adds a new interface to a Dropwizard HTTP(s) service, so
that the HTTP server can be started/stopped at any point within an
already existing daemon process (no CLI required). This allows a
developer to embed Dropwizard into already existing services with their
own main run loop and SLF4J logging infrastructure.

Code example:
```java
public class WebService extends io.dropwizard.embedded.Service<Configuration> {
    public void initialize(EmbeddedBootstrap<Configuration> bootstrap) {
        // add bundles
        bootstrap.addBundle(new ViewBundle());
    }
    protected void start(Configuration config, Environment env) {
        // add resources
        env.jersey().register(new HelloWorldResource());
    }

    public static void main(String ... args) {
        handleCommandLine(args);
        configureSlf4jLogging();
        start();
        new MainRunLoop().run();
        stop();
    }
}
```

### DIFF
--- a/dropwizard-embedded/pom.xml
+++ b/dropwizard-embedded/pom.xml
@@ -23,7 +23,7 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>cobertura-maven-plugin</artifactId>
-                <version>2.3</version>
+                <version>2.6</version>
                 <configuration>
                     <check>
                         <haltOnFailure>false</haltOnFailure><!-- optional -->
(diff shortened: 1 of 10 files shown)


carlo-rtr:
Closing this. If you want something like this, I think Nick's comment makes sense.

"If we want to support embedding Dropwizard applications it's probably better for us to refactor the bootstrapping process such that the current mode of operation "embeds" the application in the default CLI-based environment. I don't think this is a particularly good candidate for another module."

PR #633, opened 2014-06-25
ConstraintViolationException should not always be thrown for empty request entities

joschi:
The `ConstraintViolationException` with an empty request entity should only be thrown
if the entity should be validated at all and is annotated with `@Valid` or `@Validated`.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JacksonMessageBodyProvider.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JacksonMessageBodyProvider.java
@@ -66,14 +66,14 @@ public class JacksonMessageBodyProvider extends JacksonJaxbJsonProvider {
     }
 
     private Object validate(Annotation[] annotations, Object value) {
-        if(null == value) {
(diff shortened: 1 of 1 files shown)


carlo-rtr:
@joschi given https://github.com/dropwizard/dropwizard/pull/842 was merged, what should we do?

Should this be checking for ```@NotNull``` instead?

nickbabcock:
I have a couple thoughts on this. When using the validator and you call `validateParameters` with the parameter representing the request entity being null:

- If no annotations on the parameter, no validations are performed
- If the parameter has `@Valid`, no validations are performed
- If the parameter has `@NotNull`, a constraint violation will be created because the parameter is null. If the parameter was not null then cascading validations wouldn't have applied
- If the parameter has `@Valid` and `@NotNull`, a constraint violation will be created

So @carlo-rtr is right, and instead and there should only be a check for `@NotNull` (no need to check for `@Valid` or `@Validated`)

This is if we keep the same semantics of Hibernate Validator.

jplock:
@joschi should we close this in favor of #1221 as its the same diff?

joschi:
@jplock Yes, makes sense.

PR #657, opened 2014-07-12
Proxy support for Dropwizard HTTP client

arteam:
I often encounter with a situation when in closed corporate environments direct access to the Internet is blocked by security reasons. So all connections to external resources happen through a proxy server (usually with BASIC authentication).

Deploying applications in such environments becomes a huge pain, because usually you don't know in advance that an application will need proxy connection support and sometimes you just don't have possibility to change code to add this facility.

I believe, it would be quite useful to have this facility in the Dropwizard config. It frees an application developer from thoughts about environments where an application will be used and shifts this responsibility to operations guys. And they can configure proxy client support just in the application config without using external tools. 

PS I've added the docs to proxy configuration. But I'm not a skilled english speaker, so, please, fix them if I messed up something.


arteam:
I see #681 has been merged.

Is there any work needed for this PR? If Java properties configuration is sufficient then this PR could be closed or moved to backlog from 0.8 release.

joschi:
@arteam I'll move this PR out of the 0.8.0 milestone for now but I won't close it for now.

Maybe someone else wants to chime in whether the current state of affairs is sufficient or if this PR should be updated and merged.

jplock:
@arteam did you want to update this for `0.9.0` or should we close this out?

arteam:
I've rebased this PR against master.

jplock:
I've got a few projects that are restricted to operating inside of an AWS VPC and need to communicate via a proxy, so I think this would be a great addition.

jplock:
Thanks a lot for this contribution!

Issue #602, opened 2014-05-30
Unable to set context application context path in DefaultServerFactory


mikeycmccarthy:
Out of the box it seems you cannot use the DefaultServerFactory and specify an application context path. Line 103 of the SimpleServerFactory sets the applicationContextPath but the DefaultServerFactory is missing this line and the appropriate json property.

I'm just about to get round the issue by extending the DefaultServerFactory but it seems like an inconvenience when this worked out of the box before. 

I've raised this issue on the dev Google group at https://groups.google.com/forum/#!topic/dropwizard-dev/MAXglyXJCVU - I'm interested to know if this is a regression or a deliberate decision.

PR #665, opened 2014-07-24
Add application and admin context paths to DefaultServerFactory

joschi:
This PR adds the `applicationContextPath` and `adminContextPath` configuration settings to `DefaultServerFactory`, as discussed in #602.

It also reintroduces the `rootPath` configuration setting to `AbstractServerFactory` and fixes the documentation regarding its usage vs. `applicationContextPath`. The latter caused some confusion with our users.


mikeycmccarthy:
Thanks @nicktelford , makes sense.

PR #674, opened 2014-08-06
`DropwizardAppRule` more strict in resetting config overrides

stuartgunter:
Prior to this change, `DropwizardAppRule` would clear *ALL* system properties that start with `dw.`, regardless of whether they were defined as `ConfigOverride` values in the test. This caused problems with tests that relied on a combination of command-line overrides (that should not be reset) and test-specific overrides (that should be reset).

An example of where this might be useful is dynamic port allocation, where the build system (Maven, Jenkins, etc.) dynamically allocates port numbers to the app and provides these via command-line overrides (e.g. `-Ddw.cassandra.port=${SOME_DYNAMIC_PORT}`). These values should be retained (otherwise subsequent tests fail), but the configuration that was explicitly overridden using `ConfigOverride` in the test should be reset.

Note that this does _not_ revert the original value of the system property if you override an already-overridden property.

### DIFF
--- a/dropwizard-testing/src/test/java/io/dropwizard/testing/junit/DropwizardServiceRuleResetConfigOverrideTest.java
+++ b/dropwizard-testing/src/test/java/io/dropwizard/testing/junit/DropwizardServiceRuleResetConfigOverrideTest.java
@@ -1,0 +1,35 @@
+package io.dropwizard.testing.junit;
+
+import org.junit.FixMethodOrder;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runners.MethodSorters;
+
+import static io.dropwizard.testing.junit.ConfigOverride.config;
+import static io.dropwizard.testing.junit.DropwizardAppRuleTest.resourceFilePath;
+import static org.assertj.core.api.Assertions.assertThat;
+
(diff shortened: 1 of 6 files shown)


joschi:
@stuartgunter Thanks for this PR!

dropwizard-testing/src/test/java/io/dropwizard/testing/junit/DropwizardServiceRuleResetConfigOverrideTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.FixMethodOrder;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runners.MethodSorters;
+
+import static io.dropwizard.testing.junit.ConfigOverride.config;
+import static io.dropwizard.testing.junit.DropwizardAppRuleTest.resourceFilePath;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
joschi: I've changed this test to just call the Rule methods before() and after() directly. No need to test JUnit itself. ;)

PR #680, opened 2014-08-08
Allow application to replace the default pattern for all log appenders

wrprice:
Allows the single, default pattern to be customized by the application (e.g. to include certain MDC values)
without requiring individual deployments to copy-paste the same format override in their YAML files.


carlo-rtr:
@wrprice sorry it took so long for a reply. 

The theme of this looks good, but I don't like the static state. Could you just have a:
```public String getPattern(Timezone timezone)```
which by default returns what existed.

If people want custom patterns, they can subclass & override. 

Alternatively, pass the factory via constructor and store it in a normal field.

thanks for the contribution.

carlo-rtr:
Looking at this some more, I think we should just rely on the configuration patterns in the config. If you really want to override the defaults, now you can roll your own LoggerFactory

PR #681, opened 2014-08-11
Expose the ability to set custom HttpRoutePlanners

philandstuff:
This change allows the user to control proxy settings by providing a
custom HttpRoutePlanner instance.  For example, the user can provide a
[SystemDefaultRoutePlanner][1] to get proxy settings from system
properties (ie [http.proxyHost and friends][2]).

This is an alternative to #657.  The difference is that this PR delegates the configuration responsibility to apache HttpClient and the user code, rather than having DropWizard own that responsibility.

It also has the advantage that, unlike #657, the user is given the option of specifying a proxy server by system properties, including the `http.nonProxyHosts` property which specifies a pattern of hostnames which should *not* be proxied.  This is a common use case in networks where external public internet requests must go via a proxy server, but internal private network requests must not.  I currently have a project with exactly this use case and so I would not be able to use #657 to configure my proxy server.

Currently, in order to work around it, I am using code like the following:

```java
    DefaultHttpClient client = (DefaultHttpClient) builder.build(name);
    client.setRoutePlanner(new SystemDefaultRoutePlanner(ProxySelector.getDefault()));
    return new ApacheHttpClient4Handler(client, null, true);
```

This workaround is a little hacky in that I'm having to do a typecast in order to get at the `setRoutePlanner` method, and I'm assuming a little too much knowledge about the type of HttpClient that HttpClientBuilder.build() returns.

This PR has the disadvantage that, unlike #657, this would split proxy configuration apart from the rest of a DropWizard application configuration.  I'm relatively new to DW so I don't know what the project rationale is and what the preferred configuration style would be.

Finally, this isn't actually mutually exclusive with #657.  It's entirely conceivable that both could be merged and coexist -- allowing the user to provide a custom HttpRoutePlanner for maximum flexibility, but providing a DropWizardish configuration for common proxy patterns.

[1]: http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/SystemDefaultRoutePlanner.html
[2]: http://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html


joschi:
Thanks!

PR #692, opened 2014-08-27
Added `LogConfigurationTask` to allow runtime changes to log levels

stuartgunter:
Logback provides a `JmxConfigurator` for changing logging levels at runtime, but this does not currently appear to be exposed in a Dropwizard-friendly manner. I didn't want to simply use the JMX approach, as that would introduce an inconsistency in managing my DW apps (everything else is managed via config or HTTP access).

I created a `Task` implementation that essentially does the same as Logback's `JmxConfigurator` and provided this via an external library (see https://github.com/stuartgunter/dropwizard-logging-config), but feel it would be appropriate to include in the main DW library. (If there's no appetite for including this in DW, I'd be happy maintaining the library separately.)

I'm not comfortable with the way it has been included, but haven't yet come up with a nice way of doing it. I would appreciate feedback (assuming there's interest in bringing this into DW) on how it should be included at a package / module level. The summary of my concerns is:

* `Task` is defined in `dropwizard-servlets`
* `Logback` is only explicitly included in `dropwizard-logging`
* The `LogConfigurationTask` straddles both of these modules due to its dependency on both Logback and `Task`
* I don't want to create a separate Maven module just for this one class, but I don't believe it truly fits in either `dropwizard-logging` or `dropwizard-servlets`
* Would it be worth relocating `Tasks` to somewhere else (they may be used by `dropwizard-servlets` but there's nothing Servlet-specific about them - they're not all that different from named runnables with predefined method args)

Feedback would be very welcome.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/AdminEnvironment.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/AdminEnvironment.java
@@ -5,6 +5,7 @@ import com.codahale.metrics.health.HealthCheckRegistry;
 import com.codahale.metrics.health.jvm.ThreadDeadlockHealthCheck;
 import io.dropwizard.jetty.MutableServletContextHandler;
 import io.dropwizard.jetty.setup.ServletEnvironment;
+import io.dropwizard.logging.tasks.LogConfigurationTask;
 import io.dropwizard.servlets.tasks.GarbageCollectionTask;
 import io.dropwizard.servlets.tasks.Task;
 import io.dropwizard.servlets.tasks.TaskServlet;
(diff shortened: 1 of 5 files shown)


dropwizard-core/src/main/java/io/dropwizard/setup/AdminEnvironment.java
```diff
@@ -37,6 +38,7 @@ public AdminEnvironment(MutableServletContextHandler handler,
         this.healthChecks.register("deadlocks", new ThreadDeadlockHealthCheck());
         this.tasks = new TaskServlet(metricRegistry);
         tasks.add(new GarbageCollectionTask());
+        tasks.add(new LogConfigurationTask());
```
joschi: :+1:

joschi:
Thanks!

I've moved the Task into dropwizard-servlets because it's rather Dropwizard-specific and the other default task is also in that package.

Issue #711, opened 2014-09-23
MetricFilter's matches returns true if Includes are specified and Excludes are not


qinfchen:
If Includes are not empty and excludes are empty, this should be the includes-only case that 
no metrics should be reported, except those with a name listed in includes.

But !getExcludes().contains(name) would always return true which ends up reporting all the metrics.

eepstein:
I think the bug fix also has a minor bug in it.  In particular, "case 4" of the unit tests seems to be different that one would imagine with a pair of exclude and include filters.

Normally with such a pair the logic is:

     include =  !isExcluded && isIncluded

So the exclusions are applied first and then anything that makes it past the exclusions and is in the explicit set of inclusions gets included.  But the logic in place now has case 4's includes overriding the excludes, which given that the default behavior is including-everything, is strange.  One has explicitly excluding something and then support explicitly overriding the exclusion.  That's double the effort when all that's needed is: don't put the thing in the exclusion list in the first place.  

In short, exclusions should be applied first, and the items that aren't excluded are now the set from which items may be explicitly included.  No explicit inclusion means: include everything not explicitly excluded (case 3).  Case 4, as coded, is inverted.

PR #713, opened 2014-09-25
issue 711 fix matches logic in BaseReporterFactory

qinfchen:
https://github.com/dropwizard/dropwizard/issues/711
1. If include list is empty and exclude list is empty, everything should be included 
2. If include list is NOT empty and exclude list is empty, only the ones specified in the list should be included
3. if include list is empty and exclude list is NOT empty, everything should be included except the ones in the exclude list
4. if include list is NOT empty and exclude list is NOT empty, everything should be included except the ones that are not in the include list AND are in the exclude list

### DIFF
--- a/dropwizard-metrics/src/main/java/io/dropwizard/metrics/BaseReporterFactory.java
+++ b/dropwizard-metrics/src/main/java/io/dropwizard/metrics/BaseReporterFactory.java
@@ -142,8 +142,9 @@ public abstract class BaseReporterFactory implements ReporterFactory {
         return new MetricFilter() {
             @Override
             public boolean matches(final String name, final Metric metric) {
-                return (!getIncludes().isEmpty() && getIncludes().contains(name))
-                        || !getExcludes().contains(name);
+                return (getIncludes().isEmpty() && getExcludes().isEmpty())
+                       || (!getIncludes().isEmpty() && getIncludes().contains(name))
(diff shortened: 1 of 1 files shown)


eepstein:
For #4 isn't the common approach: everything that is not in the excludes list and is in the includes list should be included?   The way it's written the includes list overrules the excludes list.  Is that correct / intended?

eepstein:
I'm looking at the test cases - which do correctly test for the documented logic - and the result is unexpected.  Basically if I have an includes list of {"inc, "both"} I expect those items to show up in the output.  What's odd is that when I now add in a excludes list of {"both", "exc"}, first of all "both" stays in the result set and, even weirder, "any" suddenly shows up in the result set.  In other words, I've done a complicated configuration (4 settings) to get a simple result (1 exclusion, namely "exc").  

The logic as-is amounts to extra work.

The old code did have a bug in it.  The usual way one does this kind of exclusion/inclusion lists (Eclipse, Spring, et. al.) is:    isNotExclude && isIncluded,   which in this case looks like:

    public MetricFilter getFilter() {
        return new MetricFilter() {
            @Override
            public boolean matches(final String name, final Metric metric) {
                boolean isNotExcluded = getExcludes().isEmpty() || !getExcludes().contains(name);
                boolean isIncluded =  getIncludes().isEmpty() || getIncludes().contains(name);
                // Match name that isNotExcluded && isIncluded
                return isNotExcluded && isIncluded;
            }
        };
    }

PR #728, opened 2014-10-05
Refactoring logEndpoints, filter by Resource.from() != null 

### LINKED ISSUE #726: DropwizardResourceConfig#logEndpoints does not work for implemented interfaces
with the current 0.8.0-SNAPSHOT, the registered resource does not get logged on startup when it implements a resource interface:

```java
@Path("/")
public interface LoadCalendar {
    @GET
    @Path("/foo/{name}")
    public Response load(@PathParam("name") String name);
}

public class CalendarResourceBean implements Serializable, LoadCalendar {
    private static final long serialVersionUID = 1L;

    @Override
    public Response load(final String name) {
        return Response.ok().entity("name="+name).build();
    }
}

...

environment.jersey().register(new CalendarResourceBean());
```

This did work in 0.7.1.

Reason is that logEndpoints filters all registered resources that do not have a @Path annotation without checking if an implemented interface does.

NOTE: this only affects the log report, the resource is registered and working fine.

jangalinski:
fixes #726


joschi:
Thanks!

PR #730, opened 2014-10-05
Tests and fix for MetricsFilter on BaseReporterFactory

### LINKED ISSUE #711: MetricFilter's matches returns true if Includes are specified and Excludes are not

jangalinski:
this fixes #711 and extends #713

### DIFF
--- a/dropwizard-metrics/src/main/java/io/dropwizard/metrics/BaseReporterFactory.java
+++ b/dropwizard-metrics/src/main/java/io/dropwizard/metrics/BaseReporterFactory.java
@@ -142,8 +142,21 @@ public abstract class BaseReporterFactory implements ReporterFactory {
         return new MetricFilter() {
             @Override
             public boolean matches(final String name, final Metric metric) {
-                return (!getIncludes().isEmpty() && getIncludes().contains(name))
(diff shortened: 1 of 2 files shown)


eepstein:
I'm looking at the logic for the matches() method of the MetricFilter and see that OR logic is used for the case where both excludes and includes is set.  Shouldn't that be AND logic - which is more common:

                if (useIncl && useExcl) {
                    return getIncludes().contains(name) || !getExcludes().contains(name);
                }

becomes:

                if (useIncl && useExcl) {
                    return getIncludes().contains(name) && !getExcludes().contains(name);
                }

In terms of filtering, this is more common and treats the filters as a sequence of excludes, then includes.  With the special default of no includes means include all.

Normal operation: include all.  Add an exclude filter: exclude whatever's listed.  Add only an include filter: just include what's listed.   Add both: first exclude; then include what's left.

This common pattern makes more sense with Regexs.  But then, the whole filtering works better with those.

PR #795, opened 2014-11-25
Use Type parameter for configuration in initialize

dhs3000:
Use Type parameter for configuration in both, run and initialize so that one can for example add another ConfigurationBundle in initialize method.


carlo-rtr:
the build broke, I have reverted the change. Please submit a new pr

dhs3000:
Thanks, I created https://github.com/dropwizard/dropwizard/pull/984

PR #803, opened 2014-12-04
 Real type is not always set properly in Java

meshuga:
Not every time a real type of a class is set using getClass method. It is better to rely on getting class from a field than getting from this reference. Sometimes real class type is held in this$0 synthetic field.


meshuga:
Branch https://github.com/meshuga/dropwizard/commits/issue-804-failing-test contains failing test. Sorry for not using the base class for the test case, but I couldn't find any other way ;)

carlo-rtr:
The build is not passing. Also, can you remove the issue from the package names? Feel free to reopen once these are addressed.

PR #822, opened 2014-12-28
Simplify support for Guava Optional<T> in dropwizard-jersey

joschi:
Instead of multiple providers for each type of JAX-RS parameter (e. g. @QueryParam or @HeaderParam) we now have a single ParamConverterProvider.

Refs jersey/jersey#98, [JERSEY-2612](https://java.net/jira/browse/JERSEY-2612)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/guava/OptionalParamConverterProvider.java
```diff
@@ -0,0 +1,72 @@
...
+
+                @Override
+                public String toString(final T value) throws IllegalArgumentException {
+                    return value.toString();
+                }
+            };
+        }
+        final Set<ParamConverterProvider> converterProviders = Providers.getProviders(locator, ParamConverterProvider.class);
+        for (ParamConverterProvider provider : converterProviders) {
+            @SuppressWarnings("unchecked")
```
Trundle: Does the `@SuppressWarnings` here have any effect? Can't really see what could cause a unchecked warning right now.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/guava/OptionalParamConverterProvider.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private final ServiceLocator locator;
+
+    @Inject
+    public OptionalParamConverterProvider(final ServiceLocator locator) {
+        this.locator = locator;
+    }
+
+    @Override
+    public <T> ParamConverter<T> getConverter(final Class<T> rawType, final Type genericType, final Annotation[] annotations) {
```
Trundle: I think there should be a check whether it's really an `Optional` (e.g. `if (!Optional.class.equals(rawType)) { return null; }`). Right now, a converter is returned for any generic that has a type parameter for which some converter exists, resulting in a `ClassCastException` when being used (lines 37 and 54).

dropwizard-jersey/src/main/java/io/dropwizard/jersey/guava/OptionalParamConverterProvider.java
joschi: You're completely correct. The suppression of unchecked warnings is not necessary here.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/guava/OptionalParamConverterProvider.java
joschi: Whoops. Of course the actual type of the parameter should be checked. Thanks!

Trundle:
LGTM

carlo-rtr:
Nice!

Issue #789, opened 2014-11-21
JerseyClient not reporting metrics


mabn:
JerseyClientBuilder does not register any metrics. In particular JerseyClientBuilder.builder is not used for anything - it's possible to pass various parameters (JerseyClientConfiguration, HttpRequestRetryHandler, etc), but they are ignored.

That affects v0.8.0-rc1

arteam:
Good, #791 added metrics for `ConnectionManager`. 

But unfortunately request timers remained not reported. I guess that because Jersey [ApacheConnector](https://github.com/jersey/jersey/blob/master/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java) doesn't expose a way to set a custom `HttpRequestExecutor`, so we can't instrument the client.
We can't override or delegate this connector because it's package-protected.
What options do we have?

* Report an issue to Jersey and wait for possible resolution
* Shade Jersey `ApacheConnector` to *dropwziard-client* and maintain it separately 
* Write own Connector
* Embrace absence of request timer metrics for Jersey client

Personally I more inclined to the second option.

cmicali:
@arteam It does solve it, but it needs to be merged and DW updated for that, and given the current pace of metrics progress I'm nervous it may never happen :worried:

PR #827, opened 2014-12-30
Testing: Add listeners to DropwizardAppRule

cmicali:
This PR adds the ability for tests using **DropwizardAppRule** to hook the **Application** lifecycle.  It additionally includes a convenience method to add **Managed** objects to the lifecycle.

```java
@ClassRule
public static final DropwizardAppRule<TestConfiguration> RULE =
    new DropwizardAppRule<TestConfiguration>(MyApp.class, resourceFilePath("my-app-config.yaml"))
            .manage(new MyTestManagedObject());
```

There are many cases where tests need to add **Managed** objects to the application lifecycle and execute code in the application **onRun** method.  For example, many of our services use an ActiveMQ bundle which handles connections to a remote ActiveMQ server.  When testing these services we want to standup a local broker and override the config to use this local broker.  The changes in this PR are what make that possible.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/DropwizardAppRule.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/DropwizardAppRule.java
@@ -63,7 +63,7 @@ public class DropwizardAppRule<C extends Configuration> extends ExternalResource
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


joschi:
LGTM. Thanks!

cmicali:
A good amount of issue #789 is resolved via this merge

PR #830, opened 2015-01-01
add function to retrieve className from task and test. fixes #829

### LINKED ISSUE #829: Anonymous task implementation results in (null) on start-up log
When I do

    environment.admin().addTask(new Task("foo"){//implement execute//})

instead of implementing a concrete Task sub-class, I end up with

    INFO  [2015-01-01 15:22:44,045] io.dropwizard.setup.AdminEnvironment: tasks = 
        POST    /tasks/foo (null)

on server-start-up.

"null" normally would be the implementing class name. 

I guess, since it it just an info, a possible solution would be to just leave out the (null) part in that case.

Update: I implemented a solution that falls back to getClass().getName() when getCanonicalName() is null, so the log displays

        POST /tasks/foo (my.implementing.Class$1)

which allows me to identify the code where the Task is implemented.

jangalinski:
instead of task.getClass().getCanonicalName(), task.getClass().getName is used for anonymous implementations of task.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/AdminEnvironment.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/AdminEnvironment.java
@@ -15,8 +15,6 @@ import org.eclipse.jetty.util.component.LifeCycle;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.annotation.Nullable;
-
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
@@ -30,7 +28,6 @@ public class AdminEnvironment extends ServletEnvironment {
 
     @VisibleForTesting
     final Function<Task,String> getTaskClassName = new Function<Task, String>() {
-        @Nullable
         @Override
         public String apply(final Task task) {
             // canonicalName is null for anonymous classes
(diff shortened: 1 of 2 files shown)


dropwizard-core/src/test/java/io/dropwizard/setup/AdminEnvironmentTest.java
```diff
@@ -11,6 +11,7 @@
 import javax.servlet.ServletRegistration;
 import java.io.PrintWriter;
 
+import static org.assertj.core.api.Assertions.*;
```
joschi: Please don't use wildcard imports.

dropwizard-core/src/main/java/io/dropwizard/setup/AdminEnvironment.java
```diff
@@ -59,11 +72,13 @@ public void addTask(Task task) {
...
         final StringBuilder stringBuilder = new StringBuilder(1024).append(String.format("%n%n"));
 
+
         for (Task task : tasks.getTasks()) {
+
             stringBuilder.append(String.format("    %-7s /tasks/%s (%s)%n",
                                                "POST",
                                                task.getName(),
-                                               task.getClass().getCanonicalName()));
+                                               getTaskClassName.apply(task)));
```
joschi: I'd simply use [´MoreObjects#firstNonNull(T, T)`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/MoreObjects.html#firstNonNull(T,%20T)) in this case. No need for this `Function<Task, String>`.

jangalinski:
I wanted to keep the change testable, so i went for an extra function.  But firstnonnull is a good idea, no need to test guava here.   Will refactor and commut again.

jangalinski:
Question: why not just use getClass().getName()? No null-checks, no functions, just the class name, and done.
I mean, its only logging and when inner classes are represented with i.d.Bar$Foo instead i.d.Bar.Foo, who cares?

joschi:
Thanks for your PR!

PR #834, opened 2015-01-06
Warning about maximum pool size and unbounded queues

### LINKED ISSUE #575: ExecutorServiceBuilder should not allow maxThreads() on unbounded queues.
"If there are more than corePoolSize but less than maximumPoolSize threads running, a new thread will be created only if the queue is full." 

Setting maxThreads on an unbounded queue has no effect. Throwing an exception in the builder would help avoid reasoning errors regarding the behavior of the executor.

arteam:
Fix #575
JDK ThreadPoolExecutor behaves a rather different than users can expect.

A user expects that parameter 'maximumPoolSize' controls maximum amount
of worker threads for processing tasks. When there is more tasks than
amount of worker threads then the pool grows incrementally to that size.

But actually ThreadPoolExecutor will grow the pool only if a queue is
reached it's bound. If the queue is unbounded then it won't increase
pool size beyond 'corePoolSize' limit. This seems rather awkward because
queue type shouldn't be related with speed of it's processing.

Anyway we should give a warning to users to improve their awareness
of the situation.

### DIFF
--- a/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
+++ b/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
@@ -1,12 +1,17 @@
 package io.dropwizard.lifecycle.setup;
 
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


joschi:
Thanks!

Issue #835, opened 2015-01-06
ResourceTestRule does not always use InMemoryTestContainer


Lugribossk:
There's no code in ResourceTestRule that forces it to use InMemoryTestContainerFactory for the JerseyTest instance it creates, despite the documentation saying that. The default logic in JerseyTest will default to using GrizzlyTestContainerFactory if it is on the classpath (see https://github.com/jersey/jersey/blob/master/test-framework/core/src/main/java/org/glassfish/jersey/test/TestProperties.java#L100 and https://github.com/jersey/jersey/blob/master/test-framework/core/src/main/java/org/glassfish/jersey/test/JerseyTest.java#L502). This can happen if the test using ResourceTestRule coexists with another test, that happens to subclass JerseyTest directly to use GrizzlyTestContainerFactory for its additional features.

PR #836, opened 2015-01-06
Add option to change test container factory, and explicitly default to InMemoryTestContainer

Lugribossk:
Fixes https://github.com/dropwizard/dropwizard/issues/835

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -187,7 +187,7 @@ loads a given resource instance in an in-memory Jersey server:
         }
     }
 
-Instansiate a ``ResourceTestRule`` using its ``Builder`` and add the various resource instances you
+Instantiate a ``ResourceTestRule`` using its ``Builder`` and add the various resource instances you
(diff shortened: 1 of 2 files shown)


joschi:
It would be awesome if you could add one example to the testing documentation, e. g. add the Grizzly test container to the project POM and set it in the resource test class.

Lugribossk:
Hmm, it looks like it's a little more complicated than that to get the GrizzlyWebTestContainer working. It expects a ServletDeploymentContext with ServletProperties.JAXRS_APPLICATION_CLASS set to the class name of the ResourceConfig to use, which it then creates a new instance of. But the ResourceConfig that ResourceTestRule creates is modified to add the resources the test needs, but that change is then not reflected in the instance created elsewhere.

Any ideas for how to solve that? Otherwise I think I will change it to just default to InMemoryTestContainer, and add a note in the documentation that more advanced tests will need to be JerseyTest subclasses.

joschi:
I've added some documentation and an example on how to use the Grizzly 2.x HTTP container for resource tests supporting `@Context` injections.

PR #839, opened 2015-01-10
Remove Jersey GZip Encoder

### LINKED ISSUE #774: GZipEncoder is forcibly enabled in DropwizardResourceConfig
I guess we need configuration for this instead of forcibly
```
EncodingFilter.enableFor(this, GZipEncoder.class);
```
in `DropwizardResourceConfig` consturctor

### LINKED ISSUE #821: LoggingFilter output is gzip-encoded in 0.8.0-rc1, change from 0.7.1 behavior
In DropWizard 0.7.1, I used LoggingFilter with gzip enabled (default).

        environment.jersey().property(ResourceConfig.PROPERTY_CONTAINER_REQUEST_FILTERS, new String[]{LoggingFilter.class.getName(), PMSRequestFilter.class.getName()});
        environment.jersey().property(ResourceConfig.PROPERTY_CONTAINER_RESPONSE_FILTERS, LoggingFilter.class.getName());

In DropWizard 0.8.0-rc1, I configured the same as:

        environment.jersey().register(new LoggingFilter(Logger.getLogger(LoggingFilter.class.getName()), true));

In Jersey 1.x, LoggingFilter logged the request and response entities by default. In Jersey 2, this is configured by with the "true" param.

Whereas with 0.7.1 the entities outputted by LoggingFilter were in plain text, in 0.8.0-rc1 they are GZIP-encoded. I have not modified gzip-related configurations from their defaults defaults or added any additional gzip filters when using either DW 0.7 or 0.8.

arteam:
Resolve #774, fix #821

Jersey GZipEncoder conflicts with Jetty BiDiGzipFilter and breaks
breaks backwards compatibility of Dropwizard configuration.

It always compresses responses if a client accepts gzip encoding,
which is not the same behaviour as in BiDiGzipFilter. The latter
is configured through a YML file, has more configuration options
and could be disabled externally. As a result users can't disable
response compressing even if they want to.

Also it conflicts with Jersey LoggingFilter, which is useful
for debug purposes.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -64,7 +64,6 @@ public class DropwizardResourceConfig extends ResourceConfig {
         register(OptionalMessageBodyWriter.class);
         register(OptionalParamFeature.class);
         register(new SessionFactoryProvider.Binder());
-        EncodingFilter.enableFor(this, GZipEncoder.class);
     }
 
     public static DropwizardResourceConfig forTesting(MetricRegistry metricRegistry) {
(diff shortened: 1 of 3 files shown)


joschi:
As always: Thank you!

PR #840, opened 2015-01-10
Make endpoints information public

arteam:
List of endpoints is useful not only for logging but also as API for users.

This change was originally introduced in commit c0eb424, 
but has been lost after upgrading to Jersey. 

See PR #643 for reference


joschi:
Thanks!

PR #841, opened 2015-01-10
Make test instance constructors package-protected

arteam:
Constructors that don't accept Environment should be package-protected because they are needed only in tests. 

In application code they confuse users.


joschi:
Thanks for this PR!

I'm closing it without merging, though, because I think it should still be possible to use dropwizard-client without pulling in a complete Dropwizard environment (i. e. having to provide a valid `Environment` instance to the constructors of *ClientBuilder).

arteam:
OK, let it be so.

In the fact Apache HTTP client doesn't require complete environment at all. I just thought we confuse users with two constructors without explanation which one should be used.

PR #842, opened 2015-01-11
Add Jersey Bean Validation Support

### LINKED ISSUE #625: JacksonMessageBodyProvider throws exception when post body is null
JacksonMessageBodyProvider in Dropwizard 0.6 allowed null post bodies. As a result, the resource methods would get a "default" instance (calling the default constructor) of the parameter type.

However in 0.7, JacksonMessageBodyProvider throws an exception and the resource method never gets called. Posting {} reverts to the old behaviour.

Because our application is now expecting the old behaviour (i.e. accepted null post bodies reverting to default constructors on resource methods) in multiple places (namely many client applications) changing all the calls is likely to be a nightmare.

Is there any way we can workaround this? Expected behaviour is to get null or an instance of the parameter created with the default constructor.

Thanks.

joschi:
See https://jersey.java.net/documentation/latest/bean-validation.html for details
about the Bean Validation support in Jersey. It enables the use of Java validation annotations like `@NotNull`, `@Min`, or `@Email` in the context of JAX-RS and Jersey.

Merging this could probably fix #625 and #633 (as both, the issue and the PR, are now outdated anyway).

### DIFF
--- a/dropwizard-jersey/pom.xml
+++ b/dropwizard-jersey/pom.xml
@@ -48,6 +48,17 @@
             <version>${jersey.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.glassfish.jersey.ext</groupId>
+            <artifactId>jersey-bean-validation</artifactId>
+            <version>${jersey.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.hibernate</groupId>
+                    <artifactId>hibernate-validator</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
             <groupId>io.dropwizard.metrics</groupId>
             <artifactId>metrics-jersey2</artifactId>
             <version>${metrics3.version}</version>
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -21,6 +21,7 @@ import org.glassfish.jersey.server.monitoring.ApplicationEvent;
 import org.glassfish.jersey.server.monitoring.ApplicationEventListener;
 import org.glassfish.jersey.server.monitoring.RequestEvent;
 import org.glassfish.jersey.server.monitoring.RequestEventListener;
+import org.glassfish.jersey.server.validation.ValidationFeature;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -64,6 +65,7 @@ public class DropwizardResourceConfig extends ResourceConfig {
         register(OptionalMessageBodyWriter.class);
         register(OptionalParamFeature.class);
         register(new SessionFactoryProvider.Binder());
+        register(ValidationFeature.class);
     }
 
     public static DropwizardResourceConfig forTesting(MetricRegistry metricRegistry) {


skamille:
I'm OK with this, are we comfortable with adding a new dependency? @carlo-rtr @ryankennedy if one of you 2 +1s let's put it in

ryankennedy:
I assume the exclusion of hibernate-validator is because we're pulling it in as part of dropwizard-validation and don't want Jersey's bean validation bits to stomp the version we're pulling? Is that going to be safe to do going forward? Is there any way to move these bits to dropwizard-validation instead so all the validation-related bits stay in one place? I assume removing hibernage-validator from dropwizard-validation would suddenly cause some awkward runtime and, maybe, test missing class failures way over in dropwizard-jersey. Maven exclusions make me nervous for long-term maintenance.

I'm a little confused as to why people are POSTing empty bodies, having their resources take an argument, and complaining when it says foo = null is not a valid foo. It's triggering my smell test. Like…why would you do that? Especially the one talking about an empty POST for a login request being a valid request. I'm not sure how opinionated to be about that one. But I'm not sure that's a problem I'd aim to solve.

Out of curiosity, why aren't these validations just working right now? Is this for validation annotations in the resource method parameter list (i.e. Foo getFoo(@NotNull FooRequest req))? I assume validation annotations in the representation class (i.e. class FooRequest { @NotNull String id; }) are already working since we advertise that in the docs.

I'm fine with the new dependency if we don't see a way we'd deeply regret adding this feature down the road.

jplock:
@joschi did you want to respond to @ryankennedy before someone merges this in?

carlo-rtr:
I think the validations are useful if you aren't using a bean. There are many annotations, that make more sense than NotNull. http://docs.oracle.com/javaee/7/api/javax/validation/constraints/package-summary.html

I suspect the exclusion in org.hibernate.hibernate-validator is to avoid dependency divergence. I doubt trying to use jersey module without providing org.hibernate.hibernate-validator would work. Perhaps we should also explicitly add a dependency on org.hibernate.hibernate-validator in dropwizard-jersey. This way both dropwizard-jersey & dropwizard-validation have the right deps. We can ensure version consistency by moving the version to a property field defined in the parent pom

jplock:
@joschi did you want to address @carlo-rtr 's comment and we'll add this into `0.9.0`?

joschi:
Ah, sorry for not responding. Completely overlooked the replies on this PR.

> I assume the exclusion of hibernate-validator is because we're pulling it in as part of dropwizard-validation and don't want Jersey's bean validation bits to stomp the version we're pulling?

Yes, exactly. In the end it's excluded to make the Maven Enforcer plugin happy. The alternative would be to add Hibernate Validator to the dependencyManagement section of the parent POM to enforce a certain version of it. We should probably do that anyway.

This being said, it's rather unlikely that we remove Hibernate Validator in the foreseeable future because it's quite deeply embedded in dropwizard-validation.

> Is there any way to move these bits to dropwizard-validation instead so all the validation-related bits stay in one place?

dropwizard-jersey pulls in dropwizard-validation (see [pom.xml](https://github.com/dropwizard/dropwizard/blob/58c4d52ce212d05b691e80e5ff687c92a24d510f/dropwizard-jersey/pom.xml#L20-24)) and this feature is tightly coupled with Jersey. It's completely useless without dropwizard-jersey, so I assumed putting it in there is the right place™.

> Out of curiosity, why aren't these validations just working right now? Is this for validation annotations in the resource method parameter list (i.e. Foo getFoo(@NotNull FooRequest req))?

Yes, exactly. Some annotations (namely `@Valid` and `@Validated`) can already be used in Resource classes, others are currently limited to value classes.

If the addition of this feature doesn't impose a large performance overhead (which we should test), I don't see a reason not to add it by default.

ryankennedy:
Cool, thanks for the explanation Jochen. I'm a +1.

Ryan

PR #846, opened 2015-01-13
Add CommandRunner to dropwizard-testing

oillio:
I needed a way to run commands during integration testing.

This PR adds a CommandRunner that supports running Commands against a provided Application and configuration.  I also refactored DropwizardTestSupport so that it and CommandRunner can share code.

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -288,3 +288,12 @@ running and stop it again when they've completed (roughly equivalent to having u
             assertThat(response.getStatus()).isEqualTo(302);
         }
     }
+
+If you need to run a Command in your integration tests use the ``CommandRunner`` class.  You may provide a
+command to run, or you can provide the command name if it is already loaded in your Application.
+``CommandRunner`` supports the same parameters and configuration options as ``DropwizardAppRule``.
+
+.. code-block:: java
+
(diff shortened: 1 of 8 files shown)


arteam:
Thanks for you pull request.

I'm curious, what's wrong with running commands through
````java
DropwizardAppRule<HelloWorldConfiguration> rule;
// ...
rule.getApplication().run("command", "parameter");
````

If you want to write an integration test for a command without starting an HTTP server, you can just create an application class by hand.

Do your commands have some state you want to inspect? This seems like a bad practice for me.

oillio:
Yep, the goal is to run the Command under test without also running the ServerCommand.  I wanted a way to run a Command as close as possible to how it will be run in production.

It is common for me to have setup and configuration occurring in my Application initialize method which is required by my Commands.  It seems wasteful to need to duplicate this initialization, or do other backflips, so that I can create an Application class for command testing.

arteam:
I see your case. I will try to look more closely on this PR this week.

arteam:
I gave it more thought and came to the conclusion that the problem is not worth to solve. Al least it's not worth of creation of another abstraction level. Most commands are are simple and stateless, so they can be tested by invoking the `run` method of `Application`. I think, it's enough for most users.
If people will ask for this functionality, we can return to this PR.

Sorry.

PR #847, opened 2015-01-13
either log or throw - remove LOGGER

jangalinski:
when starting the server fails, the console stacktrace is redundant, the rethrown Exception is handled by the Application as well. So there shoul be no logging here.


joschi:
The log message makes sure that the reason for the program termination will eventually end up in the right place (e. g. a log file or a remote log host). If the statement is removed, the exception stack trace is only printed on standard out.

PR #848, opened 2015-01-14
Connection request timeout

mabn:
When number of connections per route reaches the limit (dropwizard default is 1024) a call to fetch the next connection will block. The default timeout is not set so the wait might potentially exceed the expectations. A special case is that the thread could stuck if connections are leaked somewhere instead of generating timeout exception.

This PR adds a way to set this timeout by adding an attribute to HttpClientConfiguration and support for it for both JerseyClient and HttpClient.


arteam:
+1 for this

joschi:
Thanks!

PR #851, opened 2015-01-16
fixing an issue where the "name" prefix for HttpClient metrics would not...

smolloy:
I noticed that the metrics for HttpClients, when named, did not include the name.  For example: org.apache.http.client.HttpClient./api/v1/services.get-requests even though I specified a name and the connection manager metrics correctly showed it.  I found a simple omission of this parameter in the HttpClientBuilder.  Now my metrics look like this: org.apache.http.client.HttpClient.PagerDuty./api/v1/services.get-requests and I am happy (PagerDuty == the name I used to build the client)


joschi:
Nice catch. Thanks!

PR #853, opened 2015-01-18
change dropwizard-forms packaging

Athou:
Packaging needs to be changed to jar to be able to use the new bundle.


joschi:
Argh, yes of course. Thanks!

PR #854, opened 2015-01-18
Configure Jersey client thread pool with a bounded queue

arteam:
In relation to #834 

Jersey's async thread pool is configured with 2 parameters: *minThreads* and *maxThreads*. 
The problem is that the pool will not grow up than *minThread* because its work queue is 
unbounded. It's a property of JDK *ThreadPoolExecutor*.

A solution to this problem is to set a bounded queue as the work queue and provide facility 
to users to set the queue size.

By default queue size is set to 8. It's a middle ground between creating too many threads 
and waiting too much time in the queue without spawning new threads.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -879,6 +879,7 @@ See JerseyClientConfiguration_ and HttpClientConfiguration_ for more options.
     jerseyClient:
       minThreads: 1
       maxThreads: 128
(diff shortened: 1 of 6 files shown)


jplock:
@arteam this needs to be updated for merging

arteam:
I've rebased this against master.
Changes should be compatible with the current `JerseyClientBuilder`.

PR #866, opened 2015-01-28
Add facility to override the default configuration

### LINKED ISSUE #636: Problem with setting port on command line
According to SimpleServerFactory docs:

> A startup script can override the port via {@code -Ddw.server.connector.port=$PORT}.

However, when I try:

    java -Ddw.server.connector.port=8081 -jar my.jar server config.yml

I get:

> config.yml has an error:
>  Failed to parse configuration at: server.connector; Unexpected token (END_OBJECT), expected FIELD_NAME: missing property 'type' that is to contain type id  (for class io.dropwizard.jetty.ConnectorFactory)
> at [Source: N/A; line: -1, column: -1] (through reference chain: com.acmerocket.stratum.StratumConfiguration["server"]->io.dropwizard.server.SimpleServerFactory["connector"])

Without `-Ddw.server.connector.port=8081`, the app comes up fine.

### LINKED ISSUE #865: [0.8.0-rc2] Setting the port on the command line fails.
From the docs,

```
java -Ddw.server.applicationConnectors[0].port=9090 [...]
```

Trying this with a Dropwizard 0.8.0-rc2 app fails as follows:

```
$ java -Ddw.server.applicationConnectors[0].port=9090 -jar build/libs/dropbear-all.jar server
Exception in thread "main" java.lang.IllegalArgumentException: Unable to override server.applicationConnectors[0].port; node with index not found.
```

I was rather hoping that'd set the port, instead, so I could run Dropwizard on Heroku.

An app reproducing this problem is available at [ojacobson/dropbear](/ojacobson/dropbear). To reproduce:

```
./gradlew build
java -Ddw.server.applicationConnectors[0].port=9090 -jar build/libs/dropbear-all.jar server
```

arteam:
Fix #636, Fix #865

If a YAML configuration file is not provided, Dropwizard by default uses an empty object tree as a config representation.

This approach has a flaw in case of configuration overriding. For this to work, a full object tree analogous by structure with the configuration should be present.

A better approach is to instantiate a configuration object with defaults, ask Jackson to build an object tree and pass this tree to the overriding routine. After that Jackson will create a desired configuration
from a changed tree.

In case if there is no overriding Jackson will just build the same configuration object as the default.

### DIFF
--- a/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationFactory.java
+++ b/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationFactory.java
@@ -7,7 +7,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.exc.InvalidFormatException;
 import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import com.fasterxml.jackson.databind.node.TreeTraversingParser;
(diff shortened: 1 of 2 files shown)


joschi:
Thanks!

prakharrke:
This issue still persists in v1.3.8

PR #867, opened 2015-01-28
Syntax error in docs scrubbed, and MultiPartBundle includes config interface now

pavangupta:
Configuration interface is now carried into the MultiPartForm bundle and I grabbed a quick syntax error in the docs.  Everything should match up.

### DIFF
--- a/docs/source/manual/forms.rst
+++ b/docs/source/manual/forms.rst
@@ -20,7 +20,7 @@ Then, in your application's ``initialize`` method, add a new ``MultiPartBundle``
 
     @Override
     public void initialize(Bootstrap<ExampleConfiguration> bootstrap) {
-        bootstrap.addBundle(new MultiPartBundle<ExampleConfiguration>();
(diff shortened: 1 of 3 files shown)


joschi:
Please open two separate pull requests for those issues.

Your change also doesn't pass the test suite.

joschi:
Why would `MultiPartBundle` need a `Configuration` instance?

Closing this PR (the typo in the docs has been addressed in a separate commit).

pavangupta:
Totally agreed on the ``Configuration`` not making sense as a requirement, but I guessed the docs had some reason I couldn't see.  To be clear, the docs are still suggesting an illegal inclusion of ``ExampleConfiguration`` in the ``MultiPartBundle`` declaration.

joschi:
@pavangupta Hah, you're right. The docs are wrong in this case. Thanks for mentioning it!

pavangupta:
@joschi woot, happy to help!

PR #869, opened 2015-01-29
Add the database (connection pool) configuration reference

arteam:
Currently there is no database (connection pool) configuration reference in the official documentation. There are only YML examples in JDBI and Hibernate sections.

It would be great to have this configuration in the single place along with others modules.


joschi:
Thanks!

arteam:
That was quick.
Thanks for appreciation, I'm glad to contribute.

PR #871, opened 2015-01-30
Adding a file size based FileAppenderFactory

HaVonTe1:
This adds a AppenderFactory for logging with logback. The appender creates a file size based TriggerPolicy. So log files roll over when a given file size is reached.

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/SizeBasedRollingFileAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/SizeBasedRollingFileAppenderFactory.java
@@ -1,0 +1,101 @@
+package io.dropwizard.logging;
+
+
+import io.dropwizard.logging.AbstractAppenderFactory;
+import io.dropwizard.logging.AppenderFactory;
+import io.dropwizard.logging.FileAppenderFactory;
+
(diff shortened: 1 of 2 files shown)


joschi:
I'm not sure if this functionality should be included in the existing `FileAppenderFactory` instead of creating a new one.

@dropwizard/committers Any thoughts on this?

carlo-rtr:
if you're suggesting we make the archiving strategy configurable in FileAppenderFactory, I like it. We are already ignoring some configuration if archive is set to false, so I think it make sense to have an archiveStrategy. Not present, would be have as false. TimeBased would do what's currently in master. SizeBased would do what's on this branch. wdyt?

HaVonTe1:
I can see why you dislike the idea of adding a new AppenderFactory instead of enriching the existing one. At the time I made this fork I didn't liked the idea of adding a bunch of new properties and conditions in FileAppenderFactory. But I will try to do so if you decline this PR. I realy like DW and I realy missed the SizeBased TriggerPolicy :)

carlo-rtr:
@HaVonTe1 are you still interested in pursuing this?

HaVonTe1:
I am. But it would be OK for me if you decline this PR.

PR #791, opened 2014-11-21
JerseyClientBuilder: Add back metrics and use provided configuration values

cmicali:
This PR improves the JerseyClientBuilder with:

* Add back metrics provided by  **InstrumentedHttpClientConnectionManager**
* Respect the timeout, cookie, and credential provider configuration parameters

Additionally it:

* Adds a few tests for the configuration parameters
* Moves common code shared between **HttpClientBuilder** and **JerseyClientBuilder** into an **ApacheClientBuilderBase** class

I recommend including this in 0.8, as the existing JerseyClientBuilder is a bit broken without these enhancements.

**Note**: 

This PR does **not** include the request timing that InstrumentedHttpRequestExecutor provided.  I have submitted a metrics PR (https://github.com/dropwizard/metrics/pull/705) to add this, and will submit a dropwizard PR when that is merged.

PR #874, opened 2015-02-03
Add Jersey DropwizardApacheConnector  

### LINKED ISSUE #789: JerseyClient not reporting metrics

### LINKED ISSUE #849: JerseyClient does not respect keepAlive parameter
JerseyClient does not respect keepAlive parameter - underlying HttpClient uses DefaultConnectionKeepAliveStrategy. If the server does not specify keep-alive timeout in the response headers the connection will survive in the connection pool until it's closed.

In case of idle connections they tend to be closed on the server side (which connection pool is unaware of). The result is that the next request:
- leases a (stale) connection from the pool
- sends the request
- tried to read from the underlying socket but fails
- invalidates the connection and retries
AFAIR there are 2 retry attempts (retry configuration is also ignored).

In addition it's not possible to easily fix it because the client is built inside the constructor of org.glassfish.jersey.apache.connector.ApacheConnector (which is package-protected). 
It's not possible to change ConnectionKeepAliveStrategy via reflection because the field in question is final.

I see following options:
1 - spawn a separate thread which will periodically expunge idle connections - see http://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html#d5e405
2 - write own version of ApacheConnector which builds underlying HttpClient correctly (could reuse HttpClientBuilder here). It would mean reverting pull request #791 and doing it differently, but it would fix all the things that were broken in JerseyClient in 8.0

I think the 2nd seems to be the best as it would allow to have the same code/bahavior of HttpClient in both cases (when using jersey or apache clients)

### LINKED ISSUE #857: JerseyClient fails to set Content-Encoding to gzip
Following code in JerseyClientBuilder enables gzip compression:
```
if (configuration.isGzipEnabled()) {
    client.register(new GZipDecoder());
    client.register(new ConfiguredGZipEncoder(configuration.isGzipEnabledForRequests()));
}
```

Unfortunately it fails to set Content-Encoding header when using this test code:
```
WebTarget target = client.target(uri);
Response response = null;
try {
    response = target
        .request(MediaType.APPLICATION_JSON)
        .post(Entity.entity(string, MediaType.APPLICATION_JSON));
    return response.getStatus();
} finally {
    response.close();
}
```

It's easy to spot that the header is missing when DEBUG logging is enabled. Also a warning is logged indicating that ConfiguredGZipEncoder implementation is unsupported:

```
2015-01-21 18:45:15,078Z WARN o.g.j.m.i.HeaderUtils: There are some request headers that have not been sent by connector [org.glassfish.jersey.apache.connector.ApacheConnector]. Probably you added those headers in WriterInterceptor or MessageBodyWrite
r. That feature is not supported by the connector. Please, do not modify headers in WriterInterceptor or MessageBodyWriter or use default HttpUrlConnector instead.
Unsent header changes: [Content-Encoding]
```

### LINKED ISSUE #861: SSL setup stuff is missing in JerseyClientBuilder [0.8.0-rc2]
Jersey provides the following SSL configuration:
https://jersey.java.net/documentation/latest/client.html#d0e5012

I would expect to have in JerseyClientBuilder at least `sslContext` and `hostnameVerifier` options.
Something like that: 
`
Client client = jerseyClientBuilder.sslContext(sslContext).hostnameVerifier(hostnameVerifier).build();
`

### LINKED ISSUE #863: Not being able to setup request retry count using JerseyClient
JerseyClientBuilder is not using this parameter provided in HttpClientConfiguration. 
The issue is a little bit bigger though: currently there is no way to specify request retry handler for Jersey client built with ApacheConnector (it is by default 3 retries).

arteam:
Fix #861, Fix #857, Fix #849, Fix #789, Fix #863

Add our own version that uses the pre-configured Apache HTTP client.

Dropwizard has a stable and mature Apache HTTP client with an extensive configuration. Rather than trying to simulate configuration with Jersey properties, we can just use a built Apache client and write a custom connector that does a glue work between Jersey and Apache client's abstractions.

With this approach we have a single place of configuration. That reduces burden of supporting a new feature both in Apache and Jersey clients.

This also reverts the approach taken in #791 with a base class that contains common configuration.

It would be great if someone could try to test this more thoroughly because I did only sanity tests.


jplock:
+1

pparth:
+1

dropwizard-client/src/test/java/io/dropwizard/client/JerseyClientIntegrationTest.java
```diff
@@ -0,0 +1,268 @@
...
+
+/**
+ * Integration test of {@link org.glassfish.jersey.client.JerseyClient}
+ * with {@link io.dropwizard.client.DropwizardApacheConnector}
+ */
+public class JerseyClientIntegrationTest {
+
+    private static final String TRANSFER_ENCODING = "Transfer-Encoding";
+    private static final ObjectMapper JSON_MAPPER = Jackson.newObjectMapper();
+    private static final String APPLICATION_JSON = "application/json";
```
Trundle: Can be replaced by a static import of `javax.ws.rs.core.MediaType.APPLICATION_JSON`.

dropwizard-client/src/test/java/io/dropwizard/client/JerseyClientIntegrationTest.java
```diff
@@ -0,0 +1,268 @@
+package io.dropwizard.client;
+
+import com.codahale.metrics.MetricRegistry;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
```
Trundle: Unused import (as well as `java.net.URI` and `java.util.zip.Deflater`).

dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
```diff
@@ -0,0 +1,273 @@
...
+
+        @Override
+        public long getContentLength() {
+            return buffer.length;
+        }
+
+        @Override
+        public InputStream getContent() throws IOException, IllegalStateException {
+            // Shouldn't be called
+            return null;
```
Trundle: Shouldn't it rather throw an `UnsupportedOperationException`? Alternatively, it could return `new ByteArrayInputStream(buffer)`.

arteam: Agree with that. 

It's better to plainly  throw an exception to indicate that we don't 
support this method at all.

dropwizard-client/src/test/java/io/dropwizard/client/JerseyClientIntegrationTest.java
arteam: Fixed, thanks

dropwizard-client/src/test/java/io/dropwizard/client/JerseyClientIntegrationTest.java
arteam: I forgot that constants could also be imported. 
Replaced.

Trundle:
LGTM

joschi:
Thanks, @arteam! And thanks @Trundle for reviewing.

PR #875, opened 2015-02-04
Fixed connection leak in CloseableLiquibase

eiden:
When creating an instance of `CloseableLiquibase`, a jdbc connection is fetched from the supplied datasource. This connection is never closed in `CloseableLiquibase#close`, which results in a connection leak. 

I'm not sure if catching `LiquibaseException` is the "right" thing to do, let me know if I should just propagate the exception.

### DIFF
--- a/dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibase.java
+++ b/dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibase.java
@@ -6,14 +6,10 @@ import liquibase.database.jvm.JdbcConnection;
 import liquibase.exception.LiquibaseException;
 import liquibase.resource.ClassLoaderResourceAccessor;
 import liquibase.resource.FileSystemResourceAccessor;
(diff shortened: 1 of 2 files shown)


dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibase.java
```diff
@@ -29,6 +33,12 @@ public CloseableLiquibase(ManagedDataSource dataSource, String file) throws Liqu
 
     @Override
     public void close() throws Exception {
+        try {
+            database.close();
+        } catch (LiquibaseException e){
```
joschi: I think it would be better to remove the `catch` block just let the exception propagate (the method already declares to throw `Exception`) and put the `dataSource.stop()` call into the `finally` block of the `try-finally` statement.

eiden:
@joschi Thanks for reviewing this so quick! I have changed the exception handling to just propagate `LiquibaseException`

joschi:
@eiden Thanks!

PR #876, opened 2015-02-05
Update JDBI Version and Remove Unnecessary Classmate Exclusion

brianm:
The version bump is straightforward.

Classmate *is* used by jdbi, but it is shaded and renamespaced. There is no declared dependency on classmate.


joschi:
Thanks!

PR #884, opened 2015-02-18
Updating the BaseReporterFactory to optionally support regex

ToadJam:
Updating the BaseReporterFactory to optionally support regex in its includes/excludes fields. This implementation makes a few assumptions:

- By default any existing implementations or configurations that use the BaseReporterFactory should not have their behavior changed
- The user should be able to toggle on regular expressions by changing data, without modifications to the code
- The code should be easily extended to support still-more complicated matching functionality (potentially by making the StringMatchingStrategy public and adding it as a parameter to the getFilter() call)

These assumptions influenced my design decisions.

Please let me know what you think!

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -630,6 +630,7 @@ durationUnit           milliseconds   The unit to report durations as. Overrides
 rateUnit               seconds        The unit to report rates as. Overrides per-metric rate units.
 excludes               (none)         Metrics to exclude from reports, by name. When defined, matching metrics will not be reported.
 includes               (all)          Metrics to include in reports, by name. When defined, only these metrics will be reported.
+useRegexFilters        false          Indicates whether the values of the 'includes' and 'excludes' fields should be treated as regular expressions or not.
 frequency              (none)         The frequency to report metrics. Overrides the default.
 ====================== =============  ===========
 
--- a/dropwizard-metrics/src/main/java/io/dropwizard/metrics/BaseReporterFactory.java
+++ b/dropwizard-metrics/src/main/java/io/dropwizard/metrics/BaseReporterFactory.java
@@ -11,6 +11,7 @@ import io.dropwizard.util.Duration;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 
 /**
  * A base {@link ReporterFactory} for configuring metric reporters.
@@ -47,6 +48,12 @@ import java.util.concurrent.TimeUnit;
  *         reported. See {@link #getFilter()}.</td>
  *     </tr>
  *     <tr>
+ *         <td>useRegexFilters</td>
+ *         <td>false</td>
+ *         <td>Indicates whether the values of the 'includes' and 'excludes' fields should be
+ *         treated as regular expressions or not.</td>
+ *     </tr>
+ *     <tr>
  *         <td>frequency</td>
  *         <td>1 second</td>
  *         <td>The frequency to report metrics. Overrides the {@link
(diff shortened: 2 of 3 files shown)


ToadJam:
That works for me - I'll go ahead and make the change unless someone prefers my initial solution

eepstein:
Hi Travis,

Was looking at the latest request and I don't see the simpler change.  Can you point me to your branch?

Thanks.

ToadJam:
I haven't pushed them yet - I was waiting a few days to see if there was any additional feedback. I'll get the new version up on Monday.

arteam:
I'm more fond to the original implementation. 

I think a switch is a more simple and transparent solution for users. Having 4 parameters  (_includes_, _excludes_, _includesRegex_, excludesRegex_) doesn't sound great for me. If you use regular expressions - you usually don't need ordinal includes and excludes. Also ability to use several regular expressions is a benefit over writing one big regex monster.

arteam:
@ToadJam, thanks for contributing. This looks good and I'm going to merge it.

Couple of suggestions:
* I think it's better to split the `BaseReporterFactoryTest.matches` test to 2 tests. One for base
includes and excludes and one with regular expressions. Doing so we isolate possible errors 
in the regular expression implementation to one test.  
* It would be great to update the configuration [reference](https://github.com/dropwizard/dropwizard/blob/master/docs/source/manual/configuration.rst) to keep it in sync with actual code.
* It's a kind of premature optimization, but what do you think about caching regex patterns?
Filtering is a frequent operation and compiling a regular expression requires some time. Essentially, this is a classic tradeoff between time and space and I think time matters more.
As a possible solution we could a Guava cache in `BaseReporterFactory` with a safe expiring policy. Something like that:
````java
LoadingCache<String, Pattern> patterns = CacheBuilder.newBuilder()
.expireAfterWrite(1, TimeUnit.HOURS)
.build(new CacheLoader<String, Pattern>() {
    @Override
    public Pattern load(String regex) throws Exception {
        return Pattern.compile(regex);
    }
});
....

public boolean containsMatch(ImmutableSet<String> matchExpressions, String metricName) {
    for (String regexExpression : matchExpressions) {
        if (patterns.getUnchecked(regexExpression).matcher(metricName).matches()) {
            return true;
        }
    }
    return false;
}
````

ToadJam:
I agree with your feedback @arteam and I've incorporated it in my latest commit

arteam:
Thanks! Some nitpicks and I think it's done.

* Could you make the field *_patternCache* without underscore, just *patternCache*? 
Despite that some projects (Jetty, Jackson, etc) find this style useful, we (Dropwizard) adhere 
to the [Google Java Style](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s5.1-identifier-names)
* Also this field and the `RegexStringMatchingStrategy` constructor could be private.
* Could you, please, separate the head line and body in your last git commit with an empty line?
Something like this:
````
Incorporating feedback from code review 

1.) splitting out the default/regex tests
2.) updating the configuration reference and 
3.) supporting caching of compiled regex expressions
````
It makes viewing the repository history much simpler and focused. You can just amend your last commit and push it.

ToadJam:
Thanks @arteam! I apologize, that last commit was sloppier than I would like, but I've amended the fixes and pushed them up again. Please take another look when you get the chance.

eepstein:
OK.

The thing is, if you have a switch to go between regexes and ordinal values you end up with 5 parameters.  In other words, the overall parameter count goes up.  Is there benefit ?

arteam:
Thanks, @ToadJam! Great work.

arteam:
@eepstein, Sorry, I don't understand what you mean. I beleive we have only 3 parameters 
(_includes_, _excludes_ and _useRegexFilters_). The last one just changes runtime treatement the first and second.

Anyway, the decision has been made and I've merged an implementation of the original design, that I liked. @ToadJam did a good work on this, this PR solves the problem and we proceed with that.

Sorry to dissapoint you, if you think it could be be done in other way.

eepstein:
Ah.  No, I misunderstood.  I like what @ToadJam did better: treat the existing parameters into regex expressions.  That makes good sense.  Thanks for explaining it.

eepstein:
There is, however, one thing I'd still change.  This has:

                 if (useIncl && useExcl) {
                    return stringMatchingStrategy.containsMatch(getIncludes(), name) ||
                            !stringMatchingStrategy.containsMatch(getExcludes(), name);
                }

but I think that should be:

                 if (useIncl && useExcl) {
                    return stringMatchingStrategy.containsMatch(getIncludes(), name) &&
                            !stringMatchingStrategy.containsMatch(getExcludes(), name);
                }

which if we generalize it to the case where the includes filter is empty, we no longer need to check the useIncl and useExcl variables at all.  Per #890

PR #887, opened 2015-02-21
Support for a custom name of a Hibernate bundle

### LINKED ISSUE #855: Multiple HibernateBundles in the same application registers healthcheck and metrics twice with "hibernate"
In https://github.com/dropwizard/dropwizard/blob/master/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/SessionFactoryFactory.java the line:
```final ManagedDataSource dataSource = dbConfig.build(environment.metrics(), "hibernate");```

causes issues for us because we use several different session factories for different databases. This causes invalid metrics in the old 0.7.0 dropwizard and in 0.8.0* dropwizard doesn't start because it discovers that the metric is registered twice.

One solution would be to support a configuration in DataSourceFactory which would default to hibernate to not break legacy code that depends on this metric.

arteam:
Resolve #855

By default the name of a Hibernate bundle (actually "hibernate)" is hardcoded. This name is used for the database pool metrics and the health check associated with the bundle. This causes conflicts when
several bundles used alongside in a project.

This change adds to the HibernateBundle class a protected method which returns a bundle name. This method can be overridden by users who want to set a custom name. The new name will be used for the bundle's database pool metrics and health check.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java
@@ -11,6 +11,8 @@ import io.dropwizard.setup.Environment;
 import org.hibernate.SessionFactory;
 
 public abstract class HibernateBundle<T extends Configuration> implements ConfiguredBundle<T>, DatabaseConfiguration<T> {
+    private static final String DEFAULT_NAME = "hibernate";
+
     private SessionFactory sessionFactory;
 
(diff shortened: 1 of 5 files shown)


jplock:
LGTM!

vandanagopal:
If I give a custom name to all the registered hibernate bundles, then I get the below error:
Unregistered Hibernate bundle: 'hibernate'

viswacvfk:
Are you able to register multiple hibernate bundles in the same dropwizard application? I am facing similar errors.

PR #418, opened 2013-11-02
Add support for HTTP byte ranges

davidillsley:
Accessing video from iOS requires server support for HTTP byte range requests. This pull request adds support for those requests and advertises the support for media assets.

Unit tests included.

PR #889, opened 2015-02-22
Range header support. Replaces #418

jplock:
Fixes #418


carlo-rtr:
Thanks @jplock. I like the ByteRange abstraction. What are your thoughts on making ByteRange.parse having the context of the resource length? This way we can encapsulate byte range transformation from one that has no end to one that does. We also wouldn't need hasEnd() to public anymore.

jplock:
Yea I was thinking about that.  Let me give that a try to see how it looks.  It should be simpler.

jplock:
@carlo-rtr I updated the PR with your suggestion

Issue #768, opened 2014-11-06
Security issue: Data representation leaks because validation takes places before authentication


X3Cutor:
Reported this on the mailing list and was asked to create an issue. I'm attaching some basic code to illustrate the problem.

My application
```
public class TestApplication extends Application<Configuration> {

    public static void main(String[] args) throws Exception {
        new TestApplication().run(args);
    }
    @Override
    public void initialize(Bootstrap<Configuration> configurationBootstrap) {

    }

    @Override
    public void run(Configuration configuration, Environment environment) throws Exception {
        environment.jersey().register(new BasicAuthProvider(new SecuredResourcesAuthenticator(),"SecuredEndpoints"));
        environment.getObjectMapper().disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
        ValidatedResource resource = new ValidatedResource();
        environment.jersey().register(resource);
    }
}
```

My resource
```
@Produces(MediaType.APPLICATION_JSON)
@Consumes(MediaType.APPLICATION_JSON)
@Path("/greet")
public class ValidatedResource {

    @POST
    @Timed
    @ExceptionMetered
    public Object createSolution(@Valid TransportObject validatedObject,@Auth Object userPlaceholder) throws JsonProcessingException {
        String response = "Hello";
        return response;
    }
}
```

The object with "secret" internal representation I only want authenticated people to know about
```
public class TransportObject {

    @NotNull
    @Size(min = 1)
    @Pattern(regexp="some_regexp_here")
    private String given_name;

    public String getGiven_name() {
        return given_name;
    }

    public void setGiven_name(String given_name) {
        this.given_name = given_name;
    }
}
```

The class that "handles" authentication by denying everybody
```
public class SecuredResourcesAuthenticator implements Authenticator<BasicCredentials,Object> {

    @Override
    public Optional<Object> authenticate(BasicCredentials credentials) throws AuthenticationException {
        //Deny everybody access
        return Optional.absent();
    }
}
```

I'm using dropwizard 0.7. A POST to http://localhost:8080/greet after the service has been started with an empty JSON body gives the following response:
```
{
errors: [1]
0:  "given_name may not be null (was null)"
-
}
```

After  getting that info I POST
```
{
    "given_name": ""
}
```
And get basically the full internal representation of this rather simple resource
```
{
errors: [2]
0:  "given_name must match "some_regexp_here" (was )"
1:  "given_name size must be between 1 and 2147483647 (was )"
-
}
[...]

PR #891, opened 2015-02-23
Tests to verify we are not leaking any validation info before authentication

glennmcallister:
Tests to verify that issue #768 is no longer a problem in Dropwizard
8.x, presumably after the switch to Jersey 2.x.

### DIFF
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/TransportObject.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/TransportObject.java
@@ -4,16 +4,21 @@ import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 /** 
  *
  * Some object with a "secret" internal representation we want only authenticated 
  * people to know about.
  *
  */
+@JsonIgnoreProperties(ignoreUnknown = false)
 public class TransportObject {
     @NotNull
     @Size(min = 1)
     @Pattern(regexp="ab?b")
+    @JsonProperty
     private String givenName;
     
     public String getGivenName() {
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/TransportObject.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/TransportObject.java
@@ -1,0 +1,26 @@
+package io.dropwizard.auth;
+
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import javax.validation.constraints.Size;
+
+/** 
+ *
+ * Some object with a "secret" internal representation we want only authenticated 
+ * people to know about.
+ *
+ */
+public class TransportObject {
+    @NotNull
+    @Size(min = 1)
+    @Pattern(regexp="ab?b")
+    private String givenName;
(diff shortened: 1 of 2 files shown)


arteam:
Thanks, Glenn. If other collaborators don't mind, I will take a look at this.

joschi:
It's still failing: https://gist.github.com/joschi/9ff2c3d40f08bd5a8c86

The problem is the injection order in the resource method which occurs in order of the arguments. That's why the problem doesn't happen if the parameter annotated with `@Auth` is the first method parameter or if there is a class field annotated with `@Auth` (which is being injected before any method parameter).

arteam:
I too have doubts about resolving this problem. 

I've tried to reproduce a validation error in the `AuthBeforeValidateTest` and wasn't able to do it. But in 
your test case the error is clearly showed. It's bugging me. What the difference between the tests?

glennmcallister:
And that would be why I wanted other people to look at this; I'm not that
familiar with the JerseyTest framework and how to push it through it's
paces. That being said, the tests I wrote have the @Valid parameter first,
which was a requirement of the original problem.

I'll incorporate the tests Jochen has written and proceed with resolving
this. I'm at work, where I'm in theory forbidden from working on any of
this, so hopefully I can poke at it tonight.

arteam:
I think I found the differences.

* Jochen's tests are for Jackson's deserialization and validation (the `@Valid` annotation is not actually needed). 
* Glenn's tests are in turn for Jersey's Bean Validation (Jackson's deserialization and validation
works because JSON is always correct).

But I have a feeling that `GrizzlyWebTestContainer` doesn't support bean validation, so Glenn's
tests can't throw 400 error.

For future testing I also created a simple resource in *dropwizard-example* and run it in the real Jersey container.
````java
@Path("/protected")
@Consumes(MediaType.APPLICATION_JSON)
public class ProtectedResource {

    @POST
    public String showSecret(@Valid Saying saying, @Auth User user) {
        System.out.println(saying);
        return String.format("Hey there, %s. You know the secret!", user.getName());
    }
}

public class Saying {
    private long id;

    @Length(max = 3)
    private String content;

    public Saying() {
        // Jackson deserialization
    }

    public Saying(long id, String content) {
        this.id = id;
        this.content = content;
    }

    @JsonProperty
    public long getId() {
        return id;
    }

    @JsonProperty
    public String getContent() {
        return content;
    }

    @Override
    public String toString() {
        return MoreObjects.toStringHelper(this)
                .add("id", id)
                .add("content", content)
                .toString();
    }
}
````

Tested it with curl:
````
curl -v -H 'content-type:application/json' -d '{"id":2 , "content":"trash"}' 
http://127.0.0.1:8080/protected
```` 

The response was:
````
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> POST /protected HTTP/1.1
> User-Agent: curl/7.39.0
> Host: 127.0.0.1:8080
> Accept: */*
> content-type:application/json
> Content-Length: 28
> 
* upload completely sent off: 28 out of 28 bytes
< HTTP/1.1 422 
< Date: Mon, 23 Feb 2015 18:27:51 GMT
< Content-Type: application/json
< Content-Length: 65
< 
* Connection #0 to host 127.0.0.1 left intact
{"errors":["content length must be between 0 and 3 (was trash)"]}
````

So, unfortunately, this issue doesn't seem to be fixed for me.

glennmcallister:
BTW, if I do manage to resolve this I'll squash my commits down so we have a clean pull, either here (if possible) or in a new PR.

Issue #893, opened 2015-02-23
0.8.0-rc3 IllegalStateException on 2nd PUT returning 400


brackxm:
After upgrading from 0.7.1 to 0.8.0-rc3 i get an IllegalStateException.
The first put works (returning 400).
The second put throws an exception; the client (curl) hangs.

curl:
```
curl -XPUT http://localhost:8000/1/q/w -d hello
```

log:
```
127.0.0.1 - - [23/Feb/2015:14:51:44 +0000] "PUT /1/q/w HTTP/1.1" 400 23 "-" "curl/7.37.1" 72
ERROR [2015-02-23 14:51:45,829] org.glassfish.jersey.server.ServerRuntime$Responder: An I/O error has occurred while writing a response message entity to the container output stream.
! java.lang.IllegalStateException: The output stream has already been closed.
! at org.glassfish.jersey.message.internal.CommittingOutputStream.setStreamProvider(CommittingOutputStream.java:147) ~[app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.message.internal.OutboundMessageContext.setStreamProvider(OutboundMessageContext.java:803) ~[app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.server.ContainerResponse.setStreamProvider(ContainerResponse.java:372) ~[app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.server.ServerRuntime$Responder.writeResponse(ServerRuntime.java:647) [app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.server.ServerRuntime$Responder.processResponse(ServerRuntime.java:418) [app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.server.ServerRuntime$Responder.process(ServerRuntime.java:408) [app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.server.ServerRuntime$2.run(ServerRuntime.java:306) [app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.internal.Errors$1.call(Errors.java:271) [app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.internal.Errors$1.call(Errors.java:267) [app.jar:1.13.3-SNAPSHOT]
! at org.glassfish.jersey.internal.Errors.process(Errors.java:315) [app.jar:1.13.3-SNAPSHOT]
[...]

PR #901, opened 2015-02-26
example put, trying to replicate issue 893

carlo-rtr:


joschi:
The original author of issue #893 reused the `Response` object which doesn't work with JAX-RS (as the underlying `OutputStream` can only be consumed once).

PR #908, opened 2015-02-27
Lookup environment variables in configuration files

joschi:
This pull request adds the capability to lookup environment variables from within configuration files.

Some PaaS offerings make it hard to modify the Java runtime environment for applications (e. g. set custom system properties) but make it rather easy to use environment variables (also see [The Twelve Factor App](http://12factor.net/)). The `SubstitutingSourceProvider` and `EnvironmentVariableSubstitutor` classes from this PR simplify using environment variables as source for individual configuration settings.

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -216,7 +216,46 @@ Dropwizard then calls your ``Application`` subclass to initialize your applicati
     file. If it does not exist or is not an array setting, it will get added as a simple string setting, including 
     the ',' characters as part of the string.
 
-.. _man-core-environments:
+.. _man-core-environment-variables:
+
+Environment variables
+---------------------
+
+The ``dropwizard-configuration`` module also provides the capabilities to substitute configuration settings with the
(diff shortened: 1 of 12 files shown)


docs/source/manual/core.rst
```diff
@@ -216,7 +216,46 @@ Dropwizard then calls your ``Application`` subclass to initialize your applicati
     file. If it does not exist or is not an array setting, it will get added as a simple string setting, including 
     the ',' characters as part of the string.
 
-.. _man-core-environments:
+.. _man-core-environment-variables:
+
+Environment variables
+---------------------
+
+The ``dropwizard-configuration`` module a`lso provides the capabilities to substitute configuration settings with the
```
carlo-rtr: erroneous back tick in also?

carlo-rtr:
Thank you, this is really useful. This PR looks good. Other than the comment on the docs, the only other thing is that we could have a test for proper substitution of recursive variable names. e.g. `${jre-${java.version}}` but that's more of a nitpick.

joschi:
@carlo-rtr I've addressed the issues (typo in docs, test for recursive substitution) you've mentioned.

Issue #850, opened 2015-01-15
UnitOfWorkApplicationListener.java commits transaction after exception is thrown in a request.


lhcjackson-toast:
I am on 0.8.0-rc2

After an exception is thrown in the request, RESOURCE_METHOD_FINISHED event is dispatched first, then ON_EXCEPTION event is dispatched, so we have two problems with the following code:
1. Transaction is committed when exception is thrown. It should be rolled back.
2. If the commit fails, the original exception is lost, replaced by a database exception.

My recommended fix is to check for existing exceptions in RESOURCE_METHOD_FINISHED. 
```java
@Override
public void onEvent(RequestEvent event) {
    if (event.getType() == RequestEvent.Type.RESOURCE_METHOD_START) {
        this.unitOfWork = this.methodMap.get(event.getUriInfo()
                .getMatchedResourceMethod().getInvocable().getDefinitionMethod());
        if (unitOfWork != null) {
            this.session = this.sessionFactory.openSession();
            try {
                configureSession();
                ManagedSessionContext.bind(this.session);
                beginTransaction();
            } catch (Throwable th) {
                this.session.close();
                this.session = null;
                ManagedSessionContext.unbind(this.sessionFactory);
                throw th;
            }
        }
    }
    else if (event.getType() == RequestEvent.Type.RESOURCE_METHOD_FINISHED) {
        if (this.session != null) {
            try {
                commitTransaction();
            } catch (Exception e) {
                rollbackTransaction();
                this.<RuntimeException>rethrow(e);
            }
            finally {
                this.session.close();
                this.session = null;
                ManagedSessionContext.unbind(this.sessionFactory);
            }
        }
    }
    else if (event.getType() == RequestEvent.Type.ON_EXCEPTION) {
        if (this.session != null) {
            try {
                rollbackTransaction();
            }
            finally {
                this.session.close();
                this.session = null;
                ManagedSessionContext.unbind(this.sessionFactory);
            }
        }
    }
}
```

phxql:
The order in which jersey emits the events is:

In case of exception:
* MATCHING_START
* REQUEST_MATCHED
* REQUEST_FILTERED
* RESOURCE_METHOD_START
* RESOURCE_METHOD_FINISHED
* ON_EXCEPTION
* EXCEPTION_MAPPER_FOUND
* EXCEPTION_MAPPING_FINISHED
* RESP_FILTERS_START
* RESP_FILTERS_FINISHED
* FINISHED

[...]

PR #915, opened 2015-03-04
Bugfix for Issue #850.

phxql:
This fixes bug #850. Instead of committing the transaction on the `RESOURCE_METHOD_FINISHED` event, which is emitted from Jersey before the `ON_EXCEPTION` event, the transaction is commited on `RESP_FILTERS_START` event, which is emitted after the `ON_EXCEPTION` event.


jplock:
Thanks for the fix.  I'm not familiar enough with hibernate to know if this is a correct fix or not.  Is there someone else more familiar who can comment?

arteam:
I will take a look today on this.

arteam:
Looks great and works correctly. I've squashed the commits and pushed to master.

@phxql, thanks for contribution!

PR #921, opened 2015-03-05
Dbi shortcircuit

carlo-rtr:
https://groups.google.com/d/msg/dropwizard-user/SLi4deSPXus/Q0Bz1De_vfYJ


dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIHealthCheck.java
```diff
@@ -1,23 +1,40 @@
...
         this.dbi = dbi;
         this.validationQuery = validationQuery;
+        this.duration = duration;
+        this.executorService = executorService;
     }
 
     @Override
     protected Result check() throws Exception {
+        Result result;
         try (Handle handle = dbi.open()) {
```
prb: Won't this potentially block until there is a connection available?

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
```diff
@@ -53,7 +57,14 @@ public DBI build(Environment environment,
         final String validationQuery = configuration.getValidationQuery();
```
prb: This is the query that's used to validate connections in the pool (depending on settings, while idle, on borrow, on return), and the timeout wrapping is already implemented at the statement level in the Tomcat pool that DW uses.

Presuming that we trust the pool's validation activities, then just checking the age of `org.apache.tomcat.jdbc.pool.PooledConnection#getLastValidated` probably provides an adequate healthcheck (and that healthcheck already uses the timeout and lifecycle settings for the pool).

prb:
:ship: :it:...

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
arteam: The problem that Tomcat-DBCP, which we use, doesn't support handling network
failures. Checkout this [article](https://github.com/brettwooldridge/HikariCP/wiki/Bad-Behavior:-Handling-Database-Down).

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
```diff
@@ -53,7 +57,14 @@ public DBI build(Environment environment,
...
         final DBI dbi = new DBI(dataSource);
         environment.lifecycle().manage(dataSource);
-        environment.healthChecks().register(name, new DBIHealthCheck(dbi, validationQuery));
+        ExecutorService executorService = environment.lifecycle().executorService("DBIHealthCheck")
```
arteam: Shouldn't we consider `ThreadPoolExecutor.CallerRunsPolicy` as a *rejectedExecutionHandler*?
If two health check arrived concurrently, the second one will fail even if the database is up.

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIHealthCheck.java
arteam: I support this. 
Connection opening should be perfomed in the background thread. It protect us from network failures.

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIHealthCheck.java
```diff
@@ -1,23 +1,40 @@
...
 import org.skife.jdbi.v2.DBI;
 import org.skife.jdbi.v2.Handle;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
 
 public class DBIHealthCheck extends HealthCheck {
     private final DBI dbi;
     private final String validationQuery;
-
-    public DBIHealthCheck(DBI dbi, String validationQuery) {
```
arteam: This constructor is public, so users code could depend on it. For example, someone could create a custom *DBI* instance from a data source and add a health check.

Maybe we can retain it? We could use `MoreExecutors.newDirectExecutorService()` from Guava
and zero duration.

carlo-rtr: That's a good point. If we use those as defaults, it effectively means that health check won't be time bound. The callable will execute on the thread of the caller, and when it's finished the caller will retrieve the result with a timeout of zero; but it will always be ready. This seems like a good way to support backwards compatibility. Though the DBIFactory should pass in a executor.

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIHealthCheck.java
carlo-rtr: I agree, I'll do that.

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
carlo-rtr: Some interesting stuff being discussed on this PR!

carlo-rtr:
I'll make the changes discussed, plus get the tests passing.

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
carlo-rtr: This is a good point. If we use CallerRunsPolicy, we will loose the ability to terminate the thread, so we end up with the original situation. Instead of being a single a threaded executor, what about creating a pool shared across health checks where the number of threads can be configured. We can still make it have a CallerRunsPolicy, but it will be less likely to hit it.

carlo-rtr:
I've made the changes discussed, and squashed the commits. The biggest concern I have is the centralized executorService. It's not configurable, and should Environment own it?

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
prb: Interesting note.  The JDBC driver could (and should) be configured (e.g., [socket settings](https://jdbc.postgresql.org/documentation/head/connect.html) for the Postgres driver) to alleviate the failure scenario for the Tomcat pool, but the HikariCP pool sounds like a nice piece of work.

dropwizard-db/src/main/java/io/dropwizard/db/TimeBoundHealthChecks.java
```diff
@@ -0,0 +1,19 @@
+package io.dropwizard.db;
+
+import com.codahale.metrics.health.HealthCheck;
+import io.dropwizard.util.Duration;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+
+public class TimeBoundHealthChecks {
+    public static HealthCheck.Result check(ExecutorService executorService, Duration timeBound, Callable<HealthCheck.Result> c) {
```
arteam: I think it's better to make `ExecutorService` and `Duration` as class dependencies rather than method arguments.
Doing so, we can encapsulate them in `TimeBoundHealthChecks` and then inject it to `DBIHealthCheck` and `SessionFactoryHealthCheck`. As the result we have a parametrized service that accepts Callables, rather than
a utility method. 

Something like this:
````java
public class TimeBoundHealthChecks {

    private final ExecutorService executorService;
    private final Duration timeBound;

    public TimeBoundHealthChecks(ExecutorService executorService, Duration timeBound) {
        this.executorService = executorService;
        this.timeBound = timeBound;
    }

    public HealthCheck.Result check(Callable<HealthCheck.Result> callable) {
        try {
            return executorService.submit(callable).get(timeBound.getQuantity(), timeBound.getUnit());
        } catch (Exception e) {
            return HealthCheck.Result.unhealthy(Throwables.getRootCause(e));
        } 
    }
}
````

````java
public class DBIHealthCheck extends HealthCheck {
    private final DBI dbi;
    private final String validationQuery;
    private final TimeBoundHealthChecks timeBoundHealthChecks;

    public DBIHealthCheck(ExecutorService executorService, Duration duration, DBI dbi, String validationQuery) {
        this.dbi = dbi;
        this.validationQuery = validationQuery;
        timeBoundHealthChecks = new TimeBoundHealthChecks(executorService, duration);
    }

    public DBIHealthCheck(DBI dbi, String validationQuery) {
        this(MoreExecutors.newDirectExecutorService(), Duration.seconds(0), dbi, validationQuery);
    }

    @Override
    protected Result check() throws Exception {
        return timeBoundHealthChecks.check(new Callable<Result>() {
            @Override
            public Result call() throws Exception {
                try (Handle handle = dbi.open()) {
                    handle.execute(validationQuery);
                    return Result.healthy();
                }
            }
        });
    }

}
````

File: dropwizard-core/src/main/java/io/dropwizard/setup/Environment.java [hunk 6]
```diff
@@ -71,6 +78,15 @@ public Environment(String name,
 
         this.jerseyServletContainer = new JerseyContainerHolder(new ServletContainer(jerseyConfig));
         this.jerseyEnvironment = new JerseyEnvironment(jerseyServletContainer, jerseyConfig);
+
+
+        this.healthCheckExecutorService = this.lifecycle().executorService("TimeBoundHealthCheck-pool-%d")
+                .workQueue(new ArrayBlockingQueue<Runnable>(4))
```
arteam: I think we don't need a queue with a size more than 1. There are no reason in preserving the pool size.
In case if DB is down and 2 concurrent request happen, the situation is actually worse, because the second request
wiil be forced to wait whole 5 seconds before it will be processed.

dropwizard-jdbi/src/test/java/io/dropwizard/jdbi/DBIHealthCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+            @Override
+            public Object answer(InvocationOnMock invocation) throws Throwable {
+                try {
+                    TimeUnit.SECONDS.sleep(10);
+                }
+                catch (Exception ignored) {}
+                return null;
+            }
+        }).when(handle).execute(validationQuery);
+        DBIHealthCheck dbiHealthCheck = new DBIHealthCheck(Executors.newSingleThreadExecutor(),
```
arteam: We created this executor and therefore should manually close it after the test.

dropwizard-jdbi/src/test/java/io/dropwizard/jdbi/DBIHealthCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+    public void testItTimesOutProperly() throws Exception {
+        String validationQuery = "select 1";
+        DBI dbi = mock(DBI.class);
+        Handle handle = mock(Handle.class);
+        when(dbi.open()).thenReturn(handle);
+        Mockito.doAnswer(new Answer() {
+            @Override
+            public Object answer(InvocationOnMock invocation) throws Throwable {
+                try {
+                    TimeUnit.SECONDS.sleep(10);
```
arteam: How about to play it safe and set a little bigger interval?

dropwizard-core/src/main/java/io/dropwizard/setup/Environment.java
```diff
@@ -81,6 +97,13 @@ public JerseyEnvironment jersey() {
     }
 
     /**
+     * Returns an {@Link ExecutorService} to run time bound health checks
```
arteam: Nitpick. Wrong tag - should be `@link`

arteam:
Some minor improvements and, I think, we are done.

dropwizard-db/src/main/java/io/dropwizard/db/TimeBoundHealthChecks.java
carlo-rtr: I like this idea

dropwizard-core/src/main/java/io/dropwizard/setup/Environment.java
carlo-rtr: makes sense

dropwizard-jdbi/src/test/java/io/dropwizard/jdbi/DBIHealthCheckTest.java
carlo-rtr: yes. oversight on my part.

dropwizard-jdbi/src/test/java/io/dropwizard/jdbi/DBIHealthCheckTest.java
carlo-rtr: I dont think that's necessary, different units. The answer sleeps in seconds, the short circuit happens in millis.

carlo-rtr:
thanks for the feedback

arteam:
Great, looks amazing.
Thanks for your contribution.

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
brettwooldridge: @prb [Come on over](https://github.com/brettwooldridge/HikariCP), the water's nice.  We're fully Dropwizard enabled. :+1:

PR #924, opened 2015-03-06
Exclude slf4j-api from metrics-httpclient

thesmith:
Not sure why this wasn't picked up by the enforcer.
metrics-httpclient relies on slf4j-api 1.7.7, not 1.7.10

### DIFF
--- a/dropwizard-client/pom.xml
+++ b/dropwizard-client/pom.xml
@@ -31,6 +31,10 @@
                     <groupId>commons-logging</groupId>
                     <artifactId>commons-logging</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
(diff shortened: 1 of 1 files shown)


joschi:
The slf4j-api dependency is defined in the `dependencyManagement` section of the parent POM (https://github.com/dropwizard/dropwizard/blob/e85a6b71806a6da35cfc723d9c879d912a0e289f/pom.xml#L292-296) which is why the Maven Enforcer plugin doesn't throw an error.

thesmith:
I see. When I include dropwizard-client and I don't have slf4j-api stuck to 1.7.10 in my dependencyManagement, dropwizard-client tries to bring in both versions and triggers my own enforcer run.

It would be nice for users not to have to set slf4j-api in their dependencyManagement, although it's not a big issue.

joschi:
No, you're completely right with excluding it. I just wanted to provide an explanation why Maven Enforcer didn't seem to work in this case. Thanks!

thesmith:
Gotcha. Cheers!

PR #927, opened 2015-03-07
Fix logback error, which ultimately quiets the noise on startup

carlo-rtr:
The example doesn't have a file appender which does archiving, so it wasn't triggering the issue. I've added that.

http://logback.qos.ch/codes.html
## The File property must be set before any rolling policy or triggering policy.
The File property, if present, must be placed before any rolling policy or triggering policy. Thus, in a configuration file, the File property, if present, must be declared declared before any rolling policy or triggering policy declarations.

### DIFF
--- a/dropwizard-example/example.yml
+++ b/dropwizard-example/example.yml
@@ -66,6 +66,13 @@ logging:
 
   appenders:
     - type: console
+    - type: file
+      threshold: DEBUG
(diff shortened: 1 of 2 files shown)


arteam:
That looks extremely good to me!
Thanks, Carlo, for catching and fixing that.

PR #929, opened 2015-03-08
Replace ImmutableMap with Map interface in ViewRenderer interface

joschi:
The original change was using the concrete `ImmutableMap<K, V>` implementation from Google Guava instead of the generic `Map<K, V>` interface which allows the use of other map implementations.


jplock:
LGTM

PR #938, opened 2015-03-12
Change copyright notice to be up-to-date

thesmith:
It would be way better to dynamically work out what year it is, I just don't know how to test build the docs site.

If someone knows, that would be better.

Also, should we remove Yammer from the Copyright? Maybe @ryankennedy would know.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -42,7 +42,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'Dropwizard'
-copyright = u'2010-2013, Coda Hale, Yammer Inc.'
+copyright = u'2010-2015, Coda Hale, Yammer Inc.'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the


joschi:
Neither Coda Hale nor Yammer is currently actively involved in the development of Dropwizard. So the copyright time range is basically correct, but I'm not familiar with the US rules/laws/recommendation about those…

arteam:
As I recall from a mailing list discussion, Coda specifically wanted to give credit to Yammer and retain them in the notice. I believe, we should just ask him to avoid misunderstanding.

unindented:
It makes the page look stale though.

ryankennedy:
Ben, I'd consult with the Microsoft lawyers first about removing Yammer
from the copyright. Just to make sure there's no issue in doing so. Then
again, I feel like telling you to go talk to corporate attorneys makes me
the bad person. ;)

I believe part of the discussion there was also around having Microsoft's
attorneys around to help enforce the copyright if we ever needed to.

Ryan

carlo-rtr:
Is there an easy way to ask the lawyers without going thru some crazy process? I'm personally not up for that challenge. If not, I'm inclined to say we should close this and leave the copyright as it is.

ryankennedy:
Back in the days before Yammer's legal council was let go I'd say we could
work through her. These days I don't remember who's left there that would
be good to work with.

Coda's fear still stands. Without copyright from an entity with teeth to
enforce it, I'm not sure we'd want to remove it. So unless we have a good
replacement, it might be best to leave it. Assuming, of course, that
Yammer/Microsoft would even do anything if needed...the track record isn't
good (no offense to my ex-colleagues on the list, it's an organizational
complaint not a personal one).

Ryan

joschi:
So let's just leave it as it is and add "2014-2015 Dropwizard Team"?

carlo-rtr:
I like that idea

thesmith:
+1

arteam:
That's fine for me also.

PR #939, opened 2015-03-12
Ensuring that JerseyRequest scoped ClientConfig gets propagated to HttpUriRequest

velocipedist:
**Motivation**

JerseyClient and the ApacheHttpClient are statically configured at creation time. Dropwizard configuration builders ensure those configs are in sync. However, `WebTarget` objects can have the configuration overridden per generated request. This information will be added to the `JerseyRequest`. Unfortunately, the current implementation of the `DropwizardApacheConnector` does not propagate this onto the `RequestConfig` in the created `HttpUriRequest`.

In DW 0.7.1 this was working thanks to this code: http://grepcode.com/file/repo1.maven.org/maven2/com.sun.jersey.contribs/jersey-apache-client4/1.18.1/com/sun/jersey/client/apache4/ApacheHttpClient4Handler.java#241

This is currently blocking the migration of https://github.com/yammer/tenacity project to DW 0.8.0, as we need this functionality to sync Hystrix timeouts with those of the HttpClient.

**Solution**

The solution is to update the `DropwizardApacheConnector` to check the `JerseyRequest#getConfiguration` for properties (present only if overridden) that should be propagated to the `HttpUriRequest` specific `RequestConfig` instance.

I didn't want to loose the default values for not overridden properties, so I wrapped the `CloseableHttpClient` in a delegating class that exposes the default `RequestConfig` to use as basis.

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/ConfiguredCloseableHttpClient.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/ConfiguredCloseableHttpClient.java
@@ -1,0 +1,22 @@
+package io.dropwizard.client;
+
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.impl.client.CloseableHttpClient;
+
+/* package */ class ConfiguredCloseableHttpClient {
+    private final CloseableHttpClient closeableHttpClient;
+    private final RequestConfig defaultRequestConfig;
+
+    /* package */ ConfiguredCloseableHttpClient(CloseableHttpClient closeableHttpClient, RequestConfig defaultRequestConfig) {
+        this.closeableHttpClient = closeableHttpClient;
+        this.defaultRequestConfig = defaultRequestConfig;
+    }
+
+    public RequestConfig getDefaultRequestConfig() {
+        return defaultRequestConfig;
+    }
+
+    public CloseableHttpClient getClient() {
+        return closeableHttpClient;
+    }
+}
--- a/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
@@ -1,11 +1,13 @@
 package io.dropwizard.client;
 
+import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.MoreExecutors;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
 import org.apache.http.StatusLine;
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.client.methods.RequestBuilder;
@@ -13,6 +15,7 @@ import org.apache.http.entity.AbstractHttpEntity;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.util.VersionInfo;
 import org.glassfish.jersey.apache.connector.LocalizationMessages;
+import org.glassfish.jersey.client.ClientProperties;
 import org.glassfish.jersey.client.ClientRequest;
 import org.glassfish.jersey.client.ClientResponse;
 import org.glassfish.jersey.client.spi.AsyncConnectorCallback;
@@ -21,6 +24,7 @@ import org.glassfish.jersey.message.internal.OutboundMessageContext;
 import org.glassfish.jersey.message.internal.Statuses;
 
 import javax.ws.rs.ProcessingException;
+import javax.ws.rs.core.Configuration;
 import javax.ws.rs.core.Response;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
@@ -59,14 +63,19 @@ public class DropwizardApacheConnector implements Connector {
      * Actual HTTP client
      */
     private final CloseableHttpClient client;
+    /**
+     * Default HttpUriRequestConfig
+     */
+    private final RequestConfig defaultRequestConfig;
 
     /**
      * Should a chunked encoding be used in POST requests
      */
     private final boolean chunkedEncodingEnabled;
 
-    public DropwizardApacheConnector(CloseableHttpClient client, boolean chunkedEncodingEnabled) {
+    public DropwizardApacheConnector(CloseableHttpClient client, RequestConfig defaultRequestConfig, boolean chunkedEncodingEnabled) {
         this.client = client;
+        this.defaultRequestConfig = defaultRequestConfig;
         this.chunkedEncodingEnabled = chunkedEncodingEnabled;
     }
 
@@ -125,7 +134,39 @@ public class DropwizardApacheConnector implements Connector {
             }
             builder.addHeader(headerName, jerseyRequest.getHeaderString(headerName));
         }
+
+        Optional<RequestConfig> requestConfig = addJerseyRequestConfig(jerseyRequest.getConfiguration());
+        if (requestConfig.isPresent()) {
+            builder.setConfig(requestConfig.get());
+        }
+
         return builder.build();
+    }
+
+    private Optional<RequestConfig> addJerseyRequestConfig(Configuration configuration) {
+        final Integer timeout = (Integer) configuration.getProperty(ClientProperties.READ_TIMEOUT);
+        final Integer connectTimeout = (Integer) configuration.getProperty(ClientProperties.CONNECT_TIMEOUT);
+        final Boolean followRedirects = (Boolean) configuration.getProperty(ClientProperties.FOLLOW_REDIRECTS);
+
+        if (timeout != null || connectTimeout != null || followRedirects != null) {
+            RequestConfig.Builder requestConfig = RequestConfig.copy(defaultRequestConfig);
+
+            if (timeout != null) {
+                requestConfig.setSocketTimeout(timeout);
+            }
+
+            if (connectTimeout != null) {
+                requestConfig.setConnectTimeout(connectTimeout);
+            }
+
+            if (followRedirects != null) {
+                requestConfig.setRedirectsEnabled(followRedirects);
+            }
+
+            return Optional.of(requestConfig.build());
+        }
+
+        return Optional.absent();
     }
 
     /**
--- a/dropwizard-client/src/test/java/io/dropwizard/client/ConfiguredCloseableHttpClientTest.java
+++ b/dropwizard-client/src/test/java/io/dropwizard/client/ConfiguredCloseableHttpClientTest.java
@@ -1,0 +1,35 @@
+package io.dropwizard.client;
+
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ConfiguredCloseableHttpClientTest {
+    public ConfiguredCloseableHttpClient configuredClient;
+    @Mock
+    private CloseableHttpClient closeableHttpClientMock;
+    @Mock
+    private RequestConfig defaultRequestConfigMock;
+
+    @Before
+    public void setUp() {
+        configuredClient = new ConfiguredCloseableHttpClient(closeableHttpClientMock, defaultRequestConfigMock);
+    }
+
+    @Test
+    public void getDefaultRequestConfig_returns_config_provided_at_construction() {
+        assertThat(configuredClient.getDefaultRequestConfig()).isEqualTo(defaultRequestConfigMock);
+    }
+
+    @Test
+    public void getClient_returns_config_provided_at_construction() {
+        assertThat(configuredClient.getClient()).isEqualTo(closeableHttpClientMock);
+    }
+}
--- a/dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
+++ b/dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
@@ -1,0 +1,178 @@
+package io.dropwizard.client;
+
+import com.codahale.metrics.MetricRegistry;
+import com.google.common.io.Resources;
+import io.dropwizard.Application;
+import io.dropwizard.Configuration;
+import io.dropwizard.jackson.Jackson;
+import io.dropwizard.setup.Environment;
+import io.dropwizard.testing.junit.DropwizardAppRule;
+import io.dropwizard.util.Duration;
+import org.apache.http.HttpStatus;
+import org.apache.http.conn.ConnectTimeoutException;
+import org.assertj.core.api.AbstractLongAssert;
+import org.glassfish.jersey.client.ClientProperties;
+import org.junit.Before;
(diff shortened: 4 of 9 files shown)


arteam:
Thanks for this PR.
I will take a closer look at this in the weekend.

jplock:
Can we replace the hamcrest matchers with AssertJ assertions in `ConfiguredCloseableHttpClientTest` to be inline with the rest of the code?

dropwizard-client/src/main/java/io/dropwizard/client/ConfiguredCloseableHttpClient.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.conn.ClientConnectionManager;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.params.HttpParams;
+import org.apache.http.protocol.HttpContext;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+public class ConfiguredCloseableHttpClient implements HttpClient, Closeable {
```
arteam: Do we really need to create a delegator? Why not make it just a DTO?

`RequestConfig` doesn't change behavior of the client and is needed only as a value object.
We could just create return a DTO in  the `createClient` method and provide 2 build methods.
* First - public for Apache HttpClient users, which returns `CloseableHttpClient`
* Second - package-protected for the Jersey bulder, which returns `ConfiguredCloseableHttpClient`
````java
public CloseableHttpClient build(String name) {
   return buildClientWithRequestConfiguration(name).getClient();
}

ConfiguredCloseableHttpClient buildClientWithRequestConfiguration(String name) {
    final InstrumentedHttpClientConnectionManager manager = createConnectionManager(registry, name);
    return createClient(org.apache.http.impl.client.HttpClientBuilder.create(), manager, name);
}
````

File: dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java [hunk 2]
```diff
@@ -125,9 +128,43 @@ private HttpUriRequest buildApacheRequest(ClientRequest jerseyRequest) {
...
+        Optional<RequestConfig> requestConfig = addJerseyRequestConfig(jerseyRequest.getConfiguration());
+        if (requestConfig.isPresent()) {
+            builder.setConfig(requestConfig.get());
+        }
+
         return builder.build();
     }
 
+    private Optional<RequestConfig> addJerseyRequestConfig(Configuration configuration) {
+        RequestConfig.Builder requestConfig = RequestConfig.copy(client.getDefaultRequestConfig());
```
arteam: I think it's better to make a copy lazily, only in a case if there changes in the configuration.

dropwizard-client/src/test/java/io/dropwizard/client/ConfiguredCloseableHttpClientTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.http.protocol.HttpContext;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.io.IOException;
+
+import static org.hamcrest.core.Is.is;
```
arteam: Please, use [AssertJ](http://joel-costigliola.github.io/assertj/) for consistency with the rest of the codebase.

arteam:
@velocipedist. I've reviewed this PR and gave directions, where it, in my opinion, could be improved.
But motivation of this change seems convincing for me.

Also things to notice for you:
* Please, rebase you PR against the current master ans squash the commits. Unfortunately, it's not merging now.
* If possible, add an integration test for `WebTarget` configuration, so we can test that a new configuration is really applying to a request.

dropwizard-client/src/main/java/io/dropwizard/client/ConfiguredCloseableHttpClient.java
velocipedist: It's a shame that `CloseableHttpClient` is an abstract class, not an interface.

I'm happy to apply the change you've suggested as there is probably no need to make external clients exposed to this new class. However, I would much prefer to have a uniform interface.

dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
velocipedist: fair point, will do.

dropwizard-client/src/test/java/io/dropwizard/client/ConfiguredCloseableHttpClientTest.java
velocipedist: will do

velocipedist:
I can add an integration test - we have one in our tenacity project, so it should be quick and easy enough

velocipedist:
@arteam just wrote some behavioral integration tests for this. However, this forced me to put a test-scoped dependency on `dropwizard-testing` in `dropwizard-client`. I've noticed you've got other integration tests which avoid doing that, so am not sure whether this was done on purpose and my change will not cause objections.

velocipedist:
@arteam @joschi @jplock 

I've applied the review feedback. We now have a single commit that is ready to merge.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hamcrest.CoreMatchers.any;
+
+public class DropwizardApacheConnectorTest {
+    private static final int SLEEP_TIME_IN_MILLIS = 500;
+    private static final int DEFAULT_CONNECT_TIMEOUT = 200;
+    private static final String NON_ROUTABLE_IP_ADDRESS = "10.255.255.1";
```
arteam: How about to set it as "192.0.2.0"? Network *192.0.2.0/24* is defined in [RFC 5737|http://tools.ietf.org/html/rfc5737] as a non-routable.
Currently, on my local machine, the test *connect_timeout_override_changes_how_long_it_takes_for_a_connection_to_timeout* fails without timeout.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @ClassRule
+    public static DropwizardAppRule<Configuration> APP_RULE =
+            new DropwizardAppRule<>(TestApplication.class, Resources.getResource("yaml/dropwizardApacheConnectorTest.yml").getPath());
+    public final URI TEST_URI = java.net.URI.create("http://localhost:" + APP_RULE.getLocalPort());
```
arteam: If these fields are not constants, then, please, write them in the camel case. 
Also `java.net.URI` could be shortened just to `URI`.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
```diff
@@ -0,0 +1,180 @@
...
+                .build("test'");
+    }
+
+
+    @Test
+    public void when_no_read_timeout_override_then_client_request_times_out() {
+        thrown.expect(ProcessingException.class);
+        thrown.expectCause(any(SocketTimeoutException.class));
+
+        Client client = createStandardClient();
```
arteam: How about making `Client` a field of the test and move this call to the `setup` method?

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
```diff
@@ -0,0 +1,180 @@
...
+        final long startTime = System.currentTimeMillis();
+        ProcessingException processingException = null;
+        try {
+            webTarget
+                    .request()
+                    .get(Response.class);
+        } catch (ProcessingException e) {
+            processingException = e;
+        }
+        final long endTime = System.currentTimeMillis();
```
arteam: I think we could move this logic to the exception block and throw an `AssertionError` if an excption has not happened. Something like this:
````
final long startTime = System.currentTimeMillis();
try {
    webTarget
            .request()
            .get(Response.class);
    throw new AssertionError("Should fail with ProcessingException");
} catch (ProcessingException e) {
    final long endTime = System.currentTimeMillis();
    assertThat(e).isNotNull();
    //noinspection ConstantConditions
    assertThat(e.getCause()).isNotNull();
    assertThat(e.getCause()).isInstanceOf(ConnectTimeoutException.class);
    return assertThat(endTime - startTime);
}
````

File: dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java [hunk 8]
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        @Override
+        public void run(Configuration configuration, Environment environment) throws Exception {
+            environment.jersey().register(TestResource.class);
+        }
+    }
+
+    private static AbstractLongAssert<?> assertThatGetConnectonTimeoutForTarget(WebTarget webTarget) {
+        final long startTime = System.currentTimeMillis();
```
arteam: Please, use `System.nanoTime` for interval measurement. It's a way more reliable that `System.currentTimeInMillis` and more precise.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Test
+    public void connect_timeout_override_changes_how_long_it_takes_for_a_connection_to_timeout() {
+        Client client = createStandardClient();
+
+        // before override
+        assertThatGetConnectonTimeoutForTarget(
+                client.target(URI_THAT_CANNOT_BE_CONNECTED_TO)
+        ).isLessThan(DEFAULT_CONNECT_TIMEOUT + 300);
```
arteam: Sorry, I don't understand where number 300 is taken. Is there some dependency?
Also, shouldn't this assertion be `isGreaterThan(DEFAULT_CONNECT_TIMEOUT);`?

dropwizard-client/src/test/resources/yaml/dropwizardApacheConnectorTest.yml
```diff
@@ -0,0 +1,5 @@
+# this is needed to start the application in the DropwizardApacheConnectorTest
+server:
+  applicationConnectors:
+    - type: http
+      port: 34480
```
arteam: We could use zero and OS will assign a random TCP port for the server.
Doing so, we make the test more independent from environment.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
```diff
@@ -0,0 +1,180 @@
...
+        clientConfiguration = new JerseyClientConfiguration();
+        clientConfiguration.setConnectionTimeout(Duration.milliseconds(SLEEP_TIME_IN_MILLIS / 2));
+        clientConfiguration.setTimeout(Duration.milliseconds(DEFAULT_CONNECT_TIMEOUT));
+    }
+
+    private Client createStandardClient() {
+        return new JerseyClientBuilder(new MetricRegistry())
+                .using(Executors.newSingleThreadExecutor(), Jackson.newObjectMapper())
+                .using(clientConfiguration)
+                .build("test'");
```
arteam: It could be just *test* without a quote.

dropwizard-client/src/main/java/io/dropwizard/client/ConfiguredCloseableHttpClient.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public RequestConfig getDefaultRequestConfig() {
+        return defaultRequestConfig;
+    }
+
+    public CloseableHttpClient getClient() {
+        return closeableHttpClient;
+    }
+}
```
arteam: Trivial nitpick.

Please, add a newline in the end of the added files. It'a standard for text files on *nix systems.
[Reference] (http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline)

arteam:
@velocipedist Thanks! `DropwizardApacheConnectorTest` is a great piece of work.
Couple tweaks and this PR is ready to go.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
velocipedist: So this is number is an error margin needed to make the test stable - I'll put it under a named constant.

It should not be greater, because you would otherwise not verify that there is real change.

No, it has to be less then. Basically, first we show that the following is true:
```
     total_request_duration < connect_timeout + measurement_error   (1)
```

then we show that after changing to `connect_timeout' = connect_timeout+measurement_error+a_bit` we have:

```
    total_request_duration > connect_timeout'
                                        = connect_timeout+measurement_error+a_bit
                                        > connect_timeout+measurement_error
```

Now, because we've showed the `(1)` this would a contradiction, unless have changed the connection timeout to more than `connection_timeout + a_bit`.

Hence our test shows that the setting works.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
velocipedist: So this is interesting. We found this IP on [stackoverflow](http://stackoverflow.com/a/904609) and it makes the test work and CI to pass. On the other hand, when I use the one you suggest I get 'SocketTimeoutException` rather than `ConnectionTimeoutException`. Even when I change the expected exception the test does not pass because the timeout settings are not observed.

So not sure what you want to do here. Shall we just drop the test if we cannot make it work?

velocipedist:
@arteam acted on your feedback, but for the change of the IP used in a test. Please see my detailed comment - curious to know your opinion.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
arteam: I've checked on another machine and 10.255.255.1 works fine. It seems that my local ISP filters packets to this address. Ok, let's leave it. Travic CI doesn't complain, so we can think that this
test is reliable in the current test environment.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
arteam: Got it.

arteam:
@velocipedist Thanks for so much effort on this issue!

velocipedist:
@arteam thanks for your help!

btw. are there any timelines set on dw 0.8.1 release?

arteam:
Sorry, I can't give any strict ETA on 0.8.1.
It will depend on the type of release - feature or maintenance. We haven't decided yet.

jplock:
@arteam my vote would be a maintenance release as we've already fixed 6 issues in that version so far (it'd be great to get #949 resolved as well if you know a fix that for that one).  I know this issue is blocking Yammer from updating https://github.com/yammer/tenacity to DW 0.8, which I'm using for a project, so I'd personally like to see us create a release sooner rather than later if possible.

velocipedist:
@thesmith perhaps you've got the bandwidth to look at #949 ?

thesmith:
Looks like @joschi got to it before I had a chance :)

PR #942, opened 2015-03-13
Improve extraction of full URL from HttpServletRequest.

thomasdarimont:
Added fast-path for the case where a query string is not present in the request.
Previously a potentially unused StringBuilder instance was allocated for every request.


jplock:
Thanks!

PR #943, opened 2015-03-15
Provide a default view configuration.

ojacobson:
It's awkward to demand configuration from apps, even when they don't have any need to override the defaults.


ojacobson:
I'm not actually married to the _emtpy_ map as the default configuration, but `dropwizard-views` ought to be able to do a sane thing by default. It has been able to do so, in the past.

This should retain compatibility with apps that have already overridden `getViewConfiguration`. That mechanism feels pretty sane to me.

jplock:
@carlo-rtr or @joschi any thoughts on this one?

carlo-rtr:
LGTM

PR #945, opened 2015-03-16
Auth rewrite using filters

carlo-rtr:
https://github.com/dropwizard/dropwizard/issues/768


dropwizard-auth/src/main/java/io/dropwizard/auth/Auth.java
```diff
@@ -3,14 +3,13 @@
...
  *
  * @see Authenticator
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
-@Target({ ElementType.PARAMETER, ElementType.FIELD })
+@Target({ ElementType.METHOD })
 public @interface Auth {
     /**
      * If {@code true}, the request will not be processed in the absence of a valid principal. If
```
jplock: This should probably be removed as it's not longer applicable.

jplock: Oh, I see where you are using it, but does it make sense any longer?  The `SecurityContext` is always available, but a `Principal` may not be present all of the time.

carlo-rtr: I was using the Method annotation to predicate whether the filter which sets the custom ```SecurityContext``` should run for a given resource method

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public AuthDynamicFeature(AuthHandler authHandler) {
+        this.authHandler = authHandler;
+    }
+
+    @Override
+    public void configure(ResourceInfo resourceInfo, FeatureContext context) {
+        AnnotatedMethod am = new AnnotatedMethod(resourceInfo.getResourceMethod());
+
+        if (am.isAnnotationPresent(Auth.class)) {
```
glennmcallister: Given that we could be using `resourceInfo.getResourceMethod().getParameterAnnotations()`, why do we need to abandon the exiting Auth marker annotation? I admit, I don't know how we can go from `@Context SecurityContext securityContext` in the Resource parameter back to `@Auth SomeRandomAuthType authParam` to preserve backward compatibility, but this has always been my biggest concern in this change. 

Namely, is there anyway we can do this without having to force everyone to redo their auth code?

carlo-rtr: TBH, I don't see how we do it without a big shift.

MarkHerhold: The auth stuff in my project is already going to need to be refactored to move to DW 0.8.0. Having this change at the same time as the other 0.8.x stuff is better than making devs make sweeping revisions everything twice in short succession.

+1

carlo-rtr:
I've added some updates to the documentation. In doing so, I realized the ``@Auth(required=false)`` feature may not work as expected anymore. I'll work on adding some test coverage and hash out if it's possible to support that anymore.

carlo-rtr:
I've squashed the commits, got ``@Auth(required=false) working, and added test coverage

jplock:
I think this looks good and would like to see this change made.  Since this is a big departure from the old model, does anyone else want to chime in with an opinion on merging this in?  /cc @joschi @arteam @glennmcallister

glennmcallister:
To alleviate backwards compatibility concerns, we'll have to update the
migration guide, 'cause as pointed out this is a big departure. I agree
this is the right long term approach (and it's going to fix a long standing
issue I've had with our access logs as a byproduct, I suspect) and major
props for fixing this Carlo. So +1 from me.

joschi:
With this change we could probably get rid of the `@Auth` annotation altogether and use the "standard" JEE annotations like `javax.annotation.security.RolesAllowed`, `javax.annotation.security.PermitAll` and `javax.annotation.security.DenyAll` (http://docs.oracle.com/javaee/7/api/javax/annotation/security/package-summary.html) which are supported by Jersey (https://jersey.java.net/apidocs/2.17/jersey/org/glassfish/jersey/server/filter/RolesAllowedDynamicFeature.html).

arteam:
I'm basically fine with breaking backward compatability in case of security fixes.

The only concern I have is that should we put this change on 0.9.0 or 0.8.1? I guess people will want maintanence releases for the 0.8 branch only with bugfixes/improvements. Not everyone will be happy to make 2 major updates one after another.
Maybe we could do a separate _0.8.1-security_ release with this change and merge it to upstream in 0.9.0?

jplock:
To @joschi's point, what do people think about updating the `dropwizard-auth` pom.xml to include:

```xml
<dependency>
    <groupId>javax.annotation</groupId>
    <artifactId>jsr250-api</artifactId>
    <version>1.0</version>
</dependency>
```

to ship those annotations with Dropwizard?  My vote would be to include this in `0.9.0` as it's a large change for a patch release.  I think the `@Auth` annotation should remain as it provides context on whether a principal will be available in the `SecurityContext` at all.

carlo-rtr:
@joschi I looked at what you referenced. I also found [this](http://grepcode.com/file/repo1.maven.org/maven2/org.glassfish.jersey.core/jersey-server/2.15/org/glassfish/jersey/server/filter/RolesAllowedDynamicFeature.java#RolesAllowedDynamicFeature), which is included with our existing dependencies.  If we use those annotations, I wonder if people will get confused and expect the behavior provided that filter. I also wonder if we could use the ``RolesAllowedDynamicFeature``? I think we would still need a filter which reads headers and injects a valid ``SecurityContext``, similar to the logic in ``BasicCredentialAuthHandler`` and ``OAuthCredentialHandler``. Then we could rely on ``RoleAllowedDynamicFeature`` to the filtering based on the ``SecurityContext``

carlo-rtr:
@jplock if we start using them, yes I think we should include the dependency. Currently we get a dependency in the ``test`` scope

jplock:
@carlo-rtr I think it'd be a good idea to register the `RolesAllowedDynamicFeature` to make this work holistically across both authentication and authorization flows.  We just need to make it easy to override the `SecurityContext.isUserInRole()` method as that's what those annotations use to determine if a user can execute the resource method or not.  I guess in that case, authentication and authorization would become a `ConfiguredBundle` to provide a packaged solution.

carlo-rtr:
Took another swing with ``RolesAllowedDynamicFeature``. https://github.com/dropwizard/dropwizard/pull/952

If that looks good, we can close this

jplock:
@carlo-rtr should we close this in favor of #952 now?

PR #948, opened 2015-03-18
Use same ObjectMapper for client and server in ResourceTestRule

froden:
In `ResourceTestRule` configure the client in `JerseyTest` to use the same provided `ObjectMapper` as the server to prevent serialization/deserialization issues.

See https://github.com/dropwizard/dropwizard-java8/issues/8 for details.


arteam:
Looks great! Thanks for contribution.

PR #950, opened 2015-03-23
User manual mentions outdated Maven Shade Plugin

rkettelerij:
Changed 1.6 in 2.3 which provides amongst others better error messages.

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -588,7 +588,7 @@ this:
     <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


joschi:
Thanks!

arteam:
As I recall, Coda did a similar change couple of years ago, but then downgraded *maven-shade-plugin* to 1.6. Source - de8f953. 

Is this note relevant for *maven-shade-plugin-2.3*?

rkettelerij:
I don't think it's wise to stick to 1.6 just for the sake of not having a dependency-reduced-pom.xml in your base dir. That being said, if you set createDependencyReducedPom to false the maven-shade-plugin won't create such a file.

PR #951, opened 2015-03-23
Improvements of IntegrationTest in the example application

arteam:
* The test can be executed from an IDE
* Use a test, not a real configuration
* Dynamically assign application and admin ports
* Close an HTTP client in the test
* Replace tabs to spaces
* Add an integration test for work with a database  
* Create a temporary folder
* Override the DW database config
* Migrate the DB schema
* Test that database-related resource is working


jplock:
LGTM!  Thanks!

PR #952, opened 2015-03-24
Auth with RolesAllowedDynamicFeature

carlo-rtr:
This allows evaluating auth before processing resource method params


docs/source/manual/auth.rst
```diff
@@ -130,13 +261,15 @@ For this to work properly, all chained factories must produce the same type of p
 Protecting Resources
 ====================
 
-To protect a resource, simply include an ``@Auth``-annotated principal as one of your resource
-method parameters:
+To protect a resource, simply include the ``@RolesAllowed`` annotation on your resource method.
+If you need access to the Principal, you need at a parameter to your method ``@Context SecurityContext context``
```
glennmcallister: I think you meant "you need to add a parameter to"

carlo-rtr: thanks, fixed

jplock:
@carlo-rtr do we need to add that annotation dependency or are they available for use outside of the `test` scope?  Otherwise I think this looks great.  Thanks for all of your work on it!

glennmcallister:
Agreed. This is awesome.

We do still need to decide if this warrants branching for 0.8.1 for all the
other changes made up to this point, and including this in 0.9.x since it
is a MAJOR change in how we do authentication. And now authorization as a
feature, since we never really supported that previously.

jplock:
My vote would be for `0.9.0` and that we should accelerate releasing `0.8.1` to address the issues that have been fixed there.  We could create a `0.8` branch to keep a place for creating additional `0.8.x` fixes.

arteam:
I agree with Justin on this one.
I think we should release `0.8.1` in mid-april and then move to `0.9.0` with back reports to `0.8.x`

carlo-rtr:
@jplock The example works fine. You get the dependency for free with jersey-server, which is non optional for DW. Do you think I should add it explicitly?

``
[INFO] |  +- io.dropwizard:dropwizard-jersey:jar:0.8.1-SNAPSHOT:compile
[INFO] |  |  +- org.glassfish.jersey.core:jersey-server:jar:2.17:compile
[INFO] |  |  |  +- org.glassfish.jersey.core:jersey-common:jar:2.17:compile
[INFO] |  |  |  |  +- org.glassfish.jersey.bundles.repackaged:jersey-guava:jar:2.17:compile
[INFO] |  |  |  |  \- org.glassfish.hk2:osgi-resource-locator:jar:1.0.1:compile
[INFO] |  |  |  +- javax.ws.rs:javax.ws.rs-api:jar:2.0.1:compile
[INFO] |  |  |  +- org.glassfish.jersey.media:jersey-media-jaxb:jar:2.17:compile
[INFO] |  |  |  +- javax.annotation:javax.annotation-api:jar:1.2:compile
``

jplock:
It's probably fine then, thanks for checking.  So do we want to merge this into `master` and add it to a `0.9.0` milestone?  Any other opinions?

carlo-rtr:
I kicked off a discussion on the forums for the plan for 0.8.1
https://groups.google.com/forum/#!topic/dropwizard-dev/QxAz9PZBA-E

jondavidnd1:
I was looking over the changes to the documentation for this issue, and I think the code examples may be missing some info.

It seems like this line:
```java
final Function<AuthFilter.Tuple, SecurityContext> securityContextFunction = return new SecurityContext() {
```

should be more like this:

```java
final Function<AuthFilter.Tuple, SecurityContext> securityContextFunction = new Function<AuthFilter.Tuple, SecurityContext>() {
            @Override
            public SecurityContext apply(final AuthFilter.Tuple input) {
                return new SecurityContext() {
                     ...
                };
            }
       };
```

carlo-rtr:
This should fix that. Thanks.

https://github.com/dropwizard/dropwizard/pull/999

dropwizard-auth/src/test/java/io/dropwizard/auth/ChainedAuthProviderTest.java
```diff
@@ -1,12 +1,15 @@
 package io.dropwizard.auth;
```
yunspace: Should this test be moved to package `io.dropwizard.auth.chained` to be consistent with the location of `ChainedAuthFilter`?

carlo-rtr: I think that makes sense.

PR #954, opened 2015-03-27
Enabling UnitOfWork on implementations of interfaces

akumlehn:
* patched `UnitOfWorkApplicationListener` to detect `UnitOfWork` annotation on methods that implement a resource method defined on an interface
* caused by difference between `definitionMethod` (interface) and `handlingMethod` (implementation) for a jersey resouce
* behavior:
  * listener checks `definitionMethod` for `UnitOfWork`
  * if found, this annotation wins (as before)
  * if not found, listener now checks `handlingMethod` for `UnitOfWork`
* extended tests to cover the new behavior

Minimal example to share jersey definitions among clients and the server:
* an interface defines the jersey resources (without `UnitOfWork`)
```java
@Path("user")
public interface UserResource {
    @GET
    @Path("list")
    List<User> getAllUsers();//definitionMethod
}
```

* the implementation adds the `UnitOfWork` annotation (server project, to avoid a leaked dependency).
```java
public class UserResourceImpl implements UserResource {
    @Override
    @UnitOfWork
    public List<User> getAllUsers() { //... }//handlingMethod
}
```


arteam:
This PR looks extremely good for me.
Really good explanation of the change with a bunch of tests as well.

Thanks for your thorough contribution!

carlo-rtr:
agreed, nice explanation

PR #958, opened 2015-03-29
Rethrow Hibernate exceptions as Jersey exceptions

### LINKED ISSUE #949: Custom ExceptionMapper not invoked when Hibernate rollback
Using DW 0.8.0, Java 8

I have one custom exception mapper to handle `hibernate` exception:
```java
import org.hibernate.exception.ConstraintViolationException;


@Provider
public class SQLConstraintViolationExceptionMapper implements ExceptionMapper<ConstraintViolationException> {
    ....
}
```

I registered the mapper https://github.com/paukiatwee/budgetapp/blob/master/src/main/java/io/budgetapp/BudgetApplication.java#L143

When `org.hibernate.exception.ConstraintViolationException` exception is thrown when foreign key violation occurred, `SQLConstraintViolationExceptionMapper` is no called.

You can reproduce the problem by running this integration test https://github.com/paukiatwee/budgetapp/blob/master/src/test/java/io/budgetapp/resource/CategoryResourceIT.java#L65

This might caused by https://github.com/dropwizard/dropwizard/blob/master/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java#L87 exception not handled by container.

Working fine for DW 0.7.0

arteam:
If a database commit fails, it's better not to rethrow an Hibernate exception in the original form, but
wrap it as a MappableException.

This way users can map sql exceptions to web responses and derive a business value from the
database validation checks.

Resolve #949

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
@@ -6,6 +6,7 @@ import java.util.Map;
 
 import javax.ws.rs.ext.Provider;
 
+import org.glassfish.jersey.server.internal.process.MappableException;
 import org.glassfish.jersey.server.model.Resource;
 import org.glassfish.jersey.server.model.ResourceMethod;
 import org.glassfish.jersey.server.monitoring.ApplicationEvent;
@@ -84,7 +85,7 @@ public class UnitOfWorkApplicationListener implements ApplicationEventListener {
                         commitTransaction();
                     } catch (Exception e) {
                         rollbackTransaction();
-                        this.<RuntimeException>rethrow(e);
+                        throw new MappableException(e);
                     }
                     finally {
                         this.session.close();
@@ -135,11 +136,6 @@ public class UnitOfWorkApplicationListener implements ApplicationEventListener {
                     txn.commit();
                 }
             }
-        }
-
-        @SuppressWarnings("unchecked")
-        private <E extends Exception> void rethrow(Exception e) throws E {
-            throw (E) e;
         }
     }
 
--- a/dropwizard-hibernate/src/test/java/io/dropwizard/hibernate/DataExceptionMapper.java
+++ b/dropwizard-hibernate/src/test/java/io/dropwizard/hibernate/DataExceptionMapper.java
@@ -1,0 +1,26 @@
+package io.dropwizard.hibernate;
+
+import io.dropwizard.jersey.errors.ErrorMessage;
+import org.hibernate.exception.DataException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class DataExceptionMapper implements ExceptionMapper<DataException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DataException.class);
+
+    @Override
+    public Response toResponse(DataException e) {
+        LOGGER.error("Hibernate error", e);
+        String message = e.getCause().getMessage().contains("EMAIL") ? "Wrong email" : "Wrong input";
+
+        return Response.status(Response.Status.BAD_REQUEST)
+                .entity(new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(), message))
+                .build();
+    }
+}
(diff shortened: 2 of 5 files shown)


dropwizard-hibernate/src/test/java/io/dropwizard/hibernate/DataExceptionMapper.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DataException.class);
+
+    @Override
+    public Response toResponse(DataException e) {
+        LOGGER.error("Hibernate error", e);
+        String message = e.getCause().getMessage().contains("EMAIL") ? "Wrong email" : "Wrong input";
+
+        return Response.status(Response.Status.BAD_REQUEST)
+                .entity(ImmutableMap.of("error", message))
```
joschi: For the sake of consistency the response should be built using [`ErrorMessage`](https://github.com/dropwizard/dropwizard/blob/v0.8.0/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/ErrorMessage.java).

arteam: That makes sense. Reused `ErrorMessage` for error message generation and testing.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
```diff
@@ -84,7 +85,7 @@ else if (event.getType() == RequestEvent.Type.RESP_FILTERS_START) {
                         commitTransaction();
                     } catch (Exception e) {
                         rollbackTransaction();
-                        this.<RuntimeException>rethrow(e);
+                        throw new MappableException(e);
```
joschi: I'm a little bit uncomfortable with using an exception class clearly being used only (?) internally by Jersey. Maybe we could use a second `catch` block before this one which catches [`HibernateException`](https://docs.jboss.org/hibernate/orm/4.3/javadocs/org/hibernate/HibernateException.html) (which inherits from RuntimeException) and re-throw it unchanged while we use [`Throwables#propagateIfPossible(java.lang.Throwable)`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Throwables.html#propagateIfPossible(java.lang.Throwable)) in the catch-all block (which is currently throwing the MappableException).

dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/ErrorMessage.java
```diff
@@ -1,5 +1,8 @@
 package io.dropwizard.jersey.errors;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
```
jplock: Can we also add `@JsonInclude(JsonInclude.Include.NON_NULL)` in here to avoid having null fields show up in the error responses if they aren't being set? (I ran into this with the `details` field)

joschi: :+1:

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
arteam: * I beleive Jersey already takes care of propogating an original exception, if a mapping has not been found. [Source](https://github.com/jersey/jersey/blob/master/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java#L612-619).
* Do we really have checked exceptions? `commitTransaction` doesn't throw any.

joschi: Good point. So we could simply catch `RuntimeException` instead of `Exception` and re-throw after we rolled back the transaction. No need to wrap it in either `MappableException` or `RuntimeException`. This should make life easier for us. :smile:

arteam: But that returns us to the original version. The whole idea is to wrap an exception to `MappableException`, so Jersey will try to find exception mappers.

joschi: The original version casted every Exception to `RuntimeException`. Might this have caused the problem with Jersey not trying to map the specific exception with a registered exception mapper?

arteam: Well, if we don't wrap an exception
````java
try {
    commitTransaction();
} catch (RuntimeException e) {
    rollbackTransaction();
    throw e;
}
````
Then `testSqlExceptionIsHandled` fails with 500 error.

If we delibitraly throw `DataException` from a resource then the mapper works, because Jersey internally wraps the exception as `MappingException`. But, as I beleive, it does so only for resources, but not for application listeners.

joschi:
LGTM.

PR #959, opened 2015-03-30
Serving static assets example

micheee:
Changed server type from `simple` to `default`.
Having `simple` will lead to 404 errors.
I just experienced this when toying around, so maybe someone more experienced should have a closer look at this.

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -443,7 +443,7 @@ application. To enable it, move your application to a sub-URL.
 .. code-block:: yaml
 
     server:
-      type: simple
+      type: default
       rootPath: /application/*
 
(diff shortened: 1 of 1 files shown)


arteam:
That's true. 

`default` is better, because in that case the application will be available on _/application/*_ path.
For `simple` it will be _/application/application/*_, because *rootPath* is calculated from `applicationContextPath`. And it's _/application_ by default.

micheee:
I see. Coming from the tutorial it was not obvious why the `applicationContextPath` changed there. Thanks for the clarification. Should I add your explanation to the docs?

arteam:
Yes, a note about that would be great. 

Also we could rename `/application/*` to `api/*` to reduce confusion for the simple server configuration.

PR #961, opened 2015-03-30
Add configuration section to jersey client docs

### LINKED ISSUE #960: Honor Jersey Client Timeout Defaults
According to the [latest Jersey configuration documentation][jt], a client has a read and connect timeout of infinity. Dropwizard assumes a default read and connect timeout for HttpClient and Jersey's client of 500 milliseconds. This was not obvious, especially because I was led down the path of trying to set Jersey Client Properties on the Builder, and didn't realize until later that [the docs say](https://github.com/dropwizard/dropwizard/blob/9c7adb21ac5abd5768f029ca1b2331e4b4e29ed9/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java#L103-L116)

> WARNING:The default connector ignores Jersey properties

It took me an embarrassingly long time to realize how everything was connected (and that I could change the timeouts with a yaml change), and I'd like to save some time for those in the future.

Solutions:

1. Mention in docs that the timeouts are in effect (which deviates from the default). Provide a sentence stating how to change them.
2. Change the code such that Jersey configuration defaults are actually the default for the Jersey Client. I understand if there is hesitation towards this change as this may break applications that implicitly rely on the 500 millisecond timeout (which is currently the default).

If there are any other deviations, it would probably be fruitful to document them too.

Which solution is preferable?

[jt]: https://jersey.java.net/documentation/latest/appendix-properties.html#appendix-properties-client

nickbabcock:
Thought I'd flesh out Jersey's Client documentation a bit more.

I welcome a second pair of eyes to make sure it is alright and makes sense.

I should note that it relies on the changes in #939 to be accurate.

Closes #960

### DIFF
--- a/docs/source/manual/client.rst
+++ b/docs/source/manual/client.rst
@@ -150,3 +150,23 @@ Then, in your service's ``run`` method, create a new ``JerseyClientBuilder``:
         environment.jersey().register(new ExternalServiceResource(client));
     }
 
+Configuration
+-------------
+
+The Client that Dropwizard creates deviates from the `Jersey Client Configuration`_ defaults. The
+default, in Jersey, is for a client to never timeout reading or connecting in a request, while in
+Dropwizard, the default is 500 milliseconds. There are a couple of ways to change this behavior. The
+recommended way is to modify the :ref:`YAML configuration <man- configuration-clients-http>`.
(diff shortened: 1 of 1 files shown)


arteam:
Thanks for taking your time for writing that.

Commited to master with a couple of tweaks.

nickbabcock:
Great, thank you for the tweaks!

Issue #513, opened 2014-03-26
Allow in-memory unit testing of resources using views?


chids:
Since `ViewMessageBodyWriter` contains `@Context private HttpHeaders headers;` it's not possible to unit test resources using views with the lightweight in-memory Jersey provider (using `ResourceTestRule`).

However since a `MessageBodyWriter.writeTo` receives the request headers as a `MultivaluedMap<String, Object>` argument there shouldn't be an absolute need for injecting `HttpHeaders`.

I did a very rough PoC by rewriting `ViewMessageBodyWriter.detectLocale()` into this horrid mess:
```java
    private static Locale detectLocale(MultivaluedMap<String, Object> headers) {
        InBoundHeaders h = new InBoundHeaders();
        List<Object> accepts = headers.get(ACCEPT_LANGUAGE);
        if(accepts != null && accepts.size() > 0) {
            h.put(ACCEPT_LANGUAGE, Lists.transform(accepts, Functions.toStringFunction()));
        }
        ContainerRequest cr = new ContainerRequest(new WebApplicationImpl(), null, null, null, h, null);
        for (AcceptableLanguageTag langTag : HttpHelper.getAcceptLanguage(cr)) {
            Locale x = langTag.getAsLocale();
            if (!x.toString().contains("*")) { // Freemarker doesn't do wildcards well
                return x;
            }
        }
        return Locale.getDefault();
    }
```

I'd be happy to put some more effort into it and submit a PR if others think its valuable.

carlo-rtr:
If this https://github.com/dropwizard/dropwizard/pull/966 gets merged, then it should work. I verified.

``
package com.example.helloworld.resources;

import com.codahale.metrics.MetricRegistry;
import com.google.common.collect.Lists;
import io.dropwizard.testing.junit.ResourceTestRule;
import io.dropwizard.views.ViewMessageBodyWriter;
import io.dropwizard.views.ViewRenderer;
import io.dropwizard.views.mustache.MustacheViewRenderer;
import org.glassfish.jersey.test.grizzly.GrizzlyWebTestContainerFactory;
import org.junit.ClassRule;
import org.junit.Test;

import java.util.List;

import static org.assertj.core.api.Assertions.assertThat;

public class ViewResourceTest {
    private static List<ViewRenderer> renders = Lists.<ViewRenderer>newArrayList(new MustacheViewRenderer());

    @ClassRule
    public static final ResourceTestRule resources = ResourceTestRule.builder()
[...]

Issue #922, opened 2015-03-05
BasicAuthFactory in dropwizard-testing.


phxql:
I'm trying to get the `BasicAuthFactory` in dropwizard testing to work. As stated in the documentation, I'm using `GrizzlyTestContainerFactory` as test container factory. `BasicAuthFactory` still isn't working because `@Context private HttpServletRequest request;` isn't injected. I also tried `GrizzlyWebTestContainerFactory`, this throws an exception:

```
java.lang.IllegalArgumentException: The deployment context must be an instance of ServletDeploymentContext.
	at org.glassfish.jersey.test.grizzly.GrizzlyWebTestContainerFactory.create(GrizzlyWebTestContainerFactory.java:85)
	at org.glassfish.jersey.test.JerseyTest.createTestContainer(JerseyTest.java:277)
	at org.glassfish.jersey.test.JerseyTest.setUp(JerseyTest.java:609)
	at io.dropwizard.testing.junit.ResourceTestRule$1.evaluate(ResourceTestRule.java:157)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
```

How to test basic auth protected resources?

scho:
Have you tried the way it is done here:
https://github.com/dropwizard/dropwizard/blob/master/dropwizard-auth/src/test/java/io/dropwizard/auth/oauth/OAuthCustomProviderTest.java

It worked for me, except that I couldn't assert the result in case of a non-200 response (we render json with some error information, but grizzly always renders html/text).

Another idea:
Have you tried to test it with an integration test and the [DropwizardAppRule](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/DropwizardAppRule.java)?

vultron81:
I am also having issues getting ResourceTestRule to work with BasicAuthFactory and GrizzlyTestContainerFactory for v0.8.0.  I tried to implement what was done in the OAuthCustomProviderTest.java code but no matter what I do I get a HTTP 401 error.  Also my custom Authenticator is never called.  Below is how I am initializing the ResourceTestRule:

@ClassRule
    public static final ResourceTestRule resources = ResourceTestRule.builder()
            .setTestContainerFactory(new GrizzlyTestContainerFactory())
[...]

PR #966, opened 2015-03-31
test rule supports auth

carlo-rtr:
All of the relevant discussion is in the issue
https://github.com/dropwizard/dropwizard/issues/922

this also would close https://github.com/dropwizard/dropwizard/issues/513


jplock:
LGTM!

PR #969, opened 2015-04-02
Move Jersey test dependencies to the dependency management section

arteam:
It reduces burden of upgrading dependencies to the correct version and excluding unwanted dependencies in every place.

It also excludes a duplicate dependency on javax.servlet-api 3.0.1


jplock:
LGTM!

PR #971, opened 2015-04-05
Facility to specify a database timezone

### LINKED ISSUE #968: dropwizard-jdbi: Binding joda-time DateTime objects to SQL uses JVM timezone, not the timezone in the DateTime object
There is a bug in io.dropwizard.jdbi.args.JodaDateTimeArgument that appears when the JVM timezone differs from the database timezone. This class uses:
  statement.setTimestamp(position, new Timestamp(value.getMillis()));
to bind a DateTime ("value" here) to a PreparedStatement ("statement"), which creates a Timestamp using the JVM's timezone instead of the DateTime object's timezone.

The solution seems to be to use the setTimestamp method that also takes a Calendar argument (http://docs.oracle.com/javase/7/docs/api/java/sql/PreparedStatement.html#setTimestamp%28int,%20java.sql.Timestamp,%20java.util.Calendar%29).

arteam:
Provide facility to specify the database time zone during converting Joda-Time `DateTime`
values from/to `SQL TIMESTAMP`.

It's crucial for cases when the database operates in a different timezone then the client and it doesn't handle timestamps as SQL `TIMESTAMP WITH TIME ZONE`

Resolve #968

### DIFF
--- a/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
+++ b/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
@@ -6,6 +6,7 @@ import com.codahale.metrics.jdbi.InstrumentedTimingCollector;
 import com.codahale.metrics.jdbi.strategies.DelegatingStatementNameStrategy;
 import com.codahale.metrics.jdbi.strategies.NameStrategies;
 import com.codahale.metrics.jdbi.strategies.StatementNameStrategy;
+import com.google.common.base.Optional;
 import io.dropwizard.db.DataSourceFactory;
 import io.dropwizard.db.ManagedDataSource;
 import io.dropwizard.jdbi.args.JodaDateTimeArgumentFactory;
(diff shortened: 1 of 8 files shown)


jplock:
LGTM!

dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
```diff
@@ -28,18 +31,36 @@
...
-                  NameStrategies.CONTEXT_CLASS,
-                  NameStrategies.CONTEXT_NAME,
-                  NameStrategies.SQL_OBJECT,
-                  new StatementNameStrategy() {
-                      @Override
-                      public String getStatementName(StatementContext statementContext) {
-                          return RAW_SQL;
-                      }
-                  });
+                    NameStrategies.CONTEXT_CLASS,
```
glennmcallister: I realize it's a little late now, but formatting changes like this (particularly pure whitespace changes) that have no impact typically aren't something you want to include in a PR unless you are rectifying a significant style deviation. The reason being that now it becomes more difficult to determine the original source of the change for these lines.

arteam: You are right. I just formatted the class in my IDE instinctively. 
It was a bit of sloppy and I will try to avoid such changes in future.

PR #976, opened 2015-04-08
Return back the default constructor for JodaDateTimeMapper

arteam:
`JodaTimeMapper` is a public class and should provide the default constructor because of:

* backwards compatibility
* users, who don't care about timezones
* consistency with `JodaTimeArgumentFactory`


joschi:
Makes sense.

PR #979, opened 2015-04-09
Add support for X.class to DropwizardClientRule

jkrems:
This can be helpful in cases where non-Resources are passed in that
depend on `@Context` annotations or it's preferable for other
reasons to have Jersey manage the object lifetime.


carlo-rtr:
Thanks for the contrib

PR #980, opened 2015-04-09
Expand documentation around commands

nickbabcock:
I like to use commands and I thought the documentation surrounding them were a little light.

Let me know if everything looks alright, or if there is someway to improve it.


carlo-rtr:
Thanks

PR #984, opened 2015-04-10
Use Type parameter for configuration in initialize

dhs3000:
Use Type parameter for configuration in both, run and initialize so that one can for example add another ConfigurationBundle in initialize method.

See also https://github.com/dropwizard/dropwizard/pull/795

----

Example: There is a team-wide default set of bundles that is used for all webservices. They all use the same (superset) of configuration. Instead of adding all the bundles in every Application I'd like to add a single default bundle that adds all bundles I need.
```java
class DeploymentConfiguration extends Configuration {
    boolean production = true;
}

class App extends Application<DeploymentConfiguration> {
    @Override
    public void initialize(Bootstrap<DeploymentConfiguration> bootstrap) {
        bootstrap.addBundle(new DefaultBundle());
    }

    @Override
    public void run(DeploymentConfiguration configuration, Environment environment) throws Exception {
        // ..
    }
}

class DefaultBundle implements ConfiguredBundle<DeploymentConfiguration> {
    @Override
    public void run(DeploymentConfiguration configuration, Environment environment) throws Exception {
    }

    @Override
    public void initialize(Bootstrap<? extends DeploymentConfiguration> bootstrap) {
        bootstrap.addBundle(new BundleA());
        bootstrap.addBundle(new BundleB());
    }
}

class BundleA implements ConfiguredBundle<DeploymentConfiguration> {
    @Override
    public void run(DeploymentConfiguration configuration, Environment environment) throws Exception {
        environment.servlets().addServlet(
                "someservlet",
                new SomeServlet(configuration.production)); // <-- using configuration
    }

    @Override
    public void initialize(Bootstrap<? extends DeploymentConfiguration> bootstrap) {
    }
}
```
With ```Bootstrap<?>``` in the class ```DefaultBundle``` I can't add another ```ConfiguredBundle<DeploymentConfiguration>```, that produces a compile error.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/ConfiguredBundle.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/ConfiguredBundle.java
@@ -9,7 +9,7 @@ import io.dropwizard.setup.Environment;
  *
  * @param <T>    the required configuration interface
  */
-public interface ConfiguredBundle<T> {
+public interface ConfiguredBundle<T extends Configuration> {
     /**
      * Initializes the environment.
      *
@@ -24,5 +24,5 @@ public interface ConfiguredBundle<T> {
      *
      * @param bootstrap the application bootstrap
      */
-    void initialize(Bootstrap<?> bootstrap);
+    void initialize(Bootstrap<? extends T> bootstrap);
 }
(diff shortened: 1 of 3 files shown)


carlo-rtr:
This passed on jdk 7, but not jdk8. There is a non deterministic test, fix is here:

https://github.com/dropwizard/dropwizard/pull/985

PR #985, opened 2015-04-10
DateTimeZone only accepts a subset of the IDs from TimeZone

carlo-rtr:
so build fails periodically, https://travis-ci.org/dropwizard/dropwizard/jobs/57990585


arteam:
Thanks. I ran this test so many times locally, but didn't catch any errors (seems to be environment-specific).

carlo-rtr:
no problem, easy for that to happen.

PR #986, opened 2015-04-10
Empty Strings are converted into absent Optionals

### LINKED ISSUE #983: The case of the empty string Optional in QueryParam
Taking a look at the [getting started page]

```java
@GET
@Timed
public Saying sayHello(@QueryParam("name") Optional<String> name) {
    final String value = String.format(template, name.or(defaultName));
    return new Saying(counter.incrementAndGet(), value);
}
```

The url `http://localhost:8080/hello-world?name=` will return "Hello, !".

I'm unsure how I feel about this behavior. The person I was walking through
the tutorial expressed concern that it failed the principle of least
astonishment because to him `name` is blank -- thus absent, and should result
in a return of "Hello, Stranger!"

A workaround had to be coded:

```java
@GET
@Timed
public Saying sayHello(@QueryParam("name") Optional<String> name) {
    String output = name.or(defaultName);
    if (Strings.isNullOrEmpty(output)) {
        output = defaultName;
    }

    final String value = String.format(template, output);
    return new Saying(counter.incrementAndGet(), value);
}
```

This seems to defeat the purpose of optional. I think it would be reasonable
to change the behavior of `Optional` so that a blank query parameter is
absent, even though `Optional.of("")` is completely valid.

An argument against changing the code is that it would be a breaking change.
People may have come to rely on blank query params meaning a non-absent
optional. Another argument is that the [spec defines] the query as

> either null or a string holding data

Is `name=` null or is it holding empty data?

If the change is wanted, an open question is what constitues a blank query
param? Is it just a length of 0 (my bias), or can blank query param can
contain all whitespace

EDIT: If the change is not wanted, there probably should be documentation informing that `Optional` can be blank.

[getting started page]: http://www.dropwizard.io/getting-started.html#creating-a-resource-class
[spec defines]: https://url.spec.whatwg.org/#concept-url-query

nickbabcock:
When an empty string is converted into an `Optional` for FormParam,
CookieParam, HeaderParam, or QueryParam, the optional is absent instead of
the previous behavior where the empty string would fill the optional.

Closes #983

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/guava/OptionalParamConverterProvider.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/guava/OptionalParamConverterProvider.java
@@ -2,6 +2,7 @@ package io.dropwizard.jersey.guava;
 
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
+import com.google.common.base.Strings;
 import org.glassfish.hk2.api.ServiceLocator;
 import org.glassfish.jersey.internal.inject.Providers;
 import org.glassfish.jersey.internal.util.ReflectionHelper;
@@ -38,7 +39,7 @@ public class OptionalParamConverterProvider implements ParamConverterProvider {
                 return new ParamConverter<T>() {
                     @Override
                     public T fromString(final String value) {
-                        return rawType.cast(Optional.fromNullable(value));
+                        return rawType.cast(Optional.fromNullable(Strings.emptyToNull(value)));
                     }
 
                     @Override
--- a/dropwizard-jersey/src/test/java/io/dropwizard/jersey/guava/OptionalCookieParamResourceTest.java
+++ b/dropwizard-jersey/src/test/java/io/dropwizard/jersey/guava/OptionalCookieParamResourceTest.java
@@ -36,6 +36,13 @@ public class OptionalCookieParamResourceTest extends JerseyTest {
     }
 
     @Test
+    public void shouldReturnDefaultMessageWhenMessageIsBlank() {
+        String defaultMessage = "Default Message";
+        String response = target("/optional/message").request().cookie("message", "").get(String.class);
+        assertThat(response).isEqualTo(defaultMessage);
+    }
+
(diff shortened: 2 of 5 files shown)


jplock:
Do you know if this has always been the case (in DW 0.6, 0.7, etc.)?

nickbabcock:
For 7.1, 7.0, 6.2, 6.1, yes it has always been the case. I didn't test any further.

jplock:
@carlo-rtr, @glennmcallister, @joschi or @arteam any thoughts on this one?  If this is how the older versions worked, then technically we broke it in 0.8.x, so we should probably preserve the same functionality for now.

nickbabcock:
Ack, I believe I worded my previous sentence wrong. What I meant to say was that it has always been the case that the empty string would evaluate to `Optional.of("")` and not `Optional.absent()`, so nothing was broken between versions.

wirde:
An empty String is *not* an absent String though. Especially for FormParams it is quite common that you would want a "" to actually mean "". Less obvious use for empty query parameters but I still prefer to have the ability to know that a query parameter was present but it's value was the empty String rather than not present at all.

nickbabcock:
On further thought, @wirde's point is especially valid for the `@HeaderParam` where the header is present but empty. There is a [stackoverflow answer](http://stackoverflow.com/a/12131993/433785) with more information, but I guess it is valid for a header to be present, but no value.

Side note, some of the tests I wrote actually tested the wrong endpoint, so I fixed them up

jplock:
Thinking about this more, I'm -1 on this change as there'd be no way to set a value as an empty string (if that was the desired outcome).

carlo-rtr:
Yeah, I agree with @wirde @jplock

nickbabcock:
Do you mind if I submit a pull request adding to the documentation, noting the behavior, and adding the tests to support it, or do you think it's not needed?

jplock:
@nickbabcock i think that'd be a great idea.  Thanks for doing that.

carlo-rtr:
Yeah, thanks for contributing. We really appreciate the help.

PR #987, opened 2015-04-10
Make LoggingFactory threadsafe for parallel running tests

ahus1:
Running parallel unit tests in my personal project with the current Dropwizard release leads to errors in class initialization probably due to http://jira.qos.ch/browse/SLF4J-167

The change below allows me to run the unit tests in my personal project in parallel by making the log initialization `synchronized`. The change below also avoids a second initialization as this might lead to log messages getting lost.

Activating parallel tests in all dropwizard modules will lead to test failures in some modules, therefore I left it disabled in the parent pom.xml.

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/LoggingFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/LoggingFactory.java
@@ -37,8 +37,22 @@ public class LoggingFactory {
         bootstrap(Level.WARN);
     }
 
-    public static void bootstrap(Level level) {
-        hijackJDKLogging();
+    private static Level bootstrapLevel;
+
+    /** Boostrap logging. This is synchronized to avoid logging initialization problems
+     * when test are run in parallel, see also http://jira.qos.ch/browse/SLF4J-167 for more details.
+     * @param level initial log level
+     */
+    public static synchronized void bootstrap(Level level) {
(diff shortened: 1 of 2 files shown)


arteam:
What do you think if we could extract acquring `LoggerContext` to a separate method and spin-loop
a thread that didn't start initialization? Something like that:
````java
final Logger root = getLoggerContext().getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
....

 private static LoggerContext getLoggerContext() {
        while (true) {
            ILoggerFactory iLoggerFactory = LoggerFactory.getILoggerFactory();
            if (iLoggerFactory instanceof LoggerContext) {
                return (LoggerContext) iLoggerFactory;
            }
            try {
                System.out.println("Awaiting logger");
                Thread.sleep(100);
            } catch (InterruptedException e) {
                throw new RuntimeException(e);
            }
        }
    }
````

For me that a was single workaround that helped to run tests concurrently.

carlo-rtr:
It seems like either would work. Given that bootstrapping is not a hot path, my guess is the lock contention will not introduce significant performance degradation.

arteam:
I managed to run tests in parallel in the project: https://github.com/arteam/dropwizard/commits/conc_tests. It wasn't easy and I had to ignore some tests, that are inherently not thread-safe (with system properties). I will try to make a pull request in a couple of days.

carlo-rtr:
Thanks for tackling this!

PR #989, opened 2015-04-11
Make tests on an exception converter reliable

arteam:
The current test is not reliable, because it fails when it's run from an IDE. The reason for this is that the amount of omitted frames is different for the IDE test runner and, hence,
is not reliable.

Solution is not to match whole stack trace, but split the test to two independent tests:

* First checks, that every line starts with an exclamation mark
* Second checks, that the exceptions ordering is correct and the root cause is placed first in the stack trace.


jplock:
:+1:

glennmcallister:
Agreed, nice change.

Issue #963, opened 2015-03-31
Repeatable Benchmarking


carlo-rtr:
We need a way to identify macro performance degradations. Ideally this would be part of the build process. It may also be worth creating doing micro benchmarking.

JonMR:
Is the goal to have this fully automated and part of CI or to be run manually during the RC phase of development?

Fully automated is obviously more difficult.  I don't think Travis CI gives any guarantees about VM sizing.

carlo-rtr:
I don't think it needs to be a hard requirement that it runs with the build. My only concern is that if it doesn't happen in some automated way, there may be some lag before degradations are caught. This obviously makes it harder to identify the culprit. If you have ideas, I'm all ears.

carlo-rtr:
One think we should do is make sure that empower benchmark is using the latest code.

https://www.techempower.com/benchmarks/

https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Java/dropwizard

zloster:
I've added a [pull request](https://github.com/TechEmpower/FrameworkBenchmarks/pull/2357) for updating the test code to dropwizard 1.0.3 from 0.8.1 at FrameworkBenchmark project. The results from my local test look very promising.

PR #990, opened 2015-04-11
First stab on dropwizard-benchmarks

arteam:
Add benchmarks for [Duration](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-util/src/main/java/io/dropwizard/util/Duration.java) and [Size](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-util/src/main/java/io/dropwizard/util/Size.java)

The benchmarks measure an average time to parse a string to Duration and Size representations.

### DIFF
--- a/dropwizard-benchmarks/pom.xml
+++ b/dropwizard-benchmarks/pom.xml
@@ -26,6 +26,12 @@
             <artifactId>jmh-generator-annprocess</artifactId>
             <version>${jmh.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>io.dropwizard</groupId>
+            <artifactId>dropwizard-util</artifactId>
+            <version>${parent.version}</version>
+        </dependency>
     </dependencies>
 
 </project>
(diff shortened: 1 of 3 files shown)


jplock:
This looks good to me.  @carlo-rtr did you want the benchmarks to focus more on benchmarking the entire system (Jetty, Jersey, etc.), or is this level of granularity ok?

carlo-rtr:
I was thinking more macro, like the empower benchmarks. I've updated the issue: https://github.com/dropwizard/dropwizard/issues/963

With that said, it seems like this would be useful too. I have never used this before, so I'm going to read up then opine.

arteam:
My 2 cents:

* The situation with micro-benchmarks is a very similar to unit testing. There is a little profit from one lone benchmark, but when the all code is covered by benchmarks, it's far easier to find the root of a perfomance problem early. When a developer checks-in the new code, we can run micro-benchmarks and check regression. If there is any, it's a sign of a possible performance problem.
* It's a tool for the maintainers to argue about perfomance with numbers. For example, if a maintainer thinks that a pull request decreases performance, he can consolidate his opinion with numbers.
* I don't know where it will lead us, but it's a tool that can improve our awareness in making decisions. More data is better.

jplock:
I think this is a good first step.  I'll merge it.

carlo-rtr:
Do you think we should have an abstract class with the main method? The only thing changing is the name, we could have an abstract method for getBenchMarkName()

ryankennedy:
This is really cool. Any reason not to shade the final JAR and make the Main-Class be `org.openjdk.jmh.Main`? I just pulled these changes down and tried to run them but the resulting JAR doesn't have a Main-Class attribute in the manifest and even specifying the `DurationBenchmark` class results in a `java.lang.NoClassDefFoundError: io/dropwizard/util/Duration`.

The [instructions for generating a project](http://openjdk.java.net/projects/code-tools/jmh/) using `mvn archetype:generate` generate a POM with shading and the Main-Class set up like I mentioned above. I'd be happy to go in and make the changes, but I first wanted to make sure @arteam hadn't skipped those for good reason.

PR #991, opened 2015-04-13
Add documentation and tests surrounding empty string optionals

nickbabcock:
Following the discussions in #986 and #983, this pull requests adds documentation and tests around empty string params, such as cases like `/hello-world?name=` where the param will evaluate to `Optional.of("")`

I decided to add in a sample implementation for those following along in the getting started, so `/hello-world?name=` will return "Hello, Stranger". The implementation derives from `AbstractParam`. Alternatively, a custom `ParamConverterProvider` could be registered but that would affect all String params for all endpoints (so I decided against that).

Also this pull request fixes a link that was missing (originally my fault :blush:)

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/NonEmptyStringParam.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/NonEmptyStringParam.java
@@ -1,0 +1,21 @@
+package io.dropwizard.jersey.params;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Strings;
+
+/**
+ * A parameter encapsulating optional string values with the condition that empty string inputs are
+ * interpreted as being absent. This class is useful when it is desired for empty parameters to be
+ * synonymous with absent parameters instead of empty parameters evaluating to
+ * {@code Optional.of("")}.
+ */
+public class NonEmptyStringParam extends AbstractParam<Optional<String>> {
+    protected NonEmptyStringParam(String input) {
+        super(input);
+    }
+
+    @Override
+    protected Optional<String> parse(String input) throws Exception {
+        return Optional.fromNullable(Strings.emptyToNull(input));
+    }
+}
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -448,6 +448,13 @@ is no ``name`` parameter in the query string, ``sayHello`` will be called with
 ``Optional.absent()``. (Support for Guava's ``Optional`` is a little extra sauce that Dropwizard
 adds to Jersey's existing functionality.)
 
+.. note::
+
+    If the client sends a request to ``/hello-world?name=``, ``sayHello`` will be called with
+    ``Optional.of("")``. This may seem odd at first, but this follows the standards (an application
+    may have different behavior depending on if a parameter is empty vs nonexistent). For more
+    information see :ref:`the section on parameters <man-core-resources-parameters>`
+
 Inside the ``sayHello`` method, we increment the counter, format the template using
 ``String.format(String, Object...)``, and return a new ``Saying`` instance.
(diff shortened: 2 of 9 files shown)


joschi:
Thanks for this pull request!

It might even be useful to actually add `NonEmptyString` (and a test case :wink:) to `io.dropwizard.jersey.params` and refer to that in the documentation.

nickbabcock:
I'll append a commit with the new functionality, tests, and doc changes, so you guys can mix and match what commits you want in case everything in the pull request is not needed/wanted.

nickbabcock:
I added the functionality of `NonEmptyString` in class called `NonEmptyStringParam` (to keep the convention of sibling classes). It is somewhat of a misnomer, I suppose it could be called `NonEmptyOptionalStringParam`, but that starts being a mouthful

Feel free to vet the implementation/name.

This pull request is starting to become a potpourri of commits, so let me know if I can do anything on my part to help.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/NonEmptyStringParam.java
```diff
@@ -0,0 +1,21 @@
...
+ * {@code Optional.of("")}.
+ */
+public class NonEmptyStringParam extends AbstractParam<Optional<String>> {
+    protected NonEmptyStringParam(String input) {
+        super(input);
+    }
+
+    @Override
+    protected Optional<String> parse(String input) throws Exception {
+        return Optional.fromNullable(Strings.emptyToNull(input));
```
jplock: I'm not sure how likely it would be, but should we do any string trimming on the `input` here to remove whitespace if a user passes a single space character in?

nickbabcock: Yeah I can add that functionality in, it doesn't affect my use case (the clients of my api always seem to call with empty params `?bar=&baz=&qux=`, which need to be interpreted as absent). If  but this class is to ignore whitespace as well, do you have suggestions on names? It wouldn't be just NonEmpty strings anymore.

jplock: `Blank` maybe?  I'm not even sure if it's worth the effort.  Any opinion @joschi ?

joschi: I would also create two classes, NonEmptyString and NonBlankString in this case.

Although I'm not quite sure if it's worth the effort either, especially since Jersey Bean Validations could be used for this (with `@NotEmpty` and `@NotBlank`) for non-optional parameters.

nickbabcock: > especially since Jersey Bean Validations could be used for this (with @NotEmpty and @NotBlank) for non-optional parameters

I think the key here is "non-optional parameters." Ideally I want to define optional parameters that if present, contain data (ie a string with a length greater than 0). If there is no data, then I want to the parameter to act as if it was absent.

I'm fine if my implementation (the last commit of this PR) isn't merged -- it's easy enough to define per project.

jplock: I agree with @joschi that we can probably use the Jersey Bean Validations once https://github.com/dropwizard/dropwizard/pull/842 is merged in ( @joschi can you comment on that one?).  I think this PR is fine for now, I'll merge it.

BenRomberg:
Thank you guys so much, this is exactly what I was just looking for. I just googled "queryparam optional empty string" and didn't even expect a dropwizard specific solution, but guess what, it was the first search result =)

PR #993, opened 2015-04-15
ConstraintViolations context determines HTTP status

nickbabcock:
Due to #842, validation opens up new opportunities and by happenstance, a new bug. Now validations can be done on resource query parameters, return values, etc, and this, in my opinion is a good thing.

The pitfall I ran into is that the current `ConstraintViolationExceptionMapper` with the new bean validation in place, does not follow the [jersey documentation], which states that:

> 500 (Internal Server Error): If the exception was thrown while validating a method return type.
>
> 400 (Bad Request): Otherwise.

This pull request attempts to fix the current behavior, which will return 422 on return values and `*Params`. The violations in `*Params` now return a 400 and a return value violation is a 500.

There is an issue with this implementation: I'm uncomfortable that the only sane implementation for checking if constraint violation was fired on a parameter is by regex. Please advise.

EDIT: I found a better way than regex from the [Jersey code]

[jersey documentation]: https://jersey.java.net/documentation/latest/bean-validation.html#d0e13792
[Jersey code]: https://github.com/jersey/jersey/blob/54e27f31205de49004331db2d714c1be9a52bb76/ext/bean-validation/src/main/java/org/glassfish/jersey/server/validation/internal/ValidationHelper.java

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapper.java
@@ -1,19 +1,46 @@
 package io.dropwizard.jersey.validation;
 
+import javax.validation.ConstraintViolation;
 import javax.validation.ConstraintViolationException;
+import javax.validation.ElementKind;
+import javax.validation.Path.Node;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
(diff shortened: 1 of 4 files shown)


jplock:
Thanks for catching this!  I think this change looks fine to me.  Any thoughts from your side @joschi ?

nickbabcock:
Just note, this isn't ready to merge this until I comment here with the swapped out implementation, which'll be sometime tomorrow :smile:

nickbabcock:
Okay, code is ready for review :sunglasses: 

/cc @joschi

jplock:
I think this looks good.  Thanks for the contribution!

PR #996, opened 2015-04-16
allow the opportunity to customize your logging factory

carlo-rtr:


jplock:
Should we also try and address https://github.com/dropwizard/dropwizard/pull/987 or keep that as a separate change?

carlo-rtr:
That's a good point, let's get that sorted. I'll add my 2 cents.

Btw, I probably should say that there was much debate on whether this should be configurable. See https://github.com/dropwizard/dropwizard/pull/567

My feeling is that we should only maintain the default logger. Making it flexible doesn't feel like it's going to add much over head on our maintenance. If someone is really hell bent on using xml configuration, they'll find a way.

arteam:
After #987 that issue need to rebased.

Also we need to think abount the need of bootstrapping logging in `Application`, if it's configured externally.

carlo-rtr:
I've rebased and added some new code.

* documented the bootstrap logging code
* perform bootstrap logging init in constructor of ```Application``` in favor of ```static``` block

carlo-rtr:
@arteam I missed the #987 comment. I'll wait until your code is back on master, then I'll pull it down and push the rebased version.

carlo-rtr:
@jplock @arteam should be all set for review

jplock:
LGTM!  I'll let @arteam review as well.

dropwizard-logging/src/main/java/io/dropwizard/logging/BootstrapLogging.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Utililty class to configure logging before the dropwizard yml
+ * configuration has been read, parsed, and the provided logging
+ * strategy has been applied.
+ * <p/>
+ * N.B. The methods in this class have run once semantics,
+ * multiple calls are impodent
```
arteam: Nit: s/impodent/idempotent, if this word was meant.

dropwizard-logging/src/main/java/io/dropwizard/logging/BootstrapLogging.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class BootstrapLogging {
+
+    @GuardedBy("boostrappingLock")
+    private static boolean bootstrapped = false;
+    private static Lock bootstrappingLock = new ReentrantLock();
```
arteam: Could we me make the lock final? I think it's better to always lock on final fields as a safety measure.
Moreover, this field is not mutated.

dropwizard-logging/src/main/java/io/dropwizard/logging/BootstrapLogging.java
```diff
@@ -0,0 +1,67 @@
...
+ * Utililty class to configure logging before the dropwizard yml
+ * configuration has been read, parsed, and the provided logging
+ * strategy has been applied.
+ * <p/>
+ * N.B. The methods in this class have run once semantics,
+ * multiple calls are impodent
+ */
+public class BootstrapLogging {
+
+    @GuardedBy("boostrappingLock")
```
arteam: Nit: s/boostrappingLock/bootstrappingLock

dropwizard-logging/src/main/java/io/dropwizard/logging/BootstrapLogging.java
```diff
@@ -0,0 +1,67 @@
...
+    public static void bootstrap() {
+        bootstrap(Level.WARN);
+    }
+
+    public static void bootstrap(Level level) {
+        LoggingUtil.hijackJDKLogging();
+
+        bootstrappingLock.lock();
+        try {
+            if (!bootstrapped) {
```
arteam: Maybe better to revert the condition? If logging is bootstrapped, we can return early.
Advantage is a reduction of a big *if* block with identation.

dropwizard-logging/src/main/java/io/dropwizard/logging/LoggingUtil.java
```diff
@@ -0,0 +1,71 @@
...
+import io.dropwizard.util.Duration;
+import org.slf4j.ILoggerFactory;
+import org.slf4j.LoggerFactory;
+import org.slf4j.bridge.SLF4JBridgeHandler;
+
+import javax.annotation.concurrent.GuardedBy;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class LoggingUtil {
```
arteam: I wonder should we make this class a package protected or public... I mean, do we want its methods be visible for user logger factories or should they be closed as a part of internal Dropwizard logging mechanism? I'm more apt to to the latter. What is your opinion on this?

dropwizard-logging/src/main/java/io/dropwizard/logging/LoggingUtil.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final Duration LOGGER_CONTEXT_AWAITING_TIMEOUT = Duration.seconds(10);
+    private static final Duration LOGGER_CONTEXT_AWAITING_SLEEP_TIME = Duration.milliseconds(100);
+
+    @GuardedBy("julHijackingLock")
+    private static boolean julHijacked = false;
+    private static Lock julHijackingLock = new ReentrantLock();
```
arteam: The lock could be final, because we don't mutate it.

arteam:
Apart from some nitpicks, this PR looks accurate to me.

dropwizard-logging/src/main/java/io/dropwizard/logging/LoggingUtil.java
carlo-rtr: I was thinking it should be public because so that people implementing their own logging factories have a way to get the LoggerContext in a thread safe way and have a utility for hijacking jdk logging.

carlo-rtr:
I really wish I could spell :)

carlo-rtr:
@arteam fixed the typos, made the locks final, and inverted the condition. Thanks for looking

PR #998, opened 2015-04-17
Reference correct Grizzly server in testing docs

nickbabcock:
`GrizzlyWebTestContainerFactory` is what is recommended to allow `Context` injections but the docs use `GrizzlyTestContainerFactory`.


jplock:
Thanks!

Issue #892, opened 2015-02-23
Potential XSS vulnerability in AbstractParam


prb:
This came up when a customer ran a scan against a Dropwizard app.  In `AbstractParam` ([here](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParam.java#L69)), the input parameter value is sent back unescaped.  While useful, that could enable an attacker to insert scripting into a page that displays those error messages.

prb:
This issue was patched in the application in question by overriding `AbstractParam#errorMessage` in the subclasses in question, but the easy fix in general is simply not to return the input (verbatim) in the response.

joschi:
It's already escaped as a JSON string but which escaping rules would you additionally apply? SQL? JavaScript? HTML? Python?

I don't think that's a problem you can solve in a generic way.

prb:
@joschi I agree that it's a problem without a generic solution, other than perhaps to insist (probably futilely) that UI developers write clean, safe code...  In this case, the security scanning code is attempting to assess the risk of an XSS issue in that it inspects the JSON bodies for the interaction rather than the content actually delivered to the UI layer.

One option would just be to have a global configuration parameter for all `AbstractParam` and descendants that controls whether input is included with error messages.

carlo-rtr:
I don't see a good way of configuring whether abstract params error messages include the original input other than some static side effect state. 

I'm wondering if it makes sense to show the clients the input they send? Would existing users feel we broke backwards compatibility without it? My guess is no. Perhaps we should just send them the general class of error without the input, then we debug log the input & error

jplock:
@carlo-rtr I'm +1 on that approach

jplock:
Fixed in #1022

PR #1001, opened 2015-04-19
fix #953 Possible reflective cross site scripting in ConstraintViolations

### LINKED ISSUE #953: Possible reflective cross site scripting in the io.dropwizard.validation.ConstraintViolations
We have come across a problem in when doing JSR-303 validation, due to dropwizard `ConstraintViolations` adding the InvalidValue back in the message.

The following request payload,
```json
{
   "url": "/address/<script>alert(2)</script>"
}
```
will result in the following error
```json
{
"errors": [
"url must be a valid URL (was /address/<script>alert(2)</script>)"
]
}
```

We don't want the "was" part.

Based on the investigation we found that it is being appended in the `ConstraintViolations.format()` method.

```java
public static <T> String format(ConstraintViolation<T> v) {
    if (v.getConstraintDescriptor().getAnnotation() instanceof ValidationMethod) {
        final ImmutableList<Path.Node> nodes = ImmutableList.copyOf(v.getPropertyPath());
        final ImmutableList<Path.Node> usefulNodes = nodes.subList(0, nodes.size() - 1);
        final String msg = v.getMessage().startsWith(".") ? "%s%s" : "%s %s";
        return String.format(msg,
                             Joiner.on('.').join(usefulNodes),
                             v.getMessage()).trim();
    } else {
        return String.format("%s %s (was %s)",
                             v.getPropertyPath(),
                             v.getMessage(),
                             v.getInvalidValue());
    }
}
```

Is there any way that we can avoid the `invalidValue` from being attached to the message?

If not can we have a feature switch for that?

WarFox:
* ConstraintViolations should not print the invalid value as it may cause reflective-cross-site scripting issue
* Fixed all unit tests that expect the (was {invalidValue}) in the violation message.

### DIFF
--- a/dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationFactoryTest.java
+++ b/dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationFactoryTest.java
@@ -318,7 +318,7 @@ public class ConfigurationFactoryTest {
                 assertThat(e.getMessage())
                         .endsWith(String.format(
                                 "factory-test-invalid.yml has an error:%n" +
-                                        "  * name must match \"[\\w]+[\\s]+[\\w]+([\\s][\\w]+)?\" (was Boop)%n"));
+                                        "  * name must match \"[\\w]+[\\s]+[\\w]+([\\s][\\w]+)?\"%n"));
             }
         }
     }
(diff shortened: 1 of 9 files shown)


piefel:
Since the `format()` call is also used for the trace logging in `JacksonMessageBodyProvider` where printing the invalid value is harmless but helpful, I wonder whether there couldn’t be a second format function that would exhibit the classic behaviour.

carlo-rtr:
I think we should add a debug log with the invalid value, so it can be debugged on the server side.

I think this would also address @piefel concern without having to maintain a flag or separate methods.

@joschi you commented on #892 which is related, what do you think about this?

joschi:
I'm all for it. Let's just remove the original, tainted input from the error messages.

WarFox:
I agree with @carlo-rtr about the `debug` logs, that could be of help to some projects.

Should we add the `debug` logs?

carlo-rtr:
I think so :)

PR #1002, opened 2015-04-20
Gather all endpoints and sort before logging them

softarn:
The list of endpoints logged when starting Dropwizard becomes hard to read when it grows larger. This will sort all endpoints before logging them.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -3,18 +3,15 @@ package io.dropwizard.jersey;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.jersey2.InstrumentedResourceMethodApplicationListener;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Joiner;
+import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Ordering;
 import com.google.common.collect.Sets;
 import io.dropwizard.jersey.caching.CacheControlledResponseFeature;
 import io.dropwizard.jersey.guava.OptionalMessageBodyWriter;
 import io.dropwizard.jersey.guava.OptionalParamFeature;
 import io.dropwizard.jersey.sessions.SessionFactoryProvider;
-import org.glassfish.jersey.message.GZipEncoder;
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.server.ServerProperties;
-import org.glassfish.jersey.server.filter.EncodingFilter;
 import org.glassfish.jersey.server.model.Resource;
 import org.glassfish.jersey.server.model.ResourceMethod;
 import org.glassfish.jersey.server.monitoring.ApplicationEvent;
@@ -28,6 +25,9 @@ import org.slf4j.LoggerFactory;
 import javax.ws.rs.Path;
 import javax.ws.rs.ext.Provider;
 import java.lang.annotation.Annotation;
+import java.io.Serializable;
+import java.util.Comparator;
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
@@ -110,6 +110,8 @@ public class DropwizardResourceConfig extends ResourceConfig {
 
     public String getEndpointsInfo() {
         final StringBuilder msg = new StringBuilder(1024);
+        final List<EndpointLogLine> endpointLogLines = Lists.newArrayList();
+
         msg.append("The following paths were found for the configured resources:");
         msg.append(NEWLINE).append(NEWLINE);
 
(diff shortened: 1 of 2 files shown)


dropwizard-jersey/src/test/java/io/dropwizard/jersey/DropwizardResourceConfigTest.java
```diff
@@ -74,6 +76,25 @@ public void logsEndpoints() {
...
     }
 
+    @Test
+    public void logsEndpointsSorted() {
+        rc.register(DummyResource.class);
+        rc.register(TestResource2.class);
+        rc.register(TestResource.class);
+        rc.register(ImplementingResource.class);
+
+        System.out.println(rc.getEndpointsInfo());
```
klette: Should probably be removed :)

softarn: thanks, updated

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -184,8 +187,47 @@ private String normalizePath(String basePath, String path) {
...
+            return basePath;
+        }
+
+        @Override
+        public String toString() {
+            return String.format("    %-7s %s (%s)", httpMethod, basePath, klass.getCanonicalName());
+        }
+    }
+
+    private static class EndpointComparator implements Comparator<EndpointLogLine>, Serializable {
```
arteam: Could we just make `EndpointLogLine` to implement `Comparable<EndpointLogLine>`?
If it so, we won't need to specify a comparator explicitly.

softarn: So much boiler code for equals, hashcode. That's why I chose the
comparator, want me to switch anyway?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -184,8 +187,47 @@ private String normalizePath(String basePath, String path) {
...
+        private static final long serialVersionUID = 1L;
+
+        @Override
+        public int compare(EndpointLogLine endpointA, EndpointLogLine endpointB) {
+            int basePathComparison = endpointA.getBasePath().compareTo(endpointB.getBasePath());
```
arteam: What do you think about using Guava's [ComparisonChain](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ComparisonChain.java)?
Something like this:
````java
return ComparisonChain.start()
                    .compare(this.basePath, other.basePath)
                    .compare(this.httpMethod, other.httpMethod)
                    .result();
````

Comment (softarn) [inline dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java]: Thanks for tip, didn't know about it I'll check it out!

File: dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java [hunk 4]
```diff
@@ -110,6 +112,8 @@ public void setUrlPattern(String urlPattern) {
 
     public String getEndpointsInfo() {
         final StringBuilder msg = new StringBuilder(1024);
+        final Set<EndpointLogLine> endpointLogLines = Sets.newTreeSet(new EndpointComparator());
```
arteam: Do we really need a `TreeSet`? 
Basically we don't have duplicates and don't care about speed of lookup. The only thing that we need, it's a sorted collection. We could use an ordinary `ArrayList` with some expected size to avoid resizing and then sort it. The collection is relatively small, so sorting should be fast. `TreeSet` has a very big memory footprint and it's not cache-friendly. I would not use it, if we have alternatives.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -139,7 +146,7 @@ public String getEndpointsInfo() {
      */
     private static class EndpointLogger {
         private final String rootPath;
-        private final List<String> endpoints = Lists.newArrayList();
+        private final List<EndpointLogLine> endpointLogLines = Lists.newArrayList();
```
arteam: Could we pass a link to the root collection, rather then creating a new collection for every resource?
Now this collecton is created only to be captured.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -184,8 +187,47 @@ private String normalizePath(String basePath, String path) {
...
+        private final String basePath;
+        private final Class<?> klass;
+
+        public EndpointLogLine(String httpMethod, String basePath, Class<?> klass) {
+            this.basePath = basePath;
+            this.klass = klass;
+            this.httpMethod = httpMethod;
+        }
+
+        public String getHttpMethod() {
```
arteam: We could access these fields directly from the comparator, so these getters are not needed.

dropwizard-jersey/src/test/java/io/dropwizard/jersey/DropwizardResourceConfigTest.java
```diff
@@ -74,6 +76,23 @@ public void logsEndpoints() {
...
 
+    @Test
+    public void logsEndpointsSorted() {
+        rc.register(DummyResource.class);
+        rc.register(TestResource2.class);
+        rc.register(TestResource.class);
+        rc.register(ImplementingResource.class);
+
+        assertThat(rc.getEndpointsInfo())
+                .contains("    GET     / (io.dropwizard.jersey.dummy.DummyResource)\n"
```
arteam: This test will fail on Windows, because it has different line breaks (\r\n rather than \n).
Could we wrap an expected string in `String.format` or use a regular expression?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
arteam: That's a reasonable argument. Let's leave it then.

arteam:
@softarn, thanks for the contribution!
I've added some notes about where the PR could be improved, but in overall it looks great.

softarn:
@arteam I've updated the commit. Thank you for the comments!

arteam:
Thanks, @softarn, for your efforts on that.

PR #1004, opened 2015-04-21
Defer DNS resolution in GraphiteReporter

### LINKED ISSUE #1003: GraphiteReporterFactory defer IP resolution at connect time
Can we please add support for https://github.com/dropwizard/metrics/issues/606, 

i.e:
public ScheduledReporter build(MetricRegistry registry) {
     final Graphite graphite = new Graphite(**host**, **port**);
     ...
}

arteam:
`InetSockedAddres` performs DNS resolution during construction time, which is not great if the Graphite server moves to a different IP-address.

The solution is to initialize `Graphite` gateway with a hostname and port. In such configuration it will resolve an address just before connecting to the server.

Resolve #1003

### DIFF
--- a/dropwizard-configuration/pom.xml
+++ b/dropwizard-configuration/pom.xml
@@ -30,7 +30,6 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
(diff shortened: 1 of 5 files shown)


joschi:
LGTM! :+1:

Issue #1006, opened 2015-04-22
JerseyClientBuilder should accept any combo of Environment, ExecutorService, or ObjectMapper


roccato:
JerseyClientBuilder currently requires the use of either a combo of ExecutorService and ObjectMapper or an Environment from which it creates a managed ExecutorService and the objectMapper from the Environment.

That means if you want to use a separate ObjectMapper for your client you have to go through the work of creating a an ExecutorService on your own. Not hard, but unnecessary.

It would be nice if we could build a JerseyClient with any combo of ExecutorService, ObjectMapper, or Environment.

PR #1011, opened 2015-04-23
Fix issue #1006: JerseyClientBuilder should accept any combo of Environme...

roccato:
...nt, ExecutorService, or ObjectMapper
https://github.com/dropwizard/dropwizard/issues/1006


jplock:
I think this looks good, thanks for the contribution!

PR #1012, opened 2015-04-25
Run tests in parallel

arteam:
This set of changes add facility run Dropwizard tests in parallel. 
The motivation is to perform the build faster and make the test environment more unstable,
forcing tests not to depend on the execution order.

It solves 4 issues, that curently block us from doing so:
* A [bug](http://jira.qos.ch/browse/SLF4J-167) in sl4fj. The facade doesn't behaves correctly when several threads try to acquire an uninitialized logger context. 
* Integration tests for Jersey resources use the same port, so it's not possible to run them in parallel
* Tests for overriding system properties for Dropwizard applications clash with each other and behave unreliebly when run in parallel.
* There is a [test](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-logging/src/test/java/io/dropwizard/logging/PrefixedRootCauseFirstThrowableProxyConverterTest.java#L86), that checks that the final frame of a stack trace is a main method. It's not so, when
the test is run not from the default `JUnitRunner`, but from a custom thread.

A solution to the first problem is to extract acquiring the logger context into a separate method and create a spin-loop. A thread that didn't get the initialized context is spinned in the loop with timeout while it doesn't get one. It should be noted that we can't resolve the problem just by marking the method as `synchronized`. `LoggerFactory` doesn't correctly publish own state (access to the state is not synchronized and it's represented as a non-volatile integer). So even if the logger context has been initialized by the first arrived thread, the next threads can see a stale state and subsequently get a stub. So the condition -  returned context is always correct, if it's initialized without a race - doesn't hold.

A solution to the second issue is just to make sure that all tests on Jersey resources run on random ports.

A solution to the third issue is to add to `DropwizardAppRule` facility to specify a custom system property prefix. If tests use different prefixes, they don't impact on each other and can be run on parallel. 

A solution to the fourth issue is just to remove the test.

Closes #987

### DIFF
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/ChainedAuthProviderTest.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/ChainedAuthProviderTest.java
@@ -14,6 +14,7 @@ import org.glassfish.jersey.servlet.ServletProperties;
 import org.glassfish.jersey.test.DeploymentContext;
 import org.glassfish.jersey.test.JerseyTest;
 import org.glassfish.jersey.test.ServletDeploymentContext;
+import org.glassfish.jersey.test.TestProperties;
 import org.glassfish.jersey.test.grizzly.GrizzlyWebTestContainerFactory;
 import org.glassfish.jersey.test.spi.TestContainerException;
 import org.glassfish.jersey.test.spi.TestContainerFactory;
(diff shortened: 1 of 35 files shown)


carlo-rtr:
LGTM. Thank you

carlo-rtr:
The build broke. Rolling back.

https://travis-ci.org/dropwizard/dropwizard/jobs/60241683

arteam:
Well, that's awkward. I can't reproduce the error on my local machine and the build fails on Travis randomly. It looks like I was bitten by my own initiatives :) Well, at least I reached one goal: the test environment is certainly has become more fragile. I will try to investigate a source of the error.

carlo-rtr:
You have to crack eggs when making an omelette :)

One of the issues is that Application has that static block for bootstrapping logging, that's why ``DropwizardAppRuleResetConfigOverrideTest`` is failing. I believe my https://github.com/dropwizard/dropwizard/pull/996 will eventually fix that.

PR #1014, opened 2015-04-25
Add a benchmark for DropwizardResourceConfig

arteam:
The benchmark measures how long it takes to gather information about application endpoints for representing it to the user during the application startup.

The goal is to see is there were any perfomance problems in #1002.


carlo-rtr:
LGTM

PR #1015, opened 2015-04-26
Add facility to specify a custom MetricRegistry in Bootstrap

### LINKED ISSUE #956: Allow to customize MetricsRegistry
I'd like to customize MetricRegistry. It's used to create all the timers and histograms, but I don't like the defaults.
In particular I don't like the fact that MetricBuilder for timers uses default constructor for Timer, which in turn builds a timer backed by ExponentiallyDecayingReservoir.

Why is it bad? Because ExponentiallyDecayingReservoir;
1) "[default setting] heavily biases the reservoir to the past 5 minutes of measurements". I report every 30 seconds, the remaining 4.5 min just skews the results.
2) when nothing is reported the values for percentiles stay on the same level indefinitely instead of dropping to 0 or null.

There are better options, e.g. HdrHistogramResetOnSnapshotReservoir, but right now it's not possible to plug it in.

arteam:
* Add a setter for the metric registry in `Bootstrap`, so users can provide own customized instance during initialization.
* Move system metrics registration into a separate method from the constructor.
* Invoke the method after `Application#initialize`, where users can customize `Bootstrap`.

Resolve #956

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/Application.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/Application.java
@@ -73,10 +73,15 @@ public abstract class Application<T extends Configuration> {
         bootstrap.addCommand(new ServerCommand<>(this));
         bootstrap.addCommand(new CheckCommand<>(this));
         initialize(bootstrap);
+        // Should by called after initialize to give an opportunity to set a custom metric registry
+        bootstrap.registerMetrics();
+
         final Cli cli = new Cli(new JarLocation(getClass()), bootstrap, System.out, System.err);
         if (!cli.run(arguments)) {
             // only exit if there's an error running the command
             System.exit(1);
         }
     }
+
+
 }
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
@@ -47,12 +47,14 @@ public class Bootstrap<T extends Configuration> {
     private final List<Bundle> bundles;
     private final List<ConfiguredBundle<? super T>> configuredBundles;
     private final List<Command> commands;
-    private final MetricRegistry metricRegistry;
 
+    private MetricRegistry metricRegistry;
     private ConfigurationSourceProvider configurationSourceProvider;
     private ClassLoader classLoader;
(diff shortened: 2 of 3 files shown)


arteam:
@dropwizard/committers, It would be great, if someone reviewed this patch.

carlo-rtr:
@arteam sorry it took so long to look. Thanks for looking at this.

The only thing I'm concerned with is the reflection. It's easier to find references when it's statically referenced. 

What is the exact security concern with making registerMetrics public? If Application needs to call it, then I think it should be part of the ``Bootstrap`` interface.

joschi:
I also have some trouble with using reflection in this case. The problem is that it would fail only at runtime if things broke.

Maybe `registerMetrics()` should be called implicitly when providing an alternative `MetricRegistry`, i. e. rename `setMetricRegistry()` to `registerMetricRegistry()` or `initializeMetricRegistry()` and call `registerMetrics()` in that method. The can still be a (package protected) `setMetricRegistry()` if that would be required for tests.

vanDonselaar:
Another option could be to use the [SharedMetricRegistries](https://github.com/dropwizard/metrics/blob/master/metrics-core/src/main/java/com/codahale/metrics/SharedMetricRegistries.java). It's not as nice as proper dependency injection (it's a singleton class), but at least it doesn't require reflection.

arteam:
> Maybe registerMetrics() should be called implicitly when providing an alternative MetricRegistry, > > i.e. rename setMetricRegistry() to registerMetricRegistry() or initializeMetricRegistry() and call 
> registerMetrics() in that method. The can still be a (package protected) setMetricRegistry() if that 
> would be required for tests.
The problem is that we need to stop the default JMX reporter and create a new one, which adds complexity. 

> What is the exact security concern with making registerMetrics public? If Application needs to call > it, then I think it should be part of the Bootstrap interface. 
I thought that registering system metrics should be transparent to user code. Basically users should
not see this method, because it must be called only by `Application`. For example, someone can call this method in `Application#initialize` and break the startup process.

Anyway, I am more fond to make `registerMetrics` public. But I think we can make it idempotent,
by returning sucessfully if the system metrics are already registered.

carlo-rtr:
Thanks @arteam

PR #1016, opened 2015-04-26
Improve DropwizardApacheConnectorTest

arteam:
* Expect `NoRouteException` as well as `ConnectionTimeoutException`.

If a local network is configured to reject request to non-routable addresses, then this exception is returned by JVM. We should handle such situations.
* Add a health check, so the application doesn't yell on us in logs.
* Add code style and documentation improvements.

### DIFF
--- a/dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
+++ b/dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
@@ -1,11 +1,12 @@
 package io.dropwizard.client;
 
 import com.codahale.metrics.MetricRegistry;
-import com.google.common.io.Resources;
+import com.codahale.metrics.health.HealthCheck;
 import io.dropwizard.Application;
 import io.dropwizard.Configuration;
 import io.dropwizard.jackson.Jackson;
 import io.dropwizard.setup.Environment;
+import io.dropwizard.testing.ResourceHelpers;
 import io.dropwizard.testing.junit.DropwizardAppRule;
 import io.dropwizard.util.Duration;
 import org.apache.http.HttpStatus;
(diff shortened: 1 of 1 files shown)


dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
```diff
@@ -35,27 +37,31 @@
...
 
 public class DropwizardApacheConnectorTest {
-    private static final int SLEEP_TIME_IN_MILLIS = 500;
-    private static final int DEFAULT_CONNECT_TIMEOUT_IN_MILLIS = 200;
-    private static final int ERROR_MARGIN_IN_MILLIS = 300;
-    private static final int INCREASE_IN_MILLIS = 100;
-    private static final String NON_ROUTABLE_IP_ADDRESS = "10.255.255.1"; 
-    public static final URI URI_THAT_TIMESOUT_ON_CONNECTION_ATTEMPT = URI.create("http://" + NON_ROUTABLE_IP_ADDRESS);
+
+    private static final int SLEEP_TIME = 500;
```
jplock: I think it was more clear when these had `_IN_MILLIS` in their names so that people would know what the value represented.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
```diff
@@ -138,7 +138,7 @@ public void when_configuration_overridden_to_disallow_redirects_temporary_redire
         @GET
         @Path("/long_running")
         public String getWithSleep() throws InterruptedException {
-            TimeUnit.MILLISECONDS.sleep(SLEEP_TIME_IN_MILLIS);
```
jplock: I personally prefer this approach instead of `Thread.sleep()` so we know how long the sleep value represents.

dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
carlo-rtr: I agree with @jplock on the variable name and sleeping methodology

jplock:
This looks good to me.  Thanks for fixing those minor issues.

arteam:
@jplock @carlo-rtr Thanks for the review! I addressed your feedback and rebased the commit.

arteam:
Oops, looks like I was a bit of late. Thanks for merging.

PR #1017, opened 2015-04-26
Upgrade to Apache HTTP Client 4.4.1

arteam:
There were 3 releases of Apache Http Client with many bug fixes since the last update.

Release notes: 
* [4.3.6](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310360&version=12327489)
* [4.4](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310360&version=12324872)
* [4.4.1](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310360&version=12329394)

Interesting fixes for us:
* [Stale connections in pool cause NoHttpResponseException](https://issues.apache.org/jira/browse/HTTPCLIENT-1610) 
* [Add option to create a connection eviction thread for a non-shared connection manager](https://issues.apache.org/jira/browse/HTTPCLIENT-1583)
* [SSLContextBuilder does not preserve natural order of Trust / Key managers ](https://issues.apache.org/jira/browse/HTTPCLIENT-1585)
* [SSLConnectionSocketFactory not respecting enabled protocols](https://issues.apache.org/jira/browse/HTTPCLIENT-1595)
* [Connect timeout ignored during SSL connection setup](https://issues.apache.org/jira/browse/HTTPCLIENT-1589)

### DIFF
--- a/dropwizard-client/pom.xml
+++ b/dropwizard-client/pom.xml
@@ -23,6 +23,17 @@
             <version>${jersey.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.4.1</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
             <groupId>io.dropwizard.metrics</groupId>
             <artifactId>metrics-httpclient</artifactId>
             <version>${metrics3.version}</version>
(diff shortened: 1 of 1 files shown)


jplock:
Should we instead update the apache client in metrics and use that dependency?

arteam:
I think, Dropwizard Metrics makes only maintenance releases for 3.1.* branch. All active development happens in 4.0.0 branch. The release of 4.0.0 is far away, but we have a more fast cycle of releases.
Furthermore, I think it's not right that a library for instrumentation dictates a version of
a primary dependency of the module.
But nothing stop us from opening a pull request in Dropwizard Metrics as well!

jplock:
I agree, I'm just worried about us maintaining our own version and excluding the dependency and having it possibly break things.

carlo-rtr:
I've submitted a PR against metrics. my 2cents, I think we can run the risk & upgrade.
https://github.com/dropwizard/metrics/pull/794

jplock:
Ok I'll merge it

PR #1020, opened 2015-04-27
Change register object method to take varargs in Jersey Environment

ayushvora:


jplock:
The tests are failing and I'm not sure this will work as intended.

PR #1022, opened 2015-04-27
exclude orig input value from param validation results, limit XSS vulner...

carlo-rtr:
...ability

https://github.com/dropwizard/dropwizard/issues/892


jplock:
instead of `Param`, what do you think about spelling it out as `Parameter` or `Value` instead?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/BooleanParam.java
```diff
@@ -12,8 +12,8 @@ public BooleanParam(String input) {
     }
 
     @Override
-    protected String errorMessage(String input, Exception e) {
-        return '"' + input + "\" must be \"true\" or \"false\".";
+    protected String errorMessage(Exception e) {
+        return "Param must be \"true\" or \"false\".";
```
jplock: here is where i mean

carlo-rtr:
@jplock good feedback. updated

PR #1023, opened 2015-04-29
@Auth principal injection

carlo-rtr:
Hopefully this eases the burden on the transition


jplock:
This looks good to me

PR #1025, opened 2015-05-01
Doc update: Generic checks when creating AuthFilter

yunspace:
Added `<User, Authenticator>` when calling `AuthFilter.Builder`


carlo-rtr:
Thanks for updating the docs. I'm thinking it should be this:

```   
BasicCredentialAuthFilter.Builder<User, ExampleAuthenticator> builder = 
            new BasicCredentialAuthFilter.Builder<>();
```

carlo-rtr:
actually, disregard that. looks good

PR #1027, opened 2015-05-01
Add docs on constraint violations

nickbabcock:
Documents some of the changes implemented in #993


arteam:
Looks great. Thanks again for taking your time to improve the documenation. 
We really appreciate that.

PR #1028, opened 2015-05-01
Extract constraint status for public use

nickbabcock:
Useful for those who want to derive their own `ExceptionMapper<ConstraintViolationException>` (for instance, an exception mapper that returns JSON containing constraint violations)

### DIFF
--- a/dropwizard-validation/src/test/java/io/dropwizard/validation/ConstraintViolationsTest.java
+++ b/dropwizard-validation/src/test/java/io/dropwizard/validation/ConstraintViolationsTest.java
@@ -1,0 +1,94 @@
+package io.dropwizard.validation;
+
+import org.junit.Test;
+
+import javax.validation.ConstraintViolation;
+import javax.validation.Valid;
+import javax.validation.Validation;
+import javax.validation.Validator;
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+import javax.validation.executable.ExecutableValidator;
+import java.util.Set;
+
+import static org.apache.commons.lang3.reflect.MethodUtils.getAccessibleMethod;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SuppressWarnings({"UnusedDeclaration"})
+public class ConstraintViolationsTest {
+
+    private final Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+    private final ExecutableValidator execValidator = validator.forExecutables();
+
+    @Max(3)
+    public Integer theIntTest(@Min(2) Integer x) {
+        return x + 1;
+    }
+
+    @Valid
+    public ConstraintPerson thePersonTest(@Valid ConstraintPerson person) {
+        person.setName(null);
+        return person;
+    }
+
+    @Test
+    public void invalidIntReturnIs500() {
+        Set<ConstraintViolation<ConstraintViolationsTest>> violations =
+                execValidator.validateReturnValue(
+                        this,
+                        getAccessibleMethod(getClass(), "theIntTest", Integer.class),
+                        4 // the return value
+                );
+
+        assertThat(violations).hasSize(1);
+        assertThat(ConstraintViolations.determineStatus(violations)).isEqualTo(500);
+    }
+
+    @Test
+    public void invalidIntParameterIs400() {
+        Set<ConstraintViolation<ConstraintViolationsTest>> violations =
+                execValidator.validateParameters(
+                        this,
+                        getAccessibleMethod(getClass(), "theIntTest", Integer.class),
+                        new Object[]{1} // the parameter value
+                );
+
+        assertThat(violations).hasSize(1);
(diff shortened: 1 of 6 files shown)


jplock:
Do you want to add a test around this method specifically?

nickbabcock:
Most definitely, I'll chime back in on Monday :smile:

nickbabcock:
Ok tests are in, a review would be great. I also added more in-depth tests for #993, which shows that messages returned because of constraint violations leave a lot wanting.

dropwizard-validation/src/test/java/io/dropwizard/validation/ConstraintViolationsTest.java
```diff
@@ -0,0 +1,108 @@
...
+                execValidator.validateReturnValue(
+                        new ConstraintViolationsTest(),
+                        ConstraintViolationsTest.class.getMethod("theIntTest", Integer.class),
+                        4 // the return value
+                );
+
+        assertThat(violations).hasSize(1);
+
+        Set<ConstraintViolation<?>> copy = ConstraintViolations.copyOf(violations);
+        assertThat(ConstraintViolations.determineStatus(copy)).isEqualTo(500);
```
arteam: Could we change the signature of `ConstraintViolations.determineStatus`  from
````java
public static int determineStatus(Set<ConstraintViolation<?>> violations)
````
to 
````java
public static <T> int determineStatus(Set<ConstraintViolation<T>> violations)
````

Doing so, we can avoid making a copy.

File: dropwizard-validation/src/test/java/io/dropwizard/validation/ConstraintViolationsTest.java [hunk 2]
```diff
@@ -0,0 +1,108 @@
...
+import javax.validation.constraints.Min;
+import javax.validation.executable.ExecutableValidator;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SuppressWarnings({"UnusedDeclaration"})
+public class ConstraintViolationsTest {
+
+    private final Validator validator = Validation
```
arteam: Could we use a shorten version `Validation.buildDefaultValidatorFactory().getValidator()`?
The Hibernate provider will be loaded dynamically and we don't test validation of `Optional` values.

dropwizard-validation/src/test/java/io/dropwizard/validation/ConstraintViolationsTest.java
```diff
@@ -0,0 +1,108 @@
...
+        person.setName(null);
+        return person;
+    }
+
+    @Test
+    public void invalidIntReturnIs500() throws NoSuchMethodException {
+        Set<ConstraintViolation<ConstraintViolationsTest>> violations =
+                execValidator.validateReturnValue(
+                        new ConstraintViolationsTest(),
+                        ConstraintViolationsTest.class.getMethod("theIntTest", Integer.class),
```
arteam: We could add `commons-lang3` as a test dependency to the module and use static methods of `MethodUtils`.
For example, `getAccessibleMethod(getClass(), "theIntTest", Integer.class)`. Benefits - no checked exceptions.

dropwizard-validation/src/test/java/io/dropwizard/validation/ConstraintViolationsTest.java
```diff
@@ -0,0 +1,108 @@
...
+    public ConstraintPerson thePersonTest(@Valid ConstraintPerson person) {
+        person.setName(null);
+        return person;
+    }
+
+    @Test
+    public void invalidIntReturnIs500() throws NoSuchMethodException {
+        Set<ConstraintViolation<ConstraintViolationsTest>> violations =
+                execValidator.validateReturnValue(
+                        new ConstraintViolationsTest(),
```
arteam: Will just passing `this` work?

arteam:
I've added some notes during breif review. Maybe @jplock will also add something.

dropwizard-validation/src/test/java/io/dropwizard/validation/ConstraintViolationsTest.java
nickbabcock: Would that work? A `ConstraintViolationException` `.getConstraintViolations()` returns a `Set<ConstraintViolation<?>>`, and that would make `determineStatus` too narrow  to use. Please advise.

Implemented your other suggestions.

arteam: You are right, it's noo narrow. We need to make the generic type to extend `ConstraintViolation<?>`. Wildcard then can capture  `ConstraintViolation<?>` and `ConstraintViolation<ConstraintViolationsTest>`.
````java
public static <T extends ConstraintViolation<?>> int determineStatus(Set<T> violations)
````

nickbabcock:
Thanks for the review, all comments were acted upon. Let me know if you find anything more.

jplock:
I think this looks great, thanks for the contribution and for @arteam for following up!

PR #1030, opened 2015-05-03
Add facility to plug-in custom DB connection pools

arteam:
Dropwizard by default is shipped with the Tomcat JDBC Connection pool. It's mostly fine, but has some disadvantages (performance, limited metrics, handling network failures). It would be great if we had facility to plug-in other DB pools into Dropwizard. For example, Brett Wooldridge's
wonderful [HikariCP](https://github.com/brettwooldridge/HikariCP) or [BoneCP](http://jolbox.com/).

Of cource, nothing stops us from implementing a factory, that creates a `ManagedDataSource`. The problem is that `dropwizard-jdb`i, `dropwizard-hibernate` and` dropwizard-migrations` modules require `DataSourceFactory` for the configuration. `DataSourceFactory` is closely tied with the Tomcat pool. So if we use an another pool, we are forced to write own factories for JDBI, Hibernate, etc...

A solution is to provide an interface that abstracts a configuration of a connection pool. Database manipulation modules will be linked with the interface, but not with `DataSourceFactory`. A custom pool factory just need to implement this interface, and it could be plug-in as a Tomcat pool
replacement.

If we writed Dropwizard from the scratch we would move the Tomcat pool to a separate module from `dropwizard-db`. But now we have many users and it would be a nightmare for them to migrate to a new implementation.

So we are forced to retain the Tomcat JDBC pool in `dropwizard-db`. New pool users should be advised to exclude it manually in the Maven configuration.

### DIFF
--- a/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
+++ b/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
@@ -250,7 +250,7 @@ import java.util.concurrent.TimeUnit;
  *     </tr>
  * </table>
  */
-public class DataSourceFactory {
+public class DataSourceFactory implements PooledDataSourceFactory {
     @SuppressWarnings("UnusedDeclaration")
     public enum TransactionIsolation {
         NONE(Connection.TRANSACTION_NONE),
@@ -355,6 +355,7 @@ public class DataSourceFactory {
     private Duration validationInterval = Duration.seconds(30);
 
     @JsonProperty
+    @Override
     public boolean isAutoCommentsEnabled() {
         return autoCommentsEnabled;
     }
@@ -365,6 +366,7 @@ public class DataSourceFactory {
     }
 
     @JsonProperty
+    @Override
     public String getDriverClass() {
         return driverClass;
     }
@@ -405,6 +407,7 @@ public class DataSourceFactory {
     }
 
     @JsonProperty
+    @Override
     public Map<String, String> getProperties() {
         return properties;
     }
@@ -426,6 +429,11 @@ public class DataSourceFactory {
 
     @JsonProperty
     public String getValidationQuery() {
+        return validationQuery;
+    }
+
+    @Override
+    public String getHealthCheckValidationQuery() {
         return validationQuery;
     }
 
@@ -689,11 +697,24 @@ public class DataSourceFactory {
         return Optional.fromNullable(validationQueryTimeout);
     }
 
+    @Override
+    public Optional<Duration> getHealthCheckValidationTimeout() {
+        return Optional.fromNullable(validationQueryTimeout);
+    }
+
     @JsonProperty
     public void setValidationQueryTimeout(Duration validationQueryTimeout) {
         this.validationQueryTimeout = validationQueryTimeout;
     }
 
+    @Override
+    public void asSingleConnectionPool() {
+        minSize = 1;
+        maxSize = 1;
+        initialSize = 1;
+    }
+
+    @Override
     public ManagedDataSource build(MetricRegistry metricRegistry, String name) {
         final Properties properties = new Properties();
         for (Map.Entry<String, String> property : this.properties.entrySet()) {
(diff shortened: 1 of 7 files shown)


dropwizard-db/src/main/java/io/dropwizard/db/AbstractDataSourceFactory.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Gets the configuration properties for ORM tools.
+     *
+     * @return configuration properties as a map
+     */
+    Map<String, String> getProperties();
+
+    /**
+     * Gets the timeout for awaiting a response from the database.
```
joschi: Superfluous period in this line.

dropwizard-db/src/main/java/io/dropwizard/db/AbstractDataSourceFactory.java
```diff
@@ -0,0 +1,66 @@
...
+     * during health checks.
+     *
+     * @return the timeout as {@code Duration}
+     */
+    Optional<Duration> getHealthCheckValidationTimeout();
+
+    /**
+     * Gets the SQL query, that used for checking the database availability.
```
joschi: Should probably read "Returns the SQL query which is being used for the database connection health check.".

dropwizard-db/src/main/java/io/dropwizard/db/AbstractDataSourceFactory.java
```diff
@@ -0,0 +1,66 @@
...
+import com.codahale.metrics.MetricRegistry;
+import com.google.common.base.Optional;
+import io.dropwizard.util.Duration;
+
+import java.util.Map;
+
+/**
+ * Interface of an abstract JDBC connection pool
+ */
+public interface AbstractDataSourceFactory {
```
joschi: In my opinion the prefix "Abstract" is not really a good choice for use in interfaces. Maybe "PooledDateSourceFactory" is more fitting in this case.

arteam: Sure. I didn't have anything in mind which ends in 'able', so decided to go with 'Abstract'.
`PooledDataSourceFactory` also looks good to me. Naming things is hard...

dropwizard-db/src/main/java/io/dropwizard/db/AbstractDataSourceFactory.java
arteam: Thanks. That definitely looks more literate.

arteam:
Ping. Does any additional work needed in this PR?

dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
```diff
@@ -429,6 +429,10 @@ public String getValidationQuery() {
         return validationQuery;
     }
 
+    public String getHealthCheckValidationQuery() {
```
carlo-rtr: I think we should use `@Override` on the implementations of `PooledDataSourceFactory`

carlo-rtr:
One small nitpick, 'HelloWorldConfiguration` in the example should declare the variables as `PooledDataSourceFactory`. 

Otherwise this PR looks good to me. Thanks for the hard work!

dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
arteam: Thanks, forgot about that.

arteam:
> One small nitpick, 'HelloWorldConfigurationin the example should declare the variables 
> asPooledDataSourceFactory`.

I don't think it's farsightedly, because in this case we will need to enable [polymorphic-configuration](http://dropwizard.github.io/dropwizard/0.9.0-SNAPSHOT/docs/manual/configuration.html#polymorphic-configuration) for `PooledDataSourceFactory`. And this looks like overkill, because nobody probably will want to change pool implementations in the configuration.

stantonk:
@arteam is there documentation on how to use this? I don't quite see how:

```
Of cource, nothing stops us from implementing a factory, that creates a ManagedDataSource. The problem is that dropwizard-jdbi, dropwizard-hibernate anddropwizard-migrations modules require DataSourceFactory for the configuration. DataSourceFactory is closely tied with the Tomcat pool. So if we use an another pool, we are forced to write own factories for JDBI, Hibernate, etc...
```

has been sidestepped.

stantonk:
Seems like you're still left needing something like https://github.com/DeloitteDigitalUK/dropwizard-hikaricp, which as far as I can tell, still misses out on the nice features that `DBIFactory` provides.

PR #1031, opened 2015-05-04
Don't override the JVM timezone in a Hibernate integration test

arteam:
HSQLDB correctly retrieves a timestamp from the DB in the JVM timezone, Jadira converts it to DateTime and then Jackson serializes it to JSON in UTC.

So all assertions in UTC will pass regardless of the JVM timezone.


joschi:
LGTM.

PR #1032, opened 2015-05-04
Run tests in parallel

arteam:
This is rebased and squashed version of #1012 

I can't reproduce errors locally and on Travis. I've run the build a few dozens times and it passed.
I don't want to make the build flaky, but I have a feeling that's something were wrong with CI test environment.

If the build will nevertheless fail, this PR could remain opened, while we don't find the root cause.


arteam:
Ok, for some reason the build fails only when I run it as a pull-request to the project and it doesn't when I run it from my fork on Travis.  Will investigate the stack trace...

arteam:
Figured out changing logging levels needs to be protected, because `LoggerContext#loggerContextListenerList` is not thread-safe.

carlo-rtr:
LGTM!

PR #1034, opened 2015-05-07
Auth documentation & minor refactors

carlo-rtr:
* Use Filter instead of Handler
* java docs for ChainedAuthFilter 
* limit type warning suppression in ChainedAuthFilterTest


jplock:
LGTM!

PR #1039, opened 2015-05-08
Descriptive constraint violation messages

nickbabcock:
This pull requests attempts to return user friendly error messages when constraints are violated either in the response or the request.

So instead of error messages of

```json
{"errors":["blazer.arg0.name may not be empty"]}
```
and

```json
{"errors":["blaze.<return value> length must be between 0 and 3"]}
```
more user friendly error messages are returned

```json
{"errors":["query param name may not be empty"]}
```
```json
{"errors":["server response length must be between 0 and 3"]}
```

This pull request isn't done, so don't merge (hardly any documentation and I didn't completely mull over edge cases). As it happens, I'll be off the grid for the next couple weeks, but I wanted to get some eyes on the code (hence the pull request), but if someone is extremely enthusiastic they can carry this pull request the rest of the way, else I'll pick off where I left off and incorporate any suggestions in the pull request.

The code isn't fast, concise, documented, or complete -- but it passes all the test cases I wrote for it :smile:

Opinions wanted!

### DIFF
--- a/dropwizard-benchmarks/src/main/java/io/dropwizard/benchmarks/jersey/ConstraintViolationBenchmark.java
+++ b/dropwizard-benchmarks/src/main/java/io/dropwizard/benchmarks/jersey/ConstraintViolationBenchmark.java
@@ -1,0 +1,84 @@
+package io.dropwizard.benchmarks.jersey;
+
+import io.dropwizard.jersey.validation.ConstraintMessage;
+import org.hibernate.validator.constraints.NotEmpty;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import javax.validation.ConstraintViolation;
+import javax.validation.Valid;
+import javax.validation.Validation;
+import javax.validation.Validator;
+import javax.validation.executable.ExecutableValidator;
+import javax.ws.rs.HeaderParam;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+import static org.apache.commons.lang3.reflect.MethodUtils.getAccessibleMethod;
+
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+public class ConstraintViolationBenchmark {
+
+    public static class Resource {
+        public String paramFunc(@HeaderParam("cheese") @NotEmpty String secretSauce) {
+            return secretSauce;
+        }
+
+        public String objectFunc(@Valid Foo foo) {
+            return foo.toString();
+        }
+    }
+
+    public static class Foo {
+        @NotEmpty
+        private String bar;
+    }
+
+    private ConstraintViolation<ConstraintViolationBenchmark.Resource> paramViolation;
+    private ConstraintViolation<ConstraintViolationBenchmark.Resource> objViolation;
+
+    @Setup
+    public void prepare() {
+        final Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+        final ExecutableValidator execValidator = validator.forExecutables();
+
+        Set<ConstraintViolation<ConstraintViolationBenchmark.Resource>> paramViolations =
+            execValidator.validateParameters(
+                new Resource(),
+                getAccessibleMethod(ConstraintViolationBenchmark.Resource.class, "paramFunc", String.class),
+                new Object[]{""} // the parameter value
+            );
+        paramViolation = paramViolations.iterator().next();
+
+        Set<ConstraintViolation<ConstraintViolationBenchmark.Resource>> objViolations =
+            execValidator.validateParameters(
+                new Resource(),
+                getAccessibleMethod(ConstraintViolationBenchmark.Resource.class, "objectFunc", Foo.class),
+                new Object[]{new Foo()} // the parameter value
+            );
+        objViolation = objViolations.iterator().next();
+    }
+
+    @Benchmark
+    public String paramViolation() {
+        return ConstraintMessage.getMessage(paramViolation);
+    }
+
+    @Benchmark
+    public String objViolation() {
+        return ConstraintMessage.getMessage(objViolation);
+    }
+
+    public static void main(String[] args) throws Exception {
+        new Runner(new OptionsBuilder()
+                .include(ConstraintViolationBenchmark.class.getSimpleName())
+                .forks(1)
+                .warmupIterations(5)
+                .measurementIterations(5)
+                .build())
+                .run();
+    }
+}
--- a/dropwizard-validation/src/main/java/io/dropwizard/validation/ConstraintViolations.java
+++ b/dropwizard-validation/src/main/java/io/dropwizard/validation/ConstraintViolations.java
@@ -11,19 +11,23 @@ import javax.validation.Path;
 import java.util.Set;
 
 public class ConstraintViolations {
+    private static final Joiner DOT_JOINER = Joiner.on('.');
+
     private ConstraintViolations() { /* singleton */ }
 
     public static <T> String format(ConstraintViolation<T> v) {
         if (v.getConstraintDescriptor().getAnnotation() instanceof ValidationMethod) {
-            final ImmutableList<Path.Node> nodes = ImmutableList.copyOf(v.getPropertyPath());
-            final ImmutableList<Path.Node> usefulNodes = nodes.subList(0, nodes.size() - 1);
-            final String msg = v.getMessage().startsWith(".") ? "%s%s" : "%s %s";
-            return String.format(msg,
-                                 Joiner.on('.').join(usefulNodes),
-                                 v.getMessage()).trim();
+            return validationMethodFormatted(v);
         } else {
             return String.format("%s %s", v.getPropertyPath(), v.getMessage());
         }
+    }
+
+    public static <T> String validationMethodFormatted(ConstraintViolation<T> v) {
+        final ImmutableList<Path.Node> nodes = ImmutableList.copyOf(v.getPropertyPath());
+        String usefulNodes = DOT_JOINER.join(nodes.subList(0, nodes.size() - 1));
+        String msg = usefulNodes + (v.getMessage().startsWith(".") ? "" : " ") + v.getMessage();
+        return msg.trim();
     }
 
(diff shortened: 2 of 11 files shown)


arteam:
I hacked a little on your first versoin of `ConstraintViolationExceptionMapper`
````java
@Provider
public class ConstraintViolationExceptionMapper implements ExceptionMapper<ConstraintViolationException> {

    @Override
    public Response toResponse(ConstraintViolationException exception) {
        final ImmutableList<String> errors = FluentIterable.from(exception.getConstraintViolations())
                .transform(new Function<ConstraintViolation<?>, String>() {
                    @Override
                    public String apply(ConstraintViolation<?> v) {
                        return constraintMessage(v);
                    }
                }).toList();
        return Response.status(ConstraintViolations.determineStatus(exception.getConstraintViolations()))
                .entity(new ValidationErrorMessage(errors))
                .build();
    }

    private String constraintMessage(ConstraintViolation<?> v) {
        final Optional<String> returnValueName = getMethodReturnValueName(v);
        if (returnValueName.isPresent()) {
            final String name = isValidationMethod(v) ?
                    StringUtils.substringBeforeLast(returnValueName.get(), ".") : returnValueName.get();
            return name + " " + v.getMessage();
        } else {
            if (isValidationMethod(v)) {
                return ConstraintViolations.validationMethodFormatted(v);
            } else {
                final String name = getMemberName(v).or(v.getPropertyPath().toString());
                return name + " " + v.getMessage();
            }
        }
    }

    /**
     * Gets the method return value name, if the violation is raised in it
     */
    private Optional<String> getMethodReturnValueName(ConstraintViolation<?> violation) {
        int returnValueNames = -1;

        final StringBuilder result = new StringBuilder("server response");
        for (Path.Node node : violation.getPropertyPath()) {
            if (node.getKind().equals(ElementKind.RETURN_VALUE)) {
                returnValueNames = 0;
            } else if (returnValueNames >= 0) {
                result.append(returnValueNames++ == 0 ? " " : ".").append(node);
            }
        }

        return returnValueNames >= 0 ? Optional.of(result.toString()) : Optional.<String>absent();
    }

    /**
     * Gets a method parameter (or a parameter field) name, if the violation raised in it.
     */
    private Optional<String> getMemberName(ConstraintViolation<?> violation) {
        final Path propertyPath = violation.getPropertyPath();
        final int size = Iterables.size(propertyPath);
        if (size < 2) {
            return Optional.absent();
        }

        final NodeImpl parent = (NodeImpl) Iterables.get(propertyPath, size - 2);
        final NodeImpl member = (NodeImpl) Iterables.get(propertyPath, size - 1);

        final Class<?> resourceClass = violation.getLeafBean().getClass();
        switch (parent.getKind()) {
            case PARAMETER:
                final Field field = FieldUtils.getDeclaredField(resourceClass, member.getName(), true);
                return getMemberName(field.getDeclaredAnnotations());
            case METHOD:
                final Method method = MethodUtils.getAccessibleMethod(
                        resourceClass, parent.getName(), asArray(parent.getParameterTypes()));
                return getMemberName(method.getParameterAnnotations()[member.getParameterIndex()]);
            default:
                return Optional.absent();
        }
    }

    /**
     * Derives member's name and type from it's annotations
     */
    private Optional<String> getMemberName(Annotation[] memberAnnotations) {
        for (Annotation a : memberAnnotations) {
            if (a instanceof QueryParam) {
                return Optional.of("query param " + ((QueryParam) a).value());
            } else if (a instanceof PathParam) {
                return Optional.of("path param " + ((PathParam) a).value());
            } else if (a instanceof HeaderParam) {
                return Optional.of("header " + ((HeaderParam) a).value());
            } else if (a instanceof CookieParam) {
                return Optional.of("cookie " + ((CookieParam) a).value());
            } else if (a instanceof FormParam) {
                return Optional.of("form field " + ((FormParam) a).value());
            }
        }

        return Optional.absent();
    }

    private static Class<?>[] asArray(List<Class<?>> list) {
        return list.toArray(new Class<?>[list.size()]);
    }

    private static boolean isValidationMethod(ConstraintViolation<?> v) {
        return v.getConstraintDescriptor().getAnnotation() instanceof ValidationMethod;
    }
}
````

I think, it's a little more concise and readable. If you like it, you can derive your work from this
version. Or just borrow some concepts.

I didn't touch the tests. Will try to review them in a short time and give some feedback also.

**Comment by nickbabcock** [comment]
Awesome, thanks for the alternative implementation while I was away, I'll start basing work on it and reply back once I'm satisfied.

**Comment by nickbabcock** [comment]
- Largely adapted @arteam's implementation
- Added benchmarks so someone can come along and optimize if needed

If everything looks good, I can start writing up some prose documentation.

File: dropwizard-benchmarks/src/main/java/io/dropwizard/benchmarks/jersey/ConstraintViolationBenchmark.java [hunk 1]
```diff
@@ -0,0 +1,83 @@
...
+
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+public class ConstraintViolationBenchmark {
+
+    private ConstraintViolation<ConstraintViolationBenchmark> paramViolation;
+    private ConstraintViolation<ConstraintViolationBenchmark> objViolation;
+
+    public String paramFunc(@HeaderParam("cheese") @NotEmpty String secretSauce) {
```
arteam: Maybe we could we create a static class `Resource` and move both these methods there?
It will separate the benchmark from the test data.

dropwizard-benchmarks/src/main/java/io/dropwizard/benchmarks/jersey/ConstraintViolationBenchmark.java
```diff
@@ -0,0 +1,83 @@
...
+        return secretSauce;
+    }
+
+    public String objectFunc(@Valid Foo foo) {
+        return foo.toString();
+    }
+
+    public class Foo {
```
arteam: The class could be static

dropwizard-benchmarks/src/main/java/io/dropwizard/benchmarks/jersey/ConstraintViolationBenchmark.java
```diff
@@ -0,0 +1,83 @@
...
+        final Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+        final ExecutableValidator execValidator = validator.forExecutables();
+
+        Set<ConstraintViolation<ConstraintViolationBenchmark>> paramViolations =
+            execValidator.validateParameters(
+                this,
+                getAccessibleMethod(getClass(), "paramFunc", String.class),
+                new Object[]{""} // the parameter value
+            );
+        paramViolation = Iterators.getOnlyElement(paramViolations.iterator());
```
arteam: I think it's safe just to call `iterator.next`, because we know that the violation is exist.

dropwizard-benchmarks/src/main/java/io/dropwizard/benchmarks/jersey/ConstraintViolationBenchmark.java
```diff
@@ -0,0 +1,83 @@
...
+
+        Set<ConstraintViolation<ConstraintViolationBenchmark>> objViolations =
+                execValidator.validateParameters(
+                        this,
+                        getAccessibleMethod(getClass(), "objectFunc", Foo.class),
+                        new Object[]{new Foo()} // the parameter value
+                );
+        objViolation = Iterators.getOnlyElement(objViolations.iterator());
```
arteam: Ditto

dropwizard-validation/src/main/java/io/dropwizard/validation/ConstraintViolations.java
```diff
@@ -15,17 +15,21 @@ private ConstraintViolations() { /* singleton */ }
...
-            return String.format(msg,
-                                 Joiner.on('.').join(usefulNodes),
-                                 v.getMessage()).trim();
+            return validationMethodFormatted(v);
         } else {
             return String.format("%s %s", v.getPropertyPath(), v.getMessage());
         }
     }
 
+    public static <T> String validationMethodFormatted(ConstraintViolation<T> v) {
```
arteam: We probably can reuse `Joiner` and don't use `String.format`.
````java
public static <T> String validationMethodFormatted(ConstraintViolation<T> v) {
        final ImmutableList<Path.Node> nodes = ImmutableList.copyOf(v.getPropertyPath());
        String usefulNodes = JOINER.join(nodes.subList(0, nodes.size() - 1));
        String msg = usefulNodes + (v.getMessage().startsWith(".") ? "" : " ") + v.getMessage();
        return msg.trim();
    }
````
Kindly provided by you microbenchmark gives this on my machine:
````
Benchmark                                    Mode  Cnt     Score      Error  Units
ConstraintViolationBenchmark.objViolation    avgt    5   711.064 ±   22.911  ns/op
ConstraintViolationBenchmark.paramViolation  avgt    5  7400.960 ± 2268.245  ns/op

Benchmark                                    Mode  Cnt     Score     Error  Units
ConstraintViolationBenchmark.objViolation    avgt    5   693.016 ±  10.078  ns/op
ConstraintViolationBenchmark.paramViolation  avgt    5  6585.522 ± 815.698  ns/op
````

**Comment by arteam** [comment]
Looks good to me! Apart from some remarks for benchmarks and one micro optimization, I think this change can be prepared for merging.

By the way, I guess it's the first PR where we try to use `dropwizard-benchmarks` for improving an implementation :)

**Comment by nickbabcock** [comment]
Suggestions taken, thanks for the review.

**Comment by nickbabcock** [comment]
Maybe it is inappropriate for this pull request but I couldn't resist adding in a simple caching mechanism that increases performance by up to 250x when there is a hit. 

```
Benchmark                                    Mode  Cnt     Score     Error  Units
ConstraintViolationBenchmark.objViolation    avgt    5  3416.563 ± 214.989  ns/op
ConstraintViolationBenchmark.paramViolation  avgt    5  5605.761 ±  90.774  ns/op

Benchmark                                    Mode  Cnt   Score   Error  Units
ConstraintViolationBenchmark.objViolation    avgt    5  22.430 ± 0.359  ns/op
ConstraintViolationBenchmark.paramViolation  avgt    5  22.643 ± 1.384  ns/op
```

arteam:
Thanks! Merged to master.

UPD. I missed your comment about caching and merged this too early. Fixed up this a little in a21ce52

PR #1042, opened 2015-05-12
Use double quotes for string literal and fix underline to be the same length as header text

markreddy:
Two quick fixes to the Dropwizard Core manual page


arteam:
Thanks for a sharp look on the docs.

PR #1043, opened 2015-05-12
Fix underline to be the same length as header text and various spelling mistakes

markreddy:
Some quick fixes to the release notes.


jplock:
Thanks!

PR #1051, opened 2015-05-15
Use a custom mustache resolver then a custom factory

arteam:
A more canonical way to provide a custom resource loader is through delegation, rather then inheritance.

There is the `MustacheResolver` interface, that gives ability to provide own strategy of loading resources and pass it to the `DefaultMustacheFactory` class in the constructor.

Also this removes throwing `MustacheException`, because internally Mustache already handles the case when a returned reader is null and throws a `MustacheNotFoundException` with the same semantic.


jplock:
LGTM!

PR #1052, opened 2015-05-16
Do not swallow exception cause when parsing JSON.

richmidwinter:
If an error occurs processing JSON by default we just receive a message in the response indicating the dropwizard was "Unable to process JSON". This should be logged at least as a warning so a developer/sysadmin can see the cause.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -48,7 +48,7 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
         /*
          * Otherwise, it's those pesky users.
          */
-        LOGGER.debug("Unable to process JSON", exception);
+        LOGGER.warn("Unable to process JSON", exception);
         final ErrorMessage errorMessage = new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(),
                 "Unable to process JSON", showDetails ? message : null);
         return Response.status(Response.Status.BAD_REQUEST)


arteam:
I tend to agree. Wrong user request is certainly not a debug event.

j4n1:
IMHO, wrong user request is definitively not a warning, it’s a debug event indeed. 
We will have logs bloated with exceptions caused by a bug in client. Really can’t see how this, client error, can be a warning on the server side, but maybe I’m overseeing something.

joschi:
I also see this critical. A buggy or malicious client might use this to swamp the logs with warning messages including stack traces and cause a denial of service for other clients.

Using Dropwizard's LogConfigurationTask the log level for the relevant class(es) can be set to DEBUG even during runtime in order to investigate failed requests. But generally those are not a warning event (just like @j4n1 described).

Any other opinions in @dropwizard/committers?

arteam:
I have to admit, that I didn't take into account possibility of DoS.

Anyway, I think it's important to log client errors as warnings, because it's an early indicator of misbehaving clients. I don't think that's a debug information, because, in my opinion, it's not about investigating programming errors and more about monitoring wrong input data, that clients are trying to send to the server.

I guess I rushed a little to merge this without discussion. Let's hear what other maintainers and users have to say and we can revert this if I will be in minority.

arteam:
Well, no feedback during last week. Anyway, I decided to rollback this change because of its controversy. Arguments against it seems convincing for me.

carlo-rtr:
I know I'm late, but I'll share my thoughts. I think the debug level is appropriate. Clients sending bad requests are normal, and should be considered insignificant from the server perspective. I think this answer gives a good way to think about it: http://stackoverflow.com/questions/7839565/logging-levels-logback-rule-of-thumb-to-assign-log-levels/8021737#8021737

PR #1055, opened 2015-05-18
Remove deprecated Apache HTTP client API

arteam:
* Use the default cookie policy as a replacement for deprecated "best-match"
* Set 'setValidateAfterInactivity' instead of deprecated 'staleConnectionCheckEnabled'
* Use 'NoopHostnameVerifier' as a replacement for 'AllowAllHostnameVerifier'


joschi:
LGTM.

PR #1061, opened 2015-05-20
Manage Apache HTTP and Jersey HTTP clients

arteam:
In the [documentation](http://www.dropwizard.io/manual/client.html#apache-httpclient) we say that we create managed instances of Apache HTTP client. But in the fact, it's not true. 

To follow this contract, we should tie the lifecyle of the client to the lifecycle of the server, if the user specified the environment in the builder. The rationale is the same for Jersey HTTP client.

If the user has already created own managed object for the client  - it's not an issue, because the close method of the HTTP client is idempotent.


jplock:
LGTM, I'll see if anyone else would like to review as well

carlo-rtr:
Looks good. One minor nitpick, should we add a test in `JerseyClientBuilderTest`

arteam:
I've added a check in `DropwizardApacheConnectorTest#tearDown` method. 
We can't create a mock Jersey client, so we check that the client is closed in a real environment.

PR #1062, opened 2015-05-21
Use configuration class from Application object

Toilal:
In some case, configuration class retrieved by MigrationsBundle is invalid (in my case, when MigrationsBundle is constructed from an abstract Application class where configuration is still a Generic Type).

This fix the issue by retrieving the configuration class from Application object.


arteam:
Looks reasonable. Thanks.

PR #1064, opened 2015-05-22
[Documentation] Specify boolean false parameter in EnvironmentVariableSubstitutor constructor

jnwelzel:
In order for the app startup not to fail if an env var defined in the yaml config file cannot be found I had to add a boolean false parameter to the EnvironmentVariableSubstitutor constructor


arteam:
Thanks for pointing this. Non-strict mode is actually seems like a better choice for an example.

PR #1065, opened 2015-05-22
Fix Checkstyle warnings

arteam:
Fix various Checkstyle warnings about whitespaces, formatting, modifiers visibility and common code style.

I ran [Checkstyle](http://checkstyle.sourceforge.net/) on the project other day and fixed a couple of simple warnings that were reported. 
I don't think there is much value in this change, but it doesn't break anything, makes the code more easy to read, and, in some degree, reduces the technical dept. We can fix all these style warnings in one step and don't worry about them in future.


joschi:
Looks good (and doesn't functionally change anything as far as I see).

Issue #1066, opened 2015-05-23
Missing @Path annotation causes odd NPE in 0.8.1


baank:
Referring to: https://github.com/federecio/dropwizard-swagger/issues/52

If the @Path annotation is missing from a Resource a NPE is thrown in this class:
io.dropwizard.jersey.DropwizardResourceConfig$EndpointLogger.normalizePath

Instead of throwing random, odd NPE it should clean show an ERROR log message stating that a path is missing from a resource ideally indicating which one.

thiagomoretto:
@baank AFAIK `@Path` annotation on class level is not necessarily required, so you can add `@Path` just on method level. Instead of a better message, we could fix this.

So, I've just send a pull request (https://github.com/dropwizard/dropwizard/pull/1067) with a possible fix, maybe we could discuss there if this issue was really fixed or we should do something yet.

PR #1067, opened 2015-05-24
Fix NPE when resources have the @Path annotation just on method level

thiagomoretto:
Referring this https://github.com/dropwizard/dropwizard/issues/1066 and consequently this: https://github.com/federecio/dropwizard-swagger/issues/52

This pull request fixes this referred NPE. Basically, what I've found is that resources without @Path annotation on class level the method `org.glassfish.jersey.server.model.Resource#getPath` returns null, so, causing the issue in `EndpointLogger#normalizePath`. I've just wrote a defensive code to avoid this NPE.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -45,7 +45,6 @@ public class DropwizardResourceConfig extends ResourceConfig {
         this(true, null);
     }
 
-    @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


joschi:
Thanks!

Lael-Costa:
The NPE also shows up if you're missing @GET/@POST/... annotation on a resource method that has a @Produces annotation. This is in Dropwizard 1.0.0. Thanks!

evnm:
@Lael-Costa, could you file a separate issue and provide an example stacktrace from the NPE you're seeing, please? I don't see how it could be the exact NPE tracked by this issue, since the specific case is now guarded by a null check.

PR #1068, opened 2015-05-24
Include information about the metrics annotations

mcgin:
The current documentation gives no indication of what the metrics annotations do, added some rudimentary documentation to explain the basics


arteam:
Looks good. Thanks for taking your time to write this.

PR #1070, opened 2015-05-24
Migration tests

arteam:
It's a first attempt to write tests for the `dropwizard-migrations` module.

* First of all, this change moves [CloseableLiquibaseTest](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-example/src/test/java/io/dropwizard/migrations/CloseableLiquibaseTest.java) from `dropwizard-example` to `dropwizard-migrations`. It seems to me that it was placed there for mistake. 

* It sets the test directory, Logback testing configuration and test dependencies for the module.
* It adds a test for the [DbMigrate](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-migrations/src/main/java/io/dropwizard/migrations/DbMigrateCommand.java) command. The following use cases are tested:
  * Migration of pending migrations to the DB.
  * Dry-run mode (printing the pending change set to the console).
  * Migration of limited amount of changes.
  * Help page (description of the command and available arguments)

### DIFF
--- a/dropwizard-migrations/pom.xml
+++ b/dropwizard-migrations/pom.xml
@@ -50,5 +50,13 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>net.jcip</groupId>
+            <artifactId>jcip-annotations</artifactId>
+            <version>1.0</version>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 </project>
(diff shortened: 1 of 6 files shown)


arteam:
Argh, the `Liquibase` class is actually not thread safe... #1032 doesn't give me a rest.

arteam:
I excluded the tests from running from parallel, so now they run correctly as a parth of the build.

carlo-rtr:
LGTM. Minor nitpick: the db user & pass could be constants.

PR #1072, opened 2015-05-24
Log context paths in `SimpleServerFactory` during startup

### LINKED ISSUE #1058: Warn about changes to applicationContextPath
I am trying out Dropwizard and really like it so far, but I was nearly tempted to give up during the Getting Started tutorial. I have an application that binds to port 8080, so I modified my example Dropwizard app to use another port, copying the simple server factory config from https://github.com/dropwizard/dropwizard/blob/master/dropwizard-example/example.yml for convenience. Unbeknownst to me, this changed the applicationContextPath from `"/"` to `"/application"`, and I spent several hours wondering why my application seemed to be running without error but I kept getting 404 errors when I went to the paths that the tutorial suggested, which also matched what the application log had printed out. Perhaps there could be a warning of this effect in the sample YAML file. Or better yet, perhaps the application log could print out the `applicationContextPath` on startup, and show it in the paths of resources that are registered. There's an inconsistency when the log says that the path for a resource is `GET    /hello` but the server returns a 404 status unless you request `GET    /application/hello`.

arteam:
Add logging of application and admin context paths during server startup. `DefaultServerFactory` already does [that](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-core/src/main/java/io/dropwizard/server/DefaultServerFactory.java#L160).

It should help to avoid confusion for users who use this configuration for the first time and expect that application resources are served from the root path.

Resolve #1058

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/server/SimpleServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/server/SimpleServerFactory.java
@@ -12,6 +12,8 @@ import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.util.thread.ThreadPool;
 import org.hibernate.validator.constraints.NotEmpty;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.validation.Valid;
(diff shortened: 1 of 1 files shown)


glennmcallister:
LGTM

PR #1073, opened 2015-05-24
Update an example of using query parameters

arteam:
~~* Use a POST, rather then GET, because GET requests should return some values, no 204 return code~~
~~* Change name of the parameter to from `date` to `value`~~
* Use `Optional` to handle an omitted parameter
* Return a passed date, so Jersey doesn't warn us about 204 error

### DIFF
--- a/dropwizard-example/src/main/java/com/example/helloworld/resources/HelloWorldResource.java
+++ b/dropwizard-example/src/main/java/com/example/helloworld/resources/HelloWorldResource.java
@@ -39,10 +39,18 @@ public class HelloWorldResource {
     public void receiveHello(@Valid Saying saying) {
         LOGGER.info("Received a saying: {}", saying);
     }
-    
+
     @GET
     @Path("/date")
-    public void receiveDate(@QueryParam("date") DateTimeParam dateTimeParam) {
-        LOGGER.info("Received a saying: {}", dateTimeParam.get());
+    @Produces(MediaType.TEXT_PLAIN)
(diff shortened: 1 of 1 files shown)


joschi:
I think the GET verb was perfectly fine here as the request doesn't change any state and is idempotent.

Instead of returning an empty 204 response we could return the provided date as a string (with media type `text/plain`).

dropwizard-example/src/main/java/com/example/helloworld/resources/HelloWorldResource.java
```diff
@@ -39,10 +39,14 @@ public Saying sayHello(@QueryParam("name") Optional<String> name) {
...
         LOGGER.info("Received a saying: {}", saying);
     }
-    
-    @GET
+
+    @POST
     @Path("/date")
-    public void receiveDate(@QueryParam("date") DateTimeParam dateTimeParam) {
-        LOGGER.info("Received a saying: {}", dateTimeParam.get());
+    public void receiveDate(@QueryParam("value") Optional<DateTimeParam> dateTimeParam) {
```
joschi: Why did you change the query parameter name? "date" was more specific than the very generic "value".

arteam: It seems confusing to me that the path and the query parameter have the same name.
Let's leave it then, if it just my personal perception.

arteam:
Thanks for the review. Please, take a look and let me now if I messed up anything.

joschi:
:+1:

Issue #1074, opened 2015-05-25
Saying class in the tutorial should be moved to the api package


mcgin:
In the Getting Started guide, the "Saying" representation is placed in the core package:
http://www.dropwizard.io/getting-started.html#creating-a-representation-class

However in the Organizing your project section in the manual it states representations should be placed in the api package
http://www.dropwizard.io/manual/core.html#organizing-your-project

I believe Saying should be moved to the api package in the example solution for consistency.

I'm going to start making the necessary changes and submit a pull request, but open to input as to whether this is the correct approach

PR #1075, opened 2015-05-25
Moving the Saying representation to the api package

mcgin:
Addresses https://github.com/dropwizard/dropwizard/issues/1074


arteam:
You are probably right. Thanks for pointing this.

PR #1082, opened 2015-05-27
Only cache present values in the CachingAuthenticator. Fixes #947

### LINKED ISSUE #947: CachingAuthenticator caches Optional.absent()
Currently the CachingAuthenticator loads Optional.absent() into the cache which means that if you had a failed credential for whatever reason then it gets cached as absent() by the underlying authenticator.

Instead I think this should be updated to something like:

            @Override
            public Optional<P> load(K key) throws Exception {
                cacheMisses.mark();

                //Only cache successful operations, not failures
                Optional<P> g = underlying.authenticate(key);
                if (g.isPresent()) {
                    return g;
                }

                throw new AuthenticationException("Failed to load security context into cache");
            }

Or at the very least, it should be an option to not cache these failures.

Or would be prefer that each individual authenticator throw an exception instead of returning Optional.absent() ?

jplock:


dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
```diff
@@ -60,7 +60,12 @@ public CachingAuthenticator(final MetricRegistry metricRegistry,
             @Override
             public Optional<P> load(C key) throws Exception {
                 cacheMisses.mark();
-                return underlying.authenticate(key);
+                final Optional<P> result = underlying.authenticate(key);
+                // only cache present values
+                if (!result.isPresent()) {
+                    throw new AuthenticationException("Failed to load security context into cache");
```
joschi: Won't this exception be thrown (and propagated!) every time invalid credentials are provided?

dropwizard-auth/src/test/java/io/dropwizard/auth/CachingAuthenticatorTest.java
```diff
@@ -104,4 +105,16 @@ public void calculatesCacheStats() throws Exception {
         cached.authenticate("credentials1");
         assertThat(cached.stats().loadCount()).isEqualTo(1);
     }
+
+    @Test
+    public void shouldNotCacheAbsentPrincipals() throws Exception {
+        when(underlying.authenticate(anyString())).thenReturn(Optional.<Principal>absent());
+        try {
+            cached.authenticate("credentials");
+            failBecauseExceptionWasNotThrown(AuthenticationException.class);
```
joschi: Invalid credentials should in fact never cause a `AuthenticationException`. That exception should only be used if the backend is unavailable and the credentials can't be verified.

dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
carlo-rtr: You are right, I missed that.

carlo-rtr:
Maybe we could do something like this: https://github.com/dropwizard/dropwizard/compare/dropwizard:789402598eb0e000b7e5e03091d8750970bc99cc...carlo-rtr:PR1082?expand=1

joschi:
@carlo-rtr I'm not sure that using a `CacheLoader` is sensible at all in this case. After all the assumption with a `CacheLoader` is (https://code.google.com/p/guava-libraries/wiki/CachesExplained):
> Is there some **sensible default function to load or compute a value** associated with a key? If so, you should use a CacheLoader.

Unfortunately there is no sensible default value in our case.

The `CachingAuthenticator#authenticate()` should explicitly add a successfully authenticated principal to the cache instead and omit the failed ones.

carlo-rtr:
@joschi That makes sense

jplock:
So what would you like me to do with this @joschi to address the original issue as described in #947?

carlo-rtr:
My understanding is something like this:

https://github.com/dropwizard/dropwizard/compare/dropwizard:789402598eb0e000b7e5e03091d8750970bc99cc...carlo-rtr:PR1082?expand=1

jplock:
Resubmitted as https://github.com/dropwizard/dropwizard/pull/1084

PR #1084, opened 2015-05-27
Update caching implementation to not cache absent values.

### LINKED ISSUE #947: CachingAuthenticator caches Optional.absent()

jplock:
Fixes #947

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
@@ -5,16 +5,12 @@ import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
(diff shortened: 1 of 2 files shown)


jplock:
@carlo-rtr @joschi how's this?

carlo-rtr:
It looks good to me. Let's let @joschi hit the merge

joschi:
LGTM. :+1:

PR #1085, opened 2015-05-30
Upgrade to Jackson 2.6.0-rc1

arteam:
2.6.0 is a cross-module update for Jackson, so we can remove the mess with the dependencies exclusion.

It's a release candidate, but pretty stable. This upgrade is backward-compatible, so it should not affect Dropwizard users.

### DIFF
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -41,7 +41,7 @@ v0.9.0
 * Upgraded to Freemarker 2.3.22
 * Upgraded to H2 1.4.187
 * Upgraded to Hibernate 4.3.9.Final
-* Upgraded to Jackson 2.5.4
+* Upgraded to Jackson 2.6.0
 * Upgraded to Jersey 2.19
 * Upgraded to Jetty 9.2.12.v20150709
 * Upgraded to Jetty ALPN boot 7.1.3.v20150130
--- a/dropwizard-jackson/pom.xml
+++ b/dropwizard-jackson/pom.xml
@@ -30,51 +30,23 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
-            <version>${jackson.version}</version>
+            <version>${jackson.api.version}</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
-            <exclusions>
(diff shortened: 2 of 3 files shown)


joschi:
I'd rather wait for the GA release of Jackson 2.6.0 than update to the release candidate now.

Additionally I'd keep the exclusions because with Jackson 2.6.1 we will have the same problem again.

arteam:
Ok, let's wait for the GA release. I hope @cowtowncoder will ping us, when it happens.

cowtowncoder:
I think that for actual DW release, a stable version is needed. For master branch of DW, using rc1  (or soon rc2) might not be a bad idea. But I understand hesitation to some degree.
The latest stable version now is 2.5.3.

The intent is to get official release out within couple of weeks -- didn't happen by end of may as I was hoping, but definitely will be during June.

cowtowncoder:
Minor update: rc2 is out, and I will still make rc3 before final release. It looks like GA may slip into early July; if this is problematic, RCs have been rather solid and are probably usable.

joschi:
@cowtowncoder Cool, thanks for the update!

gberberoglu:
jackson 2.6.0 is released! https://github.com/FasterXML/jackson-core/commit/80b8319500e68434ca6da91eb7b9243918cdff02

cowtowncoder:
Actually, onlu core and annotations are; databind will follow starting today. So please wait until monday 21st before update, to get latest versions of all components.

cowtowncoder:
Jackson 2.6.0 is now fully released, finally.

nickbabcock:
Since this is a dependency bump that is backwards compatible, I have no reservations for this being merged for 0.9

@arteam, did you want to update the release notes in this pull request as well?

arteam:
Sure. I've updated the release notes.

About the exclusions: I don't think we should keep them, because with 2.6.x we will have a  different dependency graph. It looks better to me to solve version conflicts on demand, rather than pull the history of exclusions, which could be out of date.  In any case, we have Maven Enforcer plugin to see if we messed up.

cowtowncoder:
Only other suggestion I have is to make sure that every Jackson component DropWizard depends on should be added as a direct dependency, and not rely on transitive dependencies. This because Maven has issues resolving transitive dependencies to latest version. Specifically I have noticed that `jackson-annotations` is sometimes omitted, causing challenges; or sometimes even `jackson-core`. Since minor versions between core components must match, this can lead to odd issues.

nickbabcock:
Ok, merging this. We can figure exclusions as needed in the future.

PR #1087, opened 2015-05-31
Add tests for DbStatusCommand

arteam:
It checks 4 cases:

* Database that's up to date with the change set
* Database that's diverged with the change set
* Verbose output
* Help page

This also moves common code between migration command tests into the `AbstractMigrationTest` class.


dropwizard-migrations/src/test/java/io/dropwizard/migrations/AbstractMigrationTest.java
```diff
@@ -0,0 +1,46 @@
...
+        try {
+            return File.createTempFile("test-example", null).getAbsolutePath();
+        } catch (IOException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        databaseUrl = "jdbc:h2:" + createTempFile();
```
joschi: You could use an in-memory database here.
http://www.h2database.com/html/features.html#in_memory_databases

arteam: Indeed. I will refactor the tests, so they are not tied to a file-based database.

joschi:
LGTM.

Issue #1038, opened 2015-05-08
0.8+ with Jersey 1.x


ghenkes:
I have an existing 3rd party servlet that is using Jersey 1.x that I would like to add in to my Dropwizard app. I’d like to stay current with Dropwizard (0.8+); however it currently uses Jersey 2.x. We have Jersey 1 transitive dependencies of existing projects that we utilize, making it difficult for us to use the latest versions of Dropwiazard (since we can’t uplift all of these dependencies yet).

In order to get this to work I managed the version of dropwizard-jersey to 0.7.1, as it still uses Jersey 1 and made a couple tweaks to dropwizard-core and dropwizard-jersey. Reading https://github.com/dropwizard/dropwizard/issues/934 and [here](https://groups.google.com/forum/#!topic/dropwizard-dev/VInOW_ebiAc) it sounds like the Jersey upgrade has caused other users problems when upgrading as well. Is there interest in having a dropwizard-jersey1 module included so consumers of dropwizard-core can choose which jersey version to use? If so, I'll contribute that and the changes needed in dropwizard-core and dropwizard-jersey.

cchesser:
+1 for offering a compatibility module for consumers to stay current with dropwizard, but allow migration from Jersey 1 to 2 (as this can be quite painful for service assembly owners). It would be nice to achieve this vs. pinning to dropwizard to 0.7.1 for the interim.

Perhaps naming could be dropwizard-jersey-glassfish (2) or dropwizard-jersey-sun (1. Or just keeping dropwizard-jersey and a dropwizard-jersey-compat /  dropwizard-jersey-legacy module.

[Migrating from Jersey 1.x to 2.0](https://jersey.java.net/documentation/latest/migration.html#mig-1.x)

joschi:
To be honest, I don't think this will happen anytime soon if at all.

There are basically a gazillion dependencies on Jersey 2 in Dropwizard, e. g. dropwizard-auth, dropwizard-client, dropwizard-core, dropwizard-jersey (obviously), dropwizard-forms, dropwizard-hibernate, dropwizard-jdbi, and dropwizard-testing all depend on classes provided by Jersey 2. So in the end it would either lead to a complete fork of Dropwizard or a impenetrable pile of dependencies like dropwizard-auth-jersey1/dropwizard-auth-jersey2 etc.

ghenkes:
[...]

PR #1090, opened 2015-06-02
Remove Jersey dependency from dropwizard-core.

ghenkes:
Decouple dropwizard-core from Jersey by providing a JerseyServletContainer in dropwizard-jersey.
This would allow consumers to implement their own dropwizard-jersey using Jersey 1.

See: https://github.com/dropwizard/dropwizard/issues/1038


arteam:
Seems harmless enough. @joshi, what do you think?

carlo-rtr:
Seems fine to me

ghenkes:
@arteam @carlo-rtr I made a small change to JerseyServletContainer to take in a DropwizardResourceConfig rather than a ResourceConfig. I also pulled in the latest changes from dropwizard into my fork, that's why you'll see those 2 merge commits.

arteam:
Applied to master with tweaks in 2a08d3

PR #1100, opened 2015-06-04
update the client docs in the manual

dareonion:
The jersey client example should use `jerseyClient` as its name to be consistent with the [configuration docs](http://www.dropwizard.io/manual/configuration.html#man-configuration-clients-jersey). While I was at it, I also trimmed some trailing whitespace.


arteam:
You have my vote. I applied the change to master with small polishing.
Thanks.

PR #1104, opened 2015-06-06
Respect deserialization of enums using toString()

### LINKED ISSUE #1098: PermissiveEnumDeserializer doesn't support Jackson Enum features
Jackson has two features for Enum handling, `READ_ENUMS_USING_TO_STRING` and `WRITE_ENUMS_USING_TO_STRING`, that indicate that Enums should be de/serialized using the `toString()` value of the Enum, which can be more convenient than adding `@JsonValue` and `@JsonCreator` annotations. These annotations are currently ignored by `PermissiveEnumDeserializer` that's registered by `FuzzyEnumModule`.

I'm a little leery of what that deserializer is doing anyway. I get that it wants to allow different variations of enum constants over the wire, but it seems odd to make it a default. I'd prefer that it be an opt-in feature through an annotation on the Enum type, e.g.:

```
@FuzzyMatch
public enum MyEnum { ... }
```

and the deserializer would only activate for types that had the annotation. If not that, then perhaps an opt-out annotation, to allow the reverse?

arteam:
If Jackson if configured with *READ_ENUMS_USING_TO_STRING* feature, `PermissiveEnumDeserializer` should respect it and let Jackson to correctly deserialize an enum using its `toString` method.

Fix #1098

### DIFF
--- a/dropwizard-jackson/src/test/java/io/dropwizard/jackson/FuzzyEnumModuleTest.java
+++ b/dropwizard-jackson/src/test/java/io/dropwizard/jackson/FuzzyEnumModuleTest.java
@@ -1,6 +1,7 @@
 package io.dropwizard.jackson;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.Before;
@@ -14,8 +15,8 @@ import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
 
 public class FuzzyEnumModuleTest {
     private final ObjectMapper mapper = new ObjectMapper();
-    
-    private enum EnumWithLowercase { lower_case_enum, mixedCaseEnum }
+
+    private enum EnumWithLowercase {lower_case_enum, mixedCaseEnum}
 
     private enum EnumWithCreator {
         TEST;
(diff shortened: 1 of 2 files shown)


dropwizard-jackson/src/test/java/io/dropwizard/jackson/FuzzyEnumModuleTest.java
```diff
@@ -83,17 +101,22 @@ public void failsOnIncorrectValue() throws Exception {
...
     @Test
     public void mapsMixedCaseEnums() throws Exception {
         assertThat(mapper.readValue("\"mixedCaseEnum\"", EnumWithLowercase.class))
                 .isEqualTo(EnumWithLowercase.mixedCaseEnum);
     }
-   
+
+    @Test
+    public void readsEnumsUsingToString() throws Exception {
+        mapper.configure(DeserializationFeature.READ_ENUMS_USING_TO_STRING, true);
```
joschi: The configuration should be changed on a copy of the object mapper. Otherwise different test methods in this class might fail (order isn't guaranteed): [`ObjectMapper#copy()`](https://fasterxml.github.io/jackson-databind/javadoc/2.5/com/fasterxml/jackson/databind/ObjectMapper.html#copy())

arteam: I don't think it's neccesary, because instances of `ObjectMapper` are not shared between the tests. Every test creates own instance of the mapper, so changes in it don't affect other tests.

carlo-rtr: @arteam your right about the new instance per test. With that said, I think setup should create the instance and configure it. In mind, that's more transparent. Otherwise, LGTM

arteam: OK, shouldn't be a problem to create a copy.

PR #1105, opened 2015-06-06
Use dropwizard hibernate validator in jersey validations

nickbabcock:
This commit enables `Optional` and `AbstractParam` parameters to be
validated in resource endpoints. This required additional code needed in
setting up the validator, so I extracted the common functionality out to a
new class `Hibernate` that should be analogous to the present `Jackson`
class.

Also this commit makes the `NonEmptyStringParam` constructor public like other
`AbstractParam`s. The second commit registers the default
`ConstraintViolationExceptionMapper` in test cases (and this can always be
overridden by the user if they have a custom exception mapper for constraint
violations)

The one area where this commit falls short is that it currently doesn't
allow a user custom hibernate configuration in the jersey validation. Thus,
request entities will be validated by the user's validator but parameters
will not (they'll still be validated by the Dropwizard configured
validator). Since I wasn't sure how best to incorporate this
functionality, I decided to leave it out of this commit.

This allows for some interesting endpoints:

```java
    @GET
    @Path("search")
    public String search(
            @QueryParam("limit")
            @DefaultValue("10")
            @UnwrapValidatedValue
            @Min(1) @Max(100)
            IntParam limit,

            @QueryParam("offset")
            @DefaultValue("0")
            @UnwrapValidatedValue
            @Min(0)
            IntParam offset) {
        // ...
    }
```

It's annotation overload (I believe `@UnwrapValidatedValue` will be
redundant in Hibernate 5.2), but it certainly does pack a lot of
functionality in the parameters.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/NonEmptyStringParamFeature.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/NonEmptyStringParamFeature.java
@@ -1,0 +1,40 @@
+package io.dropwizard.jersey.params;
+
+import com.google.common.base.Optional;
+
+import javax.ws.rs.core.Feature;
+import javax.ws.rs.core.FeatureContext;
+import javax.ws.rs.ext.ParamConverter;
+import javax.ws.rs.ext.ParamConverterProvider;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+/**
+ * A class for describing how Jersey should serialize a {@link NonEmptyStringParam}. If the
+ * parameter was not detected in the response, instead of the resulting value being null, it will
+ * evaluate to {@link Optional#absent()}
+ */
+public class NonEmptyStringParamFeature implements Feature {
+    @Override
+    public boolean configure(final FeatureContext context) {
+        context.register(new ParamConverterProvider() {
+            @Override
+            public <T> ParamConverter<T> getConverter(final Class<T> rawType,
+                                                      final Type genericType,
+                                                      final Annotation[] annotations) {
+                return (rawType != NonEmptyStringParam.class) ? null : new ParamConverter<T>() {
+                    @Override
+                    public T fromString(final String value) {
+                        return rawType.cast(new NonEmptyStringParam(value));
+                    }
+
+                    @Override
+                    public String toString(final T value) {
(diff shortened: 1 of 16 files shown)


arteam:
Mostly it looks good to me. I have 2 proposions/quesions:

* Maybe we could we rename the class `Hibernate` to more specific to reduce confusion with the ORM? `DropwizardValidator` or `Validators` sounds good to me. 
* Is there any way to move this class to the module *dropwizard-validation* without adding circular
dependencies? It would be great if we could use it across the project instead of `Validation.buildDefaultValidatorFactory().getValidator()`. Maybe we could load Jersey unwrappers undirectly(reflection?) or use separate configurations... I don't have a good solution in mind yet.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/NonEmptyStringParamFeature.java
```diff
@@ -0,0 +1,40 @@
...
+import java.lang.reflect.Type;
+
+/**
+ * A class for describing how Jersey should serialize a {@link NonEmptyStringParam}. If the
+ * parameter was not detected in the response, instead of the resulting value being null, it will
+ * evaluate to {@link Optional#absent()}
+ */
+final public class NonEmptyStringParamFeature implements Feature {
+    @Override
+    public boolean configure(final FeatureContext context) {
```
arteam: Minor nitpic: We don't require to mark all method parameters or classes as final. This applies only to local variables,

nickbabcock:
Updated the name to `Validators` and fixed some of the style.

>Is there any way to move this class to the module dropwizard-validation without adding circular dependencies? It would be great if we could use it across the project instead of Validation.buildDefaultValidatorFactory().getValidator(). Maybe we could load Jersey unwrappers undirectly(reflection?) or use separate configurations... I don't have a good solution in mind yet.

I wouldn't have the faintest clue how to achieve this. I'll take all suggestions :smile:

nickbabcock:
Looks like Travis failed for an unrelated reason, a `SocketTimeoutException: Read timed out` in the HTTP Client :confused:

joschi:
I restarted the failed build job.

arteam:
Ok, that's good enough for me. Thanks, @nickbabcock, for your gigantic efforts on this feature!

PR #1106, opened 2015-06-06
Width independent test

christophercurrie:
The test for the db migrate command help is dependent upon terminal width, so it fails on my mac laptop. This change fixes the test to instruct argparse4j to not detect terminal width during the test. Unfortunately the library keeps this setting as a static class variable, rather than a setting on the parser, so the try/catch block is needed to ensure it gets reset after the test is done.


arteam:
I think we probably don't want to enable terminal width detection in tests at all.
So it makes sense to move disabling it to a static block in `AbstractMigrationTest` or into a `@BeforeClass` method. It should make tests on command help pages more focused.

christophercurrie:
Ok, I've updated the change to use a static initializer.

arteam:
Thanks, looks good.

PR #1107, opened 2015-06-07
Tests for DbDumpCommand

arteam:
This pull requests consists from 2 changes:
* Fix descriptions of some parameters in `DbDumpCommand`
* Add tests for `DbDumpCommand`

This suite checks 4 cases:

* Dump the existed database schema to an XML change set
* Dump the schema and data to an XML change set
* Dump only data
* Print a correct help page

For checking correctness of a change set, it uses a DOM XML parser, rather than regular expressions or string manipulations. The reason for this is that the resulted XML text is not a constant and depends on environment and time of a run. To overcome this, we use the XML parser, because it provides ability to make an intelligent  assertions. In the end, it makes tests more stable and easy to maintain.

### DIFF
--- a/dropwizard-migrations/src/main/java/io/dropwizard/migrations/DbDumpCommand.java
+++ b/dropwizard-migrations/src/main/java/io/dropwizard/migrations/DbDumpCommand.java
@@ -1,5 +1,6 @@
 package io.dropwizard.migrations;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.dropwizard.Configuration;
 import io.dropwizard.db.DatabaseConfiguration;
 import liquibase.CatalogAndSchema;
@@ -39,6 +40,14 @@ import java.util.HashSet;
 import java.util.Set;
 
 public class DbDumpCommand<T extends Configuration> extends AbstractLiquibaseCommand<T> {
+
+    private PrintStream outputStream = System.out;
+
+    @VisibleForTesting
+    void setOutputStream(PrintStream outputStream) {
+        this.outputStream = outputStream;
+    }
+
     public DbDumpCommand(DatabaseConfiguration<T> strategy, Class<T> configurationClass) {
         super("dump",
               "Generate a dump of the existing database state.",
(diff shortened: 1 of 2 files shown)


dropwizard-migrations/src/test/java/io/dropwizard/migrations/DbDumpCommandTest.java
```diff
@@ -0,0 +1,234 @@
...
+        final Map<String, Object> attributes = ImmutableMap.of("output", (Object) file.getAbsolutePath());
+        dumpCommand.run(null, new Namespace(attributes), existedDbConf);
+        // Check that file is exist, and has some XML content (no reason to make a full-blown XML assertion)
+        assertThat(Files.toString(file, Charsets.UTF_8)).startsWith("<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\"?>");
+    }
+
+    @Test
+    public void testHelpPage() throws Exception {
+        createSubparser(dumpCommand).printHelp(new PrintWriter(baos, true));
+        assertThat(baos.toString("UTF-8")).isEqualTo(String.format(
```
gtrog: Unintentional String.format(), since no additional arguments are passed in?

dropwizard-migrations/src/main/java/io/dropwizard/migrations/DbDumpCommand.java
```diff
@@ -39,6 +40,14 @@
 import java.util.Set;
 
 public class DbDumpCommand<T extends Configuration> extends AbstractLiquibaseCommand<T> {
+
+    private PrintStream outputStream = System.out;
+
+    @VisibleForTesting
+    public void setOutputStream(PrintStream outputStream) {
```
carlo-rtr: should this be package private?

carlo-rtr:
Besides the small comment, looks good!

dropwizard-migrations/src/main/java/io/dropwizard/migrations/DbDumpCommand.java
arteam: Sure. Will fix that.

PR #1108, opened 2015-06-08
Update validationmethod documentation to use JsonIgnore

RawToast:
Added @JsonIgnore to example validation method. Whilst the application will work without this flag, it will fail during testing as additional paramters will be added to the created JSON StackOverflow: http://stackoverflow.com/questions/26551726/dropwizard-validationmethod-adding-a-json-property

To be safe, and promote unit testing, this should be included in the example documentation.


arteam:
Looks like it's worth to add. Thanks.

PR #1113, opened 2015-06-12
Add support for setting a custom DB pool connection validator

### LINKED ISSUE #1089: dropwizard-db 0.8.1 closed connections are returned to pool
Hello guys, we have migrated to Dropwizard 0.8.1 from 0.6.2 recently and faced the following issue while using JDBI:
after database restart (Postgres), our application wasn't been able to restore connectivity to DB, all connections were closed and every time, after failed attempt to execute some SQL statement, returned back to pool. Pool is configured with all tests (on borrow, on return, on connect, while idle) disabled. Dropwizard 0.8.1 is using Tomcat DBCP 8.0.18 vs 7.0.37 on 0.6.2. Database configuration we are using is the same for both Dropwizard versions.
Dropwizard 0.6.2 is using 7.0.37 DBCP and is able to restore from DB restart, because PoolableConnection class performs explicit check if underlying SQL connection is closed:

     public synchronized void close() throws SQLException {
        if (_closed) {
            // already closed
            return;
        }

        boolean isUnderlyingConectionClosed;
        try {
            isUnderlyingConectionClosed = _conn.isClosed();
        } catch (SQLException e) {
            try {
                _pool.invalidateObject(this); // XXX should be guarded to happen at most once
            } catch(IllegalStateException ise) {
                // pool is closed, so close the connection
                passivate();
                getInnermostDelegate().close();
            } catch (Exception ie) {
                // DO NOTHING the original exception will be rethrown
            }
            throw (SQLException) new SQLException("Cannot close connection (isClosed check failed)").initCause(e);
        }
       ...

There is no such check in DBCP 8.0.18. 

Have anybody else faced this? We don't want to enable on borrow/on return validation checks just to overcome it.

P.S.: I've asked in Tomcat mailing list why they have removed this underlying connection check, will update this issue with response.

arteam:
Some users want to specify a custom validation logic for database connections. 
Tomcat JDBC Pool provides this facility and we can support it in the YAML configuration.

Close #1089

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -988,6 +988,9 @@ evictionInterval                5 seconds                The amount of time to s
 
 validationInterval              30 seconds               To avoid excess validation, only run validation once every
                                                          interval.
+
+validatorClassName              none                     Name of a class of a custom validator implementation, which
+                                                         will be used for validating connections.
 ============================    =====================    ===============================================================
 
 .. _man-configuration-polymorphic:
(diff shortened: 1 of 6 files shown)


carlo-rtr:
LGTM

PR #1114, opened 2015-06-13
Simplify authorization

arteam:
The current implementation of authorization is too cumbersome and requires implementing a big amount of boilerplate code.

This changes introduce `Authorizer` interface, which is responsible for authorizing users. It's a replacement for specifying a function for creating a new secure context.

Advantage stem from the fact that users only implement business logic for authorization and nothing more. Also this type of API is similar to `Authenticator`, which makes API more consistent.

### DIFF
--- a/docs/source/manual/auth.rst
+++ b/docs/source/manual/auth.rst
@@ -19,7 +19,7 @@ Authenticators implement the ``Authenticator<C, P extends Principal>`` interface
 
 .. code-block:: java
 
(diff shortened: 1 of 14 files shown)


glennmcallister:
I like it, personally. Its easier to read and conceptually more compact than what we have currently.

carlo-rtr:
Agreed, I think it's a better abstraction. I'm still working thru the code, my goal is to have feedback or merge by EOD.

jplock:
This looks a lot better.  Great job.

PR #1118, opened 2015-06-19
Refactor CLI drivers for non-http use

christophercurrie:
As discussed on the mailing list, this change pulls out the CLI and configuration parsing drivers of Dropwizard applications into a library that can be used by applications that want the infrastructure of dropwizard without the dependency on Jetty. It does not yet provide any `Command` implementations other than `check`; it is up to consumers to determine what commands make sense for their application.

This is a breaking change; I know that the 0.9-rc1 is already out, but think it would be better to get this out as a part of 0.9 to allow users more opportunity to use it and provide feedback.


christophercurrie:
The build is failing, because Maven in trying to build `dropwizard-core` before `dropwizard-cli`, which is mystifying because (a) the dependencies look correct, and (b) it doesn't reproduce for me locally. If anyone has insights as to why Maven is confused, I'd appreciate the help.

mattnelson:
The version in the cli and cli-example poms needs to be updated to match the rest of the reactor. That fixed the build for me.

christophercurrie:
Thanks for pointing that out; wish there were an easy way to have Maven ignore existing cached dependencies without removing the whole repo.

christophercurrie:
This PR mysteriously failed the build on Java 7; I think a trivial change to trigger a rebuild would fix it. That said, while this approach allows a cli app to not depend on Jersey, the dependency on the Jetty Lifecycle means an HTTP server is still included. I'm going to reconsider the approach and come back to this idea post 0.9.

PR #1123, opened 2015-06-22
Null entities return error messages

nickbabcock:
The previous behavior was returning an empty error list.


jplock:
Is this related to https://github.com/dropwizard/dropwizard/pull/633 or different?

nickbabcock:
It is somewhat related. It adds the fix and a test case for what the expected response should be if a null request entity is not allowed (whether that means annotated with `@NotNull` or if just `@Valid/Validated` is what #633 is addressing). If #633 is merged, the test will still pass.

Before: `{ "errors": [] }`
After: `{ "errors": ["The request entity was empty"] }`

jplock:
LGTM!

PR #1126, opened 2015-06-24
code excerpt in docs should be 'static final', not 'final static'

dareonion:


joschi:
Thanks!

For reference: https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.3.1

PR #1125, opened 2015-06-23
Register default mappers in ResourceTestRule

nickbabcock:
The default exception mappers are not registered in `ResourceTestRule`, which may be an annoyance to users, as they would have know what the defaults are and know to re-register them if their tests relied on functionality the exception mappers provided. This pull requests fixes this situation by registering them out of the box.

The only users who would be negatively affected by this pull request are those that turn off `registerDefaultExceptionMappers` in the configuration and choose not to define new mappers to replace them. Those that override default exception mappers in the application will still override them in `ResourceTestRule` as usual.

Closes #808

PR #1129, opened 2015-06-24
Default exception mappers in test can be overridden

nickbabcock:
#1125 registered the default exception mappers in the wrong place in the `ResourceTestRule`, such that one couldn't register their own `ExceptionMapper` with the same signature and have it be used :blush:

This pull request fixes the behavior and adds a test to ensure that the default exception mappers can be overridden in `ResourceTestRule`

### DIFF
--- a/dropwizard-testing/src/test/java/io/dropwizard/testing/app/OverrideDefaultExceptionMapperTest.java
+++ b/dropwizard-testing/src/test/java/io/dropwizard/testing/app/OverrideDefaultExceptionMapperTest.java
@@ -1,0 +1,42 @@
+package io.dropwizard.testing.app;
+
+import io.dropwizard.testing.junit.ResourceTestRule;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import javax.validation.ConstraintViolationException;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Test to make sure that the default exception mappers registered in
+ * {@link ResourceTestRule} can be overridden
+ */
+public class OverrideDefaultExceptionMapperTest {
+    private static final PeopleStore dao = mock(PeopleStore.class);
+
+    @ClassRule
+    public static final ResourceTestRule resources = ResourceTestRule.builder()
+            .addResource(new PersonResource(dao))
+            .addProvider(new CustomMapper())
+            .build();
+
+    private static class CustomMapper implements ExceptionMapper<ConstraintViolationException> {
+        @Override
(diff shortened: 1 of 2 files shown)


arteam:
For some reason the test is flaky on my development box. It works steadily only if I explicitly set `@Priority(0)` on `CustomMapper`, so it's processed before `ConstraintViolationExceptionMapper`. 
Is it desired behaviour for users?

dropwizard-testing/src/test/java/io/dropwizard/testing/app/OverrideDefaultExceptionMapperTest.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Test to make sure that the default exception mappers registered in
+ * {@link ResourceTestRule} can be overridden
+ */
+public class OverrideDefaultExceptionMapperTest {
+    private static final PeopleStore dao = mock(PeopleStore.class);
+
+    @ClassRule
+    public static final ResourceTestRule resources = ResourceTestRule.builder()
+            .addResource(new PersonResource(dao))
```
arteam: You could just use `new PersonResource(mock(PeopleStore.class))`, because `PeopleStore` is used as a just dummy object.

dropwizard-testing/src/test/java/io/dropwizard/testing/app/OverrideDefaultExceptionMapperTest.java
```diff
@@ -0,0 +1,42 @@
...
+            .addProvider(new CustomMapper())
+            .build();
+
+    private static class CustomMapper implements ExceptionMapper<ConstraintViolationException> {
```
arteam: Please, move the class to the start or to the end of `OverrideDefaultExceptionMapperTest`, so it's more easy to track.

nickbabcock:
Oh wow good catch. It took me a half dozen runs to see the behavior is non-deterministic.

I was hoping not to use `@Priority` on exception mappers. There also seems [to be a couple](https://java.net/jira/browse/JERSEY-2437) [tickets open still](https://java.net/jira/browse/JERSEY-2722) about exception mappers and `@Priority` not working together.

nickbabcock:
How do you feel about annotating all the exception mappers that Dropwizard defines as `@Priority(Integer.MAX_VALUE)`? Smoke testing this implementation made all the tests (including the one in this PR) pass a half dozen times in a row, so hopefully it is deterministic. This may also allow for people to override the default exception mappers in an application without having to disable registering all of them.

arteam:
Sounds like a plan.

nickbabcock:
I tested it out and annotating Dropwizard's exception mappers with `@Priority(Integer.MAX_VALUE)` allows for the user to override any of them without setting `registerDefaultExceptionMappers` to false. I'm closing this pull request and will open a new one (hopefully today/tomorrow) that has the implementation and tests.

PR #1137, opened 2015-06-26
Bugfix for: Ensuring that JerseyRequest scoped ClientConfig gets prop…

velocipedist:
…agated to HttpUriRequest #939

The change makes the config changes introduced via `Invocation.Builder#property(propertyName, propertyValue)` effective. We want to use this pattern because it is truely request creation scoped. In contrast the current implementation is only effective if config changes are being introduced via `WebTarget#property(propertyName, propertyValue)` which unfortunatly creates a new copy of copy of `org.glassfish.jersey.client.ClientRuntime` which effectively makes it innefective for implementing dynamic configuration (we saw OutOfMemmory errors due to that).

Original change can be founde here: https://github.com/dropwizard/dropwizard/pull/939

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
@@ -24,7 +24,6 @@ import org.glassfish.jersey.message.internal.OutboundMessageContext;
 import org.glassfish.jersey.message.internal.Statuses;
 
 import javax.ws.rs.ProcessingException;
-import javax.ws.rs.core.Configuration;
 import javax.ws.rs.core.Response;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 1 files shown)


jplock:
This LGTM, any feedback @arteam ?

velocipedist:
@jplock I noticed you've added it to the 0.9.0 milestone. Does this mean there will be no 0.8.2?

jplock:
@velocipedist correct, we already cut 0.9.0-rc1, so ideally this would make it into rc2

velocipedist:
A bit OT, but will ask anyway.

@jplock what are the timelines for this? I'm asking because we found a bug with the metrics library version used by 0.8.1 where request durations are not reported correctly for async resource usage. We were hoping to patch this (unless already done) before the next release of dw.

jplock:
Unfortunately the metrics releases are a bit out of our hands.  @ryantenney has been working on a roadmap for v4.x but I don't have any visibility into its release schedule.  It may be best to work around it manually for now if you have a workaround.  We're hoping to get Dropwizard 0.9.0 out in July.  /cc @carlo-rtr

velocipedist:
FYI: @chrisgray (see the comment around release schedules and metrics)

arteam:
This also looks good to me. I've applied the patch to master with some commit message polishing.

ryantenney:
Metrics 4 isn't going to happen anytime soon.  I wanted to get a 3.2 release out, with some fixes/small features like this.  We could probably make that happen before the Dropwizard 0.9 release.

PR #1140, opened 2015-06-28
Make 0.9.0 release notes more human-readable and organized

arteam:


jplock:
This looks good to me.  As an aside, do we want to link to the actual PR's or issues in the release notes so people can reference them?

arteam:
I also thought about that and it looks reasonable. I will work on that.

PR #1141, opened 2015-06-29
use joda-time 2.8.1

profes:
It seems to be safe to upgrade from 2.7 to 2.8.1
See:
http://www.joda.org/joda-time/upgradeto281.html
http://www.joda.org/joda-time/upgradeto280.html


joschi:
Could you please update `docs/source/about/release-notes.rst` as well? Other than that it seems to be a drop-in replacement, so :+1: .

profes:
Done

PR #1144, opened 2015-06-29
Default exception mappers are overridable

nickbabcock:
This pull request annotates the default exception mappers with `@Priority(Integer.MAX_VALUE)`, which allows for a user to override one of the default exception mappers without having to set the configuration option `registerDefaultExceptionMappers` to `false`. Since the configuration option won't  be useful once this pull request is merged, it can be slated for removal at a later point.

As a corollary, we can add additional exception mappers (`ExceptionMapper<Foo>`) and not fear breaking applications where the user had also defined `ExceptionMapper<Foo>`

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -58,6 +58,8 @@ rootPath                            ``/``                                       
                                                                                      the JAX-RS resources will be served.
 registerDefaultExceptionMappers     true                                             Whether or not the default Jersey ExceptionMappers should be registered.
                                                                                      Set this to false if you want to register your own.
+                                                                                     Note that this option is deprecated as user defined ExceptionMappers can
+                                                                                     override the default Jersey ExceptionMappers.
 =================================== ===============================================  =============================================================================
 
 
--- a/docs/source/manual/validation.rst
+++ b/docs/source/manual/validation.rst
@@ -253,15 +253,17 @@ code and ensure that the error messages are user friendly.
 Extending
 =========
 
-While Dropwizard provides good defaults for error messages, one size may not fit all and so there
-are a series of extension points. To register your own
-``ExceptionMapper<ConstraintViolationException>`` you'll need to first set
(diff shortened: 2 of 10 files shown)


jplock:
Should we add a `@Deprecated` on https://github.com/dropwizard/dropwizard/blob/master/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java#L242 ?

nickbabcock:
Yes I can append a couple commits, one to deprecate, and another to update the documentation.

/cc @scho, as you expressed concern earlier in #1125 and I was curious if you had any concerns about deprecating `registerDefaultExceptionMappers`

scho:
Looks good to me!

nickbabcock:
Upgrading to Jersey 2.19 #1143 breaks this pull request (it works fine on 2.18). The investigation is on! :sunglasses:

nickbabcock:
Ok, back from the investigation with some not so good news. I don't normally like to throw in the towel, but this is a doozy.

This implementation doesn't work as [Priority annotations do not work on exception mappers (JERSEY-2722)](https://java.net/jira/browse/JERSEY-2722). 

The suggested workaround, using [`ExtendedExceptionMapper`](https://jersey.java.net/apidocs/2.13/jersey/org/glassfish/jersey/spi/ExtendedExceptionMapper.html), is not a 100% solution. I can expand on this if needed, but I was not able to find an alternative either.

Plan of action:

- Close this pull request and remove 0.9 milestone
- Revert #1125 so that tests don't break for people when moving to 0.9
- Roll up #1125 into this pull request so that everything is in one pull request
- Wait for Jersey issue to be resolved
- Re-open pull request
- Celebrate :tada:

Why did this code worked on 2.18 appears to be a total fluke (though it may, in part, be due to [JERSEY-2891](https://java.net/jira/browse/JERSEY-2891))! My apologies for any wasted time, hopefully it was informative at least :smile:

PR #1151, opened 2015-07-03
Fix hibernate error

wakandan:
FakeEntity1 was wrongl placed in io.dropwizard.hibernate package while it should have been in io.dropwizard.hibernate.fake.entities.pckg in its code


wakandan:
wrongly pushed a pr from my working copy

PR #1152, opened 2015-07-03
FakeEntity1 was wrongly placed. Moved it to the correct package

wakandan:
Just moved it to the correct package


joschi:
Thanks!

PR #1153, opened 2015-07-07
Make LoggingFactory discoverable.

dylanscott:
It looks like [#996](https://github.com/dropwizard/dropwizard/pull/996) omitted these changes - I'm not sure if that was deliberate. I just pulled down 0.9.0-rc1 to play around with adding a custom LoggingFactory and it wouldn't work without these changes.


dylanscott:
@carlo-rtr is this a reasonable change? I wasn't sure if this was deliberately left out initially.

arteam:
You are right. It's probably just an omission.

carlo-rtr:
Yes, looks like I forgot that. Thank you.

PR #1154, opened 2015-07-07
Add toString for configuration objects

kmejka:
Objects which hold configuration are serialized from json/yaml files.
Without proper ```toString()``` methods, it is impossible to log what
is the configuration of the application after it was deserialized into
the configuration class. It is especially important in regards to the
```DefaultServerFactory``` class, which holds the configuration for the
```server``` command. By adding the ```toString()``` methods, it is
possible to log during application startup what is the actual configuration
used by the app.

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java [hunk 1]
```diff
@@ -176,4 +177,15 @@ private Logger configureLevels() {
...
     }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                .add("level", level)
+                .add("loggers", loggers)
+                .add("appenders", appenders)
+                .add("loggerContext", loggerContext)
+                .add("configurationErrorsStream", configurationErrorsStream)
```
joschi: I don't think that printing the configurationErrorsStream or the loggerContext is really useful here.

kmejka: You're right, perhaps I should've written them myself, instead of using the code my IDE blurped out. I'll amend the commit.

joschi:
LGTM. :+1:

kmejka:
Great, thx for merging :)

PR #1161, opened 2015-07-11
Remove Authenticator generic parameter from AuthFilterBuilder

arteam:
Generic parameter that controls a type of a passed authenticator to the `AuthFilterBuilder` is not actually needed. Type-safety is already forced by the generic parameters C(credentials) and P(principal). The user has to provide a class that implements interface `Authenticator<C,P>`, and we don't care about concrete type of this implementation, because it doesn't affect the auth filter.

This commit also moves the common logic of building auth filters to the `AuthFilterBuilder` and adds some comments for the public API.

### DIFF
--- a/docs/source/manual/auth.rst
+++ b/docs/source/manual/auth.rst
@@ -100,7 +100,7 @@ takes instances of ``BasicCredentials``:
     public void run(ExampleConfiguration configuration,
                     Environment environment) {
         environment.jersey().register(new AuthDynamicFeature(
-                new BasicCredentialAuthFilter.Builder<User, ExampleAuthenticator>()
+                new BasicCredentialAuthFilter.Builder<User>()
                     .setAuthenticator(new ExampleAuthenticator())
                     .setAuthorizer(new ExampleAuthorizer())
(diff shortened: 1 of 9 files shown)


jplock:
Does this allow the authorizer to be optional?

arteam:
No, the authorizer and the authenticator must be provided during building.

jplock:
Is that what we want since basically right now we don't have an authorizer? Would it make sense to have a default one that always allows access?

arteam:
I thought about that. For this case, we would need then to allow users to use the `@PermitAll` annotation on resource methods. It would simplify things for users, who don't core about roles.
I can prepare another PR with optional authorizer, if this change is merged.

arteam:
Moreover, I think we handle `@PermitAll` wrong now. It means "no authorization", but we treat it like "no authentification".

jplock:
Sounds good.  I think this PR is a good cleanup anyway.  Thanks!

arteam:
Ok, I  submitted #1162

carlo-rtr:
This a great idea! Thanks

PR #1162, opened 2015-07-12
Make authorization optional

arteam:
This change consists from two parts: fixing behavior of the`@PermitAll` annotation and making
providing an authorizer in `AuthFilterBuilder` optional.

The `@PermitAll` annotation is actually intended  for marking that a resource doesn't require authorization. Currently we treat it like the resource doesn't require authentication. 

The solution is to register an auth filter for the methods marked with the `@PermitAll` annotation. The methods will require authentication, but anyone, who was given access, can use it. This simplifies the work of users who don't care about roles and authorization. They don't need create a fiction role and register `RolesAllowedDynamicFeature`.

Because users have an option to not authorize their resource methods, it makes sense to not require a custom authorizer in `AuthFilterBuilder`. By default the filter will permit access to any principal in any role.

### DIFF
--- a/docs/source/manual/auth.rst
+++ b/docs/source/manual/auth.rst
@@ -92,7 +92,8 @@ Basic Authentication
 
 The ``AuthDynamicFeature`` with the ``BasicCredentialAuthFilter`` and ``RolesAllowedDynamicFeature``
 enables HTTP Basic authentication and authorization; requires an authenticator which
-takes instances of ``BasicCredentials``:
+takes instances of ``BasicCredentials``. If you don't use authorization, then ``RolesAllowedDynamicFeature``
+is not required.
 
 .. code-block:: java
 
@@ -117,7 +118,8 @@ OAuth2
 
 The ``AuthDynamicFeature`` with ``OAuthCredentialAuthFilter`` and ``RolesAllowedDynamicFeature``
 enables OAuth2 bearer-token authentication and authorization; requires an authenticator which
-takes instances of ``String``:
+takes instances of ``String``. If you don't use authorization, then ``RolesAllowedDynamicFeature``
+is not required.
 
 .. code-block:: java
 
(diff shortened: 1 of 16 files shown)


arteam:
There is a lot of duplication in authentication provider tests. I wish we could generalize them and not copy-past the tests between different providers.

jplock:
This looks great.  One question, if a user wants to use an `Authenticator` but has no intention of doing anything related to authorization, is `@PermitAll` still required on the resources?

jplock:
Maybe a better question is, if you don't register an authorizer, is there anyway we can assume that all of the resource methods have `@PermitAll`?  That would provide a clear upgrade path for anyone pre-0.9 who is just using an authenticator.

arteam:
> Maybe a better question is, if you don't register an authorizer, is there anyway we can assume that all of the resource methods have @PermitAll? That would provide a clear upgrade path for anyone pre-0.9 who is just using an authenticator.

No, we can't. To register the filter you need to annotate the method with the `@PermitAll` annotatiion. In this case we will register the auth filter and the autentiication process will start.

To help migration from 0.8.x we could register the filter if the `@Auth` annotation is present on resources method parameters. But I'm not completely sure that's a good idea. From one point it makes things simpler if authorization is not a concern, from the other point it will increase confusion in the process of protecting resources. 

Currently we clearly point in the documentation that to protect a method one should annotate it with `javax.security` annotations. And that `@Auth` is used only for injecting the principal to method parameters. If we add additional logic to `@Auth`, there could be a question, why don't register the filter if the `@Context` annotation is present on a method? So far, I'm against it.

carlo-rtr:
Great work, definitely makes the feature more user friendly. This looks good to me, I'll merge. 

Previous versions always required the `@Auth` for authentication. So while the argument about `@Context` makes sense, I think most people will probably (and should) use `@Auth`.  So I'm a fan of registering @PermitAll if the method has a @Auth parameter. I think if we call it out on the docs, it won't be terribly confusing. I think aiding the migration out weighs the downside. wdyt?

arteam:
Weighing the fact that we have many users who use just `@Auth` and they expect smooth transition to 0.9.0, we could take the step. If we document the reasons why we do that, the process of protecting resources will look transparent enough.

carlo-rtr:
This is a start on making that change: https://github.com/dropwizard/dropwizard/compare/master...carlo-rtr:auth_registers_filter?expand=1

I'm planning to try to consolidate what I can for the tests, and then submit a PR

PR #1163, opened 2015-07-12
Improve AuthFactoryProvider

arteam:
* Make sure users pass the type inherited from `Principal`.
* Remove unchecked cast warnings.
* Remove unnecessary static class `AuthValueFactory`, which could be replaced by an anonymous class.
* Make `PrincipalClassProvider` a holder class, rather then an interface. Doing so, we don't create a new anonymous class in `Binder`.
* Documentation and style improvements.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthValueFactoryProvider.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthValueFactoryProvider.java
@@ -10,49 +10,69 @@ import org.glassfish.jersey.server.internal.inject.MultivaluedParameterExtractor
 import org.glassfish.jersey.server.internal.inject.ParamInjectionResolver;
 import org.glassfish.jersey.server.model.Parameter;
 import org.glassfish.jersey.server.spi.internal.ValueFactoryProvider;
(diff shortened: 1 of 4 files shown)


arteam:
These are cosmetic changes, but I hope they are useful and make the codebase cleaner.

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthValueFactoryProvider.java
```diff
@@ -10,49 +10,69 @@
...
 import org.glassfish.jersey.server.model.Parameter;
 import org.glassfish.jersey.server.spi.internal.ValueFactoryProvider;
-import org.jvnet.hk2.annotations.Contract;
+
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import java.security.Principal;
 
+/**
+ * Value factory provider supporting the {@link Principal} injection
```
carlo-rtr: I don't think you need "the"

carlo-rtr:
Besides the one nitpick, this looks great. Thanks

arteam:
Ok, thanks for feedback. Pushed a new version, let me know if it's good enough.

jplock:
LGTM!

PR #1172, opened 2015-07-16
Added hook for UnauthorizedHandler to be customized in the builder. 

mnrasul:
Added hook for `UnauthorizedHandler` to be customized in the `AuthFilterBuilder`.
Not sure how to test it though.


arteam:
Thanks. Applied to master in a minimal form.

PR #1177, opened 2015-07-18
Fix ZLIB decompression in BiDiGzipFilter

arteam:
To decompress deflate-encoded requests one should use the `InflaterInputStream` class, not `DeflaterInputStream`. The latter actually is used for compressing requests on the fly.

This change adds the parameter `gzipCompatibleInflation` for controlling the mode of ZLIB decompression.

Also it fixes a minor bug, when a broken inflater could stay in the thread local storage.


jplock:
LGTM, thanks!

PR #1178, opened 2015-07-21
Improve View Template Error Message

### LINKED ISSUE #1077: MustacheViewRenderer template error is misleading
When an error happens in MustacheViewRenderer the message is always "Template xxxx not found", leading the user into thinking that the renderer could not find the file when a template compilation could have happened. The exception thrown is also swallowed.

The message to the user could be more specific and the exception should be logged, so that he can see which line of my template is incorrect.

nickbabcock:
Instead of hiding behind a sometimes deceptive error message saying that the
template file was not found, ~~propagate the error message in an HTML safe
way~~ return a generic error message with the actual error going to the log
file. This makes the behavior more in line with validation errors (though this
doesn't deal with validation errors, itself)

Closes #1077

### DIFF
--- a/dropwizard-views/src/main/java/io/dropwizard/views/ViewMessageBodyWriter.java
+++ b/dropwizard-views/src/main/java/io/dropwizard/views/ViewMessageBodyWriter.java
@@ -2,18 +2,18 @@ package io.dropwizard.views;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.Produces;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.*;
 import javax.ws.rs.ext.MessageBodyWriter;
 import javax.ws.rs.ext.Provider;
-import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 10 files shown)


nickbabcock:
So based on @nathancomstock's comment, I've refocused the pull request such that instead of throwing up an error page that says that the file isn't found, return a generic page to the client saying that something went wrong. Then the developer can see the error under debug mode. 

This pull request also explains the behavior in the `views` documentation.

dropwizard-views/src/main/java/io/dropwizard/views/ViewMessageBodyWriter.java
```diff
@@ -77,11 +78,11 @@ public void writeTo(View t,
                 }
             }
             throw new ViewRenderException("Unable to find a renderer for " + t.getTemplateName());
-        } catch (FileNotFoundException e) {
-            final String msg = MessageFormat.format(MISSING_TEMPLATE_MSG, t.getTemplateName());
+        } catch (Throwable e) {
```
joschi: Catching `Throwable` instead of just `Exception` might be too broad here as it also catches non-recoverable errors like `OutOfMemoryError`.

nickbabcock:
Oi, good catch. Fixed :blush:

arteam:
LGTM!

Issue #1160, opened 2015-07-11
Http Client Connection Pool Error


ruimateus:
Hi,
I have a series of Tasks that connect to multiple targets that have an hourly cronjob running them.

The multiple tasks are using the same JerseyClient instance from JerseyClientConfiguration.

After a while, I start getting **javax.ws.rs.ProcessingException: java.lang.IllegalStateException: Connection pool shut down**

The only way to recover is restarting the application.

This only started to occur in Dropwizard 0.8.2.

Thanks for any help!

ruimateus:
After further investigation, it seems that InstrumentedHttpClientConnectionManager shuts down the pool for an unknown reason:

DEBUG [2015-07-11 14:12:44,652] org.apache.http.impl.conn.HttpClientConnectionOperator: Connecting to ****/****
DEBUG [2015-07-11 14:12:44,652] org.apache.http.impl.execchain.MainClientExec: Connection discarded
DEBUG [2015-07-11 14:12:44,652] org.apache.http.impl.conn.DefaultManagedHttpClientConnection: http-outgoing-6409: Close connection
DEBUG [2015-07-11 14:12:44,652] com.codahale.metrics.httpclient.InstrumentedHttpClientConnectionManager: Connection released: [id: 6409][route: {s}->****:443][total kept alive: 0; route allocated: 7 of 1024; total allocated: 21 of 1024]
DEBUG [2015-07-11 14:12:44,653] com.codahale.metrics.httpclient.InstrumentedHttpClientConnectionManager: Connection manager is shutting down

After this point, any HttpClient call, returns an exception ***java.lang.IllegalStateException: Connection pool shut down***

PoolingHttpClientConnectionManager shuts down on close() and finalize() method.

When does Dropwizard closes the CloseableHttpClient?

Thanks

mikeycmccarthy:
This also affected us - we had to rollback to 0.8.1.

nickbabcock:
I tried to reproduce the issue locally and couldn't get the behavior seen. Does anyone have a gist or repo demonstrating the bug?

carlo-rtr:
This probably not very helpful, but I figured I would list the places I believe close could be called:

* Jetty shuts down
* ``InstrumentedHttpClientConnectionManager`` gets garbage collected and ``finalize`` is triggered
* ``DropwizardApacheConnector#close``
* One thread calls close on the Client

carlo-rtr:
Maybe this should be true?

[...]

PR #1179, opened 2015-07-21
make connection pool shared by default

carlo-rtr:
https://github.com/dropwizard/dropwizard/issues/1160


dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
```diff
@@ -221,6 +221,7 @@ protected ConfiguredCloseableHttpClient createClient(
             final org.apache.http.impl.client.HttpClientBuilder builder,
             final InstrumentedHttpClientConnectionManager manager,
             final String name) {
+        builder.setConnectionManagerShared(configuration.isShareConnectionManagerEnabled());
```
jplock: This looks good, but do you want to move this down to around line 249 so its with the rest of the builder option setting?   This one is allow alone by itself.

nickbabcock:
I do have a question, does anyone know under what circumstances one would want `shareConnectionManagerEnabled` to be false in Dropwizard?

PR #1180, opened 2015-07-22
Revert "make connection pool shared by default"

carlo-rtr:
Reverts dropwizard/dropwizard#1179


vanDonselaar:
Excuse me, but what was the reason for reverting this? I think it's a good fix and I'm looking forward to the release. :+1:

carlo-rtr:
@vanDonselaar see https://github.com/dropwizard/dropwizard/issues/1160

PR #1181, opened 2015-07-23
Add a config option to validate inactive connections

chrisholmes:
This adds an option to `HttpClientConfiguration` named `validateAfterInactivityPeriod` which enables a 'stale' connection check with the frequency of the given period.

This check helps detect connections that have become stale (half-closed) while kept inactive in a client's connection pool.

This check is disabled by default, but is configurable as a duration.

##### Reason for introduction
The HTTP/1.1 specification indicates the responsibility is with the client to handle closed keep-alive connections gracefully, and that the server can be expected to close that session at any time with or without intention.

In the HTTP/1.1 RFC#7230 it is stated in section 6.3.1 that:

 >Connections can be closed at any time, with or without intention.
 >Implementations ought to anticipate the need to recover from asynchronous close events.

Without the 'stale' connection check in place we see regular broken connections due to this behaviour.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -789,24 +789,26 @@ See HttpClientConfiguration_  for more options.
       userAgent: <application name> (<client name>)
 
 
-========================= ======================================  =============================================================================
-Name                      Default                                 Description
-========================= ======================================  =============================================================================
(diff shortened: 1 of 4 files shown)


arteam:
Thanks for the contribution, for some reason we missed that option. Probably because the client wasn't originally designed to be long-lived.

This looks good to me.

PR #1184, opened 2015-07-23
Add missing closing parenthesis in configuration example

neurodesign:
Add missing closing parenthesis in MessageQueueFactory build method


joschi:
Good catch. Thanks!

PR #785, opened 2014-11-20
Sanitize URL pattern from rootPath

Toilal:
This is related to #784

PR #1185, opened 2015-07-24
Correct default value of rootPath

zodvik:
The documentation has default value of rootPath as `/` while it should be `/*`. Setting it to `/` causes 404 on resources


nickbabcock:
Ok, looks good to me. A couple notes.

Starting in 0.9.0, it does not matter if the `rootPath` ends with `/` or `*` (due to #785). For those that will be on earlier than 0.9.0 (which should be most at this point), the default is `/*` (it was just documented wrong in #665

Even though the code has the default `rootPath` of `/`, it does get converted to `/*` before being passed to jersey.

Good catch!

PR #1187, opened 2015-07-24
Update hibernate validator to 5.2.1.Final

nickbabcock:
Picked out a few interesting bits:

In some cases, when validating `Optional`, the `UnwrapValidatedValue` is not needed. Only one line of code in the entire code base was changed as the result of this. Since it's not entirely intuitive when `UnwrapValidatedValue` is needed, I've omitted changing the documentation because it still works with the annotation applied.

This doesn't affect dropwizard immediately, since we're not targeting jdk 8 right now, but with [new type annotations](https://blogs.oracle.com/java-platform-group/entry/java_8_s_new_type), it is possible to write code like:

```java
List<@NotEmpty String> strs; 
```

### DIFF
--- a/dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ValidatingResource.java
+++ b/dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ValidatingResource.java
@@ -52,7 +52,7 @@ public class ValidatingResource {
 
     @GET
     @Path("headCopy")
-    public String heads(@QueryParam("cheese") @NotNull IntParam secretSauce) {
+    public String heads(@QueryParam("cheese") @NotNull @UnwrapValidatedValue(false) IntParam secretSauce) {
         return secretSauce.get().toString();
     }
 
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -48,6 +48,7 @@ v0.9.0
 * Upgraded to Jadira Usertype Core 4.0.0.GA
 * Upgraded to Jersey 2.20
 * Upgraded to Jetty 9.2.13.v20150730
+* Upgraded to Hibernate Validator 5.2.1.Final
 * Upgraded to Jetty ALPN boot 7.1.3.v20150130
 * Upgraded to Jetty SetUID support 1.0.3
 * Upgraded to Liquibase 3.4.1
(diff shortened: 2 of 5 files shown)


nickbabcock:
Interesting, the travis build and maven build consistently fail due to code that wasn't changed. And I can't seem to make the same test fail in Intellij.

arteam:
Should we place ` @UnwrapValidatedValue(false)` on 
```java
@GET
@Path("headCopy")
public String heads(@QueryParam("cheese") @NotNull @UnwrapValidatedValue(false) IntParam secretSauce) {
    return secretSauce.get().toString();
}
```

nickbabcock:
That is the fix. I am worried that upgrading this dependency means that some applications may now break at runtime (only if they have insufficient tests).

nickbabcock:
Ok, fix applied. Note that in the Hibernate release, there is also JavaFx handler baked into Hibernate validator (hence why the test had to change because now there are more than three handlers). But it shouldn't be too egregious to keep the extra handler in.

nickbabcock:
Rebased against master and did a minor bump to 5.2.1.Final.

dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ValidatingResource.java
```diff
@@ -52,7 +52,7 @@ public String heads(@HeaderParam("cheese") @NotEmpty String secretSauce) {
 
     @GET
     @Path("headCopy")
-    public String heads(@QueryParam("cheese") @NotNull IntParam secretSauce) {
+    public String heads(@QueryParam("cheese") @NotNull @UnwrapValidatedValue(false) IntParam secretSauce) {
```
joschi: We should really make that change very obvious in the release/upgrade notes for Dropwizard 0.9.0. Otherwise many people won't even realize that their resources are broken.

natnan:
It would be really nice if this makes it to 0.9. As a (controversial) hardcore Optional user, I haven't been able to validate those values because of a [bug with @UnwrapValidatedValue](https://hibernate.atlassian.net/browse/HV-928), leaving the API half-baked. 

And the lack of `List<@NotEmpty String> strs; `, sigh..

arteam:
I don't see reasons, why we should not make it.

nickbabcock:
Great, I'll update the upgrade guide.

Issue #1192, opened 2015-07-27
WebApplicationException should be formatted


vzx:
In my eyes, the result of thrown `WebApplicationException` should be formatted like other unhandled bubbled exceptions, in `LoggingExceptionMapper`. Currently they are just passed through:

```java
        if (exception instanceof WebApplicationException) {
            return ((WebApplicationException) exception).getResponse();
        }
```

Any `WebApplicationException` which was thrown with a response entity set (eg. through `throw new WebApplicationException(Response.status(..).entity(...)...)`) will _not_ be passed to any `ExceptionMapper`: it's just processed in the way the thrower intended.

However, this leaves `WebApplicationExceptions` which were thrown with a message, but without any entity (eg. `throw new WebApplicationException("not found", 404)`). These end up in `LoggingExceptionMapper`, in the code branch shown above.

The problem is that even though a message was specified, the returned `Response` does not have any entity data. That means the error will end up in Jetty, which will just produce a generic HTML page for the associated HTTP status code, without the message.

It would be nicer to format these exceptions in the same other unhandled exceptions are formatted (like `{"code": 404, "message": "not found"}`).

Even if someone throws `WebApplicationException` with some specific format (eg. a `text/plain` entity), it will not be re-formatted, because these are not passed to the `ExceptionMapper`. And finally, if someone wants a different for _all_ `WebApplicationExceptions`, they can just implement their own `ExceptionMapper<WebApplicationException>` because the mappers are chosen by most specific first.

vzx:
I made a pull request for this: https://github.com/dropwizard/dropwizard/pull/1193

arteam:
Closed by #1193

PR #1193, opened 2015-07-27
#1192: Format WebApplicationException

vzx:
This would fix issue https://github.com/dropwizard/dropwizard/issues/1192.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/LoggingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/LoggingExceptionMapper.java
@@ -16,15 +16,23 @@ public abstract class LoggingExceptionMapper<E extends Throwable> implements Exc
 
     @Override
     public Response toResponse(E exception) {
+        final int status;
+        final ErrorMessage errorMessage;
+
         if (exception instanceof WebApplicationException) {
-            return ((WebApplicationException) exception).getResponse();
+            status = ((WebApplicationException) exception).getResponse().getStatus();
+            errorMessage = new ErrorMessage(status, exception.getLocalizedMessage());
+        } else {
+            final long id = ThreadLocalRandom.current().nextLong();
+            logException(id, exception);
+
+            status = Response.Status.INTERNAL_SERVER_ERROR.getStatusCode();
+            errorMessage = new ErrorMessage(formatErrorMessage(id, exception));
         }
 
-        final long id = ThreadLocalRandom.current().nextLong();
-        logException(id, exception);
-        return Response.serverError()
+        return Response.status(status)
                 .type(MediaType.APPLICATION_JSON_TYPE)
-                .entity(new ErrorMessage(formatErrorMessage(id, exception)))
+                .entity(errorMessage)
                 .build();
     }
 
--- a/dropwizard-jersey/src/test/java/io/dropwizard/jersey/errors/ExceptionResource.java
+++ b/dropwizard-jersey/src/test/java/io/dropwizard/jersey/errors/ExceptionResource.java
@@ -5,7 +5,9 @@ import com.fasterxml.jackson.databind.JsonMappingException;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 import java.io.IOException;
 
 @Path("/exception/")
@@ -20,5 +22,11 @@ public class ExceptionResource {
     @Path("json-mapping-exception")
     public void jsonMappingException() throws JsonMappingException {
         throw new JsonMappingException("BOOM");
+    }
+
+    @GET
+    @Path("web-application-exception")
+    public void webApplicationException() throws WebApplicationException {
+        throw new WebApplicationException("KAPOW", Response.Status.BAD_REQUEST);
     }
 }
--- a/dropwizard-jersey/src/test/java/io/dropwizard/jersey/errors/LoggingExceptionMapperTest.java
+++ b/dropwizard-jersey/src/test/java/io/dropwizard/jersey/errors/LoggingExceptionMapperTest.java
@@ -56,4 +56,17 @@ public class LoggingExceptionMapperTest extends JerseyTest {
                     + "\"There was an error processing your request. It has been logged (ID ");
         }
     }
+
+    @Test
+    public void formatsWebApplicationException() throws Exception {
+        try {
+            target("/exception/web-application-exception").request(MediaType.APPLICATION_JSON).get(String.class);
+            failBecauseExceptionWasNotThrown(WebApplicationException.class);
+        } catch (WebApplicationException e) {
+            final Response response = e.getResponse();
+
+            assertThat(response.getStatus()).isEqualTo(400);
+            assertThat(response.readEntity(String.class)).isEqualTo("{\"code\":400,\"message\":\"KAPOW\"}");
+        }
+    }
 }


jplock:
I like this as I tend to do this using custom exception mappers myself, but I'm slightly worried about breaking backward compatibility.

vzx:
> but I'm slightly worried about breaking backward compatibility.

Backwards compatibility in what way? The only case it will "break" is when someone throws a `WebApplicationException` without any entity (eg. just a message and status code, or even only just a status code), which end up in Jetty just returning a basic HTML page for the status code (and I can't imagine anyone relies on this). If a `WebApplicationException` has an entity, it bypasses the `ExceptionMapper` chain entirey.
You can check out my full description about it in issue https://github.com/dropwizard/dropwizard/issues/1192.

vzx:
How should I move this forward? Should I share this pull request or the issue with a certain team member, post it to some mailing list?

jplock:
@dropwizard/committers does anyone have any feedback on this one?

nickbabcock:
I think this is a great idea, and should be the default behavior. I am not too worried about people relying on this previous behavior, because there aren't many situations when one would want Jetty's error page. I also feel that most people like @jplock and myself, end up defining our own `WebApplicationException` to get similar behavior :smile:

So, this patch looks good to me. I'm sure it will save people from needing to define their own `WebApplicationException` down the road.

The thing I'm unsure of is if there is sufficient test coverage in this area. There are multiple constructors for `WebApplicationException`, and I'd like to see test cases for than just one so that the behavior is stated. For instance, a test case ensuring that if someone constructed an exception with a `Response` (with/without) an entity that it is passed through to the client.

ryankennedy:
My one concern is that we'll suddenly be taking an exception message and returning it to clients. If anyone does the following, for instance…

`throw new ForbiddenException("Don't tell Bob, but we've flipped the super secret flag that doesn't allow him to perform this action anymore.")`

Suddenly that message becomes public to Bob.

nickbabcock:
If my understanding is correct, the behavior only changes if `ForbiddenException` derives from `WebApplicationException`, in which case it makes sense for the client to see the error, right?

vzx:
@nickbabcock 
> The thing I'm unsure of is if there is sufficient test coverage in this area. There are multiple constructors for WebApplicationException, and I'd like to see test cases for than just one so that the behavior is stated. For instance, a test case ensuring that if someone constructed an exception with a Response (with/without) an entity that it is passed through to the client.

I definitely agree with this, however I'm unable to add a test case for this to `LoggingExceptionMapperTest`. Somehow in an actual Dropwizard application, a `WebApplicationException` which contains an entity never enters the `LoggingExceptionMapper#toResponse` (which can be verified quite easily manually by using a debugger). But when run from `LoggingExceptionMapperTest` the exact same exception does enter the `LoggingExceptionMapper#toResponse`, and so this test would produce a false positive.

Example of `WebApplicationException` with entity which works correctly in an actual app:
```java
@GET
@Path("web-application-exception-with-entity")
public void webApplicationExceptionWithEntity() throws WebApplicationException {
    throw new WebApplicationException("exception message", Response.status(Response.Status.NOT_FOUND)
        .type(MediaType.APPLICATION_JSON_TYPE).entity("{\"code\": 9001}").build());
}
```

Test case which fails when put into `LoggingExceptionMapperTest`:
```java
@Test
public void webApplicationExceptionWithEntityNotTouched() throws Exception {
    try {
        target("web-application-exception-with-entity").request(MediaType.APPLICATION_JSON).get(String.class);
        failBecauseExceptionWasNotThrown(WebApplicationException.class);
    } catch (WebApplicationException e) {
        final Response response = e.getResponse();

        assertThat(response.getStatus()).isEqualTo(404);
        assertThat(response.readEntity(String.class)).isEqualTo("{\"code\": 9001}");
    }
}
```

@ryankennedy 
As for `ForbiddenException`, it does derive in the end from `WebApplicationException` (through `ClientErrorException` which is usually generated in Jersey clients only). The bigger problem here is that any `ClientErrorException` is derived from `WebApplicationException` and as such "bleeds" through to the web app layer if not caught and handled correctly (this in my eyes is major flaw in JAX-RS/Jersey, but that's beside the point).
But even then, if any Jersey client throws some `ClientErrorException` with an entity (because let's say some remote server returns a rich error message), this will also be returned to Bob, regardless of the implementation `LoggingExceptionMapper` (remember any `WebApplicationException` with an entity will be passed through).

ryankennedy:
`WebApplicationException` and its subclasses "bleeding" through to the web app layer if not "caught and handled correctly" is not a major flaw in JAX-RS, it's the intended design pattern. For better or worse, this is how you get to have pretty resources methods like:

```java
@Path("/games/{id}/score")
public class SportScoreResource {
    private GameDAO gameDAO;

    @GET
    @UnitOfWork
    public SportScores getSportScores(long gameId) {
        Optional<Game> game = gameDAO.findById(gameId);
        return game.map(ApiMapper::gameToScore).orElseThrow(NotFoundException::new);
    }
}
```

Without being able to throw `NotFoundException` and have it generate a 404 the only way you'd be able to signal to JAX-RS that you want to generate a 404 would be to have your method return [javax.ws.rs.core.Response](http://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Response.html). Suddenly my JAX-RS is starting to look an awful lot like Servlets again. It's also far more difficult to test than my version, which is just a simple Java object that just happens to be annotated with some JAX-RS annotations.

```if any Jersey client throws some ClientErrorException with an entity (because let's say some remote server returns a rich error message), this will also be returned to Bob```

Yes, that is indeed the [documented behavior of `WebApplicationException`](http://docs.oracle.com/javaee/7/api/javax/ws/rs/WebApplicationException.html#WebApplicationException-javax.ws.rs.core.Response-). However, the documentation [does not indicate that an accompanying message will be sent back to the user](http://docs.oracle.com/javaee/7/api/javax/ws/rs/WebApplicationException.html#WebApplicationException-java.lang.String-).

For the record, I do actually wish there was a version of `WebApplicationException` that allowed me to quickly pass a string into the response body for surfacing in the HTTP response. I don't think we ought to be co-opting the exception message for that purpose, though.

**Comment by vzx** [comment]
@ryankennedy I'm very sorry for the confusion I caused. I misunderstood something about the `ClientErrorException`, and did not communicate my point about the design flaw clear. The `NotFoundException` is a `ClientErrorException`, which I thought were  are usually only thrown from JAX-RS clients, when a request returns any faulty response. But it's not related to JAX-RS clients specifically.

The design flaw I meant was about using a JAX-RS `Client`. For example, consider:

```java
ClientBuilder.newClient().target("..").request().get(String.class);
```

When the URL returns 404, it will throw a `NotFoundException` (and in case of some other non-successful error, it will throw some other `WebApplicationException`). So if my app A communicates with app B in this way, and doing the request to app B results in some error, a `WebApplicationException` is generated, which then bubbles up to the web layer, and is translated into an HTTP response. So if app B is internal and its error output is not intended for the whole wide world, I have to take care to catch `WebApplicationException`, and re-throw it in some other way.

> For the record, I do actually wish there was a version of WebApplicationException that allowed me to quickly pass a string into the response body for surfacing in the HTTP response. I don't think we ought to be co-opting the exception message for that purpose, though.

Personally I always expect the exception message of the `WebApplicationException` to be the message to show to the API user (and as such construct my own `ExceptionMapper` for `WebApplicationException` to do so), but perhaps this breaks other people's expectations. I'm not sure how else to implement it though because as you say there is no other way to carry a message inside the `WebApplicationException`.

ryankennedy:
Ah, now I see what you're saying, @vzx. Yeah…it does seem like a flaw that your JAX-RS resource may use a JAX-RS client to make a request, which may throw a WebApplicationException that will propagate all the way to the caller of your resource just because WebApplicationException is a descendant of RuntimeException.

Now I find myself wanting ClientWebApplicationException and ServiceWebApplicationException where only ServiceWebApplicationException gets mapped to an appropriate HTTP status code, while ClientWebApplicationExceptions get turned into a generic HTTP 500. *sigh*

I'm still slightly concerned about the solution's ability to leak information from the exception message, but I'm now clearer about your frame of reference. Thanks for the explanation. I'm a tentative +1 on seeing this merged. My only hesitation would be to have a few more @dropwizard/committers chime in on how they feel about having the exception message exposed.

arteam:
I'm in a boat with @jplock and @nickbabcock. This patch looks fine to me and I don't think we break anything with it. 

We can add additional tests in future. As I see it, the problem lies is Jersey test containers, which not always behave as the real container.

arteam:
Thanks, @vzx!

PR #1196, opened 2015-07-28
Fix DefaultLoggingFactoryTest on Windows

nickbabcock:
Also it is unnecessary to delete the log files after the test is done
because the log files are zeroed out when the test runs.

### DIFF
--- a/dropwizard-logging/src/test/java/io/dropwizard/logging/DefaultLoggingFactoryTest.java
+++ b/dropwizard-logging/src/test/java/io/dropwizard/logging/DefaultLoggingFactoryTest.java
@@ -6,13 +6,20 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.TextNode;
 import com.google.common.base.Charsets;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Files;
 import com.google.common.io.Resources;
 import io.dropwizard.configuration.ConfigurationFactory;
+import io.dropwizard.configuration.FileConfigurationSourceProvider;
+import io.dropwizard.configuration.SubstitutingSourceProvider;
 import io.dropwizard.jackson.Jackson;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.text.StrSubstitutor;
 import org.assertj.core.data.MapEntry;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 import org.slf4j.LoggerFactory;
 
 import javax.validation.Validation;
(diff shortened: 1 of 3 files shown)


arteam:
It's a bummer, that we can't run the build on Windows in Travis CI. The amount of Windows test fixes is a clear sign, that there are plenty of developers on this OS, who hack on Dropwizard.

joschi:
> Also it is unnecessary to delete the log files after the test is done because the log files are zeroed out when the test runs.

I don't agree. If the log files aren't removed after the test run, we don't know if they were created in the following test runs or if there was an error – unless we clean the complete working directory (e. g. by running `mvn clean`).

arteam:
I don't like the proposed solution, because it writes log files to the current directory with source code.

It would be better if we created a temporary file by `File.createTempFile` and passed its path to the configuration factory by a system property. The problem is that `dw.loggers.com.example.newApp.appenders.currentLogFilename` will not work, because the parser
will treat _com_, _example_, and _newApp_ as keys, but not as a whole property. And I don't know how to escape dots.

arteam:
As a sidenote, in future we should probably retrofit the parser to support escaping with something like this: `dw.loggers.com{dot}example{dot}newApp.appenders[0].currentLogFilename`

nickbabcock:
I was following the example set in other tests in `FileAppenderFactoryTest`, which have the same behavior (writes log files to current directory and doesn't delete them afterwards). I can rework the tests to see if I can take advantage of `TemporaryFolder`

nickbabcock:
The commit I just pushed fixes everything in `dropwizard-logging` that outputs logs to the current working directory. Now logs are stored in a cross platform temporary folder. It's not exactly the prettiest but leaves `logging_advanced.yml` in a sane state, so that it can be used as an example.

arteam:
What do you think about using `StrSubstitutor`?
We could replace actual locations in the yml file by variables and then initialize them in runtime.
Something like this:
````yml
level: INFO
loggers:
  "com.example.app": INFO
  "com.example.newApp":
    level: DEBUG
    appenders:
      - type: file
        currentLogFilename: ${new_app_log}
        archivedLogFilenamePattern: ${new_app_archive_log}
        logFormat: "%-5level %logger: %msg%n"
        archivedFileCount: 5
  "com.example.legacyApp":
    level: DEBUG
appenders:
  - type: console
  - type: file
    currentLogFilename: ${default_log}
    archivedLogFilenamePattern: ${default_archive_log}
    logFormat: "%-5level %logger: %msg%n"
    archivedFileCount: 5
````
```java
final File newAppLog = folder.newFile("example-new-app.log");
final File defaultLog = folder.newFile("example.log");
final StrSubstitutor substitutor = new StrSubstitutor(ImmutableMap.of(
        "new_app_log", newAppLog.getAbsolutePath(),
        "new_app_archive_log", "/tmp/example-new-app-%d.log.gz",
        "default_log", defaultLog.getAbsolutePath(),
        "default_archive_log", "/tmp/example-%d.log.gz"
));

final String filePath = new File(Resources.getResource("yaml/logging_advanced.yml")
.toURI()).getAbsolutePath();
config = factory.build(new SubstitutingSourceProvider(
new FileConfigurationSourceProvider(), substitutor), filePath);
```

nickbabcock:
Confirmed.  That is prettier code. I shortened it to use 2 substituted variables instead of 4.

arteam:
Great. This looks a lot better now.

PR #1199, opened 2015-07-30
Use inmemory jersey server for view tests

nickbabcock:
The view tests were the only ones that used the `jdk-http` provider. Once I swapped it out for `inmemory`, the tests ran 25 seconds faster.


joschi:
LGTM. :+1:

PR #1200, opened 2015-07-30
Fixed dropwizard jersey duplicate path logging

afrin216:
So this was a tiny bug I discovered while using dropwizard (which is awesome by the way) on our project. The problem being if there are two separate functions present exposing same path in jersey but supporting different media type, then dropwizard logs duplicate paths.
Eg:
```
@Get
@Path("/abc")
@Produces(MediaType.TEXT_HTML)
public Respone functionA(){
   //some code produces html output
}

@Get
@Path("/abc")
public Respone functionB(){
   //some code produces json output
}
```
Now dropwizard logs the following
```
GET /abc MyResourceClass
GET /abc MyResourceClass
```
This PR fixes that and logs just one line by using a TreeSet instead of List. Its passive and doesn't affect anything outside this class (There could be an enhancement to log the one line and list all supported content types, for which I might do a PR later).

Thanks and continue being awesome guys!

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -3,8 +3,10 @@ package io.dropwizard.jersey;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.jersey2.InstrumentedResourceMethodApplicationListener;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Objects;
 import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.Sets;
+
 import io.dropwizard.jersey.caching.CacheControlledResponseFeature;
 import io.dropwizard.jersey.guava.OptionalMessageBodyWriter;
 import io.dropwizard.jersey.guava.OptionalParamFeature;
@@ -26,8 +28,10 @@ import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.Path;
 import javax.ws.rs.ext.Provider;
+
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
+import java.util.Comparator;
 import java.util.Set;
 
 public class DropwizardResourceConfig extends ResourceConfig {
@@ -111,7 +115,7 @@ public class DropwizardResourceConfig extends ResourceConfig {
 
     public String getEndpointsInfo() {
         final StringBuilder msg = new StringBuilder(1024);
-        final Set<EndpointLogLine> endpointLogLines = Sets.newTreeSet();
+        final Set<EndpointLogLine> endpointLogLines = Sets.newTreeSet(new EndpointComparator());
 
         msg.append("The following paths were found for the configured resources:");
         msg.append(NEWLINE).append(NEWLINE);
@@ -193,12 +197,10 @@ public class DropwizardResourceConfig extends ResourceConfig {
         }
     }
 
-    private static class EndpointLogLine implements Comparable<EndpointLogLine>, Serializable {
+    private static class EndpointLogLine {
         private final String httpMethod;
         private final String basePath;
         private final Class<?> klass;
-
-        private static final long serialVersionUID = 1L;
 
         public EndpointLogLine(String httpMethod, String basePath, Class<?> klass) {
             this.basePath = basePath;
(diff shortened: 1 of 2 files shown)


dropwizard-jersey/src/test/java/io/dropwizard/jersey/DropwizardResourceConfigTest.java
```diff
@@ -118,6 +120,22 @@ public void logsNestedEndpoints() {
...
+                + "    GET     /anotherMe (io.dropwizard.jersey.DropwizardResourceConfigTest.TestDuplicateResource)%n"
+                + "    GET     /callme (io.dropwizard.jersey.DropwizardResourceConfigTest.TestDuplicateResource)%n");
+
+        final String shouldNotContainLog = String.format(
+                  "    GET     /callme (io.dropwizard.jersey.DropwizardResourceConfigTest.TestDuplicateResource)%n"
+                + "    GET     /callme (io.dropwizard.jersey.DropwizardResourceConfigTest.TestDuplicateResource)%n"
+                + "    GET     /callme (io.dropwizard.jersey.DropwizardResourceConfigTest.TestDuplicateResource)%n");
+
+        assertThat(rc.getEndpointsInfo()).contains(expectedLog);
+        assertThat(rc.getEndpointsInfo()).doesNotContain(shouldNotContainLog);
```
nickbabcock: Would it make sense to have

```java
assertThat(rc.getEndpointsInfo())
        .containsOnlyOnce()("    GET     /callme (io.dropwizard.jersey.DropwizardResourceConfigTest.TestDuplicateResource"))
```

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -213,18 +210,52 @@ public EndpointLogLine(String httpMethod, String basePath, Class<?> klass) {
...
-                .compare(endpointA.basePath, endpointB.basePath)
-                .compare(endpointA.httpMethod, endpointB.httpMethod)
+                .compare(this.basePath, other.basePath)
+                .compare(this.httpMethod, other.httpMethod)
                 .result();
         }
+
+        @Override
+        public int hashCode() {
+            final int prime = 31;
```
nickbabcock: Instead of hand writing `hashCode` can you use `Objects.hash`?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -114,7 +111,7 @@ public void setUrlPattern(String urlPattern) {
 
     public String getEndpointsInfo() {
         final StringBuilder msg = new StringBuilder(1024);
-        final List<EndpointLogLine> endpointLogLines = Lists.newArrayList();
+        final Set<EndpointLogLine> endpointLogLines = Sets.newTreeSet();
```
nickbabcock: `newTreeSet` accepts a `Comparator`, any reason why you couldn't reuse `EndpointComparator`?

nickbabcock:
Hey, thanks for the PR, I like the direction (I too think it'd be neat if supported content-types were logged with the endpoints, yet that could make for some long log lines), but I had a few questions in my code comments that should be addressed.

I do wonder if anyone depends on the current behavior to know that multiple content types were registered...hmm

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
afrin216: My bad. Reverted!

afrin216:
@nickbabcock  Addressed your comments. Thanks.

Yeah it could make the logs a little longer. But we could organise it so that only for those endpoints with restricted return types, further info such as content-types gets printed. Eg :-
````
GET /abc MyResourceClass
        Accepts  - application/json,  text/html
        Produces - application/xml
GET /abc2 MyResourceClass
GET /abc3 MyResourceClass
GET /abc4 MyResourceClass
````

Would be useful atleast during debugging.

nickbabcock:
Ohhh, I personally really like that output, but I'll see if others have opinions on that.

Also you can simplify the code even more. There is no need to override `hashCode` or `equals` :smile:

afrin216:
Eh right! Changes do look cleaner now. Thanks! :smile:

jplock:
My recommendation would be to keep this PR focused on fixing the path logging issue and if we want to change around the output format, do that in another PR targeted for a future release.

nickbabcock:
@jplock, absolutely -- keep pull requests focused!

Squashed and merged PR as part of c91851b

Thanks, @afrin216

Issue #1201, opened 2015-07-31
Clone method in ChainedAuthFactory always set required to true


mtakaki:
The class `ChainedAuthFactory` always set the parameter `required` to true in the `clone` method. This is a snippet of the method:

```
@Override
public AuthFactory<Object, T> clone(boolean required) {
    ChainedAuthFactory<T> clone = new ChainedAuthFactory<>();
    for (AuthFactory<?, T> factory : factories) {
        clone.addChainedProvider(factory.clone(true));
    }
    return clone;
}
```

Instead of calling `factory.clone(true)`, it should be `factory.clone(required)`. This makes the authentication always required when `ChainedAuthFactory` is used.

nickbabcock:
Closed by #1202

PR #1202, opened 2015-07-31
Fixing issue with ChainedAuthFactory to allow not required authentication

mtakaki:
Fixing issue #1201.

`ChainedAuthFactory` currently defaults `required` parameter to true, instead of using the annotation `required` value.


jplock:
Good catch!  Thanks.

PR #1204, opened 2015-07-31
Add logging of explicitly mentioned consumes and produces media type for jersey end points

afrin216:
This is a minor enhancement based on a discussion in yesterday's PR - <a href="https://github.com/dropwizard/dropwizard/pull/1200#issuecomment-126455960">Link</a>.

What this does it for cases where a person has two or three functions for handling different media types of the same endpont, this prints out the consumes & produces media types, if available. Note that if there are no @Produces or @Consumes annotation for an endpoint then nothing is logged, thereby making the logging appear the same to previous users of dropwizard. (I haven't had to modify any of the other unit tests, which prove this point)

Example:-
````
GET /abc MyResourceClass
        Accepts  - application/json,  text/html
        Produces - application/xml
GET /abc2 MyResourceClass
GET /abc3 MyResourceClass
GET /abc4 MyResourceClass
````

One caveat of this change is that I had to revert back to Lists from Sets (as Sets don't support get/set operation to add-on media types from different functions for the same end point) thereby rendering my PR yesterday useless :cry: .

I'm open to suggestions/modifications/even total rejection of this idea itself :wink: . No worries!

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -3,14 +3,18 @@ package io.dropwizard.jersey;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.jersey2.InstrumentedResourceMethodApplicationListener;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Objects;
 import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+
 import io.dropwizard.jersey.caching.CacheControlledResponseFeature;
 import io.dropwizard.jersey.guava.OptionalMessageBodyWriter;
 import io.dropwizard.jersey.guava.OptionalParamFeature;
 import io.dropwizard.jersey.params.NonEmptyStringParamFeature;
 import io.dropwizard.jersey.sessions.SessionFactoryProvider;
 import io.dropwizard.jersey.validation.HibernateValidationFeature;
+
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.server.ServerProperties;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java [hunk 1]
```diff
@@ -206,8 +232,48 @@ public EndpointLogLine(String httpMethod, String basePath, Class<?> klass) {
 
         @Override
         public String toString() {
-            return String.format("    %-7s %s (%s)", httpMethod, basePath, klass.getCanonicalName());
+            StringBuilder builder = new StringBuilder(String.format("    %-7s %s (%s)", httpMethod, basePath, klass.getCanonicalName()));
+            if(!consumedMediaTypes.isEmpty()){
+                builder.append(System.lineSeparator() + String.format("         Accepts  -  %s", Arrays.toString(consumedMediaTypes.toArray())));
```
mattnelson: The line header should be Consumes.

mattnelson:
What if you had two methods with the same uri that produce and consume different types?

```java
@POST
@Path("/abc")
@Consumes(MediaType.APPLICATION_XML)
@Produces(MediaType.TEXT_HTML)
public Response functionA(){
   //some code produces html output
}

@POST
@Path("/abc")
@Consumes(MediaType.APPLICATION_JSON)
public Response functionB(){
   //some code produces json output
}
```

```
GET /abc MyResourceClass
        Consumes  - application/xml
        Produces - text/html
GET /abc MyResourceClass
        Consumes  - application/json
        Produces - application/json

vs

GET /abc MyResourceClass
        Consumes  - application/xml, application/json
        Produces - text/html, application/json
```

arteam:
I would prefer to keep logging Jersey resources simple. There are solution for REST API documentation like [Swagger](http://swagger.io) or [RAML](http://raml.org).

If anyone has another opinion, feel free to start a discussion. For the time being, I am not accepting this PR.

PR #1215, opened 2015-08-12
Add additivity setting to LoggerConfiguration

smcwhtdtmc:
This is related to dropwizard#868 and dropwizard@7ab73cd

With additive=true (the default), a logger writes to its appender -and- any appenders hierarchically above it, including the root logger. For true redirection as described in issue 868, we need to be able to set additive=false so a logger writes to its appender only.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -414,6 +414,7 @@ Logging
         "io.dropwizard": INFO
         "org.hibernate.SQL":
           level: DEBUG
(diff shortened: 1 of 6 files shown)


lehcim:
Last night I was wondering how to configure Dropwizard logging with different file appenders and additivity setting. Today morning I'm reading your PR ...
Thank you very much, you made my day!

joschi:
LGTM. Thanks!

PR #1216, opened 2015-08-13
1160 add shareConnectionManagerEnabled option with 'true' by default.

krutsko:
This PR fix bug #1160. Generally, it's copied from https://github.com/dropwizard/dropwizard/issues/1179 with minor addition regarding dependency on org.apache.httpcomponents:httpclient.

Please fill free to reject it and grab the original one provided by @carlo-rtr. We really need this to be released in 0.8.3. Thank you!

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -852,24 +852,25 @@ See HttpClientConfiguration_  for more options.
       userAgent: <application name> (<client name>)
 
 
-========================= ======================================  =============================================================================
-Name                      Default                                 Description
-========================= ======================================  =============================================================================
-timeout                   500 milliseconds                        The maximum idle time for a connection, once established.
-connectionTimeout         500 milliseconds                        The maximum time to wait for a connection to open.
-connectionRequestTimeout  500 milliseconds                        The maximum time to wait for a connection to be returned from the connection pool.
-timeToLive                1 hour                                  The maximum time a pooled connection can stay idle (not leased to any thread)
-                                                                  before it is shut down.
-cookiesEnabled            false                                   Whether or not to enable cookies.
-maxConnections            1024                                    The maximum number of concurrent open connections.
-maxConnectionsPerRoute    1024                                    The maximum number of concurrent open connections per route.
-keepAlive                 0 milliseconds                          The maximum time a connection will be kept alive before it is reconnected. If set
-                                                                  to 0, connections will be immediately closed after every request/response.
-retries                   0                                       The number of times to retry failed requests. Requests are only
-                                                                  retried if they throw an exception other than ``InterruptedIOException``,
-                                                                  ``UnknownHostException``, ``ConnectException``, or ``SSLException``.
-userAgent                 ``applicationName`` (``clientName``)    The User-Agent to send with requests.
-========================= ======================================  =============================================================================
+============================= ======================================  =============================================================================
+Name                          Default                                 Description
+============================= ======================================  =============================================================================
+timeout                       500 milliseconds                        The maximum idle time for a connection, once established.
+connectionTimeout             500 milliseconds                        The maximum time to wait for a connection to open.
+connectionRequestTimeout      500 milliseconds                        The maximum time to wait for a connection to be returned from the connection pool.
+timeToLive                    1 hour                                  The maximum time a pooled connection can stay idle (not leased to any thread)
+                                                                      before it is shut down.
+cookiesEnabled                false                                   Whether or not to enable cookies.
+shareConnectionManagerEnabled true                                    Whether or not to share the connection manager.
+maxConnections                1024                                    The maximum number of concurrent open connections.
+maxConnectionsPerRoute        1024                                    The maximum number of concurrent open connections per route.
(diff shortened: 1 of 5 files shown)


nickbabcock:
Thanks for the PR and the investigation in #1160. Since there are no tests, do you have confirmation that this does, indeed, fix #1160 (ie. running 0.8.2 and 0.8.3-SNAPSHOT side by side, the bug shows in 0.8.2 and not 0.8.3-SNAPSHOT)?

Also, under what circumstances one would want `shareConnectionManagerEnabled` to be false in Dropwizard? Becaues if it should _always_ be true, we wouldn't want users to accidentally run into this hard-to-track behavior by changing this configuration option

krutsko:
@nickbabcock yes, we have running a service with the fix, and the bug not show up.

In our case with 0.8.2, we create a Jersey Client (JC) for each upstream dependency and we never close() these clients. On each request JC creates the ClientRuntime and at some point ClientRuntime get finalized and shutdown the connection manager of the entire client and the next request lead to "Connection pool shut down". So in my understanding, the notion of `shareConnectionManagerEnabled` mean "Connection Manager is shared between multiple ClientRuntimes".

I can imagine the case when `shareConnectionManagerEnabled` is false, then JerseyClient should be created/closed explictly on each request.

Anyway, I will try to prepare the test for my PR.

krutsko:
@nickbabcock I have added few tests, please take a look. Thanks!

nickbabcock:
Awesome, thanks for the tests. They look a little non-deterministic, but that is alright in my book :smile:

@arteam / team, did you want to take a look at this as well? I know that you had some concerns regarding the original solution, so you should take a look before this is merged.

arteam:
The actual problem is that Jersey expects that we create a new instance of the Apache client in 
`ConnectorProvider`, but in the fact we reuse the same one. See [this](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java#L363). 

We can't create a new instance every time, because we can't create several instances of `InstrumentedHttpClientConnectionManager` with the same `MetricRegistry`. 

Instead of enabling sharing of the connection manager, I propose not closing the client in `DropwizardApacheConnector` and managing it through Dropwizard lifecycle as was done in e4256d1.
Doing so, we don't leak the connection manager. And in the same time users can close an instance of the client/connection manager manually, if they want to.

arteam:
I've created #1232 with this approach. If it's merged, I will backport it to 0.8.3

carlo-rtr:
This approach seems good to me, thanks for looking. A couple of things


   - What you said about it expecting new instances of the client makes
   sense. Looking at ApacheConnectorProvider, that's what it's doing
   - The test case you wrote, did that actually fail without nuking the
   close() call?
   - Here <https://github.com/dropwizard/dropwizard/issues/1160> you talked
   about leaking the connection manager, and that connections would only be
   closed during major GCs. How exactly does that happen? I ask, because I was
   thinking if there is a reference that is leaked then it won't get GCed
   until the reference is gone, unless it's a weak reference.

arteam:
> The test case you wrote, did that actually fail without nuking the close() call?

Yes, it did, right after second `System.gc` call. Unfortunately, I can't write a test case that reproduces it.

> Here <https://github.com/dropwizard/dropwizard/issues/1160> you talked about leaking the connection manager, and that connections would only be closed during major GCs. How exactly does that happen? I ask, because I was thinking if there is a reference that is leaked then it won't get GCed until the reference is gone, unless it's a weak reference.

When we set a new property to `JerseyWebTarget`, Jersey creates a new `ClientConfig` with a copy of `ClientConfig.State` instance with new parameters. The old instance becomes unreachable as well as its `ClientRuntime` field, which then triggers closing the `Connector` instance.

See [this](https://github.com/jersey/jersey/blob/master/core-client/src/main/java/org/glassfish/jersey/client/JerseyWebTarget.java#L125), [this](https://github.com/jersey/jersey/blob/master/core-client/src/main/java/org/glassfish/jersey/client/ClientConfig.java#L540), [this](https://github.com/jersey/jersey/blob/master/core-client/src/main/java/org/glassfish/jersey/client/ClientConfig.java#L105), and [this](https://github.com/jersey/jersey/blob/master/core-client/src/main/java/org/glassfish/jersey/client/ClientConfig.java#L209).

So, as a consequence, if we remove  
```
.property(ClientProperties.READ_TIMEOUT, SLEEP_TIME_IN_MILLIS * 2)
```
call in the test, everything will be alright, because the state doesn't change.

krutsko:
just my 0.02€, I like the idea of fixing this by https://github.com/dropwizard/dropwizard/pull/1232 and looking forward to have this backported into 0.8.3.

carlo-rtr:
this was merged https://github.com/dropwizard/dropwizard/pull/1232 and back ported to v0.8.3

carlo-rtr:
also, @arteam thanks for the detailed explanation. That makes sense.

arteam:
Backported to 0.8.3 in faf1f94

carlo-rtr:
I just realized I missed everything else that was on https://github.com/dropwizard/dropwizard/tree/release/0.8.x

I'm going to get faf1f94 into this branch and release 0.8.4

PR #1218, opened 2015-08-17
Add support for Logback's FixedWindowRollingPolicy

dennyac:
With [Fixed Window Rolling Policy](http://logback.qos.ch/manual/appenders.html#FixedWindowRollingPolicy) (archivedLogFilenamePattern: -%i), we have control on the size (maxFileSize) and the number of log files (archiveFileCount) that will reside on the host. We will not have to worry about disk usage warnings. 

Recently Size and Time Based Archiving (archivedLogFilenamePattern: -%d-%i) with maxFileSize option has been added as part of commit: 343680b. In this archival policy we have control on the size of each log file (it will rotate the logs and increment the index after maxFileSize is reached), but the archiveFileCount only applies to the number of days worth of files that should be maintained. So on any specific day, files will keep getting rolled over indefinitely with a new index when the maxFileSize is reached. These log files will continue to accumulate as long as it is within the number of days mentioned in archiveFileCount. So we will lose control on the number and size of log files on the host, and disk usage may be a concern.

Proposed Approach: 
Include FixedWindowRollingPolicy in FileAppenderFactory. Rolling policy to be used will be based on the archivedLogFilenamePattern - 
Time Based Rolling Policy - /tmp/application-%d{yyyy-MM-dd}.log.gz
Size and Time Based Rolling Policy - /tmp/application-%d{yyyy-MM-dd}-%i.log.gz
Fixed Window Rolling Policy - /tmp/application-%i.log.gz

I hope you find this useful. I'm open to suggestions as well. Thanks.


dennyac:
PS: Errors are unrelated to this change.

dennyac:
@jplock Please let me know if you have any suggestions, changes. Thanks!

dropwizard-logging/src/main/java/io/dropwizard/logging/FileAppenderFactory.java
```diff
@@ -6,11 +6,7 @@
 import ch.qos.logback.core.FileAppender;
 import ch.qos.logback.core.Layout;
 import ch.qos.logback.core.encoder.LayoutWrappingEncoder;
-import ch.qos.logback.core.rolling.DefaultTimeBasedFileNamingAndTriggeringPolicy;
-import ch.qos.logback.core.rolling.RollingFileAppender;
-import ch.qos.logback.core.rolling.SizeAndTimeBasedFNATP;
-import ch.qos.logback.core.rolling.TimeBasedFileNamingAndTriggeringPolicy;
-import ch.qos.logback.core.rolling.TimeBasedRollingPolicy;
+import ch.qos.logback.core.rolling.*;
```
jplock: Can you revert this back to not using a wildcard?  Then I think we're good to go.  Thanks.

dennyac: Makes sense. I've made the changes you have suggested.

nickbabcock:
I think this is a great feature because I just got bit twice in a short period of time by two different applications that spammed the log files and filled up the disk.

PR #1220, opened 2015-08-17
Add the ability for some enums to still deserialize properly

brentryan:
This is just a small enhancement to the fuzzy enum deserializer so that it works when some enums (but not all) use toString()

### DIFF
--- a/dropwizard-jackson/src/main/java/io/dropwizard/jackson/FuzzyEnumModule.java
+++ b/dropwizard-jackson/src/main/java/io/dropwizard/jackson/FuzzyEnumModule.java
@@ -26,11 +26,17 @@ import java.util.List;
  * </ul>
  */
 public class FuzzyEnumModule extends Module {
+
     private static class PermissiveEnumDeserializer extends StdScalarDeserializer<Enum<?>> {
+
         private static final long serialVersionUID = 1L;
 
         private final Enum<?>[] constants;
         private final List<String> acceptedValues;
+
+        //These values are only provided if the @JsonCreator annotation is provided
+        private DeserializationConfig config;
+        private AnnotatedMethod am;
 
         @SuppressWarnings("unchecked")
         protected PermissiveEnumDeserializer(Class<Enum<?>> clazz) {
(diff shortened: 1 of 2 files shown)


arteam:
Thanks for the pull request, but I have a couple of questions:

* I think we already handle the `@JsonCreator` annotation on factory method in enums. If the annotation is present, we use the specific deserializer based on this method. Could you provide a test case, where it doesn't work as expected?
* I don't think we should try to  deserialize an enum based on the `toString` method, unless the user didn't explicitly configured the mapper to do so. It's not obvious and could break backward compatibility for users, who use `toString` just for logging purposes. If enabling this feature in the mapper is not an option, the user can create a factory method with `@JsonCreator` or register a custom deserializer.

brentryan:
@arteam Please see the commit.  I already provided a unit test that causes a failure with the current FuzzyEnumModule used in dropwizard.  The patch I provided fixes this issue.  If there's a different/better way to fix this issue then I'm fine with that as well, but this current approach should maintain backwards compatibility and allow for more flexible enums.

Even when using tools like jsonschema2pojo, you get enums generated that can look like this:
@Generated("org.jsonschema2pojo")
public enum DeployedEnvironmentType {

    DEV("development"),
    ALPHA("ALPHA"),
    LOAD("LOAD"),
    STAGING("STAGING"),
    PRODUCTION("PRODUCTION");
    private final String value;
    private static Map<String, DeployedEnvironmentType> constants = new HashMap<String, DeployedEnvironmentType>();

    static {
        for (DeployedEnvironmentType c: values()) {
            constants.put(c.value, c);
        }
    }

    private DeployedEnvironmentType(String value) {
        this.value = value;
    }

    @JsonValue
    @Override
    public String toString() {
        return this.value;
    }

    @JsonCreator
    public static DeployedEnvironmentType fromValue(String value) {
        DeployedEnvironmentType constant = constants.get(value);
        if (constant == null) {
            throw new IllegalArgumentException(value);
        } else {
            return constant;
        }
    }

}

Without my change then if you passed in "development" it wouldn't deserialize properly.

dropwizard-jackson/src/main/java/io/dropwizard/jackson/FuzzyEnumModule.java
```diff
@@ -48,22 +50,36 @@ protected PermissiveEnumDeserializer(Class<Enum<?>> clazz) {
...
             for (Enum<?> constant : constants) {
                 if (constant.name().equalsIgnoreCase(text)) {
                     return constant;
                 }
             }
 
+            //In some cases there are certain enums that don't follow the same patter across an enterprise.  So this
+            //means that you have a mix of enums that use toString(), some use @JsonCreator, and some just use the
+            //standard constant name().  This block handles finding the proper enum by toString()
+            String toStringText = jp.getText()
```
arteam: Should we really replace hyphens and underscores, if we want deserialize on the result of the `toString` method?
It could return a string with these symbols. For example, something like that:
```java
private enum CurrencyCode {
    USD("united_states_dollar"),
    EUR("euro"),
    GBP("pound_sterling");

    private final String description;

    CurrencyCode(String name) {
        this.description = name;
    }

    @Override
    public String toString() {
        return description;
    }
}
```

arteam:
Okay, I am convinced now that the problem is worth to fix and the patch would not break anything. I only have a one question, please see an inline comment.

Also, it would be great if you squashed your commits into a single change.

brentryan:
I agree.  I'll make the adjustments above and squash the commits down.

jplock:
@arteam do you want to get this into 0.9 as soon as @brentryan makes the requested changes?

arteam:
Probably not. I think we should add only bugfixes for 0.9. Will happily merge the PR to 1.0, though.

brentryan:
I'll re-create a new PR with the squash.

brentryan:
See https://github.com/dropwizard/dropwizard/pull/1316 for details on the new PR

PR #1221, opened 2015-08-17
Fix handling of empty request entities that don't have any annotation …

brentryan:
Some request bodies are optional and if you POST an empty request body with 0 annotations like @Valid then this code blows up.  This fix allows it to work as expected.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JacksonMessageBodyProvider.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JacksonMessageBodyProvider.java
@@ -69,14 +69,14 @@ public class JacksonMessageBodyProvider extends JacksonJaxbJsonProvider {
     }
 
     private Object validate(Annotation[] annotations, Object value) {
-        if (null == value) {
-            throw new ConstraintViolationException("The request entity was empty",
-                    Collections.<ConstraintViolation<Object>>emptySet());
-        }
-
         final Class<?>[] classes = findValidationGroups(annotations);
 
         if (classes != null) {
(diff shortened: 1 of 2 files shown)


nickbabcock:
Thanks for the pull request -- it is similar to #633

The proposed fix could be better, though I can be convinced otherwise. The reason being is that to be consistent with Hibernate Validator, the check should be if `@NotNull` is present

brentryan:
Sounds good.  I know we had this patched a while back off of 0.7.X code for Cvent and I decided to contribute back our fix.  But if you guys have a better option with the @NotNull check then that should work as well.  I just wanted to see a fix get into the mainline for this so that we can stop running off of our own fork if possible.

brentryan:
What's happening here? Are these things going to get into .9 release? or soon after?

jplock:
@dropwizard/committers this seems like a good fix to include in 0.9 before we tag the final release?  Any objections?

nickbabcock:
I'm fine with this being in 0.9, but I just want to state that with this and #1251, 0.8, 0.9, and 1.0 will all have different behaviors regarding null entities

brentryan:
@nickbabcock What's the status of this item?  This and the fuzzy enum change I submitted are the only things currently holding us back from not using a forked version.  Do we have an ETA for this OR a solution via #1251 ?

nickbabcock:
Barring anything unforeseen, #1251 will be the solution for 1.0 (and you'l have to add the `@NotNull` annotation on your endpoints). I'm fine with leaving this open for additional discussions until #1251 is merged.

arteam:
#1251 is merged and we can proceed here further.

PR #1224, opened 2015-08-18
Support configuration in yml of client TLS properties

chrisholmes:
This PR introduces the ability to modify the TLS behaviour of a HTTP client through a Dropwizard application's configuration file. The major benefit here is the ability to easily configure a client to connect to servers that are secured in non-trivial ways (e.g. client authentication or custom PKI).

The key features that can be configured are:

* Disable hostname verification;
* Use a custom truststore (when dealing with a custom PKI);
* Use a custom keystore (for client authentication);
* Allow self-signed certificates if necessary;
* Specify supported protocols (e.g. TLSv1.2 only);
* Specify supported ciphers.


jplock:
This looks good, but can you fix the tests?

chrisholmes:
Will do. It looks like the exceptions that get thrown are different depending on the OS used, which is a bit frustrating.

chrisholmes:
I think I'm done with changes here. Everything looks to be passing, but it does seem to fall afoul of the intermittent timeout errors in tests that master is also experiencing.

jplock:
I think this looks great.  Thanks for the contribution!

arteam:
Great contribution, indeed. I'm looking forward to see it in 0.9.0.
Should we update the release notes?

PR #1229, opened 2015-08-19
Should reset mock DAO in tear down

wakandan:
It doesn't make sense in the docs to setup mock _and_ then reset immediately. It was very confusing for me when I was setting up test


joschi:
Good catch. Thanks!

wakandan:
:+1:

PR #1232, opened 2015-08-20
Don't close the client instance in DropwizardApacheConnector

### LINKED ISSUE #1160: Http Client Connection Pool Error

arteam:
Jersey expects that we create a new instance of the Apache HTTP client every time, when we create a new instance of the connector.

So, when `ClientRuntime` is garbage collected, the client is closed and new requests, that work with a new connector, use already closed client.

The fix is to not close the client in the connector. We can rely on the fact, that the client is managed by Dropwizard environment and it will be closed, when the application will be shut down.

Also users of Apache HTTP Client keep the ability to close the client manually, if they want to.

Fix #1160

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
@@ -215,11 +215,7 @@ public class DropwizardApacheConnector implements Connector {
      */
     @Override
     public void close() {
-        try {
-            client.close();
-        } catch (IOException e) {
-            throw new ProcessingException(LocalizationMessages.FAILED_TO_STOP_CLIENT(), e);
-        }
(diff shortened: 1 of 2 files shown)


carlo-rtr:
closed / reponed to retrigger travis ci

joschi:
@carlo-rtr You can restart failed jobs directly at Travis CI: https://travis-ci.org/dropwizard/dropwizard

carlo-rtr:
thanks @joschi didn't realize that

PR #1233, opened 2015-08-21
fix timeouts by setting client configs

carlo-rtr:
similar to DropwizardApacheConnectorTest


joschi:
While it doesn't exactly fix the issue at hand (i. e. Travis CI failing builds on Java 8), it's an oversight and a mistake in the tests to fix, nonetheless. Thanks for fixing it, Carlo!

carlo-rtr:
@joschi I had hi hopes for this. I'm not sure what's the problem. Travis ci help points to stack overflow, so I've opened up a question there: http://stackoverflow.com/questions/32172925/travis-ci-sporadic-timeouts-to-localhost

If we don't make some traction on this on the next few days, I'm with @arteam on disabling the tests

PR #1235, opened 2015-08-26
Fix issue with validation message caching when a field has more than …

smcwhtdtmc:
…one constraint

Some hibernate validator constraints (@Min, and @Max come to mind) accept nulls as valid. This motivates you to slap @NotNull on the parameter in addition to @Min (or whatever). The Path -> validation message caching that goes on in ConstraintMessage.java generates bogus messages in these cases - with the right call pattern, you can pass in a non-null-but-still-invalid value and be told "query param foo may not be null."

This should fix it. I didn't quite follow the returnValueName-related stuff, but I believe I kept the behavior intact.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
@@ -6,24 +6,31 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.Iterables;
 import io.dropwizard.validation.ConstraintViolations;
 import io.dropwizard.validation.ValidationMethod;
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.reflect.FieldUtils;
-import org.apache.commons.lang3.reflect.MethodUtils;
-
-import javax.validation.ConstraintViolation;
-import javax.validation.ElementKind;
-import javax.validation.Path;
-import javax.ws.rs.*;
-import javax.ws.rs.core.Context;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.List;
-import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
+import javax.validation.ConstraintViolation;
+import javax.validation.ElementKind;
+import javax.validation.Path;
+import javax.validation.metadata.ConstraintDescriptor;
+import javax.ws.rs.CookieParam;
+import javax.ws.rs.FormParam;
+import javax.ws.rs.HeaderParam;
+import javax.ws.rs.MatrixParam;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Context;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.commons.lang3.reflect.MethodUtils;
+import org.apache.commons.lang3.tuple.Pair;
 
 public class ConstraintMessage {
-    private static final Cache<Path, String> MESSAGES_CACHE = CacheBuilder.newBuilder()
+
+    private static final Cache<Pair<Path, ? extends ConstraintDescriptor<?>>, String> MESSAGES_CACHE =
+            CacheBuilder.newBuilder()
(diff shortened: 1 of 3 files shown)


nickbabcock:
Thanks for the pull request. Off hand it looks pretty good, but I'll review it more in-depth tomorrow.

nickbabcock:
I think the implementation can be a bit simpler. I was able to get all the tests to pass (including the one you just wrote) and have the cache work with:

```java
private static final Cache<Pair<Path, ? extends ConstraintDescriptor<?>>, String> MESSAGES_CACHE =
        CacheBuilder.newBuilder()
        .expireAfterWrite(1, TimeUnit.HOURS)
        .build();

public static String getMessage(ConstraintViolation<?> v) {
    Pair<Path, ? extends ConstraintDescriptor<?>> of =
            Pair.of(v.getPropertyPath(), v.getConstraintDescriptor());

    String message = MESSAGES_CACHE.getIfPresent(of);
    if (message == null) {
        message = calculateMessage(v);
        MESSAGES_CACHE.put(of, message);
    }
    return message;
}
```

What do you think about that? I don't think there should be a need for two caches.

smcwhtdtmc:
Looks good - why didn't I think of that... ;)

nickbabcock:
No problem. Just making sure there wasn't something I missed. If you want to apply my suggestion, I'll take another look :smile:

smcwhtdtmc:
Done and pushed.

smcwhtdtmc:
Hm... CI build failed, but it looks like a flakey test. I'm not sure how to repeat the build. Any tips?

smcwhtdtmc:
While I was in there, I found another edge case... I'm quite unsure what to do about this one, but that didn't stop me from trying.

What do you think about https://github.com/smcwhtdtmc/dropwizard/commit/515b4d04b184d424daa55bd69c896f9c8e65bf01?

Here are my concerns with this approach: Is there a better way to discover resource methods than the DynamicFeature? Is post(null) even supposed to reach the validator, or is that something that should be caught earlier in the stack, like post(Entity.json(null)) is?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
```diff
@@ -6,38 +6,42 @@
...
+            CacheBuilder.newBuilder()
             .expireAfterWrite(1, TimeUnit.HOURS)
             .build();
 
-    private ConstraintMessage() {
-    }
-
-    /**
-     * Gets the human friendly location of where the violation was raised.
-     */
```
nickbabcock: No need to delete this snippet of code. My sample in the comment was abbreviated

nickbabcock:
I don't think it is an issue because the request body validation is handled by [JacksonMessageBodyProvider](https://github.com/dropwizard/dropwizard/blob/5b8582d9ae0b80f13f6bf4712acb7e1f9c4f25ea/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JacksonMessageBodyProvider.java#L71-L107).

Also I'll try restarting your build a few times, if it still fails due to unrelated tests, I'll just verify locally and merge.

smcwhtdtmc:
(If commenting on a merged pull request is bad, apologies in advance.)

JacksonMessageBodyProvider doesn't run in the scenario I'm talking about. .post(null) gets you a request without any Content-Type header. With no Content-Type header, https://github.com/jersey/jersey/blob/master/core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderInterceptorExecutor.java#L226 doesn't find a bodyReader, which gets you a null entity.

BeanValidation notices the null entity (yay), but the ConstraintValidationException gets formatted like '${resourceMethod}.argN may not be null' in the current code (boo).

You can see this in action with the test case in the branch I linked. I was suspicious that this may just be an artifact of the way the test is set up, so I tried it on a real service (0.9.0-rc3) and got the same behavior.

nickbabcock:
This does sound like a different issue (so it'd be wise to open another one).

Also it sounds like this issue affects all version of dropwizard, not just 0.9.x, correct?

nickbabcock:
Have you had a chance to look at [JERSEY-2636](https://java.net/jira/browse/JERSEY-2636). Looks like it is not WAD for a null body to be passed into a endpoint that expects a content-type.

smcwhtdtmc:
Thanks for digging that out. It looks like the same thing. I'm convinced this isn't an exception rendering problem anymore.

PR #1234, opened 2015-08-25
Add docs for dropwizard auth

wakandan:
Add some docs to test resources with `@Auth`

PR #1239, opened 2015-08-26
Fix auth testing documentation

akraxx:
The previous PR #1234 was for dropwizard 0.8.*, I updated it to work with dw 0.9 :)

I think that an empty line was missing after ```.. code-block:: java``` so the block was not displayed well.

Do we have to keep this part of the documentation in this file or should I move it to the **Testing** section?


jplock:
This looks good to me, thanks for the contribution!

Issue #1197, opened 2015-07-29
@Valid on object parameter implementing Principal without @Auth


derrikcurran:
I'm using Dropwizard **0.9.0-rc3** and I'm running into some trouble with the new auth system.

I have a User model that implements Principal and a POST resource method for signing up new users. The signup method has a validated User parameter but it's not authenticated. When I try access that resource, I get an IllegalStateException with the message: `Cannot inject a custom principal into unauthenticated request`. Is there an easy way to handle this scenario?

**EDIT:** I should note that it works just fine if I don't use `@Valid`. 

My signup method looks like this:
```java
@POST
@Path("signup")
@Timed
@UnitOfWork
public Response signup(@Valid final User user, @Context final HttpServletRequest request) throws JOSEException {
	dao.setFilteredByTenantId(false);
	user.setInterceptorBypassed(true);
	user.setRole("Admin");
	final Optional<User> newUser = dao.create(user);
	dao.resetFilters();
	
	if (newUser.isPresent()) {
		final Token token = AuthUtils.createToken(request.getRemoteHost(), newUser.get().getId());
		return Response.status(Status.CREATED).entity(token).build();
	} else {
		return JSONErrorResponse.build(Status.CONFLICT, "Failed to create new user.");
	}
}
```

StanSvec:
For some reason Jersey is trying to use AuthValueFactoryProvider for injecting of the User instance into the method even when no authentication is used. Looks like a bug.  Do you use @Auth annotation on another method or on class level for this resource class?

derrikcurran:
No @Auth annotations in that class.

StanSvec:
I've located the bug and will provide fix for this.

nickbabcock:
Closed by #1241

PR #1241, opened 2015-08-30
Fix for #1197 - Principal instance not injected when annotated

StanSvec:
Principal parameter which is not annotated with Auth annotation shouldn't be authenticated. This worked before but only when such parameter is not annotated with any annotation. Otherwise parameter injection failed. The prove that this didn't work is NoAuthPrincipalEntityTest.annotatedPrincipalEntityWithoutAuth test which fails without the fix.

Btw., still wondering why Auth annotation is allowed on method and type level..

### DIFF
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/AbstractAuthResourceConfig.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/AbstractAuthResourceConfig.java
@@ -1,0 +1,26 @@
+package io.dropwizard.auth;
+
+import com.codahale.metrics.MetricRegistry;
+import io.dropwizard.jersey.DropwizardResourceConfig;
+import org.glassfish.jersey.server.filter.RolesAllowedDynamicFeature;
+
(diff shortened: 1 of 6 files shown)


dropwizard-auth/src/test/java/io/dropwizard/auth/AbstractAuthResourceConfig.java
```diff
@@ -0,0 +1,26 @@
+package io.dropwizard.auth;
+
+import com.codahale.metrics.MetricRegistry;
+import io.dropwizard.jersey.DropwizardResourceConfig;
+import org.glassfish.jersey.server.filter.RolesAllowedDynamicFeature;
+
+import javax.ws.rs.container.ContainerRequestFilter;
+import java.security.Principal;
+
+public abstract class AbstractAuthResourceConfig extends DropwizardResourceConfig {
```
carlo-rtr: Good Idea

carlo-rtr:
Thanks for this. The failures look like they were due to our flaky tests. 

This PR looks good to me. I'll give it a day to see what others think.

PR #1247, opened 2015-09-07
Add FileAppenderFactory validations and tests

dennyac:
I have added validations and corresponding tests to FileAppenderFactory so that the error messages are understandable.


nickbabcock:
Looks good to me :+1:

PR #1252, opened 2015-09-16
Add non-JUnit integration test section

MikaelAmborn:
We use DropWizard from Scala and I couldn't find any documentation about how to write integration tests, apart from the discussion in #488.


jplock:
Looks good, thanks!

PR #1260, opened 2015-09-24
Don't require username in the data source configuration

### LINKED ISSUE #1259: Dropwizard DataSourceFactory on with just DATABASE_URL
Heroku and Dokku only offer a single DATABASE_URL when deploying an application. It would be convenient to create a DataSourceFactory using just this url.

The alternative is to parse the DATABASE_URL and pluck username and password from there, but this is error-prone and makes it hard to use different databases, like h2 in test.

Example database_url:

`DATABASE_URL=postgres://postgres:SOME_PASSWORD@172.17.0.1:5432/lolipop`

Example database-configuration:
```
database:
    driverClass: org.postgresql.Driver
    url: ${DATABASE_URL}
```

Also, see http://stackoverflow.com/questions/32705394/dropwizard-datasourcefactory-on-heroku-with-just-database-url

arteam:
Some platforms provide credentials to connect to a database inlined in the connection URL. Examples:

* `postgres://postgres:SOME_PASSWORD@172.17.0.1:5432/lolipop`
* ` jdbc:h2:file:~/sample;USER=sa;PASSWORD=123`
* `jdbc:postgresql://localhost/test?user=fred&password=secret`

JDBC drivers support such a type of configuration. We should respect it and provide users ability to use it.

It means we should not require the username field to be set in the data source configuration.

Fixes #1259

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -973,7 +973,7 @@ driverClass                     REQUIRED                 The full name of the JD
 
 url                             REQUIRED                 The URL of the server.
 
-user                            REQUIRED                 The username used to connect to the server.
+user                            none                     The username used to connect to the server.
 
 password                        none                     The password used to connect to the server.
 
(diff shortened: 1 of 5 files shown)


jplock:
LGTM!

PR #1264, opened 2015-09-28
Add settings for removeAbandonedTimeout and removeAbandoned

### LINKED ISSUE #1255: No configuration for setting PoolProperties.removeAbandoned in io.dropwizard.db.DataSourceFactory.java
There seems to be not any support for passing removeAbandoned parameter value to the tomcat jdbc pool through dropwizard 7 database config. Is there any way we can set this property?

joschi:
Fixes #1255

### DIFF
--- a/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
+++ b/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Optional;
 import com.google.common.collect.Maps;
+import com.google.common.primitives.Ints;
 import io.dropwizard.util.Duration;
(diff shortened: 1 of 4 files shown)


dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
```diff
@@ -769,6 +813,12 @@ public ManagedDataSource build(MetricRegistry metricRegistry, String name) {
         poolConfig.setUrl(url);
         poolConfig.setUsername(user);
         poolConfig.setPassword(user != null && password == null ? "" : password);
+        poolConfig.setRemoveAbandoned(removeAbandoned);
+
+        if (getRemoveAbandonedTimeout().isPresent()) {
```
arteam: Maybe we could set the timeout unconditionally? We already set the default value the same as in the pool configuration, if the user didn't specify the property.

arteam:
Could you update `DataSourceConfigurationTest`, so we can verify that deserialization works correctly and the default values are applied? Apart from that, it looks good to me.

arteam:
Thanks!

Issue #1265, opened 2015-09-28
Make AssetBundle/AssetServlet customizable


Randgalt:
AssetBundle/AssetServlet are very useful as is. But, with a few minor changes, they could be subclasses and customized.

arteam:
Closed via #1266

PR #1266, opened 2015-09-28
Dropwizard #1265 - Allow AssetBundle/AssetServlet to be customized.

Randgalt:
minor changes so that these classes can be subclassed for customization/reuse


arteam:
Looks like the build failed for some bizarre reason on SSL certificate tests. We are still fighting with Travis CI to make build stable, because it periodically fails on Java 8 without reasons. I've restarted the build and it has completed successfully.

The change looks good to me. Thanks!

Issue #1267, opened 2015-09-29
JerseyClientConfiguration.setGzipEnabled(false) not working as expected


lehcim:
I'm using dropwizard 0.9.0-rc4 to build a service which uses another Rest API. In order to debug my application, I'm trying to observe the requests and responses between a local jersey client and the remote server on a logging proxy. To make the logs human readables, I tried to disable the compression on the client side like this:

```java
JerseyClientConfiguration clientConfig = new JerseyClientConfiguration();
clientConfig.setGzipEnabled(false);
clientConfig.setGzipEnabledForRequests(false);

Client client = new JerseyClientBuilder(environment).using(clientConfig).build("myClient")
```

but the  remote server answered with a compressed answer, because the request was looking like this:

```
GET /foo/bar HTTP/1.1
Host: localhost:8080
Connection: Keep-Alive
User-Agent: myClient
Accept: application/json,application/xml
Accept-Encoding: gzip,deflate
```

Where do this 'Accept-Encoding' header come from, since gzip is disabled in the JerseyClientConfiguration ?
In short, if you disable compression at the Jersey level, then apache will enable it for you again...

* The io.dropwizard.client.JerseyClientBuilder delegates client creation to io.dropwizard.client.HttpClientBuilder, which uses org.apache.http.impl.client.HttpClientBuilder to really create the instance.

* In the apache builder, compression is enabled by default, if you don't call explicitely disableContentCompression() then the "gzip,deflate" is automatically added by org.apache.http.client.protocol.RequestAcceptEncoding ...

* Thus, to honor the .setGzipEnabled(false) call, the dropwizard JerseyClientBuilder should tell the dropwizard HttpClientBuidlder to disable the compression in the apache HttpClientBuilder... but the gzipEnabled flag is not visible in dropwizard HttpClientBuilder, because it is a member of the above JerseyClientConfiguration 

A possible way to solve this issue is to force disableContentCompression() in apache HttpClientBuilder, because Gzip is managed at the Jersey level (see io.dropwizard.jersey.gzip.GzipDecoder). It should work
in my use case, but I can't say it will not break something elsewhere.

Any thoughts ?

arteam:
It makes sense to me. 

I don't think we will break anything, because compression is enabled by default. We probably broke this feature, when we moved to `org.apache.http.impl.client.HttpClientBuilder` approach for constructing the client. It  (de)compresses by default, but `DefaultHttpClient` don't.

lehcim:
[...]

PR #1268, opened 2015-09-29
JerseyClientConfiguration.setGzipEnabled(false) not working as expected

lehcim:
Correction for issue #1267:
- force disableContentCompression() in HttpClientBuilder
- set Content-Encoding in GzipDecoder


dropwizard-jersey/src/main/java/io/dropwizard/jersey/gzip/GZipDecoder.java
```diff
@@ -39,4 +36,11 @@ public Object aroundReadFrom(ReaderInterceptorContext context) throws IOExceptio
         return context.proceed();
     }
 
+	@Override
```
jplock: mind fixing the indentation here?

arteam:
Closed via #1270

PR #1269, opened 2015-09-29
Sorry! I forgot to include this. This is the most important part. Please accept.

Randgalt:
Need to be able to override getResourceUrl() in case resource is in an external JAR


arteam:
This also looks reasonable. Thanks.

PR #1270, opened 2015-09-29
Support for disabling decompression of HTTP responses

arteam:
By default Apache HTTP client builder enables decompression of responses. If we want to disable this feature in Jersey HTTP client, we should explicitly disable it in the underlying Apache HTTP client
during building.


jplock:
Looks good to me

PR #1272, opened 2015-10-01
Fix documentation on servlet filters

douzzi:


joschi:
Good catch. Thanks for fixing it!

PR #1276, opened 2015-10-10
Add support for several Hibernate bundles

### LINKED ISSUE #1217: No two hibernateBundles can listen in parallel to @UnitOfWork
given two active hibernate-bundles, it is not possible that both are used concurrently. The sketch below is roughly what i use atm. The problem is caused by HibernateBundle#run(), which - in my case - registers a ´UnitOfWorkApplicationListener´ instance two times as jersey singletons. So, as a result ´@UnitOfWork´ wont find any session when i access the BResource methods, because jersey only registers the first call.

https://github.com/dropwizard/dropwizard/blob/50d906e17bfb29cad76c6f7a2c24da8f2fa02003/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java#L57

```java
public class BrokenApplication extends Application<MyConfiguration> {
	private final HibernateBundle<MyConfiguration> bundleA = new HibernateBundle<MyConfiguration>(A.class) {
		@Override
		public DataSourceFactory getDataSourceFactory(MyConfiguration configuration) {
			return configuration.getDatabaseFactoryA();
		}

		@Override
		protected String name() {
			return "hibernate.A";
		}
	};
	
	private final HibernateBundle<MyConfiguration> bundleB = new HibernateBundle<MyConfiguration>(B.class) {
		@Override
		public DataSourceFactory getDataSourceFactory(MyConfiguration configuration) {
			return configuration.getDataSourceFactoryB();
		}

		@Override
		protected String name() {
			return "hibernate.B";
		}

	};

	@Override
	public void initialize(final Bootstrap<MyConfiguration> bootstrap) {
		bootstrap.addBundle(bundleA);
		bootstrap.addBundle(bundleB);
	}

	@Override
	public void run(final MyConfiguration configuration, final Environment environment) {
		// each repo contains accessors to the Daos.

		final ARepository repoA = new ARepository(bundleA.getSessionFactory());
		environment.jersey().register(new AResource(repoA));

		final BRepository repo = new BRepository(bundleB.getSessionFactory());
		environment.jersey().register(new BResource(repo));
	}
}
```

arteam:
In some cases one need to simultaneously connect to several relational databases. Users want to use Hibernate to access each database. The problem that the current implementation of
`UnitOfWorkApplicationListener` supports only one session factory, which ultimately restricts ability to support several session factories.

Another problem is that users should have the ability to specify the name of datasource, against which a transaction should be opened in a resource method.

This change:

* Adds the method `value` to the `@UnitOfWork` annotation. It allows to specify the name of Hibernate bundle (session factory). By default it specifies the default bundle.

* Replaces the single session factory in `UnitOfWorkApplicationListener` to a map of session factories by names of bundles.

* The correct session factory is determined in runtime, based on the `value` method in the corresponding `@UnitOfWork` annotation.

In the result users can simultaneously use declarative transactions on several databases, access to which is managed by Hibernate.

Fixes #1217

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java
@@ -12,7 +12,7 @@ import io.dropwizard.util.Duration;
 import org.hibernate.SessionFactory;
 
 public abstract class HibernateBundle<T extends Configuration> implements ConfiguredBundle<T>, DatabaseConfiguration<T> {
-    private static final String DEFAULT_NAME = "hibernate";
+    public static final String DEFAULT_NAME = "hibernate";
 
     private SessionFactory sessionFactory;
 
@@ -54,13 +54,24 @@ public abstract class HibernateBundle<T extends Configuration> implements Config
     public final void run(T configuration, Environment environment) throws Exception {
         final PooledDataSourceFactory dbConfig = getDataSourceFactory(configuration);
         this.sessionFactory = sessionFactoryFactory.build(this, environment, dbConfig, entities, name());
-        environment.jersey().register(new UnitOfWorkApplicationListener(sessionFactory));
+        registerUnitOfWorkListerIfAbsent(environment).registerSessionFactory(name(), sessionFactory);
         environment.healthChecks().register(name(),
                                             new SessionFactoryHealthCheck(
                                                     environment.getHealthCheckExecutorService(),
                                                     dbConfig.getHealthCheckValidationTimeout().or(Duration.seconds(5)),
                                                     sessionFactory,
                                                     dbConfig.getHealthCheckValidationQuery()));
+    }
+
+    private UnitOfWorkApplicationListener registerUnitOfWorkListerIfAbsent(Environment environment) {
+        for (Object singleton : environment.jersey().getResourceConfig().getSingletons()) {
+            if (singleton instanceof UnitOfWorkApplicationListener) {
+                return (UnitOfWorkApplicationListener) singleton;
+            }
+        }
+        final UnitOfWorkApplicationListener listener = new UnitOfWorkApplicationListener();
+        environment.jersey().register(listener);
+        return listener;
     }
 
     public SessionFactory getSessionFactory() {
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWork.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWork.java
@@ -47,4 +47,10 @@ public @interface UnitOfWork {
      * @see org.hibernate.Session#setFlushMode(org.hibernate.FlushMode)
      */
     FlushMode flushMode() default FlushMode.AUTO;
+
+    /**
+     * The name of a hibernate bundle (session factory) that specifies
+     * a datasource against which a transaction will be opened.
+     */
+    String value() default HibernateBundle.DEFAULT_NAME;
 }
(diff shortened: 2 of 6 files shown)


nickbabcock:
Excellent, I like it. Backwards compatible as well!

jplock:
Should this be in `0.9.0` or `1.0.0`?  (related, we really need to release `0.9.0` if there aren't any show stoppers)

nickbabcock:
I *think* this is fine, there have been other semi-major-but-backwards-compatible features since the latest rc: [(changes since 0.9.4-rc4)](https://github.com/dropwizard/dropwizard/compare/v0.9.0-rc4...master). But I'm fine if a release wants to be cut and save some of the features for 0.9.x

lecummins:
This is great, I have been looking to be able to do this for a while. Just one question - is it possible to use two different session factories within a single resource method? I have a number of resources that need to be able to access user settings from one database and information from a different one. I would have thought that this would be possible since each of the DAOs takes in their relevant bundle that they will use.

nathan-olson:
This totally breaks if you need to use both databases within one resource call by having the UnitOfWork value coupled directly with the bundle name.  My use case is such that I need to update a legacy database with a bit of data, but also create an entry within my new database within a POST.  However, since the transaction has to be tied to one, and only one, bundle (even if I set 'transactional' to false), then I only have context into one of my databases, making it impossible to solve my use case

bhagatdinesh:
It's awesome that Dropwizard hibernate supports multiple hibernate bundle.
Could you please share document where I can specify multiple DB config and use multiple bundles. I could not find any reference of this here: http://www.dropwizard.io/0.6.2/manual/hibernate.html
I am using dropwizard-hibernate-0.9.2.jar.
I am still getting exception: org.hibernate.HibernateException: No session currently bound to execution context
Suggest If I am missing anything.
Referred following site for issue.
http://stackoverflow.com/questions/30212125/no-session-currently-bound-to-execution-context
http://stackoverflow.com/questions/42384671/dropwizard-hibernate-no-session-currently-bound-to-execution-context

Tried,

public class MyApplication extends Application<Configuration> {
private final HibernateBundle<Configuration> slaveHibernateBundle =
			new HibernateBundle<Configuration>(TableClass.class) {
				@Override
				public DataSourceFactory getDataSourceFactory(Configuration configuration) {
					return configuration.getDataSourceFactory1();
				}
				@Override
				protected String name() {
					return "slave-hibernate";
				}
	};
public static void main(String[] args) throws Exception {
		new MyApplication().run(args);
	}

@Override
	public void initialize(Bootstrap<DriverPerformanceEngineConfiguration> bootstrap) {
		bootstrap.addBundle(slaveHibernateBundle);
	}
@Override
	public void run(Configuration configuration, Environment environment) throws Exception {
GenericDBDAO<TableClass>  = new GenericDBDAO<>(slaveHibernateBundle.getSessionFactory(), "TableClass");
environment.jersey().register(new AbstractBinder() {
			@Override
			protected void configure() {
	}
}

arteam:
I think you need to set the name of the slave database in the `@UnitOfWork` annotation on your resource.

bhagatdinesh:
Yes, It is correct that 
if we do not specify a value in @UnitOfWork then it considers default hibernate bundle.
But If I specify like @UnitOfWork(value = "slave-hibernate") then it consideres only slave hibernate session.
But, I am trying to connect to both master and slave DB within the same resource "/getResult".
Is there any way to achieve this in Dropwizard-hibernate?
Please suggest an alternative if exists.

yateam:
You could work this around by splitting your resource method into several parts, each of them talks to a different database. Then annotate each of these parts (functions or services) with @UnitOfWork(value = 'needed-db'). And remove the top UnitOfWork annotation from the resource method.

it could be something like this:

```
@Path("/getResult")
@GET
//@UnitOfWork -- comment this out
Result getResult() {
   service.talkToDb1();
   service.talkToDb2();
}

class Service {
  @UnitOfWork(value = "db1")
   void talkToDb1() {}

  @UnitOfWork(value = "db2")
   void talkToDb2() {}
}
```

JGovindaraj:
@yateam - rollback will not be done in db1 when db2 fails in some situation.  
Manually opening a session and bind it with ManagedSessionContext works fine, but wanted to know if there is a better way to handle two databases.

my code/situation are also similar to bhagatdinesh posted above in this thread. let me know if need to explain

dondell:
I have tried @yateam approach but it is not working. I couldn't find still the hibernate values declared on the overriding names on hibernatebundle.

PR #1284, opened 2015-10-17
Move dependency declarations into dedicated BOM (Bill of materials) artifact

joschi:
This PR moves all full dependency declarations into a dedicated BOM (Bill of materials) artifact which simplifies creating Dropwizard modules and applications without dependency conflicts (usually checked by the Maven Enforcer plugin).

Users can simply import the Dropwizard BOM into their projects and don't have to take care of providing a matching version of the dependencies (but are still able to explicitly override those if required):

```xml
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>io.dropwizard</groupId>
        <artifactId>dropwizard-bom</artifactId>
        <version>${dropwizard.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```

This follows the best practice of several other widely used Java libraries and frameworks like JBoss and Spring.

* https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html
* http://www.infoq.com/news/2014/07/springio-platform
* http://howtodoinjava.com/2014/02/18/maven-bom-bill-of-materials-dependency/


arteam:
This looks great. Definetely should help for users who use many Dropwizard modules.

dropwizard-archetypes/java-simple/src/main/resources/archetype-resources/pom.xml
```diff
@@ -26,11 +26,22 @@
         <mainClass>\${package}.\${name}Application</mainClass>
     </properties>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.dropwizard</groupId>
+                <artifactId>dropwizard-bom</artifactId>
+                <version>${dropwizard.version}</version>
```
jplock: Does this need a `\` in front of it similar to the line below?

joschi: Yes, it does. Thanks, good catch!

jplock:
I noticed that there is a `jersey-bom` in maven central?  Would it make any sense to use that in this bom file?

joschi:
@jplock Yes, indeed. I think this makes sense and I've replaced most of the explicit Jersey dependencies in Dropwizard's BOM simply with `jersey-bom`. For some dependencies it's not possible due to dependency exclusions.

PR #1285, opened 2015-10-20
Fixed failing CliTests by Setting Default Locale

fexbraun:
Multiple tests in CliTest will fail on systems with a default locale other than English.

ArgParse4j localizes some parts of the validation error message. The tests assert the equality of the whole text. As a fix, I set the default locale to English in the setup of the tests.

FMPOV it is not necessary to save the former default locale and to switch it back after the tests because tests are executed in a separate JVM. Let me know, if you think different.

### DIFF
--- a/dropwizard-core/src/test/java/io/dropwizard/cli/CliTest.java
+++ b/dropwizard-core/src/test/java/io/dropwizard/cli/CliTest.java
@@ -11,13 +11,14 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.io.ByteArrayOutputStream;
+import java.util.Locale;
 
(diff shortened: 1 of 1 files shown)


joschi:
@fexbraun My default locale is `de_DE.UTF-8` and those tests are working. Did you specifically run into problems with those tests?

fexbraun:
Hi Jochen, 
Yes, on a new machine with a new workspace these tests were failing.

My default locale is "de_DE" when the tests are executed within eclipse, and then they fail.
When I execute the tests via mvn test the default locale is "en_US" and the tests are fine.

joschi:
Thanks for this PR! I've opted to reset the original default locale at the end of the test.

Issue #937, opened 2015-03-12
Git `autocrlf` Settings


kilemensi:
When it comes to windows development support, I've noticed we've gone back and forth about how should `autocrlf` settings be set.

Initially it was agreed that it should be set to `false` since it was just easier to set your editor/IDE to use `LF` as EOL marker than git auto-converting all source files (See #398 and #400). But after many months I've just done a full pull, clean and install and I see everything has been changed again.

Can we have the official Dropwizard recommendation for this? And may be a setting as well if it's supported (even using editorconfig)?

Personnaly, I prefer `autocrlf = false`, and set my ide/editors to use `LF` for EOL marker per project (with the help of editorconfig where applicable).

carlo-rtr:
Thanks for raising this.

This is probably a silly question, but what is the impact of the inconsistency?  

Also, do you think this article summarizes the problem correctly?
http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/

With my limited by knowledge, I'm thinking we should probably use .gitattributes so we can actually enforce it in an automated way. Probably with:

```
*.txt text
*.java text
*.yml text
*.xml text
*.md text
*.mustache text
*.ftl text
```

I assume we would have to run process to clean up files. Any of this make sense?

kilemensi:
The issue is like this:
The `FreemarkerViewRendererTest.rendersViewsWithRelativeTemplatePaths` test loads the `relative.ftl` file and tests the rendered output. But, the test itself is done in platform independent way i.e. uses `System.lineSeparator()` which in windows translates to `CRLF`.
The original `relative.ftl` however will contain `LF` in git itself and depending on how one sets the auto conversion, it will either contain `LF` as is (which is the case for me since I'm not doing any conversion) or it will contain `CRLF` for those who do (most users of GitHub for Windows app). Hence the code as is now will fail for the previous case and pass for the later.

The article does summerize the problem very well but evidently I've been using "the old way". I'm not sure how the two systems play together. If one provides .gitattributes, does that mean it doesn't matter what the core.[autocrlf, safecrlf] are set to? If yes the .gitattributes is THE BEST way to go about this.

carlo-rtr:
@joschi do you have an opinion?

joschi:
Thanks for bringing this up. I've added the `.gitattributes` file @carlo-rtr suggested to the repository.

PR #1286, opened 2015-10-20
Fixed failing Freemarker tests (on Windows OS)

fexbraun:
Two tests in FreemarkerViewRendererTest assumed that the line ending of
the system is '\n'. They tried to match a new line from the template
file with a System.lineSeparator().


joschi:
Thanks for your PR. I've chosen a slightly different approach to the problem at hand (see 518d4b5bec3ff511fee088211009d46209597bd2 and #937).

@fexbraun Please let me know if this fixes your problems on Windows too.

fexbraun:
@joschi works fine for me.

joschi:
:+1:

PR #1290, opened 2015-10-20
Add SPI for RequestLogFactory

mwhipple:
Allows for overriding of `RequestLog` implementation such as to allow use of logback-access

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Charsets;
 import com.google.common.base.Joiner;
+import com.google.common.base.Optional;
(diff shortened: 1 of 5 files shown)


arteam:
Thanks! Applied to master in 05d3fcd with some enhancements.

mattnelson:
@mwhipple Do you have an implementation using these changes for logback-access? I am also looking for this feature and would like to avoid reimplementing.

mwhipple:
@mattnelson I should be able to get something together that I can post for you in some form if not as a directly usable artifact tomorrow. The working implementation with logback-access that I have doesn't use this change and instead extends the top level configuration object (the ugliness of which inspired the PR)

mwhipple:
I'd left out extending `Discoverable` so the logback access module won't be as convenient to use pending https://github.com/dropwizard/dropwizard/pull/1412, but it should be able to be handled with an explicit registration of a Jackson module in the consuming service. I extracted most of the implementation and I'll publish the present state in the morning (should work but could stand cleanup) and work on info about the Jackson module when time allows.

mwhipple:
@mattnelson https://github.com/mwhipple/dropwizard-logback-access - still awaiting enough time to be in a presentable state and published but at least enough to start with.

PR #1293, opened 2015-10-22
Replace default validator in tests

### LINKED ISSUE #1292: Optional configuration parameter validation problem in 0.90-rc5
After swtiching to version 0.90-rc5, my server failed to load its configuration with the following error message :

```
Exception in thread "main" javax.validation.UnexpectedTypeException: HV000186: 
The constraint of type 'javax.validation.constraints.NotNull' defined on 'client.userAgent' 
has multiple matching constraint validators which is due to an additional value handler of type 
'io.dropwizard.validation.valuehandling.OptionalValidatedValueUnwrapper'. 
It is unclear which value needs validating. Clarify configuration via @UnwrapValidatedValue.
```

The configuration field which trigger the exception is a dropwizard `HttpClientConfiguration` with a `@Valid` annotation:

```java
public class PetConfiguration extends Configuration {
    @Valid    // remove this annotation to avoid HV000186
    public HttpClientConfiguration client ;
}
```
The problem is caused by the annotation `@NotNull` on the property userAgent in class HttpClientConfiguration. It can be reproduced  with a simple String parameter, like this :

```java
public class PetConfiguration extends Configuration {
    @NotNull
    @Valid    // remove this annotation to avoid HV000186 
    public Optional<String> foo = Optional.absent() ;
}
```

Everything is fine with the two configurations classes above using 0.90-rc4, I think it's a 0.90-rc5 regression, but I don't known the cause yet : to my knowledge, there was no change on HttpClientConfiguration.userAgent between rc4 and rc5 ...

nickbabcock:
The validator used in tests
(`Validation.buildDefaultValidatorFactory().getValidator()`) is not dropwizard
configured, thus bugs like #1292 slipped through. This pull request moves the
tests to the dropwizard configured validator. Note that this doesn't cover all
cases as some classes are not able to import dropwizard-jersey (where the
dropwizard configured validator lies) into the tests, so more work will need
to be done to cover all the bases.

~~This pull request also fixes #1292~~

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java
@@ -9,6 +9,7 @@ import com.google.common.collect.Maps;
 import io.dropwizard.jersey.gzip.ConfiguredGZipEncoder;
 import io.dropwizard.jersey.gzip.GZipDecoder;
 import io.dropwizard.jersey.jackson.JacksonMessageBodyProvider;
+import io.dropwizard.jersey.validation.Validators;
 import io.dropwizard.lifecycle.Managed;
 import io.dropwizard.setup.Environment;
 import org.apache.http.client.CredentialsProvider;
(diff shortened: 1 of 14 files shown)


joschi:
@nickbabcock Could you please create a separate PR for fixing `HttpClientConfiguration` for #1292? Just to keep things minimal.

nickbabcock:
Once #1295 is good, review/merge that, and then I'll see what needs to be fixed up in this PR.

joschi:
@nickbabcock You can probably just reset your branch to 6202669898f1b4c964d3f20c1e9fe2588f752cd6 and force push it.

nickbabcock:
Did just that :+1:

joschi:
LGTM. :+1: 
Thanks!

PR #1295, opened 2015-10-22
Fix HTTP Client Config User Agent Validation

### LINKED ISSUE #1292: Optional configuration parameter validation problem in 0.90-rc5

nickbabcock:
With Hibernate Validator 5.2.x, when an annotation such as `@NotNull` can be applied to either the inner or wrapper type (in this case, `String` or `Optional`), the code must explicitly define `@UnwrapValidatedValue`

Fixes #1292

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientConfiguration.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientConfiguration.java
@@ -5,6 +5,7 @@ import com.google.common.base.Optional;
 import io.dropwizard.client.proxy.ProxyConfiguration;
 import io.dropwizard.client.ssl.TlsConfiguration;
 import io.dropwizard.util.Duration;
+import org.hibernate.validator.valuehandling.UnwrapValidatedValue;
 
 import javax.annotation.Nullable;
 import javax.validation.Valid;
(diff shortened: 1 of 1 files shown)


joschi:
Thanks!

PR #1299, opened 2015-10-23
Enhanced Views In Example Application

fexbraun:
- added a reference to a css served by the AssetsBundle
- added one line to with jobTitle to html


joschi:
Thanks!

Issue #1297, opened 2015-10-23
Optional Support for Dropwizard JDBI broken 


pandaadb:
Hi,

I recently noticed there is a bug in the optional handling for JDBi and joda DateTime objects. There is a serialisation issue that prevents any Optional datetimes to be inserted into a table. 

I created following unittest to demonstrate the problem:
```java
public class TestJDBIOptional {

	protected static DBI dbi;
	protected static Environment env;

	static DataSourceFactory getDataSourceFactory() {
		DataSourceFactory dataSourceFactory = new DataSourceFactory();
		dataSourceFactory.setDriverClass("org.h2.Driver");
		dataSourceFactory.setUrl("jdbc:h2:mem:testDb");
		dataSourceFactory.setUser("sa");
		dataSourceFactory.setPassword("");
		return dataSourceFactory;
	}

	String createTable = "CREATE TABLE IF NOT EXISTS test ( id varchar(255) primary key, "
			+ " time timestamp);";

	@Before
	public void setupTests() throws IOException {
		dbi.open().createStatement(createTable).execute();
	}

	@BeforeClass
	public static void setup() {
		env = new Environment("test-env", Jackson.newObjectMapper(), null, new MetricRegistry(), null);
		dbi = new DBIFactory().build(env, getDataSourceFactory(), "test");
		dbi.registerArgumentFactory(new JodaDateTimeArgumentFactory());
		dbi.registerMapper(new JodaDateTimeMapper(Optional.absent()));
	}
	
	@Test
	public void testInsert() {
		JDBIDao onDemand = dbi.onDemand(JDBIDao.class);
		onDemand.insert(Optional.of(DateTime.now()));
	}

	@Test 
	public void testAbsent() {
		JDBIDao onDemand = dbi.onDemand(JDBIDao.class);
		onDemand.insert(Optional.absent());
	}
	
	public interface JDBIDao {
		 
		@SqlUpdate("INSERT INTO test VALUES ('test', :ts)")
		abstract int insert(@Bind("ts") Optional<DateTime> dt);
	}
}
```

arteam:
Thanks for reporting this issue. You would need to register own argument factory to combine optional and JodaTime's `DateTime` arguments. Something similar to this:

````java
import com.google.common.base.Optional;
import org.joda.time.DateTime;
import org.skife.jdbi.v2.StatementContext;
import org.skife.jdbi.v2.tweak.Argument;
[...]

PR #1301, opened 2015-10-23
Add support for custom arguments to OptionalArgumentFactory

joschi:
This PR adds support for custom arguments (e. g. `JodaDateTimeArgumentFactory`) to `OptionalArgumentFactory`.

The implementation is rather clumsy because jDBI doesn't seem to provide extension points for this kind of customization.

Refs #1297.

### DIFF
--- a/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
+++ b/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
@@ -7,8 +7,9 @@ import com.codahale.metrics.jdbi.strategies.DelegatingStatementNameStrategy;
 import com.codahale.metrics.jdbi.strategies.NameStrategies;
 import com.codahale.metrics.jdbi.strategies.StatementNameStrategy;
 import com.google.common.base.Optional;
-import io.dropwizard.db.PooledDataSourceFactory;
 import io.dropwizard.db.ManagedDataSource;
+import io.dropwizard.db.PooledDataSourceFactory;
+import io.dropwizard.jdbi.args.DelegatingArgumentFactory;
 import io.dropwizard.jdbi.args.JodaDateTimeArgumentFactory;
 import io.dropwizard.jdbi.args.JodaDateTimeMapper;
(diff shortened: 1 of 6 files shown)


nickbabcock:
Are the Travis test failures relevant to the pull request?

joschi:
@nickbabcock Yes, unfortunately they are. I'll take another look at this tomorrow if no one else has a better idea. :disappointed:

arteam:
Do we really need `DelegatingArgumentFactory`? I think, we could write `OptionalJodaTimeArgumentFactory`, register it last, and use the fact that JDBI processes argument factories in the reverse order.

Personally, I don't think that this problem is worth to solve, if there is no simple solution.
Users could easily work around it.

joschi:
@arteam I created `DelegatingArgumentFactory` as preparation if more than just one additional `ArgumentFactory` is available in Dropwizard (e. g. [`dropwizard-java8`](https://github.com/dropwizard/dropwizard-java8/tree/dropwizard-java8-0.9.0-rc5-1/dropwizard-java8-jdbi/src/main/java/io/dropwizard/java8/jdbi/args) adds more). It's a bit unfortunate that jDBI doesn't support wrapped arguments in a way that would make this PR unnecessary.

Feel free to come up with another (maybe more pragmatic) PR for this issue. I'll happily review and merge it.

joschi:
@arteam Any more reasons not to merge this? Otherwise I'd merge this PR for Dropwizard 0.9.0.

arteam:
As I said, I don't like relative complexity of the fix and that we could break something in the final release, because nobody tested its effects on real applications.

I will send a PR with an alternative implementation, but would contest that this issue should be addressed in 0.9.0. 

@dropwizard/committers, what is your opinion on this?

jplock:
I could go either way on this. I agree with @arteam that his approach in #1304 is more isolated at this time for `0.9.0`.  We could revisit this PR in `1.0.0` as a more generic implementation.

joschi:
Closed in favor of #1304.

criccomini:
FYI I prefer this to 1304. We have use cases for Optional<UUID> as well. Having to do a new argument factory for each complex type with Optional is really irritating.

PR #1302, opened 2015-10-23
Fixing failing tests on non-English OS by changing default locale

fexbraun:
Another tests which was failing on non-English system because the whole (L10ned) validation message is verified.


joschi:
Thanks!

PR #1304, opened 2015-10-27
Add support for optional JodaTime arguments in JDBI

arteam:
An alternative implementation of an argument factory that handles JodaTime arguments wrapped by Guava's Optional.

It uses the fact that JDBI processes argument factories in the specific order, so we can process optional JodaTime arguments before other optional arguments.


arteam:
It's alternative implementation of #1301

nickbabcock:
My understanding of these two implementation is that #1301 is more a general solution for types contained within `Optional` and this specifically targets `DateTime`, correct? Would this solution mean that for each future type for jdbi, a `ArgumentFactory<Type>` and an associated `ArgumentFactory<Optional<Type>>` will have to be created?

arteam:
Correct. I don't think that's a big concern: optional arguments are not so useful as optional return values. Users, who would want to use optional arguments with own custom types, will have to write a specific factory.

PR #1306, opened 2015-10-28
Enforce LF line endings on Mustache and Freemarker template files

### LINKED ISSUE #1305: Fix Tests with LF Line Endings in Windows
Currently the tests will fail Windows because of 9f14b2e and 518d4b5

Having .gitattributes set to `auto` will convert `lf` to `crlf` on windows on checkout (and `crlf` to `lf` on checkin), which makes the tests fail (as the tests only check for `\n`). There are a couple solutions

### Solution 1:

Change gitattributes for mustache and freemarker files to force `lf` endings even on windows machines.

```
*.mustache eol=lf
*.ftl eol=lf
```

Not entirely thrilled with this solution as freemarker and mustache files would be treated special. We could set `eol=lf` for all file types, though carelessly doing it may have other tests, which rely on `System.lineSeparator()` to fail.

### Solution 2:

Keep on using `System.lineSeparator()` everywhere and put up with the annoyance it may bring

joschi:
Closes #1305.

@nickbabcock I've assigned this PR to you because you seem to have a Windows build machine to actually test it. :wink:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,5 +3,5 @@
 *.yml text=auto
 *.xml text=auto
 *.md text=auto
-*.mustache text=auto
-*.ftl text=auto
+*.mustache text eol=lf
+*.ftl text eol=lf


nickbabcock:
```
git clone -b issue-1305 https://github.com/joschi/dropwizard.git
cd dropwizard
mvn package
```

And all the tests passed :smile:

Issue #1308, opened 2015-10-28
NotNull validation of BaseReporterFactory.frequency fails


evnm:
Upon upgrading to Dropwizard 0.9.0, the `db migrate` task of an application I maintain failed with the following output:

```
Exception in thread "main" javax.validation.UnexpectedTypeException: HV000186: The constraint of type 'javax.validation.constraints.NotNull' defined on 'metrics.reporters[0].frequency' has multiple matching constraint validators which is due to an additional value handler of type 'io.dropwizard.validation.valuehandling.OptionalValidatedValueUnwrapper'. It is unclear which value needs validating. Clarify configuration via @UnwrapValidatedValue.
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.getConstraintValidatorInstanceForAutomaticUnwrapping(ConstraintTree.java:266)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.getInitializedConstraintValidator(ConstraintTree.java:163)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:116)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:87)
	at org.hibernate.validator.internal.metadata.core.MetaConstraint.validateConstraint(MetaConstraint.java:73)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateMetaConstraint(ValidatorImpl.java:592)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraint(ValidatorImpl.java:555)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForDefaultGroup(ValidatorImpl.java:490)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForCurrentGroup(ValidatorImpl.java:454)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateInContext(ValidatorImpl.java:406)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateCascadedConstraint(ValidatorImpl.java:770)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateCascadedConstraints(ValidatorImpl.java:656)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateInContext(ValidatorImpl.java:415)
[...]

PR #1309, opened 2015-10-28
Add UnwrapValidatedValue annotation to `BaseReporterFactory.frequency`

evnm:
Proposed fix for #1308. I'm not familiar with how validations are tested in the Dropwizard codebase, so please advise on how I can write a test for this change.


arteam:
Looks like this is a similar issue to #1292. Hibernate Validator 5.2 now requries `UnwrapValidatedValue` on containers. For some reason, we missed that during release candidates...

jplock:
We should get this into `release/0.9.x` as well for a `0.9.1` release I think.

joschi:
@jplock Agreed, see 15b6533bde4e223d3bf01437f9eb3dacffda85f8.

PR #1310, opened 2015-10-28
Use a minimally configured Hibernate Validator everywhere

nickbabcock:
Use a minimally configured Hibernate Validator everywhere, instead of the
default unconfigured Hibernate Validator that doesn't have Optional support
enabled.

Good news, no other part of the code was affected by the behavior seen in #1309. So this pull request is more preventive than anything.

I do wonder if there is a better implementation out there (I'd rather not expose `BaseValidator` to the world, but it's the best I came up with). Let me know if a better implementation exists :wink:

### DIFF
--- a/dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationFactoryFactoryTest.java
+++ b/dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationFactoryFactoryTest.java
@@ -1,34 +1,32 @@
 package io.dropwizard.configuration;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import com.google.common.io.Resources;
 import io.dropwizard.configuration.ConfigurationFactoryTest.Example;
 import io.dropwizard.jackson.Jackson;
-
-import java.io.File;
-
-import javax.validation.Validation;
-import javax.validation.Validator;
-
+import io.dropwizard.validation.BaseValidator;
 import org.junit.Before;
 import org.junit.Test;
 
-import com.google.common.io.Resources;
+import javax.validation.Validator;
+import java.io.File;
+
+import static org.assertj.core.api.Assertions.assertThat;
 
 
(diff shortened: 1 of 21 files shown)


jplock:
I think we should also get this into `release/0.9.x`

jplock:
I'm not 100% up to speed on the validator changes, but is this in any way related to the work you did in #1251?  I agree that seeing `BaseValidator.newConfiguration().buildValidatorFactory().getValidator();` everywhere doesn't feel very DRY and wonder if there is a better way.

joschi:
Wouldn't make more sense to move [`Validators`](https://github.com/dropwizard/dropwizard/blob/v0.9.0/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/Validators.java) into `dropwizard-validation` instead of creating the new `BaseValidator` class?

nickbabcock:
@joschi : Yes it would make sense if it was possible, but `Validators` depends on Jersey for the param validation, and I don't know a way around it :frowning:

arteam:
Could we add a factory method to `BaseValidator`? Something like this:
```java
public static Validator getValidator() {
    return newConfiguration().buildValidatorFactory().getValidator();
}
```

So we can use just `BaseValidator.getValidator()` instead of `BaseValidator.newConfiguration().buildValidatorFactory().getValidator()`, which breaks the Law of Demeter.

nickbabcock:
Ok, I took @arteam's suggestion. I initially didn't have it this way as I couldn't think of a logical reasoning how a Dropwizard user would utilize the function. But I'm good either way.

arteam:
LGTM!

arteam:
Cherry-picked to 0.9.x branch - 08d8863

nickbabcock:
And to answer your question, @jplock, no the pull requests aren't related by other than dealing with validation. This one deals with validation of configuration files in tests and the other deals with validating parts of the request.

PR #1319, opened 2015-10-31
Test {Ganglia,Graphite}ReporterFactory classes

joschi:
This PR adds tests for the `GangliaReporterFactory` and `GraphiteReporterFactory` classes to check whether they can be instantiated (and validated) using a `ConfigurationFactory`.


nickbabcock:
This is great, thanks!

PR #1323, opened 2015-11-03
Guava Optional<T>: Treat null values after conversion as absent value.

Trundle:
Some converters return a null value (e.g. in case of empty input).

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/guava/OptionalParamConverterProvider.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/guava/OptionalParamConverterProvider.java
@@ -1,6 +1,5 @@
 package io.dropwizard.jersey.guava;
 
-import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import org.glassfish.hk2.api.ServiceLocator;
 import org.glassfish.jersey.internal.inject.Providers;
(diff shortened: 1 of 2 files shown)


joschi:
LGTM. Thanks!

joschi:
Also merged in `master` (0a7e0a96ca152269d2d0546e5f7ad810c72da9bd) and `release/0.8.x` (963728419fc5fa7c5410f48be6470253a0a62a72).

nickbabcock:
I don't know if this can be considered a bug (see #986). The empty string _is_ significant and shouldn't be considered `absent`.

joschi:
@nickbabcock The case you've described (and which is mentioned in #986) is unaltered by this PR, see https://github.com/dropwizard/dropwizard/blob/137aa42560312966f18ccc732c769666286135e6/dropwizard-jersey/src/test/java/io/dropwizard/jersey/guava/OptionalQueryParamResourceTest.java#L47-L51.

This PR merely enables non-string `Optional<T>` method parameters to work correctly, e. g. `Optional<Integer>` like in the unit tests accompanying this change set. Before, this resulted in a `NullPointerException` in `AbstractParamValueExtractor<T>#convert(String)`:
```
java.lang.NullPointerException: the Function passed to Optional.transform() must not return null.
```

Or maybe I misunderstood your objections?

nickbabcock:
Oh yes, you're absolutely right! I saw `queryParam("value", "")` and had a flashback. :smile:

PR #1326, opened 2015-11-04
Make HealthCheck ExecutorService configurable in the yml-file

rvs-fluid-it:
Currently the threadpool used by the HealthChecks executorservice is hardcoded.
Perhaps it is useful to make it configurable from the yml-file.

Please provide feedback ...

### DIFF
--- a/dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
+++ b/dropwizard-client/src/test/java/io/dropwizard/client/DropwizardApacheConnectorTest.java
@@ -4,6 +4,7 @@ import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.health.HealthCheck;
 import io.dropwizard.Application;
 import io.dropwizard.Configuration;
+import io.dropwizard.health.ManagedExecutorServiceFactory;
 import io.dropwizard.jackson.Jackson;
 import io.dropwizard.jersey.validation.Validators;
 import io.dropwizard.setup.Environment;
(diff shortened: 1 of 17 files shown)


joschi:
@rvs-fluid-it Do you have a specific use-case which requires to use another executor service for health checks?

rvs-fluid-it:
In some cases I deploy my Dropwizard applications on a JEE-container (see https://github.com/rvs-fluid-it/wizard-in-a-box) In that context  it is considered bad practice to start threads which are not managed by the application server.

But irrespective of the deployment scenario, I think it is a valuable feature to be able to fine tune the used threadpool (minThreads, maxThreads, ...). Also from the perspective of operations it is interesting to have all the used threadpools enlisted in the yml-file.

joschi:
> In some cases I deploy my Dropwizard applications on a JEE-container

This is not an intended usage scenario for Dropwizard. :wink: 

Currently I tend to not merge this PR because I don't see an overwhelming reason that would justify the increased complexity in Dropwizard's codebase.

@dropwizard/committers Any opinions on that?

jplock:
I tend to agree.  This would set a bad precedent and then we'd have to start supporting more use-cases for deploying Dropwizard within a container.

mattnelson:
Is the root of the issue you are trying to solve that the health check thread pool isn't using a daemon thread factory? If so, I think that would be a simple change. Additionally, I don't think that dropwizard should prevent an application from shutting down because a health check is being executed.

rvs-fluid-it:
@mattnelson 
The root issue I'm trying to solve is that the number of threads used by the health check thread pool is not configurable.

rvs-fluid-it:
@joschi @dropwizard/committers 
I agree that a not supported use case is not a good driver to change the codebase.

So back to the drawing table.
But I welcome a bit of input from you.
a) Do you find it valuable, especially from an OPS point of view, that all the thread pools witch are started by Dropwizard are fine-tunable from the yml configuration file (number of threads, ...)?
b) (Class Design) What is or should be the responsibility of the Environment class? Should it just be a dumb wrapper object or should it also contain behaviour (=instantiating of a thread pool) ? In case if it also would contain behaviour, should the behaviour then be executed in the constructor?

joschi:
Closing due to inactivity.

PR #1327, opened 2015-11-05
Adding Websockets support to dropwizard

eitan101:
I've added a bundle, including some integration tests and docs.


joschi:
Thank you for your contribution.

Since this doesn't require any changes in Dropwizard itself, I'd suggest publishing the module as a 3rd party extension and add it to http://modules.dropwizard.io/thirdparty/

eitan101:
OK. I'll do that. A few questions:

1. The bundle consists of instrumented Jetty websocketContainer. Do you think that I can create a pull request to the metrics repository?
2. Could you add a reference to the official / third party bundles in the manual ?
3. Currently the documentation does not mentioning anything about websockets. I can create a piece of documentation that show how to do it by yourself, or add link to the third party bundles?

joschi:
> The bundle consists of instrumented Jetty websocketContainer. Do you think that I can create a pull request to the metrics repository?

Sure, @ryantenney will be happy to receive a PR, and even if it's not merged into the Metrics repository, you could still publish it as a third-party module. Personally I don't think there's too much use in pulling everything and the kitchen sink into one repo if it's not necessary at all. That's why I'm reluctant to include (optional) extensions into Dropwizard.

> Could you add a reference to the official / third party bundles in the manual ?

That's what http://modules.dropwizard.io/ is for. We should include a link to that page into the documentation, though.

> Currently the documentation does not mentioning anything about websockets. I can create a piece of documentation that show how to do it by yourself, or add link to the third party bundles?

Please add the appropriate documentation to the repository of the Websockets module.

eitan101:
@ryantenney, do you think that [instrumented jetty websocketContainer](https://github.com/eitan101/dropwizard/blob/master/dropwizard-websockets/src/main/java/io/dropwizard/websockets/WebsocketBundle.java#L49) should be part of the [metrics repository](https://dropwizard.github.io/metrics/3.1.0/manual/jetty/) or a third-party metrics module ?

eitan101:
I moved the project to [third party repository](https://github.com/LivePersonInc/dropwizard-websockets).

PR #1298, opened 2015-10-23
Improved Documentation of Example Application

fexbraun:
The documentation of the example application was not up-to-date.

I improved the steps by grouping and removing optional steps.

PR #1328, opened 2015-11-05
Fixed layout of the example.rst

fexbraun:
PR #1298 broke the layout: double space before asterisk is not working in sphinx


fexbraun:
current layout (locally rendered) looks like this

Issue #1036, opened 2015-05-08
Upgrade to Jetty 9.3.0


arteam:
This is an issue for future versions (maybe 1.0), because currently we are not ready for migration.

The Jetty development team actively works on 9.3.0 version. The main feature is HTTP/2 support, but also there are also many performance improvements and bugfixes .

I've been playing around with the last milestone build (9.3.0.M2) on HTTP/2 support in Dropwizard. The current work in this [branch](https://github.com/arteam/dropwizard/tree/jetty-93). Basically, I managed to create a module and start Dropwizard as an HTTP/2 server. But during this process I found several places, where we are not compatible:

* `org.eclipse.jetty.servlet.ServletTester` was removed
* `dropwizard-metrics-jetty9` doesn't support 9.3.0
* `SPDY` is not sipported in Jetty 9.3.0
* `GzipFilter` was removed (actually deprecated, but the code is stripped) and migrated to `GzipHandler`
* In `RequestLog` the signature of the `log` method is changed.

I plan to update the list according to the latest Dropwizard and Jetty milestones and convert this
issue to a pull request when we are ready.

jplock:
Relates to https://github.com/dropwizard/metrics/pull/823

joakime:
`org.eclipse.jetty.servlet.ServletTester` was not removed.

It's available still, but as part of the `<classifier>tests</classifier>` artifact.

```
<dependency>
    <groupId>org.eclipse.jetty</groupId>
    <artifactId>jetty-servlet</artifactId>
    <version>9.3.0.v20150612</version>
    <classifier>tests</classifier>
</dependency>
```

arteam:
Yes, thanks. I also figured it out. The branch I mentioned is a bit of outdated, because I didn't push the latest changes, but locally I added *tests* dependencies on `jetty-servlet` and `jetty-http` to use `ServletTester` and `HttpTester`.

devmage:
@arteam I pulled down your fork and merged master with your branch. Only conflict was with the `dropwizard-spdy/pom.xml` that you removed, which was trivially resolved (by removing it again). All current tests passed.

arteam:
Thanks. I've rebased the changes against current master and it should be mergeable now.

I didn't have time to work on this and 9.3.* support is not ready yet. Most of the items in the list are done, except HTTP/2, so probably some applications could be run on Jetty 9.3.* with this branch.

[...]

PR #1330, opened 2015-11-08
Upgrade to Jetty 9.3.5

arteam:
This PR ugrades Dropwizard baseline to Jetty 9.3.5. Relevant issue is #1036. 
It consists from several changes:

* `dropwizard-spdy` is removed, because it's not longer supported by Jetty 9.3. Its successor is HTTP 2.0 integration, which we wiil need to implement.
* Upgaded the Jetty dependency in the BOM file. 
* Added own `InstrumentConnectionFactory`, because `metrics` library doesn't support Jetty 9.3. See https://github.com/dropwizard/metrics/pull/823
* `BiDiGzipFilter` is moved to `BiDiGzipHandler`, because Jetty doesn't provide `GzipFilter` anymore.
Some parameters, that were accessible in the filter, are no longer exist. So, we had to remove them
from the handler configuration. 
* Finally, it looks like that Jetty now requires providing a path to a trust store in HTTPS configuration.
Even with disabled peer validation. I don't know a way to get it progmatically, so I provided a path that works on Linux systems (at least on Ubuntu-based). If anyone can provide a better solution, I will gladly hear about it.


jplock:
@dropwizard/committers should we also get this merged in now? It'd be nice to start prepping 1.0.0 release candidates as we'll want to make sure its heavily tested.

nickbabcock:
Awesome pull request, I'm a little hesitant about the trust store changes. I skimmed the Jetty 9.3 source code and it appears that it should still accept a null trust store, in which case will `SSLContext.init` will search for 

> the installed security providers will be searched for the highest priority implementation of the appropriate factory

arteam:
I checked Jetty's [SslContextFactory](https://github.com/eclipse/jetty.project/blob/master/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java) one more time. It looks like that by default
Jetty loads a truststore as same as a keystore. But if we start an application, it fails with 

```java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty.```

It looks like that JRE doesn't want to trust the specified in the keystore certificate. If we enable 
```yaml
validatePeers: false
``` 
in the configuration, then the application starts up without errors. I don't know, why this worked
previously. Anyway, looks like that enabling this option fixes the problem.

arteam:
As a side-note: users of PKCS12 certificates would need to specify `trustStoreType`.

nickbabcock:
Ok interesting, I'm not sure why it worked previously either. I wanted to play around with your branch but I'm unable to in Intellij, as various compilation errors crop up even after a fresh clone (it keeps wanting to use jetty 9.2.13). `mvn package` works fine though, so I'm unsure what the issue could be.

Also any reason not to use 9.3.6.v20151106?

arteam:
> Ok interesting, I'm not sure why it worked previously either. I wanted to play around with your 
> branch |but I'm unable to in Intellij, as various compilation errors crop up even after a fresh clone 
> (it |keeps |wanting to use jetty 9.2.13). mvn package works fine though, so I'm unsure what the 
> issue |could be.

I believe, one need to perform `mvn clean install` on `dropwizard-bom` beforehand. Otherwise Maven will pull old `dropwizard-bom` snapshot from the repository.

> Also any reason not to use 9.3.6.v20151106?

No reason. I just started working on this update before than 9.3.6 came up.

nickbabcock:
The bom trick fixed the issue. :+1:

PR #1332, opened 2015-11-09
Removed unused imports

fexbraun:
The beginning of going to a 1.0 Version is maybe a good timing to get rid of a lot of the IDE warnings.


joschi:
Thanks!

PR #1333, opened 2015-11-09
Removed ancient TODOs from the code

fexbraun:
The beginning of going to a 1.0 Version is maybe a good timing to get rid of ancient list of TODOs. Important stuff pops up every day, no need to keep TODOs from two years ago.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
@@ -72,12 +72,10 @@ public class Cli {
             }
             return true;
         } catch (ArgumentParserException e) {
-            // TODO: 5/25/13 <coda> -- make ArgumentParser#handleError not depend on System.err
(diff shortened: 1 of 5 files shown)


nickbabcock:
Are any of these todos still relevant (ie, have they been addressed already?)

fexbraun:
Hi @nickbabcock, let me pull up my comment from the commit to this PR:

Justification:
In general: lean software development - don't keep long lists of ancient items which will never be done.

Concrete:
- AbstractServerFactory also tested via DefaultServerFactoryTest,
SimpleServerFactoryTest
- DefaultServerFactory already has a Test
- Environment Test cost/benefit is arguable
- getApplicationContext() is out and public now, we can't change it
- Commented code in HttpsConnectorFactory not needed for 2 years now, commented code is a burden not a help

nickbabcock:
And it looks like `Cli` doesn't depend on System.err either, so this is good to me.

Thanks :+1:

PR #1342, opened 2015-11-18
Fix nested lists and other styling in example.rst

jklukas:
The current rendered docs aren't displaying the nested lists as intended, but rather lumping each block into a single line: http://www.dropwizard.io/0.9.1/docs/manual/example.html


nickbabcock:
Good catch.

PR #1345, opened 2015-11-19
HTTP/2 and server push support

arteam:
This PR adds support for HTTP/2 over TLS, HTTP/2 plain text connections and server push to Dropwizard.  This is based on the work in Jetty 9.3 branch.

We use to have SPDY support, but this protocol is completelty deprecated and supressed by
HTTP/2. Jetty 9.3.* now provides only HTTP/2 support. One important distinction from SPDY is that HTTP/2 can work over TLS, as well as over a plain connection. The latter connection type is not supported by major browsers, but could be useful for internal network installations or for handling non-enctypted traffic from load balancers.

This PR provides two connection factories for each connection type. HTTP/2 over TLS has `http2` key, HTTP/2 over plain text - `http2c`.

Other important feature is support for the server push techonology. The server can send additional
resources along with the requested resource to clients on a single response. This could be a big
perfomance optimization for web pages that contain many static assets.

Jetty has `ServerPushFilter` which provides a good heuristic for determining primary and additinal resources. It uses URI from the `Referer` header and request timestamps to guess that resources are linked with each other. This change adds a simple factory for enabling this filter from the configuration. By default it's disabled.


dropwizard-http2/src/main/java/io/dropwizard/http2/Http2CConnectorFactory.java
```diff
@@ -0,0 +1,93 @@
...
+ *         </td>
+ *     </tr>
+ * </table>
+ * <p/>
+ * For more configuration parameters, see {@link HttpsConnectorFactory}.
+ * @see HttpConnectorFactory
+ */
+@JsonTypeName("http2c")
+public class Http2CConnectorFactory extends HttpConnectorFactory {
+
```
jplock: Should we add `@Min` and maybe `@Max` annotations onto these values?

arteam: Certainly, makes configuration more robust.

arteam:
Resubmitted as #1349

PR #1346, opened 2015-11-19
Fix typo in Getting Started section of docs

tomshen:
There's a typo in the "And Friends" section of the "Getting Started" docs. `JSR-349` should be `JSR-303`, as specified in the links below that section.


nickbabcock:
Correct me if I'm wrong, but JSR-349 is the reference implementation to JSR-303, so the link name may be wrong -- which is correct? :smile:

tomshen:
JSR-349 looks like the right link, so I changed the link name and URL instead.

nickbabcock:
Ok looks good, to me.

PR #1347, opened 2015-11-20
Entities that can't be serialized now return a 500

nickbabcock:
Jackson is good about serializing types that conform to the Bean spec, but
needs help serializing and deserializing types that don't (normally with
the use of annotations). Previous behavior would return a 400 (Bad
Request) when the server would attempt to return an entity that couldn't
be serialized. This is misleading because it gives the illusion that the
fault lies with the client instead of the server.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -2,6 +2,8 @@ package io.dropwizard.jersey.jackson;
 
 import com.fasterxml.jackson.core.JsonGenerationException;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


jplock:
I think this is a good idea and I don't know if someone would be relying on this or whatnot, so it should be find to break BC in 1.0.0

brentryan:
I'm not sure I agree entirely here. The error could still be the clients issue, right? If they send in bad data then a 400 is appropriate.

nickbabcock:
That should be covered by the `returnsA400ForNonDeserializableRequestEntities` test, but if you can find an edge case, I'd be happy to re-evaluate my implementation.

brentryan:
Na, I just read the conversation above and not the code.  It looks good to me and pasted my test locally.  Thanks!

nickbabcock:
No problem, great to have additional eyes for these topics!

mikewatt:
I've come come across a scenario where this causes a 500, which should probably be a 4xx. I believe it is related to this change, although I'm not sure where the appropriate fix is.

The JSR310 module re-throws parser exceptions as JsonMappingExceptions (https://github.com/FasterXML/jackson-datatype-jsr310/blob/7eccc43bbaac9e742f80b3897256563e28812345/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DeserializerBase.java#L60), so if the client request body contains a value that is not parseable into the necessary instance type, a 500 is generated.

Is this fix to handle this in the exception mapper, or should the JSR310 module throw a different exception, or something else entirely?

nickbabcock:
Thank you for the report mike. I can confirm that this is undesirable behavior and will have a fix out shortly.

PR #1348, opened 2015-11-21
Fix broken links to manual in javadoc

magnusr:
Would be nice if the docs for the latest version were available at `/latest/docs` or something like that we one could point to the docs without specifying a particular version, but lacking that, pointing to `/0.9.1/docs` is the best I can think of.


joschi:
Thanks!

PR #1349, opened 2015-11-21
HTTP/2 and server push support

arteam:
This PR adds support for HTTP/2 over TLS, HTTP/2 plain text connections and server push to Dropwizard.  This is based on the work in Jetty 9.3 branch.

We use to have SPDY support, but this protocol is completelty deprecated and supressed by
HTTP/2. Jetty 9.3.* now provides only HTTP/2 support. One important distinction from SPDY is that HTTP/2 can work over TLS, as well as over a plain connection. The latter connection type is not supported by major browsers, but could be useful for internal network installations or for handling non-enctypted traffic from load balancers.

This PR provides two connection factories for each connection type. HTTP/2 over TLS has `http2` key, HTTP/2 over plain text - `http2c`.

Other important feature is support for the server push techonology. The server can send additional
resources along with the requested resource to clients on a single response. This could be a big
perfomance optimization for web pages that contain many static assets.

Jetty has `ServerPushFilter` which provides a good heuristic for determining primary and additinal resources. It uses URI from the `Referer` header and request timestamps to guess that resources are linked with each other. This change adds a simple factory for enabling this filter from the configuration. By default it's disabled.


jplock:
What needs to be done to have `JerseyClient` support `HTTP/2`?  Is it just a matter of waiting until its been upgraded?

arteam:
Yes, `JerseyClient` and Apache `HttpClient` don't support HTTP/2 currently.

I know 3 Java client HTTP/2 libraries:
* Jetty [Http2Client](https://github.com/eclipse/jetty.project/tree/master/jetty-http2/http2-client)
* [OkHttp](https://github.com/square/okhttp)
* Netty [Client](https://github.com/netty/netty/tree/master/example/src/main/java/io/netty/example/http2/helloworld/client) 

I like the first one from Jetty, because:

* It provides an asynchronous API; 
* `h2` and `h2c` support; 
* Has the same core as `jetty-http2-server`. It provides a good interopability with the server implementation. Despite that HTTP/2 is a standart, there could be some idiosyncrasy .

jplock:
Yea, I agree using the Jetty Http2Client is the best choice for now.  Thanks!

brentryan:
Have you guys looked into switching to retrofit for clients instead? It uses OkHttp under the hood and has nice async features.

arteam:
> Have you guys looked into switching to retrofit for clients instead? It uses OkHttp under the hood and has nice async features.

Personally, I don't like `OkHttp` and `Retrofit` much, because they are developed for Android in the first place. Currently they run on OpenJDK/HotSpot, but maintainers could drop its support in any time.

Yes, the library has a fluent API and code quality is really good. Nevertheless, I wouldn't pull it into the Dropwizard core. But we open to add any integration to the list of 3rd party modules, if anyone develop it.

dropwizard-http2/src/main/java/io/dropwizard/http2/Http2CConnectorFactory.java
```diff
@@ -0,0 +1,101 @@
...
+        final HTTP2ServerConnectionFactory http2c = new HTTP2CServerConnectionFactory(httpConfig);
+        http2c.setMaxConcurrentStreams(maxConcurrentStreams);
+        http2c.setInitialStreamSendWindow(initialStreamSendWindow);
+
+        // The server connector should use HTTP/1.1 by default. It affords to the server to stay compatible
+        // with old clients. New clients which want to use HTTP/2, however, will make an HTTP/1.1 OPTIONS
+        // request with an Upgrade header with "h2c" value. The server supports HTTP/2 clear text connections,
+        // so it will return the predefined HTTP/2 preamble and the client and the server will switch to the
+        // new protocol.
+        return buildConnector(server, new ScheduledExecutorScheduler(), buildBufferPool(), name, threadPool,
```
jplock: Should this `ScheduledExecutorScheduler()` be `Managed` so its stopped and started alongside Jetty?

arteam: It implements `Scheduler` which in turn extends `LifeCycle`. It will be closed along with the connector.

brentryan:
@arteam why does it matter what they developed the http client for if it works better, has clean code, is resilient, and has decent perf when compared to anything else out there. It runs fine on oracle jdk as well so I'm confused by your comment.

jplock:
@arteam was there anything else you'd like to add to this, otherwise I think its ready for merging?  Any other feedback @dropwizard/committers?  This is great.  Glad to see this get into 1.0.

arteam:
I'm think that the PR is ready to merge. But I would wait for another feedback from @dropwizard/committers, because it's a quite a big change and I could miss something.

Also it would be great, if someone tested the HTTP/2 integration with a SSL cerftificate issued by CA. I tested it only on self-signed ceritificates.

arteam:
BTW, a very interesting [talk](http://unrestful.io/2015/12/01/devoxx.html) about HTTP/2 and Java clients/servers from Fabian Stäber.

nickbabcock:
Something is causing this pull request to fail consistently locally and on Travis.

> testSqlExceptionIsHandled(io.dropwizard.hibernate.JerseyIntegrationTest)
javax.ws.rs.ProcessingException: Server-side request processing failed with an error.

arteam:
Looks like I mixed up this PR with another (#1361) in some time. Now the build should pass tests.

nickbabcock:
I attempted to use this pull request with a CA cert, but there is a current problem on windows. Every test in `Http2IntegrationTest` fails with `no valid keystore`. I'm investigating, but debugging tls issues can sometimes be black magic :stuck_out_tongue: 

Additionally, [HTTPCLIENT-1672](https://issues.apache.org/jira/browse/HTTPCLIENT-1672) is tracking HTTP 2.0 status. Unfortunately, I doubt it will land anytime soon.

nickbabcock:
The fix was to add another config override for the trust store path:

```java
    @Rule
    public final DropwizardAppRule<Configuration> appRule = new DropwizardAppRule<>(
            FakeApplication.class, ResourceHelpers.resourceFilePath("test-http2.yml"),
            Optional.of("tls_http2"),
            ConfigOverride.config("tls_http2", "server.connector.keyStorePath",
                    ResourceHelpers.resourceFilePath("stores/http2_server.jks")),
            ConfigOverride.config("tls_http2", "server.connector.trustStorePath",
                ResourceHelpers.resourceFilePath("stores/http2_client.jts"))
    );
```

nickbabcock:
Ok, I tested the branch (after I applied my small fix) with a CA cert and I was able to see in the request logs HTTP 2.0!

I'm not sure if you noticed this as well, but it looks like I had to supply at least one `supportedCipherSuites` in the yaml, else Chrome and Jetty couldn't negotiate. I'm curious if the HTTP 2.0 Jetty starts with zero supported cipher suites. If so, we'll have to document this and potentially provide a sane default.

Anyways, awesome work. As soon as I can compile/test on windows, this PR is good to merge :+1:

jplock:
@arteam did you want to comment on @nickbabcock's comment above regarding `supportedCipherSuites`?  Otherwise, we should get this merged in. :+1:

arteam:
Yes, I wanted to look at this closely, but hadn't a chance. Was busy with work and  holiday stuff. I am going to give my feedback today or tomorrow.

arteam:
Looks like that HTTP/2 spec [requires](http://http2.github.io/http2-spec/index.html#rfc.section.9.2.2) the support of `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` cipher on the server side. I believe, this is the default cipher for modern browsers: see this [gist](https://gist.github.com/mikemaccana/5e12d4d2b8dcfc87b197). 

Therefore, I would argue that we should force it by default in the connector. Users, who want to allow other cipher suites, have the ability to set explicitly them in the configuration.

arteam:
@nickbabcock, could you try the new version? HTTP/2 support works for me with Firefox 43 on Linux, but I would glad to see how it's going for users in other operation systems.

nickbabcock:
It doesn't look that you applied my suggested patch of:

> The fix was to add another config override for the trust store path:
> 
> ```java
>     @Rule
>     public final DropwizardAppRule<Configuration> appRule = new DropwizardAppRule<>(
>             FakeApplication.class, ResourceHelpers.resourceFilePath("test-http2.yml"),
>             Optional.of("tls_http2"),
>             ConfigOverride.config("tls_http2", "server.connector.keyStorePath",
>                     ResourceHelpers.resourceFilePath("stores/http2_server.jks")),
>             ConfigOverride.config("tls_http2", "server.connector.trustStorePath",
>                 ResourceHelpers.resourceFilePath("stores/http2_client.jts"))
>     );
> ```

Any reason why not?

With the patch applied, HTTP/2 support works with Chrome on Windows.

I'll have to double check next week Monday that the cipher suite now works out of the box.

arteam:
> Any reason why not?

Sorry, I misinterpeted your comment. I thought about it in the context: _a HTTP/2 test requires providing a trust store_ (which had been already done), not in the context: _the current HTTP/2 test provides an invalid truststore_. You are absolutely right, `/etc/ssl/certs/java/cacerts` truststore exists
only on Unix systems. I've applied your changes.

nickbabcock:
Awesome, thank you :sunglasses:

arteam:
Great! Thanks for rigorous review and testing. Windows users will certainly appreciate it.

PR #1354, opened 2015-11-25
Add support for setting default headers in Apache HTTP client

### LINKED ISSUE #1353: HttpClientConfiguration can't configure default headers
I want to set default headers for my `HttpClient` instances (via `org.apache.http.impl.client.HttpClientBuilder.setDefaultHeaders`), but `HttpClientConfiguration` doesn't support it.

To make this happen I had to subclass both `HttpClientConfiguration` and `HttpClientBuilder`. It would be nice if it were just built in.

My solution is pretty simple. Doubtless better solutions exist. But it looks like this: 

* Subclass `HttpClientConfiguration`, adding this:
```
    @JsonProperty("defaultHeaders")
    private Collection<? extends Header> defaultHeaders;

    public void setDefaultHeaders(final Map<String, String> map) {
        final List<BasicHeader> headers = map.entrySet().stream()
            .map((e) -> new BasicHeader(e.getKey(), e.getValue()))
            .collect(Collectors.toList());
        defaultHeaders = Collections.unmodifiableList(headers);
    }

    public Collection<? extends Header> getDefaultHeaders() {
        return defaultHeaders;
    }
```
* Subclass `HttpClientBuilder` with this:
```
   private HttpClientConfiguration config;

    @Override
    public MyHttpClientBuilder using(final HttpClientConfiguration config) {
        this.config = config;
        super.using(config);
        return this;
    }

    @Override
    protected ConfiguredCloseableHttpClient createClient(
            final org.apache.http.impl.client.HttpClientBuilder builder,
            final String name) {
        if (config instanceof MyHttpClientConfiguration) {
            final MyHttpClientConfiguration myConfig =
                    (MyHttpClientConfiguration)config;
            builder.setDefaultHeaders(myConfig.getDefaultHeaders());
        }
        return super.createClient(builder, manager, name);
    }
```

arteam:
Add the ability to set predefined HTTP headers for each HTTP request of an HTTP client.

Apache `HttpClientBuilder` already supports such functionality and we should just delegate building from Dropwizard `HttpClientBuilder`.

In the result users can provide own headers without subclassing Dropwizard `HttpClientBuilder`.

Resolves #1353

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
@@ -13,6 +13,7 @@ import io.dropwizard.lifecycle.Managed;
 import io.dropwizard.setup.Environment;
 import io.dropwizard.util.Duration;
 import org.apache.http.ConnectionReuseStrategy;
+import org.apache.http.Header;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpResponse;
 import org.apache.http.auth.AuthScope;
@@ -41,6 +42,7 @@ import org.apache.http.impl.conn.SystemDefaultDnsResolver;
 import org.apache.http.protocol.HttpContext;
 
 import java.io.IOException;
+import java.util.List;
 
 /**
  * A convenience class for building {@link HttpClient} instances.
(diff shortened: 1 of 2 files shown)


jplock:
LGTM!  Thanks!

PR #1356, opened 2015-11-25
Dropwizard always adds "server" and "check" commands. This is restrictive

Randgalt:
I've started using Dropwizard for CLI apps and other types of things. This way, I can create an entire system that consumes config in the same way, etc. However, DW always adds server and check commands. I don't always want this. Add a method to not add these.


jplock:
Thanks for the contribution.  I don't see any issues with this PR.

Randgalt:
Thanks!

PR #1360, opened 2015-11-27
Simplify the ordering of Bundles and ConfiguredBundles.

ojacobson:
Applications that use both are vulnerable to surprises if

1. The order of bundle initialization matters (eg. if some bundles provide resources that can't be initialized until `run`), and

2. A `Bundle` implementation wants to depend on a resource provided by a `ConfiguredBundle`.

This simplifies the ordering: bundles are run in a strict FIFO order, regardless of bundle type.

(This bit me when a refactoring turned a `Bundle` into a `ConfiguredBundle` and silently moved it _way later_ in the initialization order.)


ojacobson:
I would assume that changing the initialization order of bundles counts as a breaking change, so I don't expect to see this out in a hurry. However, the current behaviour is pretty surprising: two calls to what appear to be the same API that are ordered the obvious when the operands are both `Bundle` or both `ConfiguredBundle` cease to be ordered the obvious way when the types differ.

joschi:
@ojacobson IMHO dependent bundles are an anti-pattern. Could you please elaborate on the use case a little bit or describe how that currently is a problem in your applications?

FWIW, I'd vote for removing `Bundle` completely and only use `ConfiguredBundle<T>`.
@dropwizard/committers Any opinion on that?

ryankennedy:
The biggest thing will be forcing prior implementations of Bundle to now depend on some `Configuration` subclass, even when they don't use a configuration. Would you just have everyone change their `Bundle` to a `ConfiguredBundle<Configuration>`?

I looked for things implementing `Bundle` in the project and I wonder how this would affect something like `[MigrationsBundle](https://github.com/dropwizard/dropwizard/blob/79cb3d6ac0d644eb415cf230ad26e694c19ab2ac/dropwizard-migrations/src/main/java/io/dropwizard/migrations/MigrationsBundle.java)`. In that case, things might actually clean up a bit if we only had `ConfiguredBundle`.

ojacobson:
@joschi Well, take the [Hibernate bundle](https://dropwizard.github.io/dropwizard/0.7.1/docs/manual/hibernate.html) as an example of bundle dependencies: code that uses the Hibernate bundle that lives directly in the app class obviously runs after the Hibernate bundle gets configured, but if that code gets refactored into a bundle of its own, for example if it's built separately or to structure the app a bit, then that bundle needs a way to get at the Hibernate bundle, or at least to get its entity manager:

```java
public abstract class UsersBundle implements Bundle {
    /* … */
    public void run(Environment environment) {
        environment.jersey().register(new UserAPI(getEntityManager())); // Looks sane, but isn't
    }

    protected abstract EntityManager getEntityManager();
}

public class ExampleApp extends Application<ExampleConfig> {
    /* … */
    public void bootstrap(Bootstrap<ExampleConfig> bootstrap) {
        bootstrap.addBundle(hibernateBundle);
        bootstrap.addBundle(new UsersBundle() {
            protected abstract EntityManager getEntityManager() { return hibernateBundle.getEntityManager(); }); // The dependency
    }
}
```

Code like this _looks_ safe, since the bundles are registered in the right order, but is actually _unsafe_, because the entity manager in the Hibernate bundle will only be created after the UsersBundle's run method has already completed.

I use exactly this pattern of refactoring functional chunks into bundles, which depend on services exposed by other bundles using overridden `get` methods, in some of my own apps. For the most part, it works really well, but bundle ordering is a constant headache. My current workaround is to push the `get` methods down into the leaf nodes of the dependency graph (into UsersAPI, in the example above), and lazily fetch things to a fault, but that's a _lot_ of code spent overriding trivial getters.

@ryankennedy It's not a huge problem to support a wrapper, at least as a transitional state. The code for that's in this pull request.

PR #1363, opened 2015-11-28
Move the reportPlugins section in the archetype

arteam:
The `reportPlugins` section has been deprecated since version 3.3 of `maven-site-plugin` in favour of the classic reporting configuration. Currently my IDE marks the former as not valid according to an XSD schema. 

Explanation is [here](http://maven.apache.org/plugins/maven-site-plugin/maven-3.html#New_Configuration_Maven_3_only_no_reports_configuration_inheritance).

Diff is kind of weird, better to view with `?w=1`

### DIFF
--- a/dropwizard-archetypes/java-simple/src/main/resources/archetype-resources/pom.xml
+++ b/dropwizard-archetypes/java-simple/src/main/resources/archetype-resources/pom.xml
@@ -125,26 +125,23 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
-                <artifactId>maven-site-plugin</artifactId>
-                <version>3.4</version>
-                <configuration>
-                    <reportPlugins>
-                        <plugin>
-                            <artifactId>maven-project-info-reports-plugin</artifactId>
-                            <version>2.4</version>
-                            <configuration>
-                                <dependencyLocationsEnabled>false</dependencyLocationsEnabled>
-                                <dependencyDetailsEnabled>false</dependencyDetailsEnabled>
-                            </configuration>
-                        </plugin>
-                        <plugin>
-                            <artifactId>maven-javadoc-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


dropwizard-archetypes/java-simple/src/main/resources/archetype-resources/pom.xml
```diff
@@ -125,26 +125,23 @@
...
+        </plugins>
+    </build>
+
+    <reporting>
+        <plugins>
             <plugin>
-                <artifactId>maven-site-plugin</artifactId>
-                <version>3.4</version>
+                <artifactId>maven-project-info-reports-plugin</artifactId>
+                <version>2.4</version>
```
joschi: The latest version of the maven-project-info-reports-plugin is 2.8.1.

joschi:
Shouldn't this PR be based on `master` instead of `release/0.9.x`?

arteam:
> Shouldn't this PR be based on master instead of release/0.9.x?

I think we should fix this for 0.9.x too. After that we could apply this change on master.

dropwizard-archetypes/java-simple/src/main/resources/archetype-resources/pom.xml
arteam: I used the same plugin version (just moved the plugin from one section to another). 
Do you think we should upgrade it?

joschi: Yes, I think the plugin should be updated in the scope of this PR. Saves a follow-up commit. :wink:

arteam: Ok, I've amended the commit. Please review.

joschi:
LGTM. :+1:

PR #1364, opened 2015-11-28
Add a development profile to Maven

arteam:
Add a development profile that allow to skip following things during project installation:

* Running Findbugs checks;
* Generating Javadoc;
* Creating a verifying a sample project from the archetype;
* Generating a shadowed artifact for the example project.

This allows to decrease the time to clean and install project from 4-5 minutes to 20-30 seconds during development.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -270,6 +270,13 @@
                 <javadoc.doclint.none>-Xdoclint:none</javadoc.doclint.none>
             </properties>
         </profile>
+        <profile>
+            <id>dev</id>
+            <properties>
+                <maven.javadoc.skip>true</maven.javadoc.skip>
+                <findbugs.skip>true</findbugs.skip>
+            </properties>
+        </profile>
     </profiles>
 
     <dependencyManagement>
--- a/dropwizard-archetypes/pom.xml
+++ b/dropwizard-archetypes/pom.xml
@@ -163,4 +163,14 @@
       </resource>
     </resources>
   </build>
+
+  <profiles>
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -270,6 +270,29 @@
                 <javadoc.doclint.none>-Xdoclint:none</javadoc.doclint.none>
             </properties>
         </profile>
+        <profile>
+            <id>dev</id>
+            <build>
```
joschi: I'd prefer simply setting Maven properties (i. e. [`maven.javadoc.skip`](https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#skip) and [`findbugs.skip`](https://gleclaire.github.io/findbugs-maven-plugin/check-mojo.html#skip)) so that these can still be overridden on the command line. Running `mvn -Pdev -Dfindbugs.skip=false` would for example skip the Javadoc generation but still run the FindBugs checks.

This change would also make the profile itself shorter:
```xml
<profile>
    <id>dev</id>
    <properties>
        <maven.javadoc.skip>true</maven.javadoc.skip>
        <findbugs.skip>true</findbugs.skip>
    </properties>
</profile>
```

arteam: Got it. That definitely looks more neat.

I used the same technique for `maven-invoker-plugin` and `maven-archetype-plugin`. Unfortunately I didn't manage to find a `skip` property for `maven-shade-plugin`.

joschi:
LGTM, thanks!

PR #1365, opened 2015-11-28
Merge dropwizard-java8 addon into mainline

### LINKED ISSUE #962: Upgrade to jdk 8

joschi:
[`dropwizard-java8`](https://github.com/dropwizard/dropwizard-java8) used to provide support for Java 8 classes like `Optional<T>` and the Java Date and Time API (JSR-310) for Dropwizard. Given that the baseline for Dropwizard is now Java 8, those additions can be maintained in the mainline repository.

The only backward-incompatible change is in the `io.dropwizard.auth.Authenticator` interface which is now using `java.util.Optional` instead of the equivalent class from Google Guava.

Closes #962

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/Authenticator.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/Authenticator.java
@@ -1,8 +1,7 @@
 package io.dropwizard.auth;
 
-import com.google.common.base.Optional;
-
 import java.security.Principal;
+import java.util.Optional;
 
 /**
  * An interface for classes which authenticate user-provided credentials and return principal
@@ -14,10 +13,10 @@ import java.security.Principal;
 public interface Authenticator<C, P extends Principal> {
     /**
      * Given a set of user-provided credentials, return an optional principal.
-     * <p/>
-     * If the credentials are valid and map to a principal, returns an {@code Optional.of(p)}.
-     * <p/>
-     * If the credentials are invalid, returns an {@code Optional.absent()}.
+     *
+     * If the credentials are valid and map to a principal, returns an {@link Optional#of(Object)}.
+     *
+     * If the credentials are invalid, returns an {@link Optional#empty()}.
      *
      * @param credentials a set of user-provided credentials
      * @return either an authenticated principal or an absent optional
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
@@ -3,7 +3,6 @@ package io.dropwizard.auth;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
-import com.google.common.base.Optional;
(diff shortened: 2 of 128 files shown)


jplock:
This looks good to me, thanks for all the work in integrating this!  Any other feedback @dropwizard/committers ?

arteam:
@joschi, what do you think about creating `dropwziard-jdbi-guava-optional` module  with Guava's 
`Optional` support instead of mixing it with Java8's `Optional`? These classes basically exclude each
other, and I don't think many folks use them together in the same codebase. It would be great to let
users choose what `Optional` support to use. I think we should encourage users to switch to Java8's
`Optional`, because it is shipped with JRE, is a more flexible type and will probably supress Guava's 
`Optional` in future. Folks, who for some reasons don't want to switch, will have to enable an 
additional module and use a custom `DBIFactory`. It could be implemented somehting like that:

```java
public class GuavaOptionalDBIFactory extends DBIFactory {

    @Override
    public DBI build(Environment environment, PooledDataSourceFactory configuration,
                     ManagedDataSource dataSource, String name) {
        final DBI dbi = super.build(environment, configuration, dataSource, name);
        dbi.registerArgumentFactory(new GuavaOptionalArgumentFactory(
                                    configuration.getDriverClass()));
        dbi.registerContainerFactory(new GuavaOptionalContainerFactory());

        final Optional<TimeZone> timeZone = databaseTimeZone();
        dbi.registerArgumentFactory(new GuavaOptionalJodaTimeArgumentFactory(timeZone));
        dbi.registerArgumentFactory(new GuavaOptionalLocalDateArgumentFactory());
        dbi.registerArgumentFactory(new GuavaOptionalLocalDateTimeArgumentFactory());
        dbi.registerArgumentFactory(new GuavaOptionalInstantArgumentFactory(timeZone));
        dbi.registerArgumentFactory(new GuavaOptionalOffsetTimeArgumentFactory(timeZone));
        dbi.registerArgumentFactory(new GuavaOptionalZonedTimeArgumentFactory(timeZone));

        return dbi;
    }
}
```

What do you think?

arteam:
I am also a bit of worried about mixing Joda-Time and JSR 310 support. Unfortunately, I don't see a way how could we modularize the support of these libraries. Removing Joda-Time support in favour of JSR310 is also not an option. I believe, for the time being, we have to force users to import Joda-Time, even if they don't use it.

joschi:
@arteam In my opinion Dropwizard should "just work" with Google Guava and Joda-Time additionally to the APIs provided by JDK 8 and higher (JSR-310 and Optional).
Swapping out the support for those libraries (which are by the way being used throughout Dropwizard itself) doesn't gain us anything in my opinion and will just put the burden to add additional dependencies on our users.
Once we got rid of Google Guava or Joda-Time from the Dropwizard core components, we can move support for them from `dropwizard-jdbi` into separate modules, but for now I'd like to keep it.

Backward compatibility with older Dropwizard versions isn't exactly required for Dropwizard 1.0.0, but it's a nice feature. :wink:

arteam:
Ok, fair enough. The rest changes look good to me. Thanks for your hard work!

joschi:
@jplock  @arteam Thanks for reviewing!

PR #1369, opened 2015-12-03
fix links to configuration manual

tonygaetani:
use project.version rather than having to change the version for each release

I got the idea from the pom.xml distribution management site


joschi:
Unfortunately that won't work without some kind of source filtering/templating in Maven (similar to resource filtering).

PR #1370, opened 2015-12-05
Encoding of properties files is ISO-8859-1

bripkens:
Java properties files are assumed to be ISO-8859-1 (latin-1) encoded.


joschi:
Thanks!

PR #1371, opened 2015-12-06
Fix Checkstyle warnings

arteam:
Another round of fixing Checkstyle warnings. In general:

* Remove star imports;
* Enforce immutability of local variables;
* Enforce the maximum line length of 128 symbols.


arteam:
The build on Travis is really flaky, but it completes successfully on my machine. Let see how it will going.

nickbabcock:
Looks good to me. I see that lambdas and streams are starting to pop up :sunglasses:

PR #1372, opened 2015-12-06
Simplify auth filters

arteam:
The current implementation of filters is rather cumbersome and doesn't really reuses the code.

This change:

* separates parsing credentials and authentication of a request;
* reuses the code of setting the security context;
* improves readability of the credentials parsing methods;
* adds additional documentation about the credentials format and authentication process.

The diff is better viewed with `?w=1`.


nickbabcock:
Nicely done, looks a lot better.

PR #1374, opened 2015-12-07
Added documentation on how to get started with Dropwizard Archetype

jansoren:


jplock:
@arteam do we need to mention anything about the dev profile you added in https://github.com/dropwizard/dropwizard/pull/1364 in these docs?

jansoren:
@arteam and @jplock - It was difficult to find the dropwizard archetype in the documentation. Is it possible to mention it in the "Getting Started"-section in http://www.dropwizard.io/ ?

dropwizard-archetypes/java-simple/src/main/resources/archetype-resources/src/main/java/__name__Application.java
```diff
@@ -3,16 +3,20 @@
...
 
+    private String applicationName;
+
     public static void main(final String[] args) throws Exception {
         new ${name}Application().run(args);
     }
 
     @Override
     public String getName() {
-        return "${name}";
```
jplock: I prefer the previous approach more as it was easier to understand and more standard.

jplock:
We also have the example app, so I don't think we have to go too in-depth in the archetype.

jplock:
@jansoren I think adding a section into the `Getting Started` (https://github.com/dropwizard/dropwizard/blob/master/docs/source/getting-started.rst) is a great idea.

jansoren:
I will make the changes and make a new pull request :)

PR #1376, opened 2015-12-07
Print help once on invalid command line arguments

### LINKED ISSUE #1357: Command help printer incorrectly reports errors
If I add a command that has required arguments, the "-h" option doesn't suppress error messages. E.g., add this command:

```java
public class MyCommand extends Command {
    public MyCommand() {
        super("test", "Do a test");
    }     

    @Override
    public void configure(Subparser subparser)
    {
        subparser.addArgument("types").choices("a", "b", "c").help("Type to use");
    }

    @Override
    public void run(Bootstrap<?> bootstrap, Namespace namespace) throws Exception {
        // ...
    }
}
```

If you run this in an app with the arguments "test -h". You get this:

```
usage: java -jar project.jar test [-h] {a,b,c}

Do a test

positional arguments:
  {a,b,c}                Type to use

optional arguments:
  -h, --help             show this help message and exit

too few arguments
usage: java -jar project.jar test [-h] {a,b,c}

Do a test

positional arguments:
  {a,b,c}                Type to use

optional arguments:
  -h, --help             show this help message and exit

Process finished with exit code 1
```

Note the "too few arguments" message. Passing "-h" should stop any argument validation and just print help.

nickbabcock:
Currently, if `-h` is provided on the commandline and invalid arguments are specified as well, the help message will be printed twice. This is not desired (the help message should only be printed once)

This change is not backwards compatible for those who need to see that `is-help` is set in the `run` command (this would only be when the case when all arguments are optional)

Closes #1357

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
@@ -75,7 +75,9 @@ public class Cli {
                 command.run(bootstrap, namespace);
             }
             return true;
-        } catch (HelpScreenException e) {
+        } catch (HelpScreenException ignored) {
+            // This exception is triggered when the user passes in a help flag.
+            // Return true to signal that the process executed normally.
             return true;
         } catch (ArgumentParserException e) {
             stdErr.println(e.getMessage());
--- a/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
@@ -11,6 +11,7 @@ import net.sourceforge.argparse4j.inf.ArgumentParser;
 import net.sourceforge.argparse4j.inf.ArgumentParserException;
 import net.sourceforge.argparse4j.inf.Namespace;
 import net.sourceforge.argparse4j.inf.Subparser;
+import net.sourceforge.argparse4j.internal.HelpScreenException;
 
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
@@ -70,11 +71,11 @@ public class Cli {
                 parser.printVersion(stdOut);
             } else {
                 final Namespace namespace = parser.parseArgs(arguments);
-                if (namespace.get("is-help") == null) {
-                    final Command command = commands.get(namespace.getString(COMMAND_NAME_ATTR));
-                    command.run(bootstrap, namespace);
-                }
+                final Command command = commands.get(namespace.getString(COMMAND_NAME_ATTR));
+                command.run(bootstrap, namespace);
             }
+            return true;
+        } catch (HelpScreenException e) {
             return true;
         } catch (ArgumentParserException e) {
             stdErr.println(e.getMessage());
@@ -138,7 +139,7 @@ public class Cli {
                         Map<String, Object> attrs, String flag, Object value)
                 throws ArgumentParserException {
             parser.printHelp(out);
-            attrs.put("is-help", Boolean.TRUE);
+            throw new HelpScreenException(parser);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


nickbabcock:
Just wanted to give a heads up that the implementation uses `HelpScreenException` in the 'internal' package of argparse4j. This matches the [help argument action][hp] in argparse4j, which is the same as the one in dropwizard, except argparse4j doesn't take an output to write the help to. So even though, this implementation uses a class in the internal package, I think it is the best implementation.

[hp]: https://github.com/tatsuhiro-t/argparse4j/blob/c986cadacd19bf5325ff0baed5efccaeadd51e9c/src/main/java/net/sourceforge/argparse4j/impl/action/HelpArgumentAction.java#L46-L65

dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
```diff
@@ -70,12 +71,12 @@ public boolean run(String... arguments) throws Exception {
...
                 final Namespace namespace = parser.parseArgs(arguments);
-                if (namespace.get("is-help") == null) {
-                    final Command command = commands.get(namespace.getString(COMMAND_NAME_ATTR));
-                    command.run(bootstrap, namespace);
-                }
+                final Command command = commands.get(namespace.getString(COMMAND_NAME_ATTR));
+                command.run(bootstrap, namespace);
             }
             return true;
+        } catch (HelpScreenException e) {
```
arteam: * Could we make a comment why `true` is returned? From the first glance it's not obvious that `HelpScreenException` is not an error. 

* Maybe it's better to name the variable `ignored` instead of `e`? I beleive that is a common name for ignored exceptions.

dropwizard-core/src/test/java/io/dropwizard/cli/CommandTest.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    private final Application<Configuration> app = new Application<Configuration>() {
+        @Override
+        public void run(Configuration configuration, Environment environment) throws Exception {
+        }
+    };
+
+    private final JarLocation location = mock(JarLocation.class);
```
arteam: Could we make `location` a local variable in the `setUp` method? I believe we don't use it in the test method(s).

dropwizard-core/src/test/java/io/dropwizard/cli/CommandTest.java
```diff
@@ -0,0 +1,82 @@
...
+    private final Bootstrap<Configuration> bootstrap = new Bootstrap<>(app);
```
arteam: Ditto, It could be a local variable in the `setUp` method.

dropwizard-core/src/test/java/io/dropwizard/cli/CommandTest.java
```diff
@@ -0,0 +1,82 @@
...
+    private final ByteArrayOutputStream stdOut = new ByteArrayOutputStream();
+    private final ByteArrayOutputStream stdErr = new ByteArrayOutputStream();
+    private final Command command = new TestCommand();
+    private Cli cli;
+
+    @Before
+    @SuppressWarnings("unchecked")
```
arteam: Do we need this annotation? I believe all operations are type-safe in the method.

dropwizard-core/src/test/java/io/dropwizard/cli/CommandTest.java
```diff
@@ -0,0 +1,82 @@
...
+    public void setUp() throws Exception {
+        when(location.toString()).thenReturn("dw-thing.jar");
+        when(location.getVersion()).thenReturn(Optional.of("1.0.0"));
+        bootstrap.addCommand(command);
+
+        this.cli = new Cli(location, bootstrap, stdOut, stdErr);
```
arteam: Probably no need for `this`, because we don't have a local variable with the same name.

arteam:
Looks good! :+1:
I've added a couple nitpicks, but overall this is ready to merge.

nickbabcock:
Looks like the nitpicks arose because I was too eager in copying and pasting the setup code from `CliTest` :blush: 

Anyways, pr should be good to go.

arteam:
Thanks for your contribution!

PR #1379, opened 2015-12-10
link to jersey 2.22.1 docs, not 1.18

philandstuff:
Since dropwizard has upgraded to jersey 2, the dropwizard-client docs should link to jersey 2 docs.


joschi:
Thanks!

PR #1380, opened 2015-12-10
Make SizeUnit more flexible

mattnelson:
Support case insensitive SizeUnit suffixes.
Support all single letter SizeUnit suffixes.

### DIFF
--- a/dropwizard-util/src/main/java/io/dropwizard/util/Size.java
+++ b/dropwizard-util/src/main/java/io/dropwizard/util/Size.java
@@ -3,39 +3,56 @@ package io.dropwizard.util;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
 
 import java.util.Locale;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


dropwizard-util/src/main/java/io/dropwizard/util/Size.java
```diff
@@ -3,39 +3,56 @@
...
+                .put("gigabyte", SizeUnit.GIGABYTES)
+                .put("gigabytes", SizeUnit.GIGABYTES)
+                .put("T", SizeUnit.TERABYTES)
+                .put("TB", SizeUnit.TERABYTES)
+                .put("TiB", SizeUnit.TERABYTES)
+                .put("terabyte", SizeUnit.TERABYTES)
+                .put("terabytes", SizeUnit.TERABYTES)
+                .build();
+        
+        final TreeMap<String, SizeUnit> treeSuffixes = Maps.newTreeMap(String.CASE_INSENSITIVE_ORDER);
```
joschi: Why not simply use an [`ImmutableSortedMap`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/ImmutableSortedMap.html) in the first place?

mattnelson: That is a better idea, updated 29c1bb1765ab913ad6abb6fa58933d4bc436c0ed

joschi:
LGTM. Thanks!

PR #1387, opened 2015-12-15
Add RegisterDefaultExceptionMappers to TestRule

nickbabcock:
This commit makes the default behavior of `ResourceTestRule` to register
all the default exception mappers, much like how the default yaml
configuration will register the default exception mappers.

If a user wants to override the default exception mappers, then they can
call `setRegisterDefaultExceptionMappers(false)` and subsequently register
their custom exception mappers.

Note that this is breaking behavior as those who registered their
exception mappers will now have to call setRegisterDefaultExceptionMappers
and those who wanted the default exception mappers should now remove their
registrations of the default exception mappers.

The  reason behind this pull request is that, in my opinion, it is unreasonable
for someone new to Dropwizard to know all the default exception mappers that
they will need to register in `ResourceTestRule` to get the same behavior as their
app.

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -215,6 +215,18 @@ easily.
 
     You can trust ``PeopleStore`` works because you've got working unit tests for it, right?
 
+Default Exception Mappers
+-------------------------
+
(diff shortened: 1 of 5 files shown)


arteam:
LGTM. I think this is a step in the right direction, because testing environment should closely resemble production.

Issue #1318, opened 2015-10-31
Support authentication of polymorphic Principal classes


evnm:
A web service that I'm implementing requires the notion of different classes of API keys for authenticating requests to different endpoints (e.g. read vs write keys). I thought that the natural way to implement this would be to form a simple class hierarchy of `java.security.Principal`s:

```
public abstract class ApiKey implements Principal { ... }
public class ReadApiKey extends ApiKey { ... }
public class WriteApiKey extends ApiKey { ... }
```

And then use the `@Auth` annotation to inject these keys in endpoints accordingly:

```
@POST public Response writeSomething(@Auth WriteApiKey apiKey, ...) { ... }
@GET public Response readSomething(@Auth ReadApiKey apiKey, ...) { ... }
```

However, Dropwizard is [currently](https://github.com/dropwizard/dropwizard/blob/23f353a896c9e5ad5e963fece3f82d227b7c0f5f/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthValueFactoryProvider.java#L55) unable to achieve this because `io.dw.auth.AuthValueFactoryProvider` enforces that only a single Principal implementation can be used per application.

I got around this deficiency by writing a custom `AbstractValueFactoryProvider` subclass that operates on a `Set<? extends Principal>` rather than a single Principal. This seems like a useful thing to have in dropwizard-auth proper, so I was wondering if others would entertain the idea of introducing it. To make sure this is feasible, I [moved the class into dropwizard-auth as a superclass of `io.dw.auth.AuthValueFactoryProvider`](https://github.com/evnm/dropwizard/commit/f1194a02271e1353945fedaa1c7a07f9a4265e53), since it's technically a generalization thereof.

Please let me know if this is a) ill-advised or b) worthy of a PR.

PixelBumper:
+1 I hit the wall because of this too.

jplock:
Fixed by #1392

PR #1392, opened 2015-12-19
Add support for authentication of polymorphic principals

evnm:
This change adds support for polymorphic `java.security.Principal` classes for use with Dropwizard's authentication mechanism. See #1318 for background.

My company has been running these custom `AuthValueFactoryProvider` and `AuthDynamicFeature` classes in our API server for months without issue. As stated in the ticket, we have a need to issue multiple classes of API keys to different API consumers and this change has suited the job well.

Note that the new `PolymorphicAuthDynamicFeature` only supports authentication, as opposed to authorization. [`AuthDynamicFeature`'s short-circuit on `RolesAllowed`/`DenyAll`/`PermitAll`](https://github.com/dropwizard/dropwizard/blob/2c3cca983f7b844a9f4bbaf93865120eae553e2a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java#L39-L41) doesn't jive with per-Principal-subclass filtering, so I opted to limit the feature to authentication.


dropwizard-auth/src/main/java/io/dropwizard/auth/AuthValueFactoryProvider.java
```diff
@@ -19,28 +16,31 @@
...
  */
 @Singleton
-public class AuthValueFactoryProvider<T extends Principal> extends AbstractValueFactoryProvider {
-
-    /**
-     * Class of the provided {@link Principal}
-     */
-    private final Class<T> principalClass;
+public class AuthValueFactoryProvider<T extends Principal>
+    extends PolymorphicAuthValueFactoryProvider
```
arteam: I would prefer `AuthValueFactoryProvider` not to be extended from `PolymorphicAuthValueFactoryProvider`.

This ties these classes rather highly without a need. I think we can find another ways to reuse the common code, if this is a concern.

dropwizard-auth/src/main/java/io/dropwizard/auth/PolymorphicAuthDynamicFeature.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.ws.rs.core.FeatureContext;
+
+/**
+ * A {@link DynamicFeature} that registers the provided auth filters
+ * to resource methods annotated with the {@link Auth} according to
+ * the type of the annotated method parameter.
+ * <p>Note that this feature only pertains to <i>authentication</i>.
+ * To enable authorization, use {@link AuthDynamicFeature}.</p>
+ */
+public class PolymorphicAuthDynamicFeature<T extends Principal> implements DynamicFeature {
```
arteam: Could you provide a minimal test case that uses this feature with several auth filters? I would prefer to see tests for every contributed functionality.

dropwizard-auth/src/main/java/io/dropwizard/auth/PolymorphicAuthValueFactoryProvider.java
```diff
@@ -0,0 +1,133 @@
...
+ * Value factory provider supporting injection of a hierarchy of
+ * {@link Principal} subclasses by the {@link Auth} annotation.
+ *
+ * @param <T> the type acting as the superclass from which injected
+ *            principals inherit
+ */
+@Singleton
+public class PolymorphicAuthValueFactoryProvider<T extends Principal>
+    extends AbstractValueFactoryProvider
+{
```
arteam: Nit: Please, remove a line break before the opening brace. In Dropwizard we follow [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style) on this matter.

dropwizard-auth/src/main/java/io/dropwizard/auth/PolymorphicAuthValueFactoryProvider.java
```diff
@@ -0,0 +1,133 @@
...
+
+        private final Set<Class<? extends T>> principalClassSet;
+
+        public Binder(Set<Class<? extends T>> principalClassSet) {
+            this.principalClassSet = principalClassSet;
+        }
+
+        @Override
+        protected void configure() {
+            bind(new PrincipalClassSetProvider<>(principalClassSet))
```
arteam: Nit: I think, it would be better to be consistent with `AuthValueFactoryProvider` and  have each bind statement (apart from the last) on one line.

dropwizard-auth/src/main/java/io/dropwizard/auth/PolymorphicAuthValueFactoryProvider.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    protected final Set<Class<? extends T>> principalClassSet;
+
+
+    /**
+     * {@link Principal} value factory provider injection constructor.
+     *
+     * @param mpep                   multivalued parameter extractor provider
+     * @param injector               injector instance
+     * @param principalClassProvider provider of the principal class
```
arteam: The parameter name in the docs doesn't match the actual parameter name `principalClassSetProvider`

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/NoAuthPrincipalEntityResource.java
```diff
@@ -13,7 +13,9 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
- * Contains resource methods which don't authenticate but use principal instance injection and thus might be affected by authentication logic.
```
arteam: Please, don't reformat source code without a need. It makes changes cleaner and focused.

arteam:
@evnm, thanks for your contribution! I reviewed it and this looks good in general. I've added some comments to the patch, see them inline.

evnm:
Thanks for the thorough review, @arteam. I've heeded your feedback and updated the diff.

In digging further into the dropwizard-auth tests, I decided to collapse `AuthBaseResourceConfig` into `AbstractAuthResourceConfig`. I don't think the usefulness of the two separate classes warrants the complexity imposed by a class hierarchy of `DropwizardResourceConfig` within a test package. So, this change now comprises two commits, the first of which does a bit of cleanup within the auth tests.

The meat of the change which adds support for polymorphic `Principal`s now reflects your comments. `PolymorphicAuthValueFactoryProvider` and `AuthValueFactoryProvider` each extend `AbstractValueFactoryProvider` directly. The meat of their respective `createValueFactory` methods is offloaded to a package-private `PrincipalContainerRequestValueFactory` class. The rest of the two provider classes is basically boilerplate.

And I added a `PolymorphicPrincipalEntityTest` which wires the new pieces up and ensures that authentication by multiple Principal subclasses in a resource actually works.

I'm not sure how (or if it's even worth it) to test `PrincipalContainerRequestValueFactory`. Taking a look at Jersey's source, `AbstractContainerRequestValueFactory` itself isn't tested, so I couldn't find a precedent for mocking `ContainerRequest` injection functionality.

Still not a fan of the "PolymorphicFoo" naming convention, but I still can't come up with anything better.

dropwizard-auth/src/main/java/io/dropwizard/auth/PrincipalContainerRequestValueFactory.java
```diff
@@ -0,0 +1,23 @@
+package io.dropwizard.auth;
+
+import org.glassfish.jersey.server.internal.inject.AbstractContainerRequestValueFactory;
+
+import java.security.Principal;
+
+/**
+ * A value factory which extracts the {@link Principal} from the
+ * current {@link ContainerRequest} instance.
```
arteam: Need to provide `import org.glassfish.jersey.server.ContainerRequest;` to use the short name.

dropwizard-auth/src/test/java/io/dropwizard/auth/basic/BasicAuthProviderTest.java
```diff
@@ -8,8 +8,13 @@
 import java.security.Principal;
 
 public class BasicAuthProviderTest extends AuthBaseTest<BasicAuthProviderTest.BasicAuthTestResourceConfig> {
-    public static class BasicAuthTestResourceConfig extends AuthBaseResourceConfig{
-        protected ContainerRequestFilter getAuthFilter() {
+    public static class BasicAuthTestResourceConfig extends AbstractAuthResourceConfig {
+        public BasicAuthTestResourceConfig() {
+            super();
```
arteam: I believe, `super()` is redundant, because the Java compiler automatically inserts a call to the default constructor of a superclass.

dropwizard-auth/src/test/java/io/dropwizard/auth/basic/BasicCustomAuthProviderTest.java
```diff
@@ -9,8 +9,13 @@
 
 public class BasicCustomAuthProviderTest extends AuthBaseTest<BasicCustomAuthProviderTest.BasicAuthTestResourceConfig> {
 
-    public static class BasicAuthTestResourceConfig extends AuthBaseResourceConfig{
-        protected ContainerRequestFilter getAuthFilter() {
+    public static class BasicAuthTestResourceConfig extends AbstractAuthResourceConfig {
+        public BasicAuthTestResourceConfig() {
+            super();
```
arteam: Ditto for `super()`

dropwizard-auth/src/test/java/io/dropwizard/auth/oauth/OAuthCustomProviderTest.java
```diff
@@ -5,25 +5,30 @@
 import io.dropwizard.auth.util.AuthUtil;
 import io.dropwizard.jersey.DropwizardResourceConfig;
 
-public class OAuthCustomProviderTest extends AuthBaseTest<OAuthCustomProviderTest.BasicAuthTestResourceConfig> {
-    public static class BasicAuthTestResourceConfig extends AuthBaseResourceConfig {
-        protected AuthFilter getAuthFilter() {
+public class OAuthCustomProviderTest extends AuthBaseTest<OAuthCustomProviderTest.OAuthTestResourceConfig> {
+    public static class OAuthTestResourceConfig extends AbstractAuthResourceConfig {
+        public OAuthTestResourceConfig() {
+            super();
```
arteam: Ditto for `super()`

dropwizard-auth/src/test/java/io/dropwizard/auth/oauth/OAuthProviderTest.java
```diff
@@ -5,9 +5,14 @@
 import io.dropwizard.auth.util.AuthUtil;
 import io.dropwizard.jersey.DropwizardResourceConfig;
 
-public class OAuthProviderTest extends AuthBaseTest<OAuthProviderTest.BasicAuthTestResourceConfig>{
-    public static class BasicAuthTestResourceConfig extends AuthBaseResourceConfig{
-        protected AuthFilter getAuthFilter() {
+public class OAuthProviderTest extends AuthBaseTest<OAuthProviderTest.OAuthTestResourceConfig>{
+    public static class OAuthTestResourceConfig extends AbstractAuthResourceConfig {
+        public OAuthTestResourceConfig() {
+            super();
```
arteam: Ditto for `super()`

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/NoAuthPrincipalEntityTest.java
```diff
@@ -41,6 +45,7 @@ protected DeploymentContext configureDeployment() {
     public static class NoAuthPrincipalInjectedResourceConfig extends AbstractAuthResourceConfig {
 
         public NoAuthPrincipalInjectedResourceConfig() {
+            super();
```
arteam: Ditto for `super()`

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/PolymorphicPrincipalEntityResource.java
```diff
@@ -0,0 +1,39 @@
+package io.dropwizard.auth.principal;
+
+import io.dropwizard.auth.Auth;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import java.lang.annotation.ElementType;
```
arteam: Redundant imports of the `java.lang.annotation` package

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/PolymorphicPrincipalEntityTest.java
```diff
@@ -0,0 +1,150 @@
...
+        @Override protected ContainerRequestFilter getAuthFilter() {
+            return new ContainerRequestFilter() {
+                @Override public void filter(ContainerRequestContext requestContext) throws IOException {
+                    throw new AssertionError("getAuthFilter result must not be invoked");
+                }
+            };
+        }
+
+        @Override protected AbstractBinder getAuthBinder() {
+            final Set<Class<? extends Principal>> principalClassSet = new HashSet<>();
```
arteam: I think using `ImmutableSet.of(JsonPrincipal.class, NullPrincipal.class)` here would be more terse.

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/PolymorphicPrincipalEntityTest.java
```diff
@@ -0,0 +1,150 @@
...
+            principalClassSet.add(JsonPrincipal.class);
+            principalClassSet.add(NullPrincipal.class);
+            return new PolymorphicAuthValueFactoryProvider.Binder<Principal>(principalClassSet);
```
arteam: We could just use the diamond type here instead explicit type argument.

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/PolymorphicPrincipalEntityTest.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        @Override protected DynamicFeature getAuthDynamicFeature(ContainerRequestFilter authFilter) {
+            final Authenticator<BasicCredentials, JsonPrincipal> jsonAuthenticator = credentials -> {
+                if (credentials.getUsername().equals(JSON_USERNAME)) {
+                    return Optional.<JsonPrincipal>of(new JsonPrincipal(credentials.getUsername()));
```
arteam: Since Java 8 the compiler can infer such explicit type arguments, so `<JsonPrincipal>` is redundant, I beleive.

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/PolymorphicPrincipalEntityTest.java
```diff
@@ -0,0 +1,150 @@
...
+                } else {
+                    return Optional.empty();
+                }
+            };
+
+            final Authenticator<BasicCredentials, NullPrincipal> nullAuthenticator = credentials -> {
+                if (credentials.getUsername().equals(NULL_USERNAME)) {
+                    return Optional.<NullPrincipal>of(new NullPrincipal());
```
arteam: Ditto for `<NullPrincipal>`

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/PolymorphicPrincipalEntityTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+            final BasicCredentialAuthFilter jsonAuthFilter = new BasicCredentialAuthFilter.Builder<JsonPrincipal>()
+                .setAuthenticator(jsonAuthenticator)
+                .buildAuthFilter();
+
+            final BasicCredentialAuthFilter nullAuthFilter = new BasicCredentialAuthFilter.Builder<NullPrincipal>()
+                .setAuthenticator(nullAuthenticator)
+                .buildAuthFilter();
+
+            return new PolymorphicAuthDynamicFeature(ImmutableMap.of(
```
arteam: To avoid a compiler warning about an unchecked call, we should provide type argument `<Principal>` for `PolymorphicAuthDynamicFeature`.

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/NoAuthPolymorphicPrincipalEntityTest.java
```diff
@@ -0,0 +1,128 @@
...
+        @Override protected ContainerRequestFilter getAuthFilter() {
+            return new ContainerRequestFilter() {
+                @Override public void filter(ContainerRequestContext requestContext) throws IOException {
+                    throw new AssertionError("Authentication must not be performed");
+                }
+            };
+        }
+
+        @Override protected AbstractBinder getAuthBinder() {
+            final Set<Class<? extends PrincipalImpl>> principalClassSet = new HashSet<>();
```
arteam: Ditto. I think using `ImmutableSet.of(...)` is a more terse and clearer.

arteam:
Thanks, Evan. Your reasons about collapsing `AuthBaseResourceConfig` and introdcing `PrincipalContainerRequestValueFactory` seem sensible for me. I've added comments about minor details of the implementation, but overall this is ready to merge.

Also, could you be so kind to update the [release notes](https://github.com/dropwizard/dropwizard/blob/master/docs/source/about/release-notes.rst) for 1.0 version? This would make easier for users
to track this improvement.

dropwizard-auth/src/test/java/io/dropwizard/auth/basic/BasicAuthProviderTest.java
evnm: You're right. I got a bit overzealous with these `super()`s while trying to get resource injection to compile with the new class hierarchy. Will remove these.

dropwizard-auth/src/test/java/io/dropwizard/auth/principal/PolymorphicPrincipalEntityTest.java
evnm: You're right. Another vestige of a battle with the compiler. Will remove.

evnm:
Change should now be good to go. I took the liberty of adding myself to the contributors page, too. Hopefully not too gauche.

Thanks again for the thoughtful review, @arteam!

arteam:
Thanks for contributing to Dropwizard!

> Hopefully not too gauche.

Not at all. But we usually do this automatically with a [Ruby script|https://github.com/dropwizard/dropwizard/blob/master/docs/list_contributors.rb]. We run it time to time to actualize the list.

BenRomberg:
Don't know if this is the right place to say "thank you", but this contribution just saved me a lot of work! Thanks a lot @evnm @arteam =)

PR #1394, opened 2015-12-24
Add ability to have config inheritance and property resolvers in configs

brentryan:
This PR provides a way for users to use inheritance in their configuration files.  I've seen several developers asking for this within dropwizard and we've been using this for a long time at Cvent to reduce as much boiler plate as we can from our configurations when each service has to support dev,alpha,staging,load, and production environment configs.

To use this for your service you'd initialize your service like this:

        MergedConfigurationFactoryFactory<T> cff
                = new MergedConfigurationFactoryFactory<>();
        MultiConfigurationSourceProvider sp = new MultiConfigurationSourceProvider();

        bootstrap.setConfigurationFactoryFactory(cff);
        bootstrap.setConfigurationSourceProvider(sp);
        bootstrap.addCommand(new DisplayMergedConfigCommand<>(this));

Your yaml config files would have something like this in them:

        parentConfigurationFile: configs/base.yaml

This allows all of your configuration files to reference a parent configuration where most "common" settings are placed.

With this PR you also have the ability to do replacements within the config using ${property} syntax.  This is useful so that if you have a configuration property like "environmentName" then you can reference this property anywhere else in your configuration without having to specify the same thing multiple times.


brentryan:
Also, this is optional and shouldn't affect prior implementations.  If you guys thing you'd like to include this or something like this in dropwizard them let me know what else is missing from this PR that could help facilitate getting this done and documented.

jplock:
Can the tasks be broken out into a separate PR? I think that is useful, but I'm personally not sure I'd want to support any type of configuration hierarchy as it makes debugging less explicit and error prone.

janolaveide:
I think this is would be very useful. I have in fact written my own Config
provider that merges 2 config files using the yaml merge directive, but I
think this PR solves the common/local requirement better.

Den torsdag 24. desember 2015 skrev Justin Plock <notifications@github.com>
følgende:

brentryan:
@jplock I can break that out into a separate task if you want.  However, it's completely relevant to this PR so not sure I understand why.  The command and task that I've added are useful for troubleshooting/verifying configurations when needed.

We've been using this system with 150+ developers for over a year now without issues around debugging and I think it's worked out well to reduce boilerplate plus still allow for some flexibility.

In our next version, we're going to get rid of configuration files entirely so they come from an externally managed centralized system so that we can control all of our configurations in 1 place across all of our applications.  And some changes will take effect at runtime and/or trigger a graceful reload of the application.  Hopefully we'll get that done in 2016.

nickbabcock:
Things that are of an immediate concern with the implementation:

- Possibility of the config task to "leak" passwords, especially if the password were referenced through environment variables. Even though the admin port should be protected and the task is optional, this does open up another attack surface that may be non-obvious.
- Looks to not handle cyclical references.
- I'd prefer an implementation of less copy + paste
- No tests/docs
- Not sure if having everything in the core module is appropriate.

More on the last point. I believe that this topic will invite a lot of bikeshedding and considering that this pull request is composed of strictly additions, I think the best first step is to make this a dropwizard module (http://modules.dropwizard.io/). There it can undergo further development and scrutiny.

Let me know what you think.

jplock:
@nickbabcock great idea on the module. +1 on that.

brentryan:
Thanks guys. I'll look into moving this to a module when I get some free time.

@nickbabcock I agree with the less copy/paste but we'd need to refactor some of the other dropwizard core classes to allow for this and now that this will be a module I'm not too sure it matters. If you'd like I can take a stab at another PR to dropwizard that fixes this then I can refactor this code before putting it into a module.

nickbabcock:
I'd much rather you get the module out and working then iterate on refactoring :smile:

brentryan:
Closing PR as I'll move over to a module.

PR #1398, opened 2015-12-30
Lazy initialization of "static" fields should be "synchronized"

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S2444 - Lazy initialization of "static" fields should be "synchronized"
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid:S2444
Please let me know if you have any questions.
Kirill Vlasov

### DIFF
--- a/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/jersey/LoggingDBIExceptionMapper.java
+++ b/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/jersey/LoggingDBIExceptionMapper.java
@@ -29,7 +29,7 @@ public class LoggingDBIExceptionMapper extends LoggingExceptionMapper<DBIExcepti
     }
 
     @VisibleForTesting
-    static void setLogger(Logger newLogger) {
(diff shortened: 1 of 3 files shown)


dropwizard-migrations/src/test/java/io/dropwizard/migrations/DbDumpCommandTest.java
```diff
@@ -43,7 +43,7 @@
     private TestMigrationConfiguration existedDbConf;
 
     @BeforeClass
-    public static void initXmlParser() throws Exception {
+    public static synchronized void initXmlParser() throws Exception {
```
arteam: I'm quite sure this is a false positive. `@BeforeClass` methods will be invoked only once per class, this is guaranteed by JUnit. I would just moved the variable assignments from this method directly to theirs declaration.

kirill-vlasov: I totally agree, file is updated.

PR #1400, opened 2015-12-30
Disable parallel tests by default

arteam:
Run them in parallel only in the dev mode. Looks like that Travis CI is really flaky with a parallel test execution. The build often fails without a reason and this really confuses new contributors. Better
to have a slow and stable build then fast and unreliable.


nickbabcock:
Yeah let's see if this fixes things.

PR #1403, opened 2016-01-04
Use h2 and h2c as the HTTP/2 connector names

arteam:
`h2` and `h2c` are official identificators for the HTTP/2 over TLS and HTTP/2 clear text protocols. It would be more consistent with the spec and less confusing for users, if we used these them instead `http2` and `http2c`.

See [HTTP/2 implementation list](https://github.com/http2/http2-spec/wiki/Implementations) for the reference.


nickbabcock:
Makes sense to me.

PR #1405, opened 2016-01-06
Fix: allow validation of Param nullable instances

nickbabcock:
Before if you had a resource that was like:

```java
@GET
@Path("nullable-int-param")
public String nullableIntParam(@QueryParam("num") @Max(3) IntParam secretSauce) {
    return secretSauce == null ? "I was null" : secretSauce.get().toString();
}
```

And you passed in `null`, the validator will fail due to an null
dereference in the unwrapper class.

This fixes the behavior by checking if the param is null before attempting
to unwrap

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ParamValidatorUnwrapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ParamValidatorUnwrapper.java
@@ -17,7 +17,7 @@ public class ParamValidatorUnwrapper extends ValidatedValueUnwrapper<AbstractPar
 
     @Override
     public Object handleValidatedValue(final AbstractParam<?> abstractParam) {
-        return abstractParam.get();
+        return abstractParam == null ? null : abstractParam.get();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


jplock:
LGTM!

joschi:
@nickbabcock As a matter of fact, the `*Param` parameters should never be `null` in the first place. The whole reason for their existence is that those parameters should never be missing and that Jersey should return an HTTP response with status 400 (see https://github.com/dropwizard/dropwizard/blob/v0.9.1/docs/source/manual/core.rst#parameters). Otherwise one could simply use the primitive types or their equivalents (e. g.  `int` or `Integer` instead of `IntParam`).

So if that doesn't work anymore, we should either fix it or get rid of the `*Param` classes altogether in favor of the more basic types and validating annotations like `@NotNull`.

jplock:
Personally I like the `*Param` classes as they provide a standard interface for converting a string into something else (I've used them to parse comma-separated parameters into a `Set<Integer>` for example).  My apologies for merging this so quickly if this is now indeed broken, but I'd like the functionality to remain.

nickbabcock:
From the [jersey docs](https://jersey.java.net/documentation/latest/jaxrs-resources.html#d0e2193) (emphasis mine):

> If the `@DefaultValue` is not used in conjunction with @QueryParam and the query parameter is not present in the request then value will be an empty collection forList, Set or SortedSet, _null for other object types_, and the Java-defined default for primitive types.

The `*Param` classes do seem to add nice behavior though, as if one uses primitives, and Jersey can't convert it, Jersey will return a 404, which, to me, is misleading. If there are alternatives for achieving the same behavior, it'd be good to evaluate the pros/cons of each.

PR #1406, opened 2016-01-06
Use lambdas and remove redundant type parameters

nickbabcock:
Since we're java 8 now, I figured I'd pick up some low hanging fruit by replacing explicit anonymous classes with lambdas and removing. I've opted to not convert code to use the Streams API because I know some find that more controversial.

The downside is that it may make backporting fixes to 0.{8,9}.x more difficult, but I'm optimistic it won't become a significant issue.


arteam:
LGTM

arteam:
I believe having a good VCS which allows versatile branching strategies, like `git`, is the reason not to fear such changes.

PR #1410, opened 2016-01-13
Add support Jetty's System.inheritedChannel option

matsumana:


dropwizard-example/example.yml
```diff
@@ -24,6 +24,16 @@ database:
 #    type: http
 #    port: 8080
 
+# run on a single port and server.connector.inheritChannel=true if your application launch by the "Server::Starter"
```
arteam: I believe it would be better to document this option in the [configuration reference](https://github.com/dropwizard/dropwizard/blob/master/docs/source/manual/configuration.rst) instead of the example configuration file. We prefer to keep it simple for users who start with Dropwizard.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
```diff
@@ -185,6 +185,8 @@ public static ConnectorFactory admin() {
 
     private String bindHost = null;
 
+    private boolean inheritChannel = false;
```
arteam: Could you add a description of this option in `HttpConnectorFactory` javadoc? It would be consistent with the rest of options, which are documented.

arteam:
Thanks for contributing! I've added some comments about documentation, but in general the patch looks good. Could you update the [release notes](https://github.com/dropwizard/dropwizard/blob/master/docs/source/about/release-notes.rst)? This would make easier for users to track this change. Please, also squash your commits into the single commit.

matsumana:
I fixed it.
Please review.

docs/source/manual/configuration.rst
```diff
@@ -277,6 +277,8 @@ Name                     Default             Description
 ======================== ==================  ======================================================================================
 port                     8080                The TCP/IP port on which to listen for incoming connections.
 bindHost                 (none)              The hostname to bind to.
+inheritChannel           false               If your application launch by the `Server::Starter`_, please set ``true``.
```
arteam: I would prefer the description to be a little more generic. What do you think about 
"Whether this connector uses a channel inherited from the JVM. Use it with `Server::Starter`, to launch an instance of Jetty on demand"? 

Also the property need to be set in the HTTP connector YAML configuration.

matsumana:
I agree.
Please review again.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
```diff
@@ -50,6 +50,11 @@
  *         <td>The hostname to bind to.</td>
  *     </tr>
  *     <tr>
+ *         <td>{@code inheritChannel}</td>
+ *         <td>false</td>
+ *         <td>If your application launch by the <a href="https://github.com/kazuho/p5-Server-Starter">Server::Starter</a>, please set 'true'</td>
```
arteam: Please, update the Javadoc too. It should be consistent with the reference docs.

arteam: Also, please don't forget to [rebase and squash](http://rebaseandsqua.sh) your commits.

matsumana:
I fixed it.
Please review again.

arteam:
Thanks, @matsumana, for patience and thorough contribution!

matsumana:
Thank you for merge!
When will 1.0.0 be released?

arteam:
I can't give you a strict ETA, because 1.0.0 is still being actively developed. But major features have already been contributed, so, I think, we could make a release candidate in the end of January. I will discuss this with the rest of the core team.

For the time being, you can use Dropwizard snapshots from the Sonatype's OSS [repository](https://oss.sonatype.org/content/repositories/snapshots). We publish them automatically after every succesfull build in the master branch. 

Thank you for using Dropwizard.

matsumana:
I see. Thank you!

PR #1411, opened 2016-01-13
Make DropwizardAppRule start and stop app only once, even when called recursively

qualidafial:
We've found on our projects that builds go significantly faster if we use DropwizardAppRule at the suite level.

This however makes it difficult to run individual tests in isolation.

This pull request makes DropwizardAppRule usable in a recursive way, but only starts and stops the application once for the outer call. With this change we are able to run the entire test suite, or a single test in a single class, and the application will be started/stopped once and only once.

There are usage examples in the javadoc changes.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/DropwizardAppRule.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/DropwizardAppRule.java
@@ -11,6 +11,7 @@ import io.dropwizard.testing.DropwizardTestSupport;
 import org.junit.rules.ExternalResource;
 
 import javax.annotation.Nullable;
+import java.util.concurrent.atomic.AtomicInteger;
 
 
 /**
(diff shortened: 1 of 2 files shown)


dropwizard-testing/src/main/java/io/dropwizard/testing/junit/DropwizardAppRule.java
```diff
@@ -21,12 +22,51 @@
  * override the {@link #newApplication()} method to provide your application instance(s).
  * </p>
  *
+ * <p>
+ * Using DropwizardAppRule at the suite level can speed up test runs, as the application is only started and stopped
+ * once for the entire suite:
+ * </p>
+ *
+ * <pre>
+ * @RunWith(Suite.class)
```
arteam: I believe, the `@` symbol need to be reprseneted like HTML number `&#064;` or be wrapped by the [@literal](http://docs.oracle.com/javase/1.5.0/docs/tooldocs/solaris/javadoc.html#{@literal}) tag. Otherwise, the `javadoc` tool interpretes `@RunWith` like a Java annotation.

qualidafial: Yep, I missed that. Will fix.

qualidafial: Fix pushed.

arteam:
Thanks for contributing to Dropwizard!

PR #1412, opened 2016-01-14
Allow RequestLogFactory implementation discovery

mwhipple:
Extend Discoverable for standard ServiceLoader based extensibility.


arteam:
Thanks! It slipped from #1290.

PR #1415, opened 2016-01-14
Add a Logback Access RequestLog implementation

ghenkes:
This change builds on top of the #1290 by applying the proposed changes from #580 with some tweaks. It still needs more javadoc and unit tests, but I thought I would get it out here to see if there is interest in this change.

A couple questions:
1. Should a warning message be logged if the LogFormat is configured but the SLF4JRequestLog is being used that states that the supplied format is being ignored?
2. Should a warning message be logged if both the LogFormat and the TimeZone are configured in the appender config that states that the TimeZone will be ignored and can be configured as part of the LogFomat?


ghenkes:
I've added more documentation and tests.

mwhipple:
I've also started a (presently very thrown together) external implementation of this functionality to be loaded as a plugin. If this is going to be merged into the main project then I'll kill mine now, otherwise I'd potentially be interested in finding a way to merge the code.

arteam:
@ghenkes, thanks for this terrific pull request! I'm interested in a logback-access integration, but I would prefer to see it as an external module, if it's possible. I'm not a fan of the idea of providing several ways to configure access logging in the core - for me it's very confusing. Speaking of the implementation, I would prefer to see it more focused without changing other Dropwizard logging classes. I don't know if it's possible, because I don't have much experience with `logback-access`, but I expected its integration to be more simple. 

I am going to experiment with this PR and also take a look at @mwhipple's work, and will get back to you, when have more information about the matter.

mwhipple:
As a disclaimer (that I'll be adding to the README) my project is code that I quickly ripped out of an app this week and would need a bit of attention before being releasable on its own.

mattnelson:
Customized request logging has been a frequent feature request. I would like to propose that the current request log implementation (Slf4jRequestLog) has reached end of life. As long as the new logback access implementation defaults to the same format pattern as the current implementation it would be a passive change and allow dropwizard request logging to be configured without the use of external bundles.

arteam:
> As long as the new logback access implementation defaults to the same format pattern as the current implementation it would be a passive change and allow dropwizard request logging to be configured without the use of external bundles.

Agree. That would be the best option, if we could make the switch transparent.

mwhipple:
To match the existing pattern exactly may require some filtering or similar since it builds missing values using the literal ` - `, so when there are two consecutive missing  the dashes are separated by two spaces rather than one. That was the only thing that changed in our logs after adopting the equivalent logback access pattern.

docs/source/manual/configuration.rst
```diff
@@ -105,13 +105,15 @@ Request Log
 .. code-block:: yaml
 
     server:
-      requestLog: 
-        timeZone: UTC
+      requestLog:
+        - type: SLF4J
```
jplock: Will users have to add this `type` reference to all of their existing configuration files if this gets merged in?  Is there anyway to default the value to `SLF4J` to ease upgrading?  If we can't support that, we need to ensure we update all of the various configuration files spread throughout the `dropwizard-example` and documentation (like the getting started pages).

jplock:
Per @arteam's comment about maintaining separate logging implementations, does it make sense to deprecate the SLF4J logging implementation completely in favor of only supporting `logback-access` (I don't know what this entails as I've never used `logback-access`).  Would it be possible to summarize what users will be able to do with this new implementation that was previously not supported?

One of the key things that I enjoy about Dropwizard is that it's opinionated in the way of following best practices.  I'm not saying we shouldn't introduce integration points so users can extend default functionality how they see fit, but for central core features (like logging), I'd prefer to see Dropwizard ship with a recommended approach.

In any case, I know logging has been a hot topic on the forums and issues list, so clearly there is some value in rethinking our default implementation.

mwhipple:
To speak for myself only: the functionality that motivated my team's switch to logback-access was to allow debug logging of full requests and responses and later to log additional information from the request headers (the latter may be available with DW also).

ghenkes:
Thanks all for the discussion!

@arteam 
> Speaking of the implementation, I would prefer to see it more focused without changing other Dropwizard logging classes. I don't know if it's possible, because I don't have much experience with logback-access, but I expected its integration to be more simple.

For the most part the changes to the other classes are to allow for reuse of the existing appenders (console, file, syslog) supplied by dropwizard.

@All
Seems like we are all in agreement that we should investigate replacing the current request logging with Logback Access. As far as what my team was looking for it was the ability to customize the request log format, specifically being able to add request headers. The current SLF4J version is backed by AbstractNCSARequestLog in Jetty. It's [log method](https://github.com/eclipse/jetty.project/blob/jetty-9.3.6.v20151106/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractNCSARequestLog.java#L100) is what decides the format of the log. By using Logback Access' [AccessPatternLayout](http://logback.qos.ch/manual/layouts.html#AccessPatternLayout) it gives users greater flexibility to control the log format (see the link for options).

As far as being able to fully replicate the SLF4J log format by default I'll start with examples of each.
##### SLF4J:
```
10.0.2.2 - - [18/Jan/2016:14:17:09 +0000] "GET /healthcheck HTTP/1.1" 200 148 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36" 185
```
##### Logback Access 
```
10.0.2.2 - - [18/Jan/2016:14:19:01 +0000] "GET /healthcheck HTTP/1.1" 200 148 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36" 182
```
when using the format:
```
%h %l %u [%t{dd/MMM/yyyy:HH:mm:ss Z,UTC}] "%r" %s %b "%i{Referer}" "%i{User-Agent}" %D
```
Here's a table showing the fields and values for both examples:

Type | Client Host | Remote Log Name | User | Date | Requested URL | Status Code | Response's Content Length | Referer | User-Agent | Elapsed Time
----- | -------------- | ---------------------- | ------ | ------ | -------------------- | --------------- | ---------------------------------- | --------- | ------------- | --------
SLF4J | 10.0.2.2 | - | - | [18/Jan/2016:14:17:09 +0000] | "GET /healthcheck HTTP/1.1" | 200 | 148 | "-" | "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36" | 185
Logbook Access | 10.0.2.2 | - | - | [18/Jan/2016:14:19:01 +0000] | "GET /healthcheck HTTP/1.1" | 200 | 148 | "-" | "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36" | 182

A couple notes:
1. Remote Log Name will always be just a ```-``` in both formats.
2. Client Host in SLF4J flexes to the X-Forwarded-For header if it is set (and preferProxiedForAddress is true, which dropwizard is hardcoding). I don't see a way of doing this through configuration in Logback Access, but it maybe to possible to replicate this behavior by defining a [custom converter](http://logback.qos.ch/manual/layouts.html#customConversionSpecifier).

##### Next steps:
I'll go ahead and remove the SLF4J implementation and replace that with Logback Access. I'll set it up to use the log format I showed above by default. I'll start investigating the custom converter for the Client Host, unless you guys don't think it's worth maintaining that functionality.

jplock:
@ghenkes can you handle the removal of slf4j in a separate PR?

ghenkes:
I've refactored the code a little so the Logback Access logging will now default to the format I mentioned above. This refactoring also allowed me to switch a few things back to the way they were originally, so this change has less impact on the existing code.

I've left the SLF4J pieces in based on @jplock's comment. (@jplock I assume your question meant leave it in this PR and once this change is merged do a second PR to remove the SLF4J implementation.)

Next I'll start investigating the custom converter for the Client Host.

jplock:
@ghenkes yes that's correct.  We'll leave SLF4J in place in this PR and remove it in a later PR.

ghenkes:
Here's the behavior of the code currently in this pull request with respect to the client host field in the access log, based on the x-forwarded-for header and the [useForwardedHeaders config](http://www.dropwizard.io/0.9.2/docs/manual/configuration.html#http):

Type | X-Forwarded-For Header Present | useForwardedHeaders Property | Client Host Field 
----- | -------------- | ---------------------- | ----------
Both | no | n/a |  IP address
Both | yes | true (default) | header value 
SLF4J | yes | false | **header value** 
Logbook Access | yes | false | **IP address**

I **bolded** the difference between the implementations. This is what I was going to attempt to align using the custom converter, however the more I think about it the more I feel that the current SLF4J version behavior is incorrect. The dropwizard user specifically chose not to use forwarded headers, yet they are still being applied to the access logs. Thoughts?

mattnelson:
Really looking forward to this feature being available.

Dug into the history of `useForwardedHeaders` and `setPreferProxiedForAddress` to see if I could find some context for these properties when they were introduced. They were both introduced with the Jetty 9 uplift for 0.7. Did not find anything definitive to confirm the existing functionality. I agree with @ghenkes that it seems like the existing Slf4jRequestLog implementation is wrong. An X-Forwarded-For header with a bogus value would mask the true of identity of a malicious user.

arteam:
> The dropwizard user specifically chose not to use forwarded headers, yet they are still being applied to the access logs

Agree, that looks like a bug. Users should have the option to disable handling `X-Forwarded-*` headers. Currently `Slf4jRequestLogFactory` forces it by setting `Slf4jRequestLog#preferProxiedForAddress` to `true`.

arteam:
As I can undersdant this PR, the idea is to extend the current model of appenders from `ILoggingEvent` events from `logback-classic` to `IAccessEvent` from `logback-access`.

This is accomplished by introducing 3 interfaces:
* `LayoutFactory`. It's implemented by `DropwizardLayoutFactory` (for standard log messages) and 
`LogbackAccessRequestLayoutFactory` (for access logs)
* `FilterFactory`. It's implemented by `ThresholdFilterFactory`, that allows to configure logging threshold (for standard log messages) and `NullFilterFactory` (for access logs)
* `AsyncAppenderFactory`. It's implemented by `AsyncLoggingEventAppenderFactory` for standard log messages and by `AsyncAccessEventAppenderFactory` for access logs.

For me this looks good on the whole, but I have a couple of proposals:
* I would move all classes related to `logback-access` to a separate module from `dropwizard-jetty`.
These classes more relate to `logback-access` rather to Jetty. I beleive, this will provide better 
encapsulation of dependencies and a more focused package layout
* I would move every additional factory to separate packages `io.dropwizard.logging.filter`, `io.dropwizard.logging.layout` and `io.dropwizard.logging.async`. I beleive it will provide more
clarity to the logging package layout.
* Also some comments inline

docs/source/manual/configuration.rst
```diff
@@ -69,7 +69,7 @@ GZip
 .. code-block:: yaml
 
     server:
-      gzip:
```
arteam: Please, don't reformat lines that are not relevant to the change. 
It makes the change less focused and more hard to review.

docs/source/manual/configuration.rst
arteam: Agree. We should mark the `logback-access` implementation as default, so users should not set the type explicetly.
But this should be not relevant after removing of the `slf4j` request log.

docs/source/manual/configuration.rst
```diff
@@ -224,7 +226,7 @@ Name                      Default                   Description
 applicationConnectors     An `HTTP connector`_      A set of :ref:`connectors <man-configuration-connectors>` which will
                           listening on port 8080.   handle application requests.
 adminConnectors           An `HTTP connector`_      An `HTTP connector`_ listening on port 8081.
-                          listening on port 8081.   A set of :ref:`connectors <man-configuration-connectors>` which will
```
arteam: Ditto for avoiding unrelated changes. Also, a couple of places below.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/LogbackAccessRequestLogFactory.java
```diff
@@ -0,0 +1,84 @@
...
+ *         <td>Description</td>
+ *     </tr>
+ *     <tr>
+ *         <td>{@code appenders}</td>
+ *         <td>a default {@link ConsoleAppenderFactory console} appender</td>
+ *         <td>The set of {@link AppenderFactory appenders} to which requests will be logged.</td>
+ *     </tr>
+ * </table>
+ */
+@JsonTypeName("logback-access")
```
arteam: I believe, we should mark implementation as the default in `RequestLogFactory`.

dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
```diff
@@ -61,18 +68,42 @@
...
+    protected TimeZone timeZone = TimeZone.getTimeZone("UTC");
+
     @Min(1)
     @Max(Integer.MAX_VALUE)
     private int queueSize = AsyncAppenderBase.DEFAULT_QUEUE_SIZE;
 
     private int discardingThreshold = -1;
 
+    @JsonProperty
+    public String getLogFormat() {
```
arteam: We already have a getter and setter for `logFormat`. I believe moving them higher doesn't provide an additional value.

dropwizard-logging/src/main/java/io/dropwizard/logging/AppenderFactory.java
```diff
@@ -23,16 +22,21 @@
  * @see SyslogAppenderFactory
  */
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-public interface AppenderFactory extends Discoverable {
+public interface AppenderFactory<E> extends Discoverable {
```
arteam: We could probably mark that `E` should extends `DeferredProcessingAware`

dropwizard-logging/src/main/java/io/dropwizard/logging/AsyncAppenderFactory.java
```diff
@@ -0,0 +1,16 @@
+package io.dropwizard.logging;
+
+import ch.qos.logback.core.AsyncAppenderBase;
+
+/**
+ * Factory used to create an {@link AsyncAppenderBase} of type E
+ * @param <E> The type of log event
+ */
+public interface AsyncAppenderFactory<E> {
```
arteam: Ditto for marking `E` as extending from `DeferredProcessingAware`

dropwizard-logging/src/main/java/io/dropwizard/logging/LayoutFactory.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.TimeZone;
+
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.core.pattern.PatternLayoutBase;
+
+/**
+ * An interface for building Logback {@link PatternLayoutBase} layouts
+ * @param <E> The type of log event
+ */
+public interface LayoutFactory<E> {
```
arteam: Dittor for extending `E` from `DeferredProcessingAware`

dropwizard-logging/src/main/java/io/dropwizard/logging/filter/FilterFactory.java
```diff
@@ -0,0 +1,17 @@
+package io.dropwizard.logging.filter;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.core.filter.Filter;
+
+/**
+ * An interface for building Logback {@link Filter Filters}
+ * @param <E> The type of log event
+ */
+public interface FilterFactory<E> {
```
arteam: Ditto for `E extends DeferredProcessingAware`

ghenkes:
@arteam I've implemented your requested changes; added new dropwizard-request-logging module, added packages for the factories and impls, cleaned up code based on your other comments

jplock:
I'm not really a fan of introducing a separate `dropwizard-request-logging` module for this.  Could the classes in there be combined with `dropwizard-logging` instead and switch the package to `io.dropwizard.logging.request`?  Does that make sense @arteam?

arteam:
> I'm not really a fan of introducing a separate dropwizard-request-logging module for this. Could the classes in there be combined with dropwizard-logging instead and switch the package to io.dropwizard.logging.request? Does that make sense @arteam?

I still believe we need a separate module for request logging, because the access logging is different from the classic logging. It works with different interfaces and abstractions and actually acts as a bridge between a web server request log abstractions and Logback appenders. By introducing a new module, we encapsulate all classes related to the access logging into a module, which could be excluded, if someone doesn't like this implementation.

arteam:
@ghenkes Thank you very much! I am going to review this in nearest days and give a feedback.

jplock:
@arteam that's a good point.   Thanks.

arteam:
Thanks, @ghenkes, for the tremendous contribution! Much appreciated.

ashic:
We're trying to log some http headers as part of access logs. It looks like this should support that. Is there a way to get this working with 0.9.2, or is there an alternative until 1.0 is out?

mwhipple:
@ashic What I've done in older versions is create a custom server factory and change the implementation there.  Extend `DefaultServerFactory`, overwrite the `addRequestLog` method to provide logback access (possibly grabbing the updated module), and then do the Discoverable/META-INF dance

omarkj:
Would there be interest in this being expanded so request attributes, which can be set using the `ContainerRequestContext.setProperty` would be usable in logback as well, or is there another preferred way of getting request-scoped data into logs?

mwhipple:
does the reqParameter format conversion not presently work properly?

omarkj:
@mwhipple Sorry -- I realized shortly after sending this that `%reqAttributre{KEY}` can be used to retrieve attributes (as opposed to request parameters).

PR #1417, opened 2016-01-16
Register RequestLogFactory as Discoverable

mwhipple:
Last missing piece for #1290 (as verified by external implementation, sorry for the scattered dot connection). Also included in #1415 by this is a much smaller change.


arteam:
Makes sense, otherwise users have to do this manually. I forgot that we use 2-step mechanism for discovering implementations. For some reason I thought that users just have to add a `io.dropwizard.jetty.RequestLogFactory` file with names of custom classes and `DiscoverableSubtypeResolver` will load them.

mwhipple:
That's what I'd expected to happen which led me to discover this. It makes sense that it's more explicit now that I know about it.

PR #1418, opened 2016-01-16
Refactor ResourceTestRule

arteam:
Currently the `ResourceTestResourceConfig` class is kind of a mess. It uses several inner classes, a static map of test rules, and doesn't have any comments. It's hard to see what's going on there.

This change:
* moves the inner class `ResourceTestResourceConfig` to a separate class named `DropwizardTestResourceConfig`. This name is more meaningful and suitable in my opinion;
* introduces the class `ResourceTestJerseyConfiguration` which contains a configuration passed by a user via `ResourceTest.Builder`. It allows to decouple `DropwizardTestResourceConfig` from `ResourceTestRule`.
* this class is also used for passing the current `ResourceConfig` configuration via a servlet context instead of the `ResourceTestRule` class in a Grizzly web test container. This removes additinal responsibility from the `ResourceTestRule` class.
* adds a minimal documentation for every class, describing their goals.

As a result, this change should make easier to make future improvements to the `ResourceTestRule` class, because of separation of concerns, better class names and added documentation.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestJerseyConfiguration.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestJerseyConfiguration.java
@@ -1,0 +1,40 @@
+package io.dropwizard.testing.junit;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.glassfish.jersey.test.spi.TestContainerFactory;
+
+import javax.validation.Validator;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A configuration of a Jersey testing environment.
+ * Encapsulates data required to configure a {@link ResourceTestRule}.
+ * Primarily accessed via {@link DropwizardTestResourceConfig}.
+ */
+class ResourceTestJerseyConfiguration {
+
+    final Set<Object> singletons;
+    final Set<Class<?>> providers;
+    final Map<String, Object> properties;
+    final ObjectMapper mapper;
+    final Validator validator;
+    final TestContainerFactory testContainerFactory;
+    final boolean registerDefaultExceptionMappers;
+
+    ResourceTestJerseyConfiguration(Set<Object> singletons, Set<Class<?>> providers, Map<String, Object> properties,
+                                    ObjectMapper mapper, Validator validator, TestContainerFactory testContainerFactory,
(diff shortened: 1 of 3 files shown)


dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestRule.java
```diff
@@ -45,6 +32,9 @@
         BootstrapLogging.bootstrap();
     }
 
+    /**
+     * A builder for {@link ResourceTestRule} that allows to configure a Jersey testing environment.
```
evnm: How about "A {@link ResourceTestRule} builder which enables configuration of a Jersey testing environment."?

dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestJerseyConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.glassfish.jersey.test.spi.TestContainerFactory;
+
+import javax.validation.Validator;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A configuration of a Jersey testing environment.
+ * It's used for configuring {@link DropwizardTestResourceConfig} from {@link ResourceTestRule}
```
evnm: I think this statement could be made a bit more clear for those who aren't familiar with the code it refers to. How about "Encapsulates data required to configure a {@link ResourceTestRule}. Primarily accessed via {@link DropwizardTestResourceConfig}."

evnm:
Looks good to me. A nice separation of concerns.

dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestRule.java
arteam: Yes, that sounds more natural.

dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestJerseyConfiguration.java
arteam: I agree, that's a more clear statement.

arteam:
@evnm, thanks for the review!

PR #1419, opened 2016-01-17
Add a test for parsing HTTPS connector configuration

arteam:
We currently don't have any tests for this. This change add a check, that the configuration, that we provide in the reference at http://dropwizard.io, is actually correctly parsed.


jplock:
Looks good to me!  Thanks!

PR #1420, opened 2016-01-17
Add ConfigurationParsingException tests

arteam:
I found out that we currently don't have tests for handling errors during parsing and binding YAML configuration files. This changes fixes this by adding several tests:

* A test that we print the `Did you mean?` suggestions on an unrecognized field in the configuration class. The suggestions are sorted based on an edit distance from the input. The limit is 5 suggestions.
* A test that we print a detailed error message, when there are type data type errors during binding a YAML tree to the configuration class.
* A test that we print a line number, a position of an error, and a parser error message, if a YAML file is malformed.

As a result, this provides the ability to catch regressions during Jackson and SnakeYAML updates.


jplock:
This looks good to me, thanks!

Issue #1416, opened 2016-01-15
Add a means of configuring the Jersey Client produced by ResourceTestRule


evnm:
Unless I'm mistaken, there isn't a way to inject functionality into [the `configureClient(ClientConfig config)` method of `ResourceTestRule`](https://github.com/dropwizard/dropwizard/blob/cc5f7a8eef8621cfc8d2ff012b525069d032238a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestRule.java#L215-L220). In cases where tests require custom Jersey client configuration (e.g. when [testing PATCH endpoints](http://stackoverflow.com/questions/29499106/using-patch-with-jersey-client-api-for-unit-testing)), the developer isn't left with a straightforward recourse.

I'd like to propose adding a method to `ResourceTestRule.Builder` akin to the following:

```
public Builder setClientConfigurator(j.u.function.Consumer<ClientConfig> configurator) {
    this.configurator = configurator;
    return this;
}
```

This `configurator` function would then be applied in the `configureClient` method linked-to above. Thus code like the following could be used to configure clients used in tests:

```
@ClassRule
public static final ResourceTestRule resources = ResourceTestRule.builder()
    .addConfigurator(clientConfig ->
        clientConfig.property(HttpUrlConnectorProvider.SET_METHOD_WORKAROUND, true));
    ...
```

This feature would be included in the 1.0.0 milestone, given the use of the Java 8 Consumer interface.

arteam:
Yes, I remember I wanted similar functionality other day and I didn't found a way to configure `JerseyClient`. I ended up configured `WebTarget` instead. I would welcome such contribution. 

One point, though: I've got a pull request brewing [here](https://github.com/arteam/dropwizard/commits/refactor_resource_test) about refactoring `ResourceTestRule`. Currently this class is quite a mess and I would like to clean up it beforehand. I am going to send a pull request today, so after that, I think, you may start to work on this feature.

arteam:
See #1418

jplock:
Closing as #1418 was merged

arteam:
No, this issue should not be closed. #1418 is just a foundation for a pull request to implement a solution for this issue.

jplock:
Oh whoops, sorry about that.

jplock:
Ok, I think this can be closed now that #1421 has been merged.  :)

PR #1421, opened 2016-01-18
dropwizard-testing: Add Jersey Client configurability to ResourceTestRule

evnm:
Addresses #1416.

Adds a `setClientConfigurator` method to the `ResourceTestRule.Builder` class which allows configuration of the Jersey Clients used in test rules. Adds a test to `PersonResourceTest` which ensures that custom providers installed by this new builder method are reflected in the produced clients.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestRule.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestRule.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Consumer;
 
 /**
  * A JUnit {@link TestRule} for testing Jersey resources.
@@ -42,6 +43,7 @@ public class ResourceTestRule implements TestRule {
         private final Map<String, Object> properties = new HashMap<>();
         private ObjectMapper mapper = Jackson.newObjectMapper();
         private Validator validator = Validators.newValidator();
+        private Consumer<ClientConfig> clientConfigurator = c -> {};
         private TestContainerFactory testContainerFactory = new InMemoryTestContainerFactory();
         private boolean registerDefaultExceptionMappers = true;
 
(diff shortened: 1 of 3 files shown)


dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestRule.java
```diff
@@ -37,11 +38,14 @@
      */
     public static class Builder {
 
+        private static Consumer<ClientConfig> NOOP_CONSUMER = c -> {};
```
arteam: I think the lambda expression could be just inlined with the consumer declaration, because we don't reference it in other places.

dropwizard-testing/src/test/java/io/dropwizard/testing/app/PersonResourceTest.java
```diff
@@ -21,6 +24,14 @@
  * Tests {@link ResourceTestRule}.
  */
 public class PersonResourceTest {
+    @Provider
```
arteam: I believe the `@Provider` annotation is redundant, because we explicitly register this exception mapper.

dropwizard-testing/src/test/java/io/dropwizard/testing/app/PersonResourceTest.java
```diff
@@ -21,6 +24,14 @@
...
+    public class DummyExceptionMapper implements ExceptionMapper<WebApplicationException> {
```
arteam: The class should be static, because it doesn't access any inner state of `PersonResourceTest`.
Also I beleive we could relax the class access level from public to private, because it's accessed only in `PersonResourceTest`.

arteam:
A couple of nitpicks, but generally looks good!

dropwizard-testing/src/main/java/io/dropwizard/testing/junit/ResourceTestRule.java
evnm: Sounds good. Too bad Java doesn't provide something like `Consumer.noop()` akin to `Function.identity()`.

evnm:
Updated the branch with the following:

- Rebase on current master
- Inline the noop `Consumer`
- Rm unnecessary `@Provider` annotation
- Make `DummyExceptionMapper` private and static

arteam:

PR #1422, opened 2016-01-18
Fix configuration factory test on windows

nickbabcock:
Looks like the exception message contains newlines, so this test failed on Windows. Not sure if this is the desired behavior, because as of now there is a mix of line endings -- I just got the test to pass.


arteam:
Oh, it looks like that the Jackson parser doesn't produce system-dependent newlines in error messages. Thanks for fixing this.

Issue #1375, opened 2015-12-07
Inconsistent `maxThreads` default for JerseyClient


natnan:
I might have misunderstood this completely but it felt wrong enough to create an issue here :).

Dropwizard's default `maxThreads` count is defaulted to 1024. While jersey client's is `128`. So, if you have a resource where you make some asynchronous request to another service and burst this resource with 1000 requests simultaneously, 872 of the requests will fail because of jersey client's thread pool reaching its limit. So, maybe, both `maxThreads` fields must have the same default value?

jplock:
Instead of changing the default (as users may be relying on that), would you mind submitting a pull request to clarify the values in the documentation?

natnan:
Sure, I wouldn't mind. But what kind of clarification should I add? Some warning like "warning: If you don't set this (`jerseyClient.maxThread`) high enough to match `server.maxThreads`, some requests may fail due to congestion on jersey client thread pool."?

jplock:
I think your use-case only applies if you are creating another request from inside an existing resource.  So that's the main point of clarification I think.

natnan:
Ah yes, of course. If it's used by a `Managed` service of sort, it's not relevant.

PR #1423, opened 2016-01-18
Add a note to jersey maxThread configuration documentation

natnan:
https://github.com/dropwizard/dropwizard/issues/1375
It's a little difficult for me to explain the situation in the documentation, but I did my best :). I can update it as you see fit.

I could also move it to above like this note:
image


jplock:
Thanks!

PR #1424, opened 2016-01-18
ValidationMethod violation message is itself

### LINKED ISSUE #1237: Send Validation Errors with Property Name
As of now there is no way to map Resource properties to their corresponding validation methods. You can use annotation ValidationMethod, which will return a set with error messages, but there is no way to find the cause (propertyname) of the error. Validator has a method validateProperty (where you can pass the property name which can come from an added field in Validationmethod annotation), but JacksonMessageBodyProvider uses the validate method. 
The propertyPath field of the Violation is of no use, as it is substring of the method name, For example, if your validation method name is isValidName, the property path will be "validName".. It should be just name.

nickbabcock:
Previously, when a ValidationMethod caused a constraint violation and the
violation was caused by a nested object using the `@Valid` annotation, the
message would contain the path of the nested object. This may be
misleading to the client who sees the message because the nested code path
may not have a related field in the JSON they sent (for example, the field
could be a misnomer).

The fix is return the message exactly as it is transcribed in the
`@ValidationMethod` annotation. Thus, it is now the user's responsibility
to create descriptive enough messages, such that the client can discern
the error through any amount of nesting.

This change does break APIs that relied on the message containing the code path
to the object that contained the ValidationMethod violation. ValidationMethods not used
in nested objects are not effected.

This fixes #1237, such that dropwizard doesn't prepend the path of the error in the message

### DIFF
--- a/dropwizard-validation/src/main/java/io/dropwizard/validation/ConstraintViolations.java
+++ b/dropwizard-validation/src/main/java/io/dropwizard/validation/ConstraintViolations.java
@@ -1,45 +1,22 @@
 package io.dropwizard.validation;
 
-import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Ordering;
 
 import javax.validation.ConstraintViolation;
-import javax.validation.ElementKind;
-import javax.validation.Path;
 import java.util.HashSet;
 import java.util.Set;
 
 public class ConstraintViolations {
-    private static final Joiner DOT_JOINER = Joiner.on('.');
-
     private ConstraintViolations() { /* singleton */ }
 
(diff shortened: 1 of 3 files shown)


dropwizard-validation/src/main/java/io/dropwizard/validation/ConstraintViolations.java
```diff
@@ -25,21 +20,7 @@ private ConstraintViolations() { /* singleton */ }
     }
 
     public static <T> String validationMethodFormatted(ConstraintViolation<T> v) {
```
arteam: What do you think about removing this method completely and using `v.getMessage()` explicitly instead? After this change, there isn't any formatting in the method, so its name now looks a little confusing.

nickbabcock:
Ok, latest commit takes your suggestions.

arteam:
Thanks!

PR #1425, opened 2016-01-19
Fixed infinite recursion bug in AuthFilter::authenticate

minisu:
The current implementation of `AuthFilter.authenticate` does this:

```java
requestContext.setSecurityContext(new SecurityContext() {
                @Override
                public boolean isSecure() {
                    return requestContext.getSecurityContext().isSecure();
                }
            });
```
Calling `isSecure()` on the filtered requestContext will recurse infinitely causing a StackOverflowError.

This PR fixes this bug and adds a test. The test is quite complex for such a small fix, but I can't think of a simpler way, due to the JAX-RS class design. Because of this, I'm fine with this being merged with or without the test.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
@@ -138,7 +138,9 @@ public abstract class AuthFilter<C, P extends Principal> implements ContainerReq
                 return false;
             }
 
-            boolean secure = requestContext.getSecurityContext().isSecure();
+            final SecurityContext securityContext = requestContext.getSecurityContext();
+            final boolean secure = securityContext != null && securityContext.isSecure();
+            
             requestContext.setSecurityContext(new SecurityContext() {
                 @Override
                 public Principal getUserPrincipal() {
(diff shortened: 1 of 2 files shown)


dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
```diff
@@ -138,6 +138,7 @@ protected boolean authenticate(ContainerRequestContext requestContext, C credent
                 return false;
             }
 
+            boolean secure = requestContext.getSecurityContext().isSecure();
```
jplock: can you make this `final`?

jplock: Also, do we need to check that a `SecurityContext` has been set before calling this to avoid a NPE?

minisu: > can you make this `final`?

Sure.

> Also, do we need to check that a SecurityContext has been set before calling this to avoid a NPE?

Good question. I searched a bit and found that such a null check is done in the Jersey examples of how to implement a [ContainerRequestFilter](https://jersey.java.net/documentation/latest/filters-and-interceptors.html#d0e9666). So yes, I think we should do it too.

minisu:
@jplock I fixed the issues that you commented on.

jplock:
Thanks, I guess I'm a little confused under what scenario `secure` will be true in this case?

minisu:
> Thanks, I guess I'm a little confused under what scenario secure will be true in this case?
`secure` will be true if the incoming `requestContext` is secure (which happens for example if the incoming request is made over HTTPS). We want our filter to keep the state of `requestContext.getSecurityContext().isSecure()`.

jplock:
Thanks!

PR #1429, opened 2016-01-25
Updated to Hibernate 5.0.7

ant3:
Hibernate 5 has been out for a while; there are a few patch releases and it seems fairly stable, so I updated the dropwizard-hibernate module to use Hibernate 5.

This is a big advantage to people who want to use Java 8 time classes such as Instant, because they won't have to include extra Jadira jars to make hibernate persist the new types.  I'm not aware of any downside since those not using Java 8 can still use the Jadira support for Joda time etc. as before.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/SessionFactoryHealthCheck.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/SessionFactoryHealthCheck.java
@@ -1,5 +1,7 @@
 package io.dropwizard.hibernate;
 
+import static org.hibernate.resource.transaction.spi.TransactionStatus.*;
+
 import com.codahale.metrics.health.HealthCheck;
 import com.google.common.util.concurrent.MoreExecutors;
 import io.dropwizard.db.TimeBoundHealthCheck;
@@ -8,7 +10,6 @@ import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
 
-import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 
 public class SessionFactoryHealthCheck extends HealthCheck {
@@ -49,7 +50,7 @@ public class SessionFactoryHealthCheck extends HealthCheck {
                     session.createSQLQuery(validationQuery).list();
                     txn.commit();
                 } catch (Exception e) {
-                    if (txn.isActive()) {
+                    if (txn.getStatus() == ACTIVE || txn.getStatus() == MARKED_ROLLBACK) {
                         txn.rollback();
                     }
                     throw e;
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/SessionFactoryHealthCheck.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/SessionFactoryHealthCheck.java
@@ -1,7 +1,5 @@
 package io.dropwizard.hibernate;
 
-import static org.hibernate.resource.transaction.spi.TransactionStatus.*;
-
 import com.codahale.metrics.health.HealthCheck;
 import com.google.common.util.concurrent.MoreExecutors;
 import io.dropwizard.db.TimeBoundHealthCheck;
@@ -50,7 +48,7 @@ public class SessionFactoryHealthCheck extends HealthCheck {
                     session.createSQLQuery(validationQuery).list();
                     txn.commit();
                 } catch (Exception e) {
-                    if (txn.getStatus() == ACTIVE || txn.getStatus() == MARKED_ROLLBACK) {
+                    if (txn.getStatus().canRollback()) {
                         txn.rollback();
                     }
                     throw e;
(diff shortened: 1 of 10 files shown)


dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/SessionFactoryHealthCheck.java
```diff
@@ -49,7 +50,7 @@ protected Result check() throws Exception {
                     session.createSQLQuery(validationQuery).list();
                     txn.commit();
                 } catch (Exception e) {
-                    if (txn.isActive()) {
+                    if (txn.getStatus() == ACTIVE || txn.getStatus() == MARKED_ROLLBACK) {
```
joschi: Why not use [`TransactionStatus#canRollback()`](https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/resource/transaction/spi/TransactionStatus.html#canRollback--) here?

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
```diff
@@ -104,7 +107,7 @@ private void rollbackTransaction() {
             return;
         }
         final Transaction txn = session.getTransaction();
-        if (txn != null && txn.isActive()) {
+        if (txn != null && (txn.getStatus() == ACTIVE || txn.getStatus() == MARKED_ROLLBACK)) {
```
joschi: Why not use [`TransactionStatus#canRollback()`](https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/resource/transaction/spi/TransactionStatus.html#canRollback--) here?

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
```diff
@@ -114,7 +117,7 @@ private void commitTransaction() {
...
```
joschi: Why not use [`TransactionStatus#canRollback()`](https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/resource/transaction/spi/TransactionStatus.html#canRollback--) here?

dropwizard-hibernate/src/test/java/io/dropwizard/hibernate/SessionFactoryFactoryTest.java
```diff
@@ -50,6 +51,11 @@ public void setUp() throws Exception {
         config.setUser("sa");
         config.setDriverClass("org.hsqldb.jdbcDriver");
         config.setValidationQuery("SELECT 1 FROM INFORMATION_SCHEMA.SYSTEM_USERS");
+        
+        HashMap<String, String> props = new HashMap<>();
```
joschi: Using the builder methods of `ImmutableMap` might save a keystroke or two here.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
ant3: Good point :-).  I hadn't noticed that method.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
ant3: Yep, could use it here too.

dropwizard-hibernate/src/test/java/io/dropwizard/hibernate/SessionFactoryFactoryTest.java
ant3: I tend to steer away from Guava stuff through force of habit because it isn't normally worth bringing in the extra dependency just to tidy up the code a bit (IMHO) - but of course Dropwizard already uses it so there isn't really any cost.  In short, yes.

ant3:
Happy to make changes, but will wait for feedback before committing.

ant3:
Well I made the changes anyway - but happy to adjust if there are further comments.

joschi:
LGTM. Thanks!

alexitooi:
hi @ant3,
unfortunately we have found a downside of upgrading hibernate from 4.x to 5.x in dropwizard!
we have been using an identifier generator which is removed in hibernate 5.x and so far we could not figure out a way on how to make our existing db work with the new identifier generators :(
we used the "hilo" strategy which hibernate implemented with "org.hibernate.id.TableHiLoGenerator" (creates a table with the next_hi value) and it seems that we can not migrate this...do you have any hints? is there probably a fork of DW 1.0 with hibernate 4.x ?

ant3:
So you were using https://docs.jboss.org/hibernate/orm/4.1/javadocs/org/hibernate/id/TableHiLoGenerator.html.  I haven't used it myself, but the javadoc for hibernate version 4.1 has it marked as deprecated.  The link above suggests that you should move to using SequenceStyleGenerator (even in Hibernate 4.1).  This class still exists in Hibernate 5 - https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/id/enhanced/SequenceStyleGenerator.html.

Is there any reason that you cannot use SequenceStyleGenerator?

alexitooi:
hi, yeah we realized that just now that it was deprecated...unfortunately we don't see a way to migrate to the newly introduced generators since also hibernate user guide states in one hidden sentence it is not compatible. do you might know of a fork of DW with hibernate 4.X in place?

ant3:
I'm not aware of a fork.  You can of course use the older version, or create your own fork, but I would be surprised if there is no way for you to migrate.  Which 'hidden sentence' are you talking about?

PR #1430, opened 2016-01-26
Added naming to migration bundles

souravmitra:
In similar lines to HibernateBundles, having the possibility of unique names. This PR attempts to do the same for MigrationBundles, so that multiple migration bundles can be provided for.

### DIFF
--- a/dropwizard-migrations/src/main/java/io/dropwizard/migrations/MigrationsBundle.java
+++ b/dropwizard-migrations/src/main/java/io/dropwizard/migrations/MigrationsBundle.java
@@ -7,7 +7,7 @@ import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 
 public abstract class MigrationsBundle<T extends Configuration> implements Bundle, DatabaseConfiguration<T> {
-    public static final String DEFAULT_NAME = "db";
+    private static final String DEFAULT_NAME = "db";
 
     @Override
     @SuppressWarnings("unchecked")
@@ -16,7 +16,7 @@ public abstract class MigrationsBundle<T extends Configuration> implements Bundl
         bootstrap.addCommand(new DbCommand<>(name(), this, klass));
     }
 
-    protected String name() {
+    public String name() {
         return DEFAULT_NAME;
     }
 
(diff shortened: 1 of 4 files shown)


dropwizard-migrations/src/main/java/io/dropwizard/migrations/MigrationsBundle.java
```diff
@@ -7,11 +7,17 @@
...
+
     @Override
     @SuppressWarnings("unchecked")
     public final void initialize(Bootstrap<?> bootstrap) {
         final Class<T> klass = (Class<T>) bootstrap.getApplication().getConfigurationClass();
-        bootstrap.addCommand(new DbCommand<>(this, klass));
+        bootstrap.addCommand(new DbCommand<>(name(), this, klass));
+    }
+
+    protected String name() {
```
jplock: can you make this `public`?

dropwizard-migrations/src/main/java/io/dropwizard/migrations/MigrationsBundle.java
```diff
@@ -7,11 +7,17 @@
 import io.dropwizard.setup.Environment;
 
 public abstract class MigrationsBundle<T extends Configuration> implements Bundle, DatabaseConfiguration<T> {
+    public static final String DEFAULT_NAME = "db";
```
jplock: can you make this `private`?

jplock:
I think this is a good improvement.  Thanks for the contribution!  Just left some minor comments.

dropwizard-migrations/src/main/java/io/dropwizard/migrations/MigrationsBundle.java
souravmitra: @jplock : Thanks for the comments. Do you see usage of name() outside subclasses of MigrationBundle?

jplock: No, but to stay consistent with `initialize()` and `run()` and how its defined in https://github.com/dropwizard/dropwizard/blob/master/dropwizard-core/src/main/java/io/dropwizard/Application.java

souravmitra: Made the change. Thanks

dropwizard-migrations/src/main/java/io/dropwizard/migrations/MigrationsBundle.java
souravmitra: Done

jplock:
Sorry, one more minor change, would you mind updating https://github.com/dropwizard/dropwizard/blob/master/docs/source/about/release-notes.rst with a note about this new functionality and reference to this PR?

souravmitra:
@jplock Added to the release notes. Thanks.

jplock:
Looks great, thanks for contributing!

PR #1434, opened 2016-01-29
Remove SLF4J request logging

ghenkes:
As requested in #1415 this is a follow up PR that removes the SLF4J request logging now that dropwizard has switched to using Logback Access for the request logs.


arteam:
LGTM!

PR #1435, opened 2016-01-29
Allow for supplying Filters to Logging Appenders

ghenkes:
This change adds a seam to allow consumers to supply Logback Filters to apply to logging Appenders. These filters will be applied, in order, after the threshold filter. This allows for further refinement of what will be logged. For example, I have an endpoint that is called on a schedule to check availability that I would like to exclude from the access logs by supplying a filter for this URI.

You could also add appenders with the following:
* A filter that will only log failures based on status code.
* A filter that will log requests with slow responses.
* A filter that will log large requests/responses.

### DIFF
--- a/dropwizard-logging/src/test/java/io/dropwizard/logging/DefaultLoggingFactoryTest.java
+++ b/dropwizard-logging/src/test/java/io/dropwizard/logging/DefaultLoggingFactoryTest.java
@@ -1,10 +1,12 @@
 package io.dropwizard.logging;
 
 import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.spi.ILoggingEvent;
 import com.codahale.metrics.MetricRegistry;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.TextNode;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Files;
 import com.google.common.io.Resources;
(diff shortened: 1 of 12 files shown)


arteam:
This seems to me like a reasonable improvement. It allows to extend logging behavior without changing the default logging configuration. It fits nicely with Dropwizard's philosophy of keeping things simple for the average user, but at the same time providing a means of control for those, who want to customize the framework.

dropwizard-logging/src/test/java/io/dropwizard/logging/DefaultLoggingFactoryTest.java
```diff
@@ -71,6 +71,7 @@ public void canParseNewLoggerFormat() throws Exception {
         assertThat(fileAppenderFactory.getCurrentLogFilename()).isEqualTo("${new_app}.log");
         assertThat(fileAppenderFactory.getArchivedLogFilenamePattern()).isEqualTo("${new_app}-%d.log.gz");
         assertThat(fileAppenderFactory.getArchivedFileCount()).isEqualTo(5);
+        assertThat(fileAppenderFactory.getFilterFactories()).containsExactly(new TestFilterFactory(), new SecondTestFilterFactory());
```
arteam: This currently produces a compiler warning about an unchecked call. What do you think about replacing this
check by a serie of checks?
```java
final ImmutableList<?> filterFactories = fileAppenderFactory.getFilterFactories();
assertThat(filterFactories).hasSize(2);
assertThat(filterFactories.get(0)).isExactlyInstanceOf(TestFilterFactory.class);
assertThat(filterFactories.get(1)).isExactlyInstanceOf(SecondTestFilterFactory.class);
```

As a benefit, we could get rid of `hashCode` and `equals` implementations for `TestFilterFactory` and `SecondTestFilterFactory`.

arteam: After another look, I beleive we could even use `ImmutableList<FilterFactory<ILoggingEvent>>` if we generify `appenderFactory` and `fileAppenderFactory`.

arteam:
@ghenkes, I reviewed the implementation, looks great. One note in a unit test, but generally this looks ready to merge!

ghenkes:
@arteam I've applied your suggestions in the unit test. Thanks!

arteam:
Thanks for your contribution!

Issue #1433, opened 2016-01-28
Hibernate validation not happening for @BeanParam with groups


sureshrengasamy:
Issue is [fixed](https://github.com/dropwizard/dropwizard/issues/1227) for validating @BeanParam object but it is not working when validator groups are present.

```
public class BaseClass {

    @NotNull(groups=SignupValidator.class)
    @QueryParam("startDate")
    protected LocalDate startDate;

    public LocalDate getStartDate() {
        return startDate;
    }
}

@Path("path")
public class QuantityResource {

    @GET
    public ResponseEntity working(@Validated(SignupValidator.class) @BeanParam BaseClass req) {
        ...    
    }

}
```

nickbabcock:
I'm marking this as a bug, as currently `Validated` only works on request entities, and one shouldn't have to declare a dummy request object just so they can get the `Validated` annotation to work. I'm proposing a fix that will let the `Validated` annotation be on any parameter but if more than one `Validated` annotation is supplied with a different set of groups then it's a runtime error because the underlying hibernate validator is unable to perform validation against parameters with different groups.

sureshrengasamy:
Thank you

arteam:
Closed via #1436

PR #1436, opened 2016-01-29
Allow Validated to annotate non-entity objects

nickbabcock:
This is a fix for #1433.

Previously, the `Validated` annotation would only be allowed on entity parameters of a resource, even though a `BeanParam` could be validated. This created a problem where one would want to specify a validation group on an endpoint that only accepted a `BeanParam`, as they would be forced to annotate a dummy request entity with `Validated`.

This pull request also handles the situation when the user supplies two or more entities with `Validated`. These annotations must be equivalent (and thus redundant), else an exception will be thrown as the underlying validator cannot validate parameters under different validation groups.

I would like critiques of the implementation because as of right now, a `WebApplicationException` is thrown, as a `RuntimeException` will result in Jersey interpreting it as a `ProcessingException`, and `WebApplicationException` works around it. Alternatively, we could define a new exception type but that would require adding a new exception mapper as well, which I feel like is more code bloat than I care for.

As a side note, it should definitely be documented (looking at me) that a single parameter annotated with `Validated` will be effect how the rest of the parameters are validated with the same group. Some would find the following surprising:

```java
@PUT
public FooBar putFoo(@Validated(Group.class) @Valid @BeanParam MyParams params,
                     @Valid @NotNull FooBar entity) {
    // ...
}
```

In this case, both `params` and `entity` are validated under the `Group` class. The same would apply if `entity` was annotated and `params` wasn't.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/DropwizardConfiguredValidator.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/DropwizardConfiguredValidator.java
@@ -1,5 +1,6 @@
 package io.dropwizard.jersey.validation;
 
+import com.google.common.collect.ImmutableList;
 import io.dropwizard.validation.ConstraintViolations;
 import io.dropwizard.validation.Validated;
 import org.glassfish.jersey.server.internal.inject.ConfiguredValidator;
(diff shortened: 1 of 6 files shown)


arteam:
Thanks for a detailed description of the issue! I find the implementation is quite terse and, what more important, it fixes the problem. Let's merge the implementation and refactor it, if needed.

PR #1437, opened 2016-01-30
Support for @RolesAllowed on the class level

tocr:


psamsotha:
I'm thinking maybe you should add `@PermitAll` also. Currently to get authentication with no authorization (using the `AuthDynamicFeature`), you need `@PermitAll` on each method. It would also make sense to add the class-level support. This is also consistent with how the `RolesAllowedDynamicFeature` is implemented.

tocr:
@psamsotha Good point, thank you. 
Added `@PermitAll` to the list of checked class-level annotations.

dropwizard-example/src/test/java/com/example/helloworld/resources/ProtectedClassResourceTest.java
```diff
@@ -0,0 +1,83 @@
...
+        assertThat(secret).startsWith("Hey there, guest. You know the secret!");
+    }
+
+    @Test
+    public void testProtectedBasicUserEndpointPrincipalIsNotAuthorized403() {
+        try {
+            RULE.getJerseyTest().target("/protected").request()
+            .header(HttpHeaders.AUTHORIZATION, "Basic Z3Vlc3Q6c2VjcmV0")
+            .get(String.class);
+            failBecauseExceptionWasNotThrown(WebApplicationException.class);
```
jplock: would a `ForbiddenException` work here or is just a `WebApplicationException` thrown?

dropwizard-example/src/test/java/com/example/helloworld/resources/ProtectedResourceTest.java
```diff
@@ -58,6 +58,21 @@ public void testProtectedEndpointNoCredentials401() {
     }
 
     @Test
+    public void testProtectedEndpointBadCredentials401() {
+        try {
+            RULE.getJerseyTest().target("/protected").request()
+                .header(HttpHeaders.AUTHORIZATION, "Basic c25lYWt5LWJhc3RhcmQ6YXNkZg==")
+                .get(String.class);
+            failBecauseExceptionWasNotThrown(WebApplicationException.class);
```
jplock: Same question here but using `NotAuthorizedException`

tocr:
You are right. I replaced the unnecessarily general `WebApplicationException` in the tests

jplock:
Thanks for the contribution!

PR #1444, opened 2016-02-04
ExecutorServiceBuilder: Set default maximumPoolSize to 1; corePoolSiz…

alexheretic:
This is somewhat related to issue #575 
For unbound queues the current warning triggers when you have a maximumPoolSize of 1, which is actually fine. Howevew, it fails to warn when you have a maximumPoolSize of Integer.MAX_VALUE but this is also misleading. Even worse the default functionality reads like its a Executors.newCachedThreadPool() but actually is limited to a single thread, ie can't run concurrent tasks.

I've added a simple test to prove if a given Executor can run concurrent tasks.

This change improves the warning to trigger in the correct situations.
The change sets the default maximumPoolSize to 1, which is less misleading and won't cause a functionality change. Unless a user had done:
```java
environment.lifecycle().executorService("some-executor")
                .workQueue(new SynchronousQueue<>())
                .build()
```
Before this commit this would result in a unbound cached thread pool, now you'd need to do
```java
environment.lifecycle().executorService("some-executor")
                .maxThreads(Integer.MAX_VALUE)
                .workQueue(new SynchronousQueue<>())
                .build()
```
Though if you're manually setting your BlockingQueue implementation I'd suggest the user probably won't rely on default settings anyway.

--commit-message
> 
ExecutorServiceBuilder: Set default maximumPoolSize to 1; corePoolSize < maximumPoolSize with an unbound work queue has misleading functionality including when the maximumPoolSize is Integer.MAX_VALUE (current default). Note: This is a minor change in functionality as before the default settings would allow only a single threaded pool.
ExecutorServiceBuilder: Creating a Executors.newSingleThreadExecutor() & Executors.newCachedThreadPool() like executor should not cause a warning message
ExecutorServiceBuilder: Simply setting maxThreads to > 1 with an unbound work queue now causes a warning in all cas

### DIFF
--- a/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
+++ b/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
@@ -1,14 +1,16 @@
 package io.dropwizard.lifecycle.setup;
 
+import com.google.common.base.Throwables;
+import io.dropwizard.util.Duration;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.exceptions.verification.WantedButNotInvoked;
 import org.slf4j.Logger;
 
-import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.*;
 
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
 
 public class ExecutorServiceBuilderTest {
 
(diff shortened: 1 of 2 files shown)


evnm:
Perhaps I'm missing some background on this, but what motivated the change in the first place? Has the current code been observed to cause problems or is this more an issue of unclear/inaccurate documentation?

It seems to me that this change would contradict the very purpose of thread-pooling in the default case. Is it wise to max out pools at a single thread by default? Assuming that the average Dropwizard user probably doesn't customize their `ExecutorService`(s), I could imagine this change resulting in frustrated surprise when folks' underlying executors effectively become `Executors.newSingleThreadExecutor()`.

alexheretic:
That's just it, before the change while the default queue is unbounded this means that the maximumPoolSize is irrelevant (other than being lower than corePoolSize will cause exception)

**Before the change the default executor _was_ effectively a single thread pool.**

I put the `assertCanExecuteAtLeast2ConcurrentTasks` method in the tests to make this clear. The following usage of ExecutorServiceBuilder is misleading
```java
@Test
public void thisIsNotACachedThreadPool() {
    ExecutorService exe = executorServiceBuilder
        .maxThreads(Integer.MAX_VALUE)
        .build();
    assertCanExecuteAtLeast2ConcurrentTasks(exe); // fails
}
```
The above test will fail with or without my src changes. (Without my changes this is the default)
Indeed the following will also fail:
```java
@Test
public void thisIsNotDualThreadPool() {
    ExecutorService exe = executorServiceBuilder
        .maxThreads(2)
        .build();
    assertCanExecuteAtLeast2ConcurrentTasks(exe); // fails
}
```
Unbounded queue => maxThreads doesn't have an effect
```java
@Test
public void thisIsADualThreadPool() {
    ExecutorService exe = executorServiceBuilder
        .minThreads(2)
        .maxThreads(2)
        .build();
    assertCanExecuteAtLeast2ConcurrentTasks(exe); // passes
}
@Test
public void thisIsStillADualThreadPool() {
    ExecutorService exe = executorServiceBuilder
        .minThreads(2)
        .maxThreads(Integer.MAX_VALUE) // no different in effect from maxThreads 2
        .build();
    assertCanExecuteAtLeast2ConcurrentTasks(exe); // passes
}
```

To get what it looks like it was doing (but wasn't) you'd need to:
```java
ExecutorService exe = executorServiceBuilder
            .maxThreads(Integer.MAX_VALUE)
            .workQueue(new SynchronousQueue<>())
            .build();
```

I opted to change it the least that I could so that it would warn in the dodgy cases, and not-warn in the not-dodgy cases.

I hope that makes sense, try running the tests to see for yourself.

dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
```diff
@@ -20,26 +22,116 @@
...
+
+        try { verify(log).warn(anyString()); }
+        catch (WantedButNotInvoked error) {
+            // no warning has been given so we should be able to execute at least 2 things at once
+            assertCanExecuteAtLeast2ConcurrentTasks(exe);
+        }
+    }
+
+    /**
+     * Tries to run 2 tasks that on the executor that rely on each others side-effect to complete. If they fail to
```
evnm: "Tries to run two tasks on the executor that..."

evnm:
I see now. Thank you for clarifying.

From the [`ThreadPoolExecutor` javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html):

> Using an unbounded queue (for example a LinkedBlockingQueue without a predefined capacity) will cause new tasks to wait in the queue when all corePoolSize threads are busy. Thus, no more than corePoolSize threads will ever be created.

Code LGTM. Could you break up the commit message, though, please? Would prefer a to-the-point subject line and then the rest of your description in a body paragraph. This makes `git log` output a lot easier to scan on the command-line.

alexheretic:
How about this one?

evnm:
👍

However, we'll have to wait until an actual committer can chime in.

arteam:
Looks good to me! I think the change is justified. Now `ExecutorServiceBuilder` doesn't give a warning for the fixed single thread pool configuration and gives a warning when a user specifies the maximum amount of threads without setting the minimum amount and a bounded queue.

Thanks, @alexheretic, for the contribution with thorough test cases. And thanks, @evnm, for the review!

Issue #1393, opened 2015-12-21
Support DropwizardClientRule with custom ObjectMapper config


rowanhill:
Hi,

As far as I can tell, it's not possible to apply any custom configuration to the ObjectMapper used by DropwizardClientRule's FakeApplication (since the application and its environment are kept private). Is this by design?

I'd like to register a Module and set some SerializationFeatures on the ObjectMapper used - without that, the JSON the client under test sends cannot be deserialized.

It looks like at the moment I'm forced to run a full integration test with DropwizardAppRule (which is a shame, as it's slower and much more difficult to manipulate). Is there any alternative, or if not, is there any argument against exposing the application and/or environment from DropwizardClientRule?

Thanks,
Rowan

jplock:
I submitted https://github.com/dropwizard/dropwizard/pull/1446 to address this

PR #1446, opened 2016-02-05
Expose the ObjectMapper and Environment from the FakeApplication

jplock:
Fixes https://github.com/dropwizard/dropwizard/issues/1393


evnm:
To minimize the change, we could simply expose `getEnvironment()`. Consumers could then access the `ObjectMapper` via `Environment.getObjectMapper()`.

Not sure if we care, though. LGTM!

jplock:
@evnm I thought about that, but then I saw that `DropwizardTestSupport` exposes them both this way anyway so I figured it'd be ok.

evnm:
Yep, seems consistent to me.

arteam:
I think that's fine. Users will have more control over the testing environment.

arteam:
Thanks, @jplock and @evnm!

PR #1448, opened 2016-02-06
Improve Support for Optional{Double,Int,Long} in Jersey

### LINKED ISSUE #1447: Java 8 - OptionalValidatedValueUnwrapper supports Optional but no classes exist for OptionalInt, OptionalLong, etc
The Java 8 module `OptionalValidatedValueUnwrapper` supports only `java.util.Optional` and there should be additional classes created for `OptionalInt`, `OptionalLong`, and `OptionalDouble`. This is necessary so `@UnwrapValidatedValue` works and can unwrap `OptionalInt` variables for things like `@Min`, `@Max`, etc. Until this is resolved we need to use `Optional<Integer>`, `Optional<Long>`, `Optional<Double>` for validation to work.

joschi:
This PR adds `ValidatedValueUnwrapper`, `ParamConverter`, and `MessageBodyWriter` implementations for `OptionalDouble`, `OptionalInt`, and `OptionalLong` to be consistent with the already existing `GuavaOptionalValidatedValueUnwrapper` and `OptionalValidatedValueUnwrapper` classes.

Fixes #1447

### DIFF
--- a/dropwizard-core/src/test/java/io/dropwizard/setup/BootstrapTest.java
+++ b/dropwizard-core/src/test/java/io/dropwizard/setup/BootstrapTest.java
@@ -12,6 +12,9 @@ import io.dropwizard.jackson.Jackson;
 import io.dropwizard.jersey.validation.NonEmptyStringParamUnwrapper;
 import io.dropwizard.jersey.validation.ParamValidatorUnwrapper;
 import io.dropwizard.validation.valuehandling.GuavaOptionalValidatedValueUnwrapper;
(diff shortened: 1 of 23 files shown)


jplock:
Thanks @joschi!  Do you think we need equivalent `Optional{Int,Long,Double}MessageBodyWriter` and `Optional{Int,Long,Double}ParamConverterProvider`?  I know that's not directly related to fixing #1447 but then you could use those as query parameters as well.

joschi:
@jplock Good point. I'll add those.

nickbabcock:
Looks great, thanks for the thorough test cases.

PR #1452, opened 2016-02-07
Handle redirect in WebApplicationException response

tumbarumba:
The change to "[Format WebApplicationException](https://github.com/dropwizard/dropwizard/commit/930443a5bc5760c3319bedf7b36e13810bf0dd8e#diff-235351c1b64afb2dba6188b2bd0c5386)" introduced a regression: namely, if the response contained within the WebApplicationException is a redirect, the redirect no longer works. Instead, a JSON entity is returned instead.

This change restores the previous behaviour for redirects in the WebApplicationException response.


evnm:
LGTM

nickbabcock:
Good catch, if there ever will be another 0.9.x, this would be good candidate to include.

PR #1459, opened 2016-02-14
test case to demonstrate possible failure, was expecting managed obje…

randyp:
It appears Server/EnvironmentCommand does not shutdown managed executors when an exception is thrown in application.run, thus causing dropwizard applications to hang even though the application has shutdown.

Historically the behavior I expected was for managed objects to be shutdown, even if they had not been started. If this is no longer the expected behavior, my apologies, just close this issue/pr.

Technically this is affecting all managed objects, not just managed executors, but is presenting itself most seriously through executors.

This is just a unit test demonstrating the issue.

### DIFF
--- a/dropwizard-core/src/test/java/io/dropwizard/cli/ServerCommandTest.java
+++ b/dropwizard-core/src/test/java/io/dropwizard/cli/ServerCommandTest.java
@@ -25,6 +25,14 @@ public class ServerCommandTest {
         }
     }
 
+    private static class NeedsToBeStopped extends AbstractLifeCycle {
+        boolean stopped = false;
+        @Override
+        protected void doStop() throws Exception {
+            stopped = true;
+        }
+    }
+
     private final MyApplication application = new MyApplication();
     private final ServerCommand<Configuration> command = new ServerCommand<>(application);
     private final Server server = new Server(0) {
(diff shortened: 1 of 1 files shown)


evnm:
This is tough. In principle, any condition that prevents clean process shutdown should be avoided. But since `ExecutorService`s don't really have a "start" state transition (i.e. they're essentially initialized as being in the "started" state), it's somewhat of a hack to stuff them into managed lifecycles.

One workaround would be to make `ExecutorServiceManager`'s constructor take an `ExecutorServiceBuilder` rather than an `ExecutorService` and then invoke `ExecutorServiceBuilder#build` in the manager's `start` method. This would be a breaking change though, so is off the table for now.

For any applications that are prone to this bug, we could add an alternate constructor to `ExecutorServiceManager` which enacts this lazy executor construction. Given this, I think we could add a test case like the one you've written that passes when the builder-accepting constructor is used.

@randyp, have you run into this bug in production? Or did you just happen to notice it in development? I think it may be one of those cases where the rough edge is preferable to the kludge that would be required to fix it.

randyp:
@evnm Apologies for delayed response.

Noticed this while developing (external to core) dropwizard libraries which heavily use ExecutorService. However, my instinct is that a one-off change to work around ExecutorService should be avoided. The issue is with **all managed objects**.

Here is my reasoning as a user:
# I expect all managed objects to be stopped on server shutdown.
# Startup errors should perform best effort server shutdown.
# Therefore startup errors should stop any managed object which hash been started.

However, let me try to walk the talk and propose a change which addresses this issue, I guess.

randyp:
Closing this PR since I no longer think this test case is valid.

evnm:
Worst case, I think we re-file this as an issue and slap the 2.0.0 milestone on it. At present, there aren't any plans for a 2.x release train, but I created the milestone to track breaking nice-to-have changes like this that we could feasibly revisit in the future when a major version bump looms.

randyp:
I agree.

PR #1466, opened 2016-02-20
Support serialisation of lazy loaded POJOs

ant3:
Currently any hibernate loaded objects must be fully loaded before the resource method ends, otherwise the Jackson serialiser fails.  This forces developers to annotate associations for eager loading, or walk the object graph to force hibernate to load associations - neither is very nice.

However, the Hibernate5Module can handle lazy loading proxies, but doesn't do this by default (not sure why not) - and developers could enable it.  The problem is that UnitOfWorkApplicationListener closes the hibernate session before the results have been seriazised, so even if developers turn on the lazy-loading support it won't work.

The changes I have made cause the session to be closed later and switch on lazy-loading support by default, as well as adding a test for this functionality.

Possible problems: 
Closing the session later may have knock-on effects?  It doesn't seem likely but I'm no expert.

The test is a little messy and there may be a better way of doing it.  Also, I have a somewhat messy commit history (3 instead of 1).  I accidentally committed to my master branch :-(.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
@@ -75,7 +75,7 @@ public class UnitOfWorkApplicationListener implements ApplicationEventListener {
                 UnitOfWork unitOfWork = methodMap.get(event.getUriInfo()
                         .getMatchedResourceMethod().getInvocable().getDefinitionMethod());
                 unitOfWorkAspect.beforeStart(unitOfWork);
-            } else if (event.getType() == RequestEvent.Type.RESP_FILTERS_START) {
+            } else if (event.getType() == RequestEvent.Type.FINISHED) {
                 try {
                     unitOfWorkAspect.afterEnd();
                 } catch (Exception e) {
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java
@@ -1,15 +1,18 @@
 package io.dropwizard.hibernate;
 
+import org.hibernate.SessionFactory;
+
 import com.fasterxml.jackson.datatype.hibernate5.Hibernate5Module;
+import com.fasterxml.jackson.datatype.hibernate5.Hibernate5Module.Feature;
 import com.google.common.collect.ImmutableList;
+
 import io.dropwizard.Configuration;
 import io.dropwizard.ConfiguredBundle;
-import io.dropwizard.db.PooledDataSourceFactory;
 import io.dropwizard.db.DatabaseConfiguration;
+import io.dropwizard.db.PooledDataSourceFactory;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 import io.dropwizard.util.Duration;
-import org.hibernate.SessionFactory;
 
 public abstract class HibernateBundle<T extends Configuration> implements ConfiguredBundle<T>, DatabaseConfiguration<T> {
     public static final String DEFAULT_NAME = "hibernate";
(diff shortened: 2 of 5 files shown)


jplock:
Looks ok to me, but I personally don't have much experience with Hibernate.  Would any of the other @dropwizard/committers mind reviewing?

joschi:
@ant3 I'm a bit hesitant to merge this PR because it changes the default behavior without providing the possibility to change it back to the old behavior.
Would it make sense to add a parameter to the `HibernateBundle` to configure whether it's using eager or lazy loading by default?

arteam:
I concur with @joschi. An additional parameter that enables lazy loading would make sense.

Personally, I am not a big fan of lazy loading, because it's a very brittle abstraction. I think, it would be reasonable to ask users to explicitly enable this feature and keep it disabled by default.

ant3:
Yes, these are good points.  I should have put in a parameter (and will do), to enable/disable lazy loading.  So with lazy loading enabled both eager and lazy loading will work and with it disabled only eager loading will work (as now).

I think lazy loading should be enabled by default as it makes things 'just work' and anyone familiar with hibernate will expect it.  Those who want to use eager loading will be unaffected either way.

I understand arteam's point about lazy loading often causing problems, but this is almost always due to hibernate sessions being closed at the wrong time.  This can be a great cause of frustration to new hibernate users and does sometimes cause hard to fix bugs.  However, the beauty of this use-case is that it is a fully encapsulated layer boundary, so we know exactly when to close the session and we can be sure the hibernate entities won't be referenced after the method has exited*.

Of course, I'm happy to go with the majority view, but that's my opinion :-).

* - The Entities could be referenced if the resource method spawned another thread which used them after the method exited - but I think this is a crazy edge-case, and ultimately crazy developers can break any framework.

ant3:
Any comments?  Is this what you were thinking of?

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
```diff
@@ -75,7 +75,7 @@ public void onEvent(RequestEvent event) {
                 UnitOfWork unitOfWork = methodMap.get(event.getUriInfo()
                         .getMatchedResourceMethod().getInvocable().getDefinitionMethod());
                 unitOfWorkAspect.beforeStart(unitOfWork);
-            } else if (event.getType() == RequestEvent.Type.RESP_FILTERS_START) {
+            } else if (event.getType() == RequestEvent.Type.FINISHED) {
```
nickbabcock: Can you explain this change and why it's needed?

ant3:
Nick: The serialisation of objects (returned by a resource method) to JSON (or whatever protocol) happens after the RESP_FILTERS_START event, so if we close the session at that point it is closed when we do the serialisation and hence we cannot handle lazy loading of objects.  This wasn't a problem before as lazy loading was explicitly disabled and any objects that were not loaded (lazy proxies) were serialised as nulls.  Now I have added the option to handle lazy loading we need to ensure that the Hibernate session stays open until after the serialisation to the wire protocol (e.g. JSON).

nickbabcock:
Thanks for the explanation, I just wanted to make sure that it wasn't a change by accident because the hibernate module seems prone to mistakes.

> Those who want to use eager loading will be unaffected either way.

Sounds like backwards compatibility is maintained, which is :+1: from me.

nickbabcock:
Merging this under the impression that backwards compatibility is kept, and this seems like a nice feature for those who want it. I'm assuming there will also be another rc for 1.0.0 at least for this and #1527

ant3:
Thanks Nick.  You are correct.  If the user sets Lazy Loading to off then the only thing that happens differently (compared to previous codebase) is that the session gets closed later (as you previously noticed).

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
apgeorge: Hi, I think this may have introduced a bug. Now the session will be closed and the transaction committed only after the serialization has happened. So if there are any database exceptions during the transaction commit (column val exceed/constraints etc..) they happen only after the serialized response has been sent back. So the client assumes things are fine, but actually the server logs show the exception.

I guess we will still need to end the transaction or commit it in the RESP_FILTERS_START event. And session can be closed in FINISHED event.

apgeorge:
Have raised the below issue introduced by the above changes. https://github.com/dropwizard/dropwizard/issues/1604

PR #1469, opened 2016-03-01
add example for Configured Bundles docs

mclarke47:
I'd like to add an example to the docs to show the structure of a simple configured bundle. I recently did some work creating a configured bundle and I think this would have helped.


arteam:
LGTM. Thanks for the your docs contribution!

PR #1480, opened 2016-03-16
Upgraded to Jetty 9.3.8.v20160314

arteam:
This is a bugfix release. The release notes are [here](https://github.com/eclipse/jetty.project/blob/jetty-9.3.x/VERSION.txt)


arteam:
Some tests are falling. It looks like that the update introduced some regression. Will try to invistigate that.

joschi:
Thanks!

Issue #1478, opened 2016-03-16
new logback versions do not log to file 


chas678:
Dropwizard 0.9.2 uses Logback 1.1.3.  Attempting to use newer versions of Logback 1.1.5 + 1.1.6 mean that logs are longer written to file when configured to do so in the YAML which violates the principle of least astonishment.

## Steps to reproduce

1. change pom to use logback version 1.1.5 or 1.1.6
1. Run a service locally with a file appender configured in the YAML - for example: 

```yaml
logging:
  level: INFO
  loggers:
    "com.example": DEBUG
  appenders:
    - type: console
      target: stdout
      timeZone: UTC
      logFormat: "%-5p [%d{ISO8601,UTC}] %7X{userUUID} %c: %m%n%rEx"
    - type: file
      threshold: ALL
      currentLogFilename: /tmp/example-service/server.log
      archivedLogFilenamePattern: /tmp/app/example-service/server-%d.log.gz
      archivedFileCount: 20
      timeZone: UTC
      logFormat: "%-5p [%d{ISO8601,UTC}] %7X{userUUID} %c: %m%n%rEx"
```
## Expected results
* Log goes to console and also to file at ` /tmp/example-service/server.log`

## Actual results
* log goes to console, but no file is created in ` /tmp/example-service/server.log`

## Suspicion/Guess as to cause
Logback has changed the way MDC is inherited by threads .

## Refs
* Logback release notes : http://logback.qos.ch/news.html

xiaodong-xie:
I don't think it's related to MDC.

Here is my analysis:
1. It's because `RollingFileAppender` is more strict from 1.1.5, as it will check whether the `triggeringPolicy` is started or not.
2. In class `FileAppenderFactory`, we first `stop` then `start` the appender, so the `triggeringPolicy` is stopped. 
3. I managed to pass all test cases in `dropwizard-logging` module by just removing line 202 `appender.stop();`.

So my question is, is `appender.stop();` really necessary? 

Here is my PR: https://github.com/dropwizard/dropwizard/pull/1481

ghenkes:
I looked into this some last week as well and came to the same conclusion as @xiaodong-xie, that `appender.stop()` should be removed.

[...]

PR #1481, opened 2016-03-16
Fix issue when upgrading logback to 1.1.5 beyond

xiaodong-xie:
Fix https://github.com/dropwizard/dropwizard/issues/1478

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/FileAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/FileAppenderFactory.java
@@ -199,7 +199,6 @@ public class FileAppenderFactory<E extends DeferredProcessingAware> extends Abst
         appender.setPrudent(false);
         appender.addFilter(levelFilterFactory.build(threshold));
         getFilterFactories().stream().forEach(f -> appender.addFilter(f.build()));
-        appender.stop();
         appender.start();
 
         return wrapAsync(appender, asyncAppenderFactory);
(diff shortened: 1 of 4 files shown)


dropwizard-logging/src/main/java/io/dropwizard/logging/FileAppenderFactory.java
```diff
@@ -199,7 +199,6 @@ public boolean isMaxFileSizeSettingSpecified() {
         appender.setPrudent(false);
         appender.addFilter(levelFilterFactory.build(threshold));
         getFilterFactories().stream().forEach(f -> appender.addFilter(f.build()));
-        appender.stop();
```
xiaodong-xie: If this line was not removed, test `DefaultLoggingFactoryTest::testConfigure` will fail

xiaodong-xie:
If we only upgrade logback to 1.1.5, build will pass: https://travis-ci.org/dropwizard/dropwizard/builds/116483726

But logback 1.1.6 introduced some other more strict stuff: (http://logback.qos.ch/codes.html#earlier_fa_collision)
```
File/FileNamePattern option has the same value "..." as that given for appender [...] defined earlier.

If a FileAppender/RollingFileAppender defined earlier has the same File option as the current appender, then those two appenders are in collision as FileAppender instances cannot share the same output target. To prevent loss of data, the current appender will not start. Make sure that each appender has a unique File option.

By analogy the same restriction applies to the FileNamePattern option of RollingFileAppender. Make sure that each RollingFileAppender has a unique FileNamePattern option
```

So if we want to upgrade 1.1.6, we need to tweak `yaml/logging_advanced.yml` file...

arteam:
The fix looks good, thanks you very much for contributing! 

I think we should not hesitate to upgrade to 1.1.6. As I see it, the release brings several improvements (e.g. disabling packaging data) and many bug fixes, which should beneficial for Dropwizard users. Judging by the rate of releases, the logback development community actively fixes other discovered bugs. I think we can rely on it and expect that potential bugs, with which Dropwizard users may encounter, will be fixed.

xiaodong-xie:
@arteam, thanks for merging... :)

Issue #125, opened 2012-07-02
Support multiple config files


tomakehurst:
We'd find it really useful to be able to specify two config files at startup, one for defaults, one for environment-specific overrides.

Would you be interested in including this if we supply a patch?

vvondra:
A nice touch would be able to override config files, or even by default be able to import them from one another.

config extended by config_prod.yml and config_staging.yml

sirianni:
Layering config files is also important for separating out sensitive values (e.g. mounting a Kubernetes `Secret` in addition mounting a Kubernetes `ConfigMap`).  Or using the [Vault sidecar injector](https://www.vaultproject.io/docs/platform/k8s/injector).

Almost all modern java microservice frameworks have this feature.

joschi:
@sirianni You can override any setting in the configuration file with a system property or with environment variables.

See also: [Using Secrets as environment variables](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables) and [Dropwizard -> Configuration -> Environment variables](https://www.dropwizard.io/en/release-2.0.x/manual/core.html#environment-variables).

vvondra:
> Layering config files is also important for separating out sensitive values (e.g. mounting a Kubernetes `Secret` in addition mounting a Kubernetes `ConfigMap`). Or using the [Vault sidecar injector](https://www.vaultproject.io/docs/platform/k8s/injector).
> 
> Almost all modern java microservice frameworks have this feature.

it's three years since I touched this, but we found a way to plug in this behaviour with our own code: https://github.com/vvondra/dropwizard-config-importable

PR #1484, opened 2016-03-19
Support importing and overriding multiple configuration files

vvondra:
Hi,

this is a quick PoC implementation of https://github.com/dropwizard/dropwizard/issues/125
The issue was closed, but I think the feature could be introduced without adding too much complexity. I deliberately ignored some details such as other paths then files in the quick prototype, but those could be added by specifying the type of the import.

The merging of the configs is done early on the JsonNode level, with same semantics as implementations of extending objects like `$.extends`.

Stuff to add:
- [ ] Circular deps check
- [ ] Different resource path types (URL, resource, file path)
- [ ] Test coverage

Would this feature be considered to be accepted? I'll gladly put in more work on it.

The use case would be:

`base.yml`:
```yml
database:
    driverClass: org.postgresql.Driver
    logValidationErrors: true
```

`dev.yml`:
```yml
import:
   - base.yml

database:
   user: postgres
```

`prod.yml`:
```yml
imports:
   - base.yml

database:
   user: ${PROD_DB_USER}
   logValidationErrors: false
```

Running the app with `prod.yml` would result in configuration loaded as:
```yml
database:
    driverClass: org.postgresql.Driver
    user: ${PROD_DB_USER}
    logValidationErrors: false # Overridden the value from base
```

### DIFF
--- a/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationFactory.java
+++ b/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationFactory.java
@@ -29,6 +29,8 @@ import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import io.dropwizard.jackson.JsonUtil;
+
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -87,8 +89,12 @@ public class ConfigurationFactory<T> {
      * @throws ConfigurationException if there is an error parsing or validating the file
      */
     public T build(ConfigurationSourceProvider provider, String path) throws IOException, ConfigurationException {
+        return build(loadConfiguration(provider, path), path);
+    }
+
+    private JsonNode loadConfiguration(ConfigurationSourceProvider provider, String path) throws IOException, ConfigurationParsingException {
         try (InputStream input = provider.open(requireNonNull(path))) {
-            final JsonNode node = mapper.readTree(yamlFactory.createParser(input));
+            JsonNode node = mapper.readTree(yamlFactory.createParser(input));
 
             if (node == null) {
                 throw ConfigurationParsingException
@@ -96,7 +102,22 @@ public class ConfigurationFactory<T> {
                         .build(path);
             }
 
-            return build(node, path);
+            if (node.has("imports") && node.get("imports").isArray()) {
+                for (JsonNode configImport : node.get("imports")) {
+                    if (configImport.isTextual()) {
+                        JsonNode imported = loadConfiguration(
+                            provider,
+                            new File(new File(path).getParent(), configImport.asText()).toString()
+                        );
+                        node = JsonUtil.merge(node, imported);
+                    }
+                }
+
+                node = ((ObjectNode) node).without("imports");
+            }
+
+
+            return node;
         } catch (YAMLException e) {
             final ConfigurationParsingException.Builder builder = ConfigurationParsingException
                     .builder("Malformed YAML")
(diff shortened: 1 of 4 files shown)


arteam:
Thanks for working on this!

> Would this feature be considered to be accepted?
Yes, I would accept it. In production installations a configuration file can become quite big (logging, metrics, database pool, own service configuration, etc..), which may be inconvenient for configuration management. Also, some administrators prefer to have several small configuration files then one big to mitigate risks of manual errors. I think in such scenarios this feature could be quite useful.

@dropwizard/committers, does anyone have concerns with this issue?

ryankennedy:
My primary concern would be the one Coda already mentioned in #125, too much complexity in the configuration process. A past employer had this really awful Spring XML-based configuration thing where the configuration would flatten: configuration from a server, configuration on disk, configuration in your JAR file, configuration in some other JAR file, etc. It basically made it an impossible task to know where a given configuration value was coming from because there were so many possibilities.

Some questions:

- Should the import depth be limited? Setting a depth limit of 1 would allow a base layer and an environment layer (the prod/dev example given) and would also prevent cycles by design.
- Should there be a /configuration resource in the admin endpoint allowing someone to see the current configuration and the source of each configured value (i.e. default, base.yml, prod.yml, or command line flag)?
- Is "import" really the right verb for what's happening? "inherit from" would be more accurate.
- Related, what's the inheritance order? I assume defaults in the Configuration classes are the base value, followed by the "lowest" imported definition of a value, and finally topped off with any command line defined values?

jplock:
-1 from me.  I think it's making the configuration process too complex if it is used.

vvondra:
@ryankennedy 
 - import depth wouldn't be limited, circular deps would be detected by keeping a simple list of visited imports and throwing an exception during configuration loading if a cycle is detected
 - I borrowed naming from Symfony2 in PHP (which probably borrowed it from Spring) http://symfony.com/doc/current/cookbook/configuration/configuration_organization.html#different-directories-per-environment
 - the inheritance order is pretty straightforward, we have all the import resolution logic **before** looking at the Configuration classes so that part of the code stays exactly as simple as before with the possibility of loading just one configuration file. (also answering @jplock )
 - it essentially just allows you to split the config file into multiple ones (anytime you import a file, any other properties you define override fields in the imported one)

vvondra:
An alternative solution helping us would be making it possible to provide a custom `ConfigurationFactory`, right now it's very tough since the class has no interface and if you write a custom `ConfigurationFactoryFactory`, you must extend the existing implementation. But extending is an issue since a) its fields are private (making them protected would add surface for BC breaks) b) you must override quite a few methods and it's hard to guarantee the extended implementations won't change.

brentryan:
FWIW, we at Cvent have implemented a very similar solution with our dropwizard services. We really couldn't live without this. However, our next evolution of this is to have 1 template.yaml that we use to generate via build/deploy pipeline a test/staging/prod config using a template system like handlebars. Then the next step will be to evolve this into a runtime config system.

nickbabcock:
My 2¢. I like this feature and I believe that some of variation should land, but I have some opinions on the matter.

- I don't like specifying multiple imports because, in a conflict, is it the last import wins or first? This problem can be alleviated with conventions (ie, "max of one import per file"), or mitigated through @ryankennedy's suggestion about some runtime (or even startup) indication about what the final form the configuration file resolved to. Though care would have to be considered if dealing with sensitive config values.
- Not sure if I like the ability for resources to be imported through URLs (I'd prefer to use only file paths) because I've been burned when an app calls out to an external IP only to be slowed down by a 60 second reverse DNS lookup. I can be convinced otherwise, though.

Overall I like the approach, but I think taking a simple approach would win out because we don't want to incorporate everything under the sun, which would be the job for a configuration management tool (eg. Chef).

vvondra:
Since there are opinions for both yes and nay on this PR, I'd like to propose a much smaller change extracting an interface for the ConfigurationFactory for easier user implementation of such behaviour

https://github.com/dropwizard/dropwizard/pull/1507

vvondra:
With #1507 merged, I'll close this and implement it on our own.

I still feel this could be useful though, if somebody has ideas how to make the behaviour predictable enough, I'd love to hear them.

PR #1485, opened 2016-03-19
Revise code example in docs for testing non-junit

nitram509:
Hi,

after reading the docs of 0.9.2, I've tried the given example for non-junit tests.
In my case I'm using TestNG.
Unfortunately the give code doesn't compile and there where 'BindException'
because port 8080 was already in use.

So I've added an overriding config to get a random port and I fixed the wrong var-name.
Additionally I removed the parameter from JerseyClientBuilder,
because 'environment' and 'test-name' aren't supported.
My tests are running fine without them, but I have to admit,
that I'm not using any "special" Jackson mappings.

Feedback is always welcome.

Kind regards
Martin

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -382,7 +382,7 @@ By creating a DropwizardTestSupport instance in your test you can manually start
 
         @Test
         public void loginHandlerRedirectsAfterPost() {
-            Client client = new JerseyClientBuilder(SUPPORT.getEnvironment()).build("test client");
+            Client client = new JerseyClientBuilder().build();
 
             Response response = client.target(
                      String.format("http://localhost:%d/login", SUPPORT.getLocalPort()))
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -326,7 +326,10 @@ The ``DropwizardClientRule`` takes care of:
 Integration Testing
 ===================
 
-It can be useful to start up your entire app and hit it with real HTTP requests during testing.
+It can be useful to start up your entire application and hit it with real HTTP requests during testing.
+The ``dropwizard-testing`` module offers helper classes for your easily doing so.
+The optional ``dropwizard-client`` module offers more helpers, e.g. a custom JerseyClientBuilder,
+which is aware of your application's environment.
 
 JUnit
 -----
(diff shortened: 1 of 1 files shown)


docs/source/manual/testing.rst
```diff
@@ -365,24 +365,27 @@ By creating a DropwizardTestSupport instance in your test you can manually start
...
         @AfterClass
         public void afterClass() {
-          SUPPORT.after();
+            SUPPORT.after();
         }
 
         @Test
         public void loginHandlerRedirectsAfterPost() {
-            Client client = new JerseyClientBuilder(SUPPORT.getEnvironment()).build("test client");
+            Client client = new JerseyClientBuilder().build();
```
jplock: Are you using the dropwizard provided JerseyClientBuilder? That should support environment.

docs/source/manual/testing.rst
```diff
@@ -365,24 +365,27 @@ By creating a DropwizardTestSupport instance in your test you can manually start
     public class LoginAcceptanceTest {
 
         public static final DropwizardTestSupport<TestConfiguration> SUPPORT =
-                new DropwizardTestSupport<TestConfiguration>(MyApp.class, ResourceHelpers.resourceFilePath("my-app-config.yaml"));
+                new DropwizardTestSupport<TestConfiguration>(MyApp.class,
+                    ResourceHelpers.resourceFilePath("my-app-config.yaml"),
+                    ConfigOverride.config("server.applicationConnectors[0].port", "0") // create random port
```
jplock: This line shouldn't be necessary if there is a separate config file specifically for testing that sets port=0

nitram509: OK, got it.
I would like to use the original config file of the application to avoid maintaining two config files.
That's why I found overriding more charming.

Maybe just change the comment to make clear, that this is optional?

docs/source/manual/testing.rst
nitram509: Not sure what you mean.
I add 'dropwizard-testing' to my classpath with scope test.
When searching for "JerseyClientBuilder" there is just one class.

See source:
https://github.com/MeteoGroup/jbrotli/blob/master/jbrotli-servlet-examples/dropwizard/src/test/java/org/meteogroup/jbrotli/servlet/HelloBrotliApplicationTest.java

docs/source/manual/testing.rst
jplock: Yea, I think a comment is a good idea

docs/source/manual/testing.rst
jplock: It's in the dropwizard-client module which is probably what these docs were assuming were being used.

docs/source/manual/testing.rst
nitram509: May I change the comment?
Or what do you propose how to adapt this PR?

docs/source/manual/testing.rst
nitram509: Thanks for pointing this out.
I will test and adopt the docs.

docs/source/manual/testing.rst
jplock: Maybe make the comment "If not using a separate testing-specific configuration file, use a randomly selected port"

nitram509:
@jplock 
revised the comment and explicitly named the modules.

What do you think, is it more clear now?

jplock:
Looks great! Thanks for the contribution.

Issue #1488, opened 2016-03-23
Missing rollbackOnReturn in DataSourceFactory


patrox:
While trying to set `rollbackOnReturn` to true in the config, I discovered that such field is not supported,
altough a bit similiar `commitOnReturn` is supported:

```
 * Unrecognized field at: database.rollbackOnReturn
    Did you mean?:
      - commitOnReturn
      - validationQuery
      - validationInterval
      - useFairQueue
      - removeAbandoned
        [30 more]
```

Both of them are available in the properties which can be set for the Tomcat JDBC Pool, which is used as a backing pool:

https://tomcat.apache.org/tomcat-9.0-doc/jdbc-pool.html#Tomcat_JDBC_Enhanced_Attributes

1.) Are there any reasons why it's not supported ?
2.) As I actually would like to have it - can I create a PR which will add it ?

jplock:
No reason I can think of. Please submit a PR and we'll gladly review. Thanks!

patrox:
Testing it right now - once the tests will pass, I'll post a PR.

patrox:
@jplock when you'll have spare 5 minutes, could you please have a look at: https://github.com/dropwizard/dropwizard/pull/1489 ?

I've updated tests & javadoc, to cover the new field.

PR #1489, opened 2016-03-23
Add DataSourceFactory.rollbackOnReturn flag

patrox:
Resolves: #1488

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -69,7 +69,7 @@ GZip
 .. code-block:: yaml
 
     server:
-      gzip: 
+      gzip:
         bufferSize: 8KiB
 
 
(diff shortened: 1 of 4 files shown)


jplock:
I think we need a setter in the DataSourceFactory. Otherwise looks good. Is this enabled or disabled by default today?

jplock:
Do you also need to add it when the pool is built?

patrox:
It's disabled by default, just like `DataSourceFactory.commitOnReturn`

patrox:
> Do you also need to add it when the pool is built?

yes, I missed that - fixed

joschi:
@patrox Please avoid mere whitespace changes in `docs/source/manual/configuration.rst`. Other than that the changes look good to me.

patrox:
@joschi Sorry for that - I guess that IDEA reformatted code during save - should I revert these changes ?

joschi:
@patrox It seems to be caused by our current [`.editorconfig`](https://github.com/dropwizard/dropwizard/blob/master/.editorconfig) file. Just keep it that way.

joschi:
Thanks!

PR #1490, opened 2016-03-23
Fixed broken or outdated links

rbarbey:
- link to JSR 349 did not render correctly, removed hyphen (which seems
  to be the official variant anyway)
- fixed broken link to Gant (codehaus shut down)
- use updated link to Guava (Google Code shut down)


jplock:
Thanks!

PR #1493, opened 2016-03-24
Replace guava's Optional with java.util's one

patrox:


jplock:
This is great, thank you!  Any feedback from @dropwizard/committers about this?  This may break some things, but probably worthwhile to get into 1.0.0-rc2.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java
```diff
@@ -58,7 +58,7 @@ public final void run(T configuration, Environment environment) throws Exception
         environment.healthChecks().register(name(),
                                             new SessionFactoryHealthCheck(
                                                     environment.getHealthCheckExecutorService(),
-                                                    dbConfig.getValidationQueryTimeout().or(Duration.seconds(5)),
+                                                    dbConfig.getValidationQueryTimeout().orElseGet(() -> Duration.seconds(5)),
```
nickbabcock: Why not `orElse(Duration.seconds(5))`?

patrox: To avoid evaluating it, if it's not needed.

joschi: This is being called *once*, so I don't think that performance optimization is relevant here.

joschi:
LGTM except for that minor comment on `HibernateBundle`. :+1:

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java
patrox: force of habit - will change that

joschi:
Thanks!

PR #1498, opened 2016-03-29
Fix handling multiple headers with the same name

### LINKED ISSUE #1494: dropwizard-client 1.0.0rc1 can't handle multiple Set-Cookie headers
I'm having trouble with the implementation of the DropwizardApacheConnector. Specifically, its use of Collections.singletonList on line 98 of this file 

  https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java

When receiveing a response with multiple cookies, they will all have a Set-Cookie header (with different data). Only the first such header will succeed. Subsequent headers can't be added since the headerValues, as far as I can tell, then always will be a Collections.singletonList, thus causing line 100 to fail since it can't add to such a "singleton" list. This results in headerValues.add causing an UnsupportedOperationException.

In order to move forward, I hacked around the problem for the time being as shown below. Please let me know if I'm totally missing the point here. If not, hopefully this helps helps someone else and shows a possible solution.

```
	for (Header header : apacheResponse.getAllHeaders()) {
		final List<String> headerValues = jerseyResponse.getHeaders().get(header.getName());
		if (headerValues == null) {
			List<String> list = new ArrayList<>(1);
			list.add(header.getValue());
			jerseyResponse.getHeaders().put(header.getName(), list);
		} else {
			headerValues.add(header.getValue());
		}
	}
```
-JM

patrox:
~~As a simple fix, an immutable, single-element list (`Collections.singletonList`) has been wrapped in a mutable one, so it can be extended (if there will be anymore headers with the same name).~~

In this case `Lists.newArrayList` was used to create a mutable `List`, so it can be extended in case of multiple headers with the same name, i.e. `Set-Cookie`.

Fixes #1494

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
@@ -1,5 +1,6 @@
 package io.dropwizard.client;
 
+import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.MoreExecutors;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
@@ -26,7 +27,6 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Future;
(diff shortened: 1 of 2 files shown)


jplock:
True, I don't know if we've settled one way or the other about Guava or not, but I don't see that being removed as a dependency any time soon.  Thanks!

nickbabcock:
Thanks for the fix and the test!

PR #1499, opened 2016-03-29
fix for metrics library url at getting started section

burakdd:
[Metrics for metrics](http://www.dropwizard.io/0.9.2/docs/getting-started.html#metrics-for-metrics)

Metrics project url `http://metrics.codahale.com/` does not point to valid url, how about pointing to `http://metrics.dropwizard.io/`?


jplock:
Thanks!

PR #1502, opened 2016-03-30
Added the ability to set application/admin contextPath during Application.run()

qinfchen:
This commit allows users to set application/admin contextPath in the following
priority order:

1. the yaml configuration
2. set the context path from the application/admin context in the application's `run` block
3. the default value from SimpleServerFactory or DefaultServerFactory

This should not have breaking changes for users who have configured the context path for their apps via the yaml file.


nickbabcock:
What's the use case for this?

qinfchen:
Closing it. Since the main use case is to have context path available during application run, but if the user configured via yaml... that won't be available with this change :(

PR #1503, opened 2016-03-31
configures the environment with the context paths before the application.run()

qinfchen:
This commit configures the context paths for the application before the
application `run()` instead of setting it prior to starting the server.
This allows applications to obtain the context paths during the run
method.

This shouldn't break any changes to the users, but note that
ServerFactory now needs to call `configure` explicitly to set the
context paths in the environment

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/server/SimpleServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/server/SimpleServerFactory.java
@@ -103,8 +103,6 @@ public class SimpleServerFactory extends AbstractServerFactory {
         final ThreadPool threadPool = createThreadPool(environment.metrics());
         final Server server = buildServer(environment.lifecycle(), threadPool);
 
-        LOGGER.info("Registering jersey handler with root path prefix: {}", applicationContextPath);
-        environment.getApplicationContext().setContextPath(applicationContextPath);
         final Handler applicationHandler = createAppServlet(server,
                                                             environment.jersey(),
                                                             environment.getObjectMapper(),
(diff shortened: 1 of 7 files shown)


arteam:
Looks good! It seems that this improvement doesn't break anything and provides a way to get access to context paths in an application class. I remember some users asked for this feature in the mailing list.

arteam:
Oh, for some reasons the build fails on Travis after merging. I've reverted the commit for the being.
Will try to try find a source of the error.

patrox:
@arteam shouldn't this commit be reverted as well: https://github.com/dropwizard/dropwizard/commit/171bbe63d836a53d966845283ba4f777cfb0fecc ?

arteam:
@patrox Let's hope we will find the error cause and this improvement will be merged anyway :)

dropwizard-core/src/main/java/io/dropwizard/server/SimpleServerFactory.java
```diff
@@ -136,4 +132,13 @@ public Server build(Environment environment) {
 
         return server;
     }
+
+    @Override
+    public void configure(Environment environment) {
```
jplock: since this is the same in both `SimpleServerFactory` and `DefaultServerFactory`, should we move this to the `AbstractServerFactory` instead?

qinfchen: the context path logic is slightly different between the two. It would be nice to move them into `AbstractServerFactory`

qinfchen:
@arteam found the issue in the original test. `configure` needs to be called to setup the environment for other tests. Separate PR is created. Sorry about that

PR #1504, opened 2016-03-31
configures the context paths before the application.run()

qinfchen:
https://github.com/dropwizard/dropwizard/pull/1503 including a build fix


qinfchen:
hmm. There is no coverage for EnvironmentCommand `run` method. Since i added a line in there, the coverage went down... :(.

jplock:
Thanks for the contribution!

Issue #1505, opened 2016-03-31
OffsetDateTimeMapperTest is failing


patrox:
While testing something completely unrelated, I stumbled upon two failed tests, which IMO seem to fail because of DST offset, as:

```
  OffsetDateTimeMapperTest.mapColumnByIndex:44 expected:<...12-03T10:15:30.375+0[2]:00> but was:<...12-03T10:15:30.375+0[1]:00>
  OffsetDateTimeMapperTest.mapColumnByName:25 expected:<...12-03T10:15:30.375+0[2]:00> but was:<...12-03T10:15:30.375+0[1]:00>
```

In the above output you can see that the [`actual`](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jdbi/src/test/java/io/dropwizard/jdbi/args/OffsetDateTimeMapperTest.java#L22-L22) value:

> 2007-12-03T10:15:30.375+01:00 // doesn't include DST offset

is **NOT** equal to the [`expected`](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jdbi/src/test/java/io/dropwizard/jdbi/args/OffsetDateTimeMapperTest.java#L25-L25) value:

> 2007-12-03T10:15:30.375+02:00 // includes DST offset

Should I take care of that ?

patrox:
the same applies to [`OffsetDateTimeArgumentTest.apply`](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jdbi/src/test/java/io/dropwizard/jdbi/args/OffsetDateTimeArgumentTest.java#L20-L20)

jplock:
Yes please! That'd be great. Thanks

patrox:
@jplock, @joschi could you please have a look at the above PR addressing the described issue ?

patrox:
Addressed in #1506

PR #1506, opened 2016-04-01
Fix tests affected by the DST change

patrox:
The problems with these three tests were caused by the fact, that the date used in them has been forcing
non-DST timezone (i.e. `+01:00` in most of the Europe), but then it was compared to the same date, but in current (that is DST) timezone (i.e. `+02:00`).


patrox:
Closes #1505.

jplock:
Looks good to me, thanks!

PR #1507, opened 2016-04-03
ConfigurationFactory has extracted interface for own implementation

vvondra:
Recently I opened up #1484 for discussion but there doesn't seem to be a consensus about the behaviour of the configuration imports.

This PR proposes to extract an interface from the current `ConfigurationFactory<T>` so instead of landing the feature from #1484, users can implement their own logic. https://github.com/dropwizard/dropwizard/pull/1484#issuecomment-199465047

I tried to be conservative with the PR, extracting the two generic methods on which the framework relies and adding adefault for building from `java.io.File`. Tests continue to use the default implementation and no big changes there. I chose to name the default implementation `YamlConfigurationFactory`, since it's an inherent part of it.


jplock:
Thanks for this. I think a default interface method makes sense.

vvondra:
Added, it also makes the patch backwards compatible

dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationFactoryFactoryTest.java
```diff
@@ -27,7 +27,7 @@ public void setUp() throws Exception {
      @Test
      public void createDefaultFactory() throws Exception {
          ConfigurationFactory<Example> factory = factoryFactory.create(Example.class, validator, Jackson.newObjectMapper(), "dw");
-         final Example example = factory.build(validFile);
+         final Example example = factory.build(new FileConfigurationSourceProvider(), validFile.getPath());
```
jplock: Is this change required now that we have the default method in the interface?

dropwizard-core/src/test/java/io/dropwizard/cli/ConfiguredCommandTest.java
```diff
@@ -5,8 +5,7 @@
 import static org.mockito.Mockito.when;
 import io.dropwizard.Application;
 import io.dropwizard.Configuration;
-import io.dropwizard.configuration.ConfigurationFactory;
-import io.dropwizard.configuration.ConfigurationFactoryFactory;
+import io.dropwizard.configuration.*;
 import io.dropwizard.setup.Bootstrap;
```
jplock: Can you use explicit imports everywhere instead of wildcards?

dropwizard-client/src/test/java/io/dropwizard/client/proxy/HttpClientConfigurationTest.java
```diff
@@ -4,9 +4,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.io.Resources;
 import io.dropwizard.client.HttpClientConfiguration;
-import io.dropwizard.configuration.ConfigurationFactory;
-import io.dropwizard.configuration.ConfigurationParsingException;
-import io.dropwizard.configuration.ConfigurationValidationException;
+import io.dropwizard.configuration.*;
```
jplock: Please don't use wildcard imports

dropwizard-db/src/test/java/io/dropwizard/db/DataSourceFactoryTest.java
```diff
@@ -1,8 +1,8 @@
 package io.dropwizard.db;
 
 import com.codahale.metrics.MetricRegistry;
-import io.dropwizard.configuration.ConfigurationFactory;
-import io.dropwizard.configuration.ResourceConfigurationSourceProvider;
+
+import io.dropwizard.configuration.*;
```
jplock: Don't use wildcard imports

dropwizard-configuration/src/main/java/io/dropwizard/configuration/YamlConfigurationFactory.java
```diff
@@ -0,0 +1,239 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A factory class for loading YAML configuration files, binding them to configuration objects, and
```
patrox: minor: to many *and*s

dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationFactoryFactoryTest.java
vvondra: Reverted back to the previous one.

vvondra:
@jplock  @patrox  updated

jplock:
Looks good to me.  Any other @dropwizard/committers want to offer any feedback?

jplock:
Thanks!

PR #1513, opened 2016-04-06
Register the default MetricRegistry as the "default"

jplock:
I end up doing this a lot in my code so I can reference the `MetricRegistry` throughout my app.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/Environment.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/Environment.java
@@ -1,6 +1,7 @@
 package io.dropwizard.setup;
 
 import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.SharedMetricRegistries;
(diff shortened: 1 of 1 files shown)


mattnelson:
I've done this many times as well, but also tried to add first class default support to the metrics registry. I wasn't fond of having to rely on a magic string for the default registry.

https://github.com/dropwizard/metrics/pull/849

jplock:
Good call. Let's see where that can go.

jplock:
Is there any harm in merging this pull request and then refactoring whenever we can upgrade the metrics library?

joschi:
@jplock Nope, it's fine I think. :+1:

jplock:
thanks!

erikvanoosten:
Will the same be done for the health check registry in `SharedHealthCheckRegistries`?

joschi:
@erikvanoosten Good catch. I've addressed this in #1722.

PR #1515, opened 2016-04-10
Feature/improve command reuse and inheritability

jakeswenson:
Improving command reuse by:

- making `EnvironmentCommand`s  `run` non final, so that inheritors can use the configured bootstrap (instead of having to cast the wildcard bootstrap)
- adding a way to configure the file argument further (like adding a default)
- and adding a protected constructor to the server command to call it differently

Currently at work we end up copying the guts of `ServerCommand` into a new command so that we can override the argument names.


jakeswenson:
Sorry just realized I should have started a discussion first. I can start that if necessary.

jplock:
Seems good to me and doesn't break backward compatibility :+1:

jakeswenson:
Thanks @jplock!

Issue #1037, opened 2015-05-08
On demand initialization of database connection pools


tblasche:
I'm currently facing the problem that DW fails to start when there are unreachable databases configured as data sources.

On startup process, DW tries to init the connection pools for all configured data sources. If there is e.g. an unreachable database configured, DW will fail to start with an exception.

**Desirable behaviour:**
DW starts even if there are connection pools which could not be set up. Everytime the application tries to borrow a connection from the connection pool, DW retries to init this connection pool.
Even better: Connection pools are just tried to be initialized when they are used for the first time (i.e. kind of lazy initialization; also can help saving resources).

**Background:**
If e.g. a non-critical database is down in production environment which is only required by let's say 1 of 10 resources, I want DW to start properly so I can use 9 of 10 resources correctly and only requests to the one resource which requires the down database will fail.

arienkock:
I suspect an easy way to support this is to allow the initialSize property of the DatasourceFactory to be 0. During startup the connection pool initializes (in my version this happens at `org.apache.tomcat.jdbc.pool.ConnectionPool.init(ConnectionPool.java:464) [tomcat-jdbc-8.0.28.jar:na]`) and that wouldn't happen if the initial size was 0.

PR #1517, opened 2016-04-13
Issue 1037

arienkock:
#1037 Simple fix to make it possible to instantiate an empty connection pool. For cases where you still want the webapp to load, e.g. you want a service to be able to start and show an error page when a DB isn't available.

Includes tests. Please ignore the changes to `travis.yml`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ sudo: false
 language: java
 jdk:
   - oraclejdk8
-#after_success:
-#  - bash .travis_after_success.sh
+after_success:
+  - bash .travis_after_success.sh
 cache:
   directories:
     - $HOME/.m2
@@ -18,7 +18,7 @@ cache:
 # NOTE: CI_DEPLOY_USERNAME is set to dropwizardci, the username
 # we've set up with Sonatype, which only has permission to push
 # to the snapshot repo.
-#env:
-#  global:
-#    - secure: "EAuz7bCKj4r438IEC2y73WVrFwXirflbXA4HhpwVmAFWNqC9LIIpkWcO5GVp773HsZvJBcJjJriP+aKRkImV8AyMgjCeEUv2dlezvbkIIz38vKyw9MWaPIyZ3uS9RCuL7OwGf5BeJ1DvHFYdMBaspZd+EmCYr7abnHdqs+Tm/W8="
-#    - secure: "RI0QcuKMsij3sgRm+Bjhu3X217U6UslvSzcRv13iLLwrTj73zhGi5PF/+kj8Qh1HMQw0oQRR6M8qPqGy82KcjiGbpgPgSy1rVAvkYg+Yw1k7v4l7Vgyj7TNsAM3pqHyojx2jgRjpQsgw/WXQmiahWV6OCOUzdbhEUwVzIXI+vtk="
+env:
+  global:
+    - secure: "EAuz7bCKj4r438IEC2y73WVrFwXirflbXA4HhpwVmAFWNqC9LIIpkWcO5GVp773HsZvJBcJjJriP+aKRkImV8AyMgjCeEUv2dlezvbkIIz38vKyw9MWaPIyZ3uS9RCuL7OwGf5BeJ1DvHFYdMBaspZd+EmCYr7abnHdqs+Tm/W8="
+    - secure: "RI0QcuKMsij3sgRm+Bjhu3X217U6UslvSzcRv13iLLwrTj73zhGi5PF/+kj8Qh1HMQw0oQRR6M8qPqGy82KcjiGbpgPgSy1rVAvkYg+Yw1k7v4l7Vgyj7TNsAM3pqHyojx2jgRjpQsgw/WXQmiahWV6OCOUzdbhEUwVzIXI+vtk="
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -2,8 +2,8 @@ sudo: false
 language: java
 jdk:
   - oraclejdk8
-after_success:
-  - bash .travis_after_success.sh
+#after_success:
+#  - bash .travis_after_success.sh
```
nickbabcock: I'm assuming these lines are commented out by mistake, right? :smile:

arienkock: Just to make CI work on my branch. As mentioned in the comments, changes to this file can be ignored.

dropwizard-db/src/test/java/io/dropwizard/db/DataSourceFactoryTest.java
```diff
@@ -47,6 +48,18 @@ private ManagedDataSource dataSource() throws Exception {
     }
 
     @Test
+    public void testInitialSizeIsZero() throws Exception {
+        factory.setUrl("nonsense invalid url");
+        factory.setInitialSize(0);
+        ManagedDataSource dataSource = factory.build(metricRegistry, "test");
+        try {
+            dataSource.start();
```
joschi: Surrounding the statement with try-catch doesn't gain anything. The test will fail anyway, if any uncaught exception occurs.

dropwizard-db/src/test/java/io/dropwizard/db/DataSourceConfigurationTest.java
```diff
@@ -107,6 +109,17 @@ public void testInlineUserPasswordConfiguration() throws Exception {
         assertThat(ds.getUser()).isNull();
         assertThat(ds.getPassword()).isNull();
     }
+    // Test added for https://github.com/dropwizard/dropwizard/issues/1037
```
joschi: Please remove that comment.

dropwizard-db/src/test/java/io/dropwizard/db/DataSourceConfigurationTest.java
```diff
@@ -107,6 +109,17 @@ public void testInlineUserPasswordConfiguration() throws Exception {
...
+    @Test
+    public void testInitialSizeZeroIsAllowed() throws Exception {
+        try {
+            DataSourceFactory ds = getDataSourceFactory("yaml/empty_initial_pool.yml");
```
joschi: Surrounding the statement with try-catch doesn't gain anything. The test will fail anyway, if any uncaught exception occurs.

.travis.yml
nickbabcock: Ah yes, but if you do it for this pull request, it as no chance of being accidentally merged and saves us time. We're lazy too!

dropwizard-db/src/test/java/io/dropwizard/db/DataSourceConfigurationTest.java
arienkock: Failing in the catch case makes it possible to distinguish between expected behavior (assertion exceptions) and unexpected problems (e.g. build env stuff)

joschi: In general I'd agree, but that argument is kind of moot in a test consisting of a single statement (and one statement for assertions).

arienkock:
In case it was lost in the exchange: the travis config on my branch was reverted as suggested.

nickbabcock:
This looks good. There are a few suggestions in the code review that should be applied.

Do you also have a link to documentation on this behavior for tomcat's jdbc pool for future reference?

arienkock:
Sure. From the Javadoc of the ConnectionPool constructor: [Instantiate a connection pool. This will create connections if initialSize is larger than 0.](https://tomcat.apache.org/tomcat-8.0-doc/api/org/apache/tomcat/jdbc/pool/ConnectionPool.html#ConnectionPool(org.apache.tomcat.jdbc.pool.PoolConfiguration))

Just pushed changes w.r.t. code review.

nickbabcock:
Did you miss one of the code review items?

arienkock:
In the last commit I removed the try catch and code comment before the unit test.

dropwizard-db/src/test/java/io/dropwizard/db/DataSourceFactoryTest.java
nickbabcock: @arienkock what about this?

nickbabcock:
And just so I understand clearly, those that are using Dropwizard and JDBI -- the [DBIHealthCheck](https://github.com/dropwizard/dropwizard/blob/144fe0cc49b14f9cb279c2d794baa1c74b8bec61/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIHealthCheck.java), which needs a connection to the database, will fail, and this is the desired result. The application will still start, but will essentially be in an unhealthy state.

arienkock:
Correct, the argument being that if the app can _run_ without a DB (and be in an unhealthy state) why would it not be able to _start_. Useful in cases where you know a DB connection is of bad quality and you don't want a deployment to fail.

* I just removed the other try-catch from the unit test

PR #1527, opened 2016-04-19
Jackson Invalid Format Exception is 400 response

nickbabcock:
Previously if a client would supply a string value where an Integer should
be, for example, a 500 error would be generated. This fix will now return
a 400 for all invalid conversions from JSON to Java Object.

As discovered by @mikewatt in #1347

CC'ing others from #1347: @jplock, @brentryan

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -3,7 +3,6 @@ package io.dropwizard.jersey.jackson;
 import com.fasterxml.jackson.core.JsonGenerationException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonMappingException;
-import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -12,6 +11,7 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
+import java.util.List;
 
 @Provider
 public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProcessingException> {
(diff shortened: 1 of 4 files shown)


jplock:
Makes sense to me 👍

mikewatt:
Hi Nick,

Thanks for quickly taking a look at this. I think there are still cases where deserializers throw JsonMappingException (not subclasses) which would still result in a 500. For example, JSR310DeserializerBase. 

I'm not all that familiar with the code, but it seems pretty hard to distinguish a serialization exception from a deserialization exception given a JsonMappingException. Possibly checking if exception.getLocation() is non-null, but that seems a bit dodgy.

I think for our case, we will end up implementing a custom ExceptionMapper and looking at the exception.getCause() to determine the response.

nickbabcock:
@mikewatt, I'm not going to give up just yet :smile:

I just pushed a change that will handle the issue with JSR310DeserializerBase by getting the root cause. But as we've seen previously this may not work in 100% of use cases. If you find additional cases, let me know, but I am also at a loss if there is an official way to determine when it's a programmer vs client error, as this relates to not only deserialization but also serialization (ie, the object I'm serializing as a `GET` response can't construct JSON somehow)

In lieu of a better solution, I may defer to @cowtowncoder if he's available. How to determine, given a `JsonProcessingException` whether it's a programmer or client error?

cowtowncoder:
@nickbabcock I would look into subtypes of `JsonProcessingException`. There are low-level problems with broken (non-well-formed) JSON (`JsonParsingException` I think), `JsonMappingException` for databinding (with more subtypes). There has been some on-going work into further dividing kinds of mapping exceptions into things like "invalid format for this datatype" (`InvalidFormatException`), "property name doesn't ring a bell" (`InvalidFormatException`), which may help. It's on-going battle, however, since handlers (deserializers) need to throw appropriate exceptions.

mikewatt:
Sounds like the most appropriate fix (in this specific case) is changing the JSR310 deserializers to throw InvalidFormatException as necessary. With the original commit in this pull request, the exception mapper is probably doing the best it can.

nickbabcock:
Thanks for the input, Tatu, greatly appreciated. That's what I thought too, but then I came up with a two pronged approach that seems to handle everything I throw at it.

- Check if we're looking at an `JsonMappingException`
- If so, get the last field that caused the error.
- If the field is `null` then we know that it was not the contents of the field but the type of the field that caused the error and thus a server error.

Is there an exception to this rule? I couldn't find one when testing the jsr310 module. I'm optimistic :smile:

cowtowncoder:
@nickbabcock That sounds reasonable at high level, although given that production of exceptions is initiated by individual deserializers it is likely there is variation. But I can and want to help there, so bug reports/PRs would be valued for cases where things do not work as well as they should.

@mikewatt would be great to get a PR or just pointers to cases that do not work that way, would be happy to fix/improve them. I am close to releasing 2.7.4, so getting fixes in soon would be great.

PR #1529, opened 2016-04-20
fix link to issue in release notes

stve:
Just a small formatting fix I found while looking at the release notes for v1.0.0. 

Here's the before and after:
image 2016-04-20 at 11 07 28 am


joschi:
Thanks!

Issue #1475, opened 2016-03-09
Client code generator


jansoren:
Hi,
I'm struggling to find a client code generator for Dropwizard apis that generates java code. Does this exist, or should we create one?

nickbabcock:
Sorry, I'm not familiar with such a tool. If you find a tool or make one yourself, we'll be more than happy to add it to the docs.

jansoren:
Just finished creating the code generator and refactored to only use jersey as you suggested @nickbabcock. You can find the plugin here - [REST API Codegenerator Maven Plugin](https://jansoren.github.io/restapi-codegen-maven-plugin/)

nickbabcock:
Cool, is the plugin on maven central or another site? I tried installing the plugin locally as well and ran into the following error:

> 'No mojo definitions were found for plugin: no.jansoren:restapi-codegen-maven-plugin.'

Some current questions (it's ok if not implemented, just curious):

- Query/Form/Path/Cookie parameters?
- HTTP headers?
- Basic Authentication?
- Non-JSON endpoints?

jansoren:
The plugin is not published on maven central yet. 
For now I've only covered the basics, and will later expand the plugin as I work on my application. Hopefully someone will create some pull requests :-)

I do not understand why you get that error. I just cleaned my `.m2` folder, cloned the plugin and ran `mvn clean install`. Did you add the plugin without building the jar locally?

jansoren:
The [REST API Codegenerator Maven Plugin](https://github.com/jansoren/restapi-codegen-maven-plugin) now generates ReactJS code aswell 🙂

PR #1535, opened 2016-05-04
Implemented Dropwizard Codegenerator Maven Plugin #1475

jansoren:
This implementation is hopefully the beginning of a great code-generator for Dropwizard API's. I would appreciate a code review.

You can find a working example here - https://github.com/jansoren/akka-persistence-java-example/

Feedback on the example would also be appreciated :+1:


jansoren:
@nickbabcock - as you mentioned in #1475 I have now started implementing the code-generator.

nickbabcock:
I think this is a fantastic start, but I have questions/concerns/suggestions

1. This should be developed outside the core repo because it is a plugin and should not be tied to a specific version of Dropwizard. This way the plugin can also be developed a quicker rate.
2. Looks like this can be extrapolated for any service that uses Jersey resources, no? There doesn't seem to be anything Dropwizard specific about the implementation.
3. I think it would be better to use Jersey's client instead of `no.bouvet.jsonclient` (sorry! :smile:) because as of right now the scope is too narrow (no query params, headers, authentication, non-json endpoints). This way, you can probably look at annotations and support more than just simple JSON endpoints.
4. How to handle [`StreamingOutput`](https://jersey.java.net/apidocs/latest/jersey/javax/ws/rs/core/StreamingOutput.html)?

jansoren:
Thanks for the great feedback :+1:  I will look into it.

PR #1538, opened 2016-05-09
Add an example of using the GraphiteReporter to the example application

jplock:
@postwait mentioned this on twitter, but the example application didn't attempt to send any metrics to Graphite.

### DIFF
--- a/dropwizard-metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporterFactory.java
+++ b/dropwizard-metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporterFactory.java
@@ -10,8 +10,8 @@ import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.google.common.annotations.VisibleForTesting;
 import io.dropwizard.metrics.BaseReporterFactory;
 import io.dropwizard.validation.OneOf;
+import io.dropwizard.validation.PortRange;
 import org.hibernate.validator.constraints.NotEmpty;
-import org.hibernate.validator.constraints.Range;
 
 import javax.validation.constraints.NotNull;
 
@@ -32,7 +32,7 @@ import javax.validation.constraints.NotNull;
  *     </tr>
  *     <tr>
  *         <td>port</td>
- *         <td>8080</td>
+ *         <td>2003</td>
  *         <td>The port of the Graphite server to report to.</td>
  *     </tr>
  *     <tr>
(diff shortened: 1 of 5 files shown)


postwait:
This works for me and in my derivative example.  Thanks!

dropwizard-metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporterFactory.java
```diff
@@ -32,7 +32,7 @@
  *     </tr>
  *     <tr>
  *         <td>port</td>
- *         <td>8080</td>
+ *         <td>2003</td>
```
jplock: I don't know why these were defaulting to `8080`, but I updated them to the standard graphite port of `2003`.

nickbabcock:
Am I the only one that finds it odd that the example application overrides the default `Configuration` metrics. Normally, I don't override the configuration so I don't have to set up the reporter, which, imo, is nicer.

jplock:
@nickbabcock I don't even use the reporters, so I'm not exactly sure the best way to do this.  I did notice that we define the graphite reporter in the config file at https://github.com/dropwizard/dropwizard/blob/master/dropwizard-example/example.yml#L94 but then to @postwait's point, nothing was starting it.

nickbabcock:
#1319 overrode the configuration (I merged it, funny enough).

But to add a (working) graphite reporter, one doesn't need to touch any java code:

- Add dependency to pom

```xml
<dependency>
    <groupId>io.dropwizard</groupId>
    <artifactId>dropwizard-metrics-graphite</artifactId>
</dependency>
```

- and then add the reporter to the yaml

```yaml
metrics:
  reporters:
    - type: graphite
      host: localhost
      port: 2003
      prefix: my-awesome-service
```

So at this point, I think the example application is _wrong_, unless @joschi remembers why it was done 8 months ago.

jplock:
@nickbabcock i went ahead and fixed the `example.yml` file.  How does this look?

dropwizard-metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporterFactory.java
joschi: Probably a copy & paste error. 😉

joschi:
LGTM. 👍 
I'll let @nickbabcock merge the PR since he's spent more time with it and should have the final say.

nickbabcock:
Ah, much better, no? :wine_glass: 

Mentioned port change in bf52889 because someone, somewhere is probably sending using the previous default port

jplock:
👍

Issue #1539, opened 2016-05-10
JSON deserialization failures should trigger a 400


vy:
When Jackson fails to deserialize a JSON object due to object instantiation, it fails with a 500. But I think this invalid formatting error should have triggered a 400 instead. Consider the following example:

```java
public class DemoModel {

    public URL url;

}

@Path("/demo")
@Consumes(MediaType.APPLICATION_JSON)
public class DemoResource {

    @POST
    public Response test(DemoModel model) {
        return Response.ok().build();
    }

}
```

Requesting this endpoint with an invalid URL input results in a 500:

```
curl -v http://localhost:8080/demo -H "Content-Type: application/json" -d '{"url": "no-scheme.com"}'
```

Relevant stack track is as follows:

```
java.net.MalformedURLException: no protocol: no-scheme.com
	at java.net.URL.<init>(URL.java:586)
	at java.net.URL.<init>(URL.java:483)
	at java.net.URL.<init>(URL.java:432)
	at com.fasterxml.jackson.databind.deser.std.FromStringDeserializer$Std._deserialize(FromStringDeserializer.java:212)
	at com.fasterxml.jackson.databind.deser.std.FromStringDeserializer.deserialize(FromStringDeserializer.java:122)
	at com.fasterxml.jackson.databind.deser.SettableBeanProperty.deserialize(SettableBeanProperty.java:490)
	at com.fasterxml.jackson.databind.deser.impl.FieldProperty.deserializeAndSet(FieldProperty.java:101)
	at com.fasterxml.jackson.module.afterburner.deser.SuperSonicBeanDeserializer.deserialize(SuperSonicBeanDeserializer.java:156)
	at com.fasterxml.jackson.databind.ObjectReader._bind(ObjectReader.java:1535)
	at com.fasterxml.jackson.databind.ObjectReader.readValue(ObjectReader.java:944)
	at com.fasterxml.jackson.jaxrs.base.ProviderBase.readFrom(ProviderBase.java:811)
	at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$TerminalReaderInterceptor.invokeReadFrom(ReaderInterceptorExecutor.java:256)
	at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$TerminalReaderInterceptor.aroundReadFrom(ReaderInterceptorExecutor.java:235)
[...]

PR #1540, opened 2016-05-10
Treat Jackson ObjectMapper#readValue failures as a 400

vy:
See #1539 for the related issue.

### DIFF
--- a/dropwizard-jersey/src/test/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapperTest.java
+++ b/dropwizard-jersey/src/test/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapperTest.java
@@ -5,14 +5,12 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
 import io.dropwizard.jersey.DropwizardResourceConfig;
-import io.dropwizard.jersey.validation.Validators;
 import io.dropwizard.logging.BootstrapLogging;
 import org.glassfish.jersey.client.ClientConfig;
 import org.glassfish.jersey.test.JerseyTest;
 import org.glassfish.jersey.test.TestProperties;
 import org.junit.Test;
 
-import javax.validation.Validator;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.Application;
 import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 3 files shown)


dropwizard-jersey/src/test/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapperTest.java
```diff
@@ -40,10 +38,10 @@ protected void configureClient(ClientConfig config) {
     }
 
     @Test
-    public void returnsA500ForNonDeserializableRepresentationClasses() throws Exception {
+    public void returnsA400ForNonDeserializableRepresentationClasses() throws Exception {
         Response response = target("/json/broken").request(MediaType.APPLICATION_JSON)
                 .post(Entity.entity(new BrokenRepresentation(ImmutableList.of("whee")), MediaType.APPLICATION_JSON));
-        assertThat(response.getStatus()).isEqualTo(500);
+        isA400ForNonDeserializableRequest(response);
```
nickbabcock: Why should this now be a 400?

vy: Because a non-deserializeable input means "bad content", not a "service error"?

nickbabcock: Nope, not in this case. It should be a 500 error because the client meets the input spec, except the server is unable to deserialize it because of the [BrokenRepresentation](https://github.com/dropwizard/dropwizard/blob/bf528890def6be0394f038be98282c541dab689f/dropwizard-jersey/src/test/java/io/dropwizard/jersey/jackson/BrokenRepresentation.java) (ie, the BrokenRepresentation can be fixed with a `@JsonCreator`, for instance)

vy: I am not 100% happy with the `catch (Throwable cause)` approach, but I thought it would be better to come up with a patch to start the discussion.

So I understand your point: a BrokenRepresentation should be a 500. On the other hand, I want my bogus URLs (see #1539) to be treated as 400. Any suggestions on how to differentiate them?

situokko:
If the issue is with URL-types, maybe use String field for the url instead, and validate proper url-string with regexp validation annotation and add separate helper-getter that returns the string as URL-object?

vy:
@situokko, no, the issue is not about URL-types -- it was just for demonstration purposes -- but about how shall we treat non-JsonProcessingException errors thrown by ObjectMapper#readValue(), i.e., the deserializer. I mean, when the user input does not result in a valid instance instantiation, should it really be the server's fault? And where/how shall we draw the line?

situokko:
@vy In a way I agree with you, I would want my server to return 500 only in cases where there is bug in my server code. Now it is very easy to get 500 just by sending invalid requests (invalid JSON contents or field values). 

Another way to think of it as bad design to have such data-types for JSON-object that throw exceptions when constructed with illegal content, such as invalid URL.

Currently I am using separate "Model" (data) Pojos for server's internal handling and converting them to "View" Pojos that have more or less the same content, but in just the right format for JSON serialization&deserialization. It causes me to have twice as many classes (=> redundancy), but in the end less hassle.

PR #1542, opened 2016-05-10
Update documentation on hibernate resource method example

sreyseng:
Add Path annotation, as method expects PathParam
(will get NullPointerException otherwise)


jplock:
Thanks!

PR #1543, opened 2016-05-10
add support for customizing the name of the migrations file

tburch:
I have a use case where I'd like to run multiple `DropwizardAppRule` instances in an integration test and each DropwizardAppRule is from a different fat jar dependency. Each DropwizardAppRule will need to migrate its own database, however the `migration.xml` files collide because they have the same name. This PR allows the migrations filename to be set in the `MigrationsBundle` by overriding `getMigrationsFileName()`, which defaults to `migrations.xml`.


nickbabcock:
Makes sense to me, thanks for the feature :+1:

PR #1549, opened 2016-05-19
Allow file appender infinite log files

### LINKED ISSUE #1491: FileAppenderFactory doesn't support infinite archive option
`FileAppenderFactory` has

```java
    @Min(1)
    private int archivedFileCount = 5;
```

`TimeBasedRollingPolicy` specifies 

```java
  static final int INFINITE_HISTORY = 0;
```

But I can never set infinite history due to the constraint.

nickbabcock:
Closes #1491

I think there is no reason not to allow dropwizard users the ability to keep infinite files as long as it is warned and the default remains something sensible.

This also fixes the erroneous description that the number of archived files have a maximum (they don't)


joschi:
LGTM. 👍

PR #1551, opened 2016-05-20
Add CacheBuilderSpecSerializer

tburch:
If a configuration has a default value for a CacheBuilderSpec (e.g. `CacheBuilderSpec cacheBuilderSpec = CacheBuilderSpec.disableCaching();`) and the value is not overridden in the configuration file, we need to serialize the object before parsing the configuration so that it can be properly instantiated in the `Configuration` class.


joschi:
LGTM. Thanks! 👍

PR #1552, opened 2016-05-21
Add customizable constructors to JodaDateTimeMapper

### LINKED ISSUE #1548: dropwizard-jdbi: JodaDateTimeMapper needs index- and name-based constructors to map columns other than the 1st one
the JodaDateTimeMapper is derived from TypedMapper which has a default constructor that uses column 1 for its mapping.

Our use case needs to map more than 1 column via a JodaDateTimeMapper, which won't work with the current code.

I'd suggest the only things necessary are the following constructor signatures:
```
    public JodaDateTimeMapper() {
// to prevent breaking existing usages.
    }

    public JodaDateTimeMapper(int index) {
        super(index);
    }

    public JodaDateTimeMapper(String name) {
        super(name);
    }

```

joschi:
Closes #1548


nickbabcock:
Looks good. I'll merge this. Not sure how we want to mark this because it only affects < 1.0

PR #1554, opened 2016-05-23
Permit override of default failed command behavior

thomaslee:
The current behavior of configuration errors etc. in `Application.run(String...)` is to call `System.exit(1)`. This can play havoc with things like integration tests, such that a hard exit will terminate the testing process.

This change allows folks to override the default behavior to do something other than the `System.exit(1)` call.


joschi:
Thanks! LGTM. 👍

PR #1555, opened 2016-05-23
Update README.md

mabuthraa:
update jar name as defined in pom.xml:    <version>1.0.0-rc3-SNAPSHOT</version>, otherwise jar wont be found!


joschi:
Thanks!

PR #1556, opened 2016-05-23
Add Maven profiles for selecting alpn-boot version

joschi:
The version of `alpn-boot` is dependent on the version of the JVM being used, otherwise the build will fail.

Since not everyone is using the same version of the JVM, this PR adds profiles to `dropwizard-http2` for selecting the correct version of `alpn-boot`.

This is only required for running tests locally and on Travis CI.

### DIFF
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -290,12 +290,6 @@
                     </exclusion>
                 </exclusions>
             </dependency>
-            <dependency>
-                <groupId>org.mortbay.jetty.alpn</groupId>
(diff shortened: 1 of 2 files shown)


joschi:
@nickbabcock Yes, but that's the status quo anyway, e. g. I can choose whether I want broken builds for dropwizard-http2 on my local machine with Java 8u92 or on Travis CI with Java 8u31.

This change benefits only developers of Dropwizard as the only place where we use alpn-boot is in the tests for dropwizard-http2 (and dropwizard-spdy in the past).

Issue #1557, opened 2016-05-24
ServerCommand.run causes NullPointerException in catch block


BusComp:
see io.dropwizard.cli.ServerCommand#run, line 46
`server.stop()` cause NullPointerException when previous start fails.
```
protected void run(Environment environment, Namespace namespace, T configuration) throws Exception {
        final Server server = configuration.getServerFactory().build(environment);
        try {
            server.addLifeCycleListener(new LifeCycleListener());
            cleanupAsynchronously();
            server.start();
        } catch (Exception e) {
            LOGGER.error("Unable to start server, shutting down", e);
            server.stop();
            cleanup();
            throw e;
        }
    }
```
Start fails; jetty not started
```
2016-05-24T13:47:00.642+0200  [main] ERROR i.d.cli.ServerCommand: Unable to start server, shutting down
org.eclipse.jetty.util.MultiException: Multiple exceptions
	at org.eclipse.jetty.server.Server.doStart(Server.java:347)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
	at io.dropwizard.cli.ServerCommand.run(ServerCommand.java:43)
	at io.dropwizard.cli.EnvironmentCommand.run(EnvironmentCommand.java:41)
	at io.dropwizard.cli.ConfiguredCommand.run(ConfiguredCommand.java:76)
	at io.dropwizard.cli.Cli.run(Cli.java:75)
	at io.dropwizard.Application.run(Application.java:79)
...
Caused by: java.sql.SQLRecoverableException: I/O-Fehler: The Network Adapter could not establish the connection
	at oracle.jdbc.driver.T4CConnection.logon(T4CConnection.java:419)
	at oracle.jdbc.driver.PhysicalConnection.<init>(PhysicalConnection.java:536)
	at oracle.jdbc.driver.T4CConnection.<init>(T4CConnection.java:228)
	at oracle.jdbc.driver.T4CDriverExtension.getConnection(T4CDriverExtension.java:32)
```
Within the catch block `server.stop()` is invoked. The root cause is replaced by an NullPointerException and `cleanup();` will never be reached...
```
2016-05-24T13:47:00.662+0200  [main] INFO  o.e.j.s.ServerConnector: Stopped application@6644bdf5{HTTP/1.1,[http/1.1]}{0.0.0.0:7060}
Exception in thread "main" java.lang.NullPointerException
[...]

PR #1560, opened 2016-05-25
Surrounded server.stop() and cleanup with a try/catch block, see #1557

BusComp:
small fix for Issue#1557, feel free to adapt/modify this bugfix


joschi:
Thanks! LGTM. 👍

PR #1562, opened 2016-05-26
Fix deprecation notices in http2 servers

nickbabcock:
The latest bump in jetty [deprecated some functions](https://github.com/eclipse/jetty.project/blob/c99c02e2f59cc4c65cc9b893710e48eeeb3bef0b/jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/AbstractHTTP2ServerConnectionFactory.java#L109-L117)


joschi:
@nickbabcock @jplock Shouldn't we rename the corresponding configuration setting (https://github.com/nickbabcock/dropwizard/blob/bbdd2fa3a4265b88096ab6f85bcfb1df7fc037a3/dropwizard-http2/src/main/java/io/dropwizard/http2/Http2CConnectorFactory.java#L59-L61) too in this case?

nickbabcock:
Good catch. Yes it should change. If you want, go ahead and make the change, else I can have a PR in an hour.

PR #1565, opened 2016-05-27
General Code Improvements

nickbabcock:
Using Intellij automated tools refactored some of the code. No behavior was changed (ie, no streams)

- Replace tabs with spaces
- Remove redundant type operations
- Prefer `Optional.empty` to `Optional.ofNullable(null)`
- Prefer try-with-resources
- Prefer lambdas to anonymous classes
- Simplify if conditionals
- Optimize imports


joschi:
The build on Travis CI fails with (https://travis-ci.org/dropwizard/dropwizard/builds/133273498#L4421-L4423):
> ```
[ERROR] /home/travis/build/dropwizard/dropwizard/dropwizard-views/src/main/java/io/dropwizard/views/ViewBundle.java:[112,48] incompatible types: inferred type does not conform to upper bound(s)
[ERROR] inferred: java.util.Map<? extends java.lang.Object,? extends java.lang.Object>
[ERROR] upper bound(s): java.util.Map<java.lang.String,java.lang.String>,java.lang.Object
```

Additionally, the `*WhenPresentButNull` tests in `GuavaOptionalValidatedValueUnwrapperTest` and `OptionalValidatedValueUnwrapperTest` are now identical to the `*WhenAbsent`/`*WhenEmpty` tests and should be removed (they were testing `Optional` to do the right thing with `null` values instead of the value unwrapper anyway).

nickbabcock:
Very bizarre that the build fails on Travis, it passes on my machine. May have to do with the jdk version (I'm using jdk1.8.0_66). If it fails again on Travis I'll remove the trouble commits

nickbabcock:
Yeah looks the old version of the JDK on travis is giving some headaches:

https://github.com/travis-ci/travis-ci/issues/3259
https://github.com/travis-ci/travis-ci/issues/4042

jplock:
This workaround may help https://github.com/openzipkin/brave/commit/0efe6cc5327fe8e5441b898e15d0731463ed6f12

joschi:
@nickbabcock I wouldn't invest too much time into making a more recent JDK usable in Travis CI and revert the offending commit instead.

joschi:
LGTM. 👍

Issue #1563, opened 2016-05-26
Enhancement to Lifecycle Setup Classes


isaki-x:
Greetings,

I noticed that the executor service factories provided in LifecycleEnvironment do not allow you to specify a ThreadFactory directory; instead you must invoke "threadFactory" on the returned builder object. However, the builder creates a default thread factory in its constructor which is waste of time and resources if the code has a ThreadFactory implementation it wants to use instead of your defaults.

I've forked your codebase and created the simplest patch possible to create the functionality I am asking for (which is to specify the thread factory up front to save startup time). This simply adds a constructor the executor builders that allows specifying a ThreadFactory (the existing constructors invoke these new constructors) and exposes this via two new factory methods in LifecycleEnvironment.

Ideally, the thread factory should only be created on the invocation of "build" but this requires a bit of work and may change the behavior for existing users who reuse builders. As such, I opted for the light touch approach specified above.

- isaki-x

joschi:
Closed in #1570.

PR #1570, opened 2016-05-28
Enhancements to lifecycle-setup classes.

isaki-x:
- Added constructors for executor service builders to allow specifying a
thread factory so that a thread factory is not built only to be thrown
away by a later call to the method to set the thread factory; the
ability to override the thread factory has not been removed for backward
compatibility.
- Exposed factory methods that create the builders based on already
existing thread factories, thus invoking the new constructors created in
the above comment.
- Issue #1563


isaki-x:
#1563

nickbabcock:
Looks like Travis is not happy with this PR.

isaki-x:
Oddly enough Eclipse happily builds the original pull request and passes all the unit tests (I just double checked to make sure I hadn't overlooked an error). Regardless, I have pushed a fix.

nickbabcock:
In the linked issue, it's mentioned that the current method:

> is [a] waste of time and resources

Can this be substantiated in any way. I realize that objectively it can be excessive but I'm not sure how tangible this benefit will be. This is more for my curiosity than anything

joschi:
LGTM. :+1:

PR #1571, opened 2016-05-29
Enable lazy-loading of ConfigOverride values via Supplier

### LINKED ISSUE #1532: Allow lazy loading of ConfigOverride values.
This code below is a current workaround the fact that values must be known ahead of time. We are running into an issue where we want to have port numbers be provided dynamically, but they won't be known ahead of time.

```
    static ConfigOverride dynamicConfig(String name, Supplier<String> valueSupplier) {
        ConfigOverride override = Mockito.mock(ConfigOverride.class);

        Mockito.doAnswer((inv)-> {
            ConfigOverride.config(name, valueSupplier.get()).addToSystemProperties();
            return null;
        }).when(override).addToSystemProperties();

        Mockito.doAnswer((inv)-> {
            ConfigOverride.config(name, valueSupplier.get()).removeFromSystemProperties();
            return null;
        }).when(override).removeFromSystemProperties();

        return override;
    }
```

We could easily change the current implementation of ConfigOverride to allow a Supplier instead of a value to work around this specific issue.

It seems like the ConfigOverride implementation isn't really great in terms of changing global system properties rather than having better scoping.

choliver:
Resolves #1532

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/ConfigOverride.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/ConfigOverride.java
@@ -1,7 +1,6 @@
 package io.dropwizard.testing;
 
-import com.google.common.base.Supplier;
-import com.google.common.base.Suppliers;
+import java.util.function.Supplier;
 
 public class ConfigOverride {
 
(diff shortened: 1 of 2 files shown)


nickbabcock:
Can you switch the implementation to Java 8's supplier? We're trying to be Java 8 first :smile:

nickbabcock:
Looks good to me :+1:

PR #1572, opened 2016-05-29
Add Resource Bean Validation

pkwarren:
Update ConstraintMessage.isRequestEntity to be resilient against
exceptions by using the `Iterables.get` method which accepts a default
value. This is required in case the property path contains fewer than 2
items.

### DIFF
--- a/dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapperTest.java
+++ b/dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapperTest.java
@@ -533,4 +533,16 @@ public class ConstraintViolationExceptionMapperTest extends JerseyTest {
         assertThat(response.readEntity(String.class))
                 .containsOnlyOnce("examples may not be empty");
     }
+
+    @Test
+    public void testInvalidFieldQueryParam() {
+        final Response response = target("/valid/bar")
+            .queryParam("sort", "foo")
+            .request()
+            .get();
+
+        assertThat(response.getStatus()).isEqualTo(422);
+        assertThat(response.readEntity(String.class))
+            .containsOnlyOnce("sortParam must match \\\"^(asc|desc)$\\\"");
+    }
 }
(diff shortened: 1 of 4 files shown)


nickbabcock:
In situations like these, I find a test much more power if the bug can be exposed by adding or modifying an endpoint in [ValidatingResource](https://github.com/dropwizard/dropwizard/blob/ceff3051889b38ac02856d2a7b231360abde76a3/dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ValidatingResource.java) and the corresponding test case in [ConstraintViolationExceptionMapperTest](https://github.com/dropwizard/dropwizard/blob/ceff3051889b38ac02856d2a7b231360abde76a3/dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapperTest.java) because it adds a real world flair to bug.

pkwarren:
@nickbabcock - Thanks for the pointer - I didn't look around enough to find the right home for the test. I found this while porting some test code from 0.9.2 to 1.0.0-rc2 and so I'm not sure if it is reproducable in the wild - it is more a correctness fix similar to the existing lines found here: https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java#L115-L118

I was able to come up with an example which failed, but it required enabling validation for the resource itself (instead of just the method parameters). I don't know if this is desired behavior (although it could be useful).

dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapperTest.java
```diff
@@ -533,4 +533,14 @@ public void invalidEmbeddedListEntities() {
...
     }
+
+    @Test
+    public void testInvalidFieldQueryParam() {
+        final Response response = target("/valid/bar")
+            .queryParam("sort", "foo")
+            .request()
+            .get();
+
+        assertThat(response.getStatus()).isEqualTo(422);
```
nickbabcock: Can you add an additional assertion about the response message

pkwarren: Updated.

nickbabcock:
Looks good to me. I renamed the PR to represent the added feature :smile:

Merging this now, yet some additional changes should be done in this area:

- `"sortParam must match \\\"^(asc|desc)$\\\"` should be `"query param sort must match \\\"^(asc|desc)$\\\"`
- May need additional checks for possible mismatch in validation groups between the `Invocable` and bean members

PR #1578, opened 2016-06-01
Remove @Ignore from DropwizardApacheConnectorTest

nickbabcock:
I rebased this branch on the master and re-ran it on Travis twice and it passed (twice!).

So here's to being optimistic :tada:


jplock:
Tests passed, so I guess let's go with it for now.  👍

PR #1582, opened 2016-06-03
Add note to Doc : migrate about Migrating with DB firebird

mabuthraa:
FireBird will throw SQLException("Validation Query Failed, enable logValidationErrors ...").. To fix that its necessary to set  validationQuery to  "select 1 from rdb$database"


jplock:
@dropwizard/committers does anyone else have any feedback on this one? I know its just a documentation update, but do we want to be calling out db-specific information in here?  I wasn't sure if there was a better way to present this information.

nickbabcock:
I find it ok, if documenting exceptions starts to proliferate and become the norm, we may try and figure out a better way to document these.

PR #1583, opened 2016-06-03
Remove OptionalValidatedValueUnwrapper

nickbabcock:
As now that Hibernate validator implements it internally, ours is no
longer needed (or even used, as Hibernate prefers its own).

I'm marking this as for the 1.1, let me know if anyone disagrees (ie. don't merge this until at least 1.0 is out (or create 1.1 branch?))


joschi:
Good idea, thanks! 👍

Issue #1584, opened 2016-06-03
make HealthCheckRegistry pluggable


gjesse:
Similar to how the ObjectMapper and MetricsRegistry can be customized/swapped in the bootstrap, I would like to be able do the same with HealthCheckRegistry.

We are attempting add some additional information to the existing health check output, and under the current setup it requires us to pass around a separate registry, which is cumbersome.

jplock:
Will be fixed in 1.0.0 (or the next 1.0.0-rc release)

PR #1585, opened 2016-06-03
allow setting a custom health check registry per #1584

gjesse:


jplock:
Looks good to me.  We're reaching the point where `Environment` should probably have a builder-style interface due to all of the constructor parameters, but I think this is fine for now.

PR #1590, opened 2016-06-09
+ take the tcp backlog from OS

tootedom:
Hi there.  

I've noticed that the acceptQueueSize of the http connector, does not take the OS setting as detailed in the documentation:

http://www.dropwizard.io/0.9.2/docs/manual/configuration.html
````
acceptQueueSize	(OS default)	The size of the TCP/IP accept queue for the listening socket.
````

Meaning that the tcp backlog on the socket isn't that of the OS setting (somaxcon).

When you do not specify the `acceptQueueSize` on the `ServerConnector`, Jetty is reverting to the default JDK setting as set in the `java.net.ServerSocket`, example (jdk8 - I check jdk6 has the same default of 50):
````
    public ServerSocket(int port, int backlog, InetAddress bindAddr) throws IOException {
        setImpl();
        if (port < 0 || port > 0xFFFF)
            throw new IllegalArgumentException(
                       "Port value out of range: " + port);
        if (backlog < 1)
          backlog = 50;
        try {
            bind(new InetSocketAddress(bindAddr, port), backlog);
        } catch(SecurityException e) {
            close();
            throw e;
        } catch(IOException e) {
            close();
            throw e;
        }
    }

````

This is currently the case for our dropwizard apps in production (I performed an strace to 100% check this was the case - `strace -e trace=network -f `  - an `ss -lt` will also show it you)

````
[pid  9831] bind(56, {sa_family=AF_INET6, sin6_port=htons(6000), inet_pton(AF_INET6, "::", &sin6_addr), sin6_flowinfo=0, sin6_scope_id=0}, 28) = 0
[pid  9831] listen(56, 50) 
````

This pull requests uses the same logic that Netty uses, to read the somaxconn setting from the environment, and set the `acceptQueueSize` to the OS value if readable.  or the defaults of 128 for *nix systems and 200 for windows.

Let me know if you need more information.
thanks
/dom

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2012 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.dropwizard.jetty;
+
+import java.io.*;
+import java.nio.charset.Charset;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Locale;
+
+/**
+ * This class is taken from the Netty project, and all credit goes to them.
+ * It has been modified, to remove dependencies on other classes, and to convert to methods, rather than a
+ * static value.
+ */
+public class NetUtil {
+    public static final int DEFAULT_TCP_BACKLOG_WINDOWS = 200;
+    public static final int DEFAULT_TCP_BACKLOG_LINUX = 128;
+    public static final String TCP_BACKLOG_SETTING_LOCATION = "/proc/sys/net/core/somaxconn";
+
+
+
+    /**
+     * The SOMAXCONN value of the current machine.  If failed to get the value,  {@code 200}  is used as a
+     * default value for Windows or {@code 128} for others.
+     */
+    public static int getTcpBacklog() {
+        return getTcpBacklog(getDefaultTcpBacklog());
+    }
+
+    /**
+     * The SOMAXCONN value of the current machine.  If failed to get the value, <code>defaultBacklog</code> argument is
+     * used
+     */
+    public static int getTcpBacklog(int tcpBacklog) {
+        // Taken from netty.
+
+        // As a SecurityManager may prevent reading the somaxconn file we wrap this in a privileged block.
+        //
+        // See https://github.com/netty/netty/issues/3680
+        return AccessController.doPrivileged(new PrivilegedAction<Integer>() {
+            @Override
+            public Integer run() {
+                // Determine the default somaxconn (server socket backlog) value of the platform.
+                // The known defaults:
+                // - Windows NT Server 4.0+: 200
+                // - Linux and Mac OS X: 128
+                int somaxconn = tcpBacklog;
+                File file = new File(TCP_BACKLOG_SETTING_LOCATION);
+                BufferedReader in = null;
+                try {
+                    // file.exists() may throw a SecurityException if a SecurityManager is used, so execute it in the
+                    // try / catch block.
+                    // See https://github.com/netty/netty/issues/4936
+                    if (file.exists()) {
+                        in = new BufferedReader(new InputStreamReader(new FileInputStream(file), Charset.forName("UTF-8")));
+                        somaxconn = Integer.parseInt(in.readLine());
+                    }
+                } catch (SecurityException | IOException | NumberFormatException e) {
+                    // file.exists() may throw a SecurityException, in this
+                    // case we are just returning the default somaxconn that was passed in.
+                } finally {
+                    if (in != null) {
+                        try {
+                            in.close();
+                        } catch (Exception e) {
+                            // Ignored.
+                        }
+                    }
+                }
+                return somaxconn;
+            }
+        });
+
+    }
+
+    public static boolean isWindows() {
+        boolean windows = System.getProperty("os.name", "").toLowerCase(Locale.US).contains("win");
+        return windows;
+    }
+
+    public static int getDefaultTcpBacklog() {
+        return isWindows() ? DEFAULT_TCP_BACKLOG_WINDOWS : DEFAULT_TCP_BACKLOG_LINUX;
+    }
+}
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
@@ -16,8 +16,11 @@
 
 package io.dropwizard.jetty;
 
-import java.io.*;
-import java.nio.charset.Charset;
+import com.google.common.base.Charsets;
+import com.google.common.io.Files;
+
+import java.io.File;
+import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java [hunk 1]
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public static int getTcpBacklog(int tcpBacklog) {
+        // Taken from netty.
+
+        // As a SecurityManager may prevent reading the somaxconn file we wrap this in a privileged block.
+        //
+        // See https://github.com/netty/netty/issues/3680
+        return AccessController.doPrivileged(new PrivilegedAction<Integer>() {
+            @Override
+            public Integer run() {
```
nickbabcock: How do you feel about the following implementation:

```java
try {
    final File file = new File(TCP_BACKLOG_SETTING_LOCATION);
    return Integer.parseInt(Files.toString(file, Charsets.UTF_8));
} catch (SecurityException | IOException | NumberFormatException e) {
}

return tcpBacklog;
```

It's a bit shorter

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
```diff
@@ -0,0 +1,99 @@
...
+                // - Linux and Mac OS X: 128
+                int somaxconn = tcpBacklog;
+                File file = new File(TCP_BACKLOG_SETTING_LOCATION);
+                BufferedReader in = null;
+                try {
+                    // file.exists() may throw a SecurityException if a SecurityManager is used, so execute it in the
+                    // try / catch block.
+                    // See https://github.com/netty/netty/issues/4936
+                    if (file.exists()) {
+                        in = new BufferedReader(new InputStreamReader(new FileInputStream(file), Charset.forName("UTF-8")));
```
joschi: Why not simply use [`Files#readAllBytes()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#readAllBytes-java.nio.file.Path-) here?

tootedom:
Do you not then hit issues due to the SecurityManager, if you don't execute in the AccessController:
- https://github.com/netty/netty/issues/4936
- https://github.com/netty/netty/issues/3680

(Sorry I missed a test in HttpConnectorFactoryTest, I'm just adding that).

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
joschi: You can also use [`StandardCharsets#UTF_8`](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/StandardCharsets.html#UTF_8) here instead of `Charset.forName()`.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public static int getTcpBacklog() {
+        return getTcpBacklog(getDefaultTcpBacklog());
+    }
+
+    /**
+     * The SOMAXCONN value of the current machine.  If failed to get the value, <code>defaultBacklog</code> argument is
+     * used
+     */
+    public static int getTcpBacklog(int tcpBacklog) {
```
joschi: This whole thing will only work on Linux (or operating systems having `/proc/sys/net/core/somaxconn`), right?

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
```diff
@@ -0,0 +1,99 @@
...
+        });
+
+    }
+
+    public static boolean isWindows() {
+        boolean windows = System.getProperty("os.name", "").toLowerCase(Locale.US).contains("win");
+        return windows;
+    }
+
+    public static int getDefaultTcpBacklog() {
```
nickbabcock: How do you feel about condensing the two functions into:

```java
public static int getDefaultTcpBacklog() {
    return SystemUtils.IS_OS_WINDOWS ? DEFAULT_TCP_BACKLOG_WINDOWS : DEFAULT_TCP_BACKLOG_LINUX;
}
```

We already have a dependency on apache commons lang in core modules, so this shouldn't add any weight (but you'll need to reference the dependency in `dropwizard-jetty`'s pom)

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.dropwizard.jetty;
+
+import java.io.*;
```
joschi: Please don't use wildcard imports.

tootedom:
How about the following?

````
return AccessController.doPrivileged(new PrivilegedAction<Integer>() {
            @Override
            public Integer run() {
                // Determine the default somaxconn (server socket backlog) value of the platform.
                // The known defaults:
                // - Windows NT Server 4.0+: 200
                // - Linux and Mac OS X: 128
                int somaxconn = tcpBacklog;
                
                try {
                    final File file = new File(TCP_BACKLOG_SETTING_LOCATION);
                    somaxconn = Integer.parseInt(Files.toString(file, Charsets.UTF_8));
                } catch (SecurityException | IOException | NumberFormatException e) {
                    // file.exists() may throw a SecurityException, in this
                    // case we are just returning the default somaxconn that was passed in.
                } finally {
                    return somaxconn;
                }
            }
        });
````

File: dropwizard-jetty/src/test/java/io/dropwizard/jetty/NetUtilTest.java [hunk 6]
```diff
@@ -0,0 +1,81 @@
...
+
+    @Test
+    public void testOsSetting() {
+        if(!NetUtil.isWindows() && isTcpBacklogSettingReadable()) {
+            assertNotEquals(-1, NetUtil.getTcpBacklog(-1));
+        }
+    }
+
+
+    public boolean isTcpBacklogSettingReadable() {
```
joschi: You can use JUnit assumptions together with this method.

https://github.com/junit-team/junit4/wiki/Assumptions-with-assume

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
tootedom: Yes /proc/sys/net/core/somaxconn is only on *nix systems, but the method will return on systems without the /proc filesystem, returning default 128 for *nix systems (i.e. OSX) and 200 for windows

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
tootedom: sorry,will fix.

nickbabcock:
Haha, little bit chaotic with all the responses, I'll let @joschi do another sweep, but once everything is sorted out I think this would be a fine addition for 1.1

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
tootedom: Are you sure you want apache commons lang pulled into dropwizard-jetty pom.xml?

nickbabcock: I'll let @joschi make the call. It's the same implementation as [netty's](https://github.com/netty/netty/blob/56a2f64665494ee0d52fba7f391de0bea6144e29/common/src/main/java/io/netty/util/internal/PlatformDependent.java#L798-L804), so I'm ok with either.

joschi: I think it's fine as-is.

tootedom:
Thanks guys, will post updates in a few mins, based on the comments, just giving it all final review/test.

tootedom:
Hi there,
Hopefully I covered the comments.  But if I've missed anything, let me know.

thanks
/dom

joschi:
@tootedom LGTM. Thanks!

PR #1591, opened 2016-06-10
Remove a first sync call workaround in HTTP2 integration tests

arteam:
It looks like that the latest versions of Jetty fixed the bug which the inability to perform the first call asynchronously. Let's remove the workaround for this bug and keep tests more simple.


joschi:
LGTM. Thanks!

PR #1592, opened 2016-06-10
Document the Dropwizard release process

arteam:
This change proposes a simple draft of a document, which describes the technical aspects of making a release of Dropwizard. This information is mostly exist in the form emails between Dropwizard maintainers and personal experience, so it's quite easy for a maintainer to forget some steps during a release and make a mistake. It would be great, if we wrote it down in the repository and make accessible to the community. This should make Dropwizard a more transparent project and allow to improve our release process.

@dropwizard/committers, please give your feedback on this document. This is a rough draft which contains only the basic information. What other aspects we should address? Maybe something should be stripped out?

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,0 +1,45 @@
+# Dropwizard Release Process
+
+This document describes the technical aspects of the Dropwizard release process.
+
+## Who is responsible for a release?
+
+A release can be performed by any member of the Dropwizard project with the commit rights to the repository.
+The person responsible for the release MUST seek an approval for it in the `dropwizard-committers`
+mailing list beforehand.
+
+## Prerequisites for a new maintainer
+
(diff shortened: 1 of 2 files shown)


jplock:
I never realized all of the steps involved in generating the documentation.  Should we include a simple shell script in the repo that takes the version as a parameter to execute those steps?

arteam:
Probably make sense for generating documentation, which can be automated, but changing version and setting the release date is still a manual process, unfortunately.

nickbabcock:
I test the documentation locally by using `make html` inside the `docs` directory 

It looks like releasing the documentation that is the tedious part :stuck_out_tongue:

arteam:
After second thought, I decided to automate some part of preparing Dropwizard documentation in a shell script. @dropwizard/committers, please review.

jplock:
Looks great to me.  Thanks for creating the script.

PR #1593, opened 2016-06-10
Support access_token query parameter for OAuth

vvondra:
The OAuth Bearer token spec allows to pass the access token not only in the Authorization header with the Bearer prefix, but also as an access_token query parameter.

See https://tools.ietf.org/html/rfc6750#section-2.3

While it's preferred to use the header for increased security, the
token can be helpful at times.

 - biggest advantage is you can avoid CORS pre-flight requests with the query parameter since they are not triggered when no non-simple header is present
 - it can be useful when testing/developing where you can simulate requests in a browser tab without a REST client


joschi:
@vvondra LGTM. Thanks!

PR #1599, opened 2016-06-17
Allow periods in keys to be escaped in ConfigOverrides

tjwilson90:
This in particular makes it possible to configure loggers (whose names frequently contain periods) with config overrides.


yinzara:
Can we please merge this into master for the next release?  It's a feature I really could use.

arteam:
Thanks for your contribution! Applied on top of the master branch.

yinzara:
Thanks! I look for it in the next release!

PR #1603, opened 2016-06-19
Add helper functions for getting listening ports and IP addresses

jplock:
I've had to do this a few times now in [dropwizard-consul](https://github.com/smoketurner/dropwizard-consul/blob/master/consul-core/src/main/java/com/smoketurner/dropwizard/consul/core/ConsulServiceListener.java#L46) and [dropwizard-discovery](https://github.com/dropwizard/dropwizard-discovery/blob/master/src/main/java/io/dropwizard/discovery/core/CuratorAdvertisementListener.java#L41) and we provide this in [DropwizardTestSupport](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java#L204), so I figured we could add these helper methods to the `ServerLifecycleListener` directly.

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/LocalIpFilter.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/LocalIpFilter.java
@@ -1,0 +1,32 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


nickbabcock:
Weird that a jdbi test failed (when this pr doesn't touch jdbi code), so I restarted travis.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/LocalIpFilter.java
```diff
@@ -0,0 +1,32 @@
+/**
```
jplock: I only added the license header here because I basically copied this code out of Curator.  However, since both Curator and Dropwizard are under the Apache 2.0 license, maybe this is unnecessary?

jplock:
Any feedback or comments on this one @dropwizard/committers ?  Could this be something we get into 1.0.0 or wait until 1.1.0?

arteam:
I think it's better to wait for 1.1.0 and push only bug fixes to 1.0.0, should they come up.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/LocalIpFilter.java
arteam: Probably a notice with Apache Curator's name would be sufficient .

nickbabcock:
Want to rebase this on current master?

arteam:
Rebased and pushed to master. Thanks!

Issue #838, opened 2015-01-08
Shouldn't Tasks get their parameters from the body rather than the url


RichoDemus:
Currently Tasks get their parameters from the URL of the request rather than the POST body of the request. I'm no HTTP expert but I believe it would be more correct to get them from the body instead

arteam:
I don't the original reasoning of this design, but one benefit of query parameters against a POST body is that they are a part of URI. Because of that they are logged in HTTP access log of any web server (as well in Dropwizard).

This fact simplifies the job of people who maintains applications. They can easily derive from logs which parameters were passed to tasks. With a POST body you need to do logging manually, which can lead to non standard logging implementations or forgetting to log at all.

scottaj:
I recently ran into a situation where having access to the request body would have been useful. I wanted to make a data loading task task that could take a large blob of JSON data. I was surprised that `Task`s could not access the POST request body.

Would you be opposed to a PR that provided access to the request body to the `Task` object? I think there are probably a lot of situations where it could be useful. The URL parameters would could still be used as the main configuration method that makes sense to me, but it would make providing larger amounts of data to a task easier.

nickbabcock:
@scottaj, sorry for the late response, but yes a PR that allows access to the request body would be welcomed.

nickbabcock:
Closed by #1606

PR #1606, opened 2016-06-22
Feature/expose post body to tasks

scottaj:
This a PR for #838.

It provides the POST body of a task request to the task's execute method. For a first pass, I went with the simplest possible implementation, just providing the post body of the request as a string. This seemed consistent with how tasks handle query params already, just providing the strings and leaving any transformations to a more useful format to the user.

As it is implemented now, this is a **BREAKING** change. Anyone who updates Dropwizard to a version with this change will get compilation errors and need to add an additional `String` parameter to all their `Task` subclasses.

I went with the breaking API change style because it was the simplest, but I would like feedback on whether that seems like a reasonable path forward. it seems like the alternatives would be:

 1. Make a subclass of task that accepts a post body e.g. `TaskWithBody` that has a different execute method. Dispatch to either `Task` or `TaskWithBody` in the task executor. (might actually make sense to switch it around and have the body version be the superclass)
 1. Instead of passing the body directly to execute, just add a `getBody` method to the `Task` class
 1. Something better I'm not thinking of

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -638,8 +638,6 @@ You can also extend ``PostBodyTask`` to create a task which uses the body of the
 .. code-block:: java
 
     public class EchoTask extends PostBodyTask {
-        private final Database database;
-
         public EchoTask() {
             super("echo");
         }
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -615,7 +615,7 @@ record runtime information about your tasks. Here's a basic task class:
         }
 
         @Override
-        public void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
+        public void execute(ImmutableMultimap<String, String> parameters, String body, PrintWriter output) throws Exception {
             this.database.truncate();
         }
     }
@@ -627,7 +627,7 @@ You can then add this task to your application's environment:
     environment.admin().addTask(new TruncateDatabaseTask(database));
 
 Running a task can be done by sending a ``POST`` request to ``/tasks/{task-name}`` on the admin
-port. For example::
+port. The task will receive any query parameters and the post body as arguments. For example::
 
     $ curl -X POST http://dw.example.com:8081/tasks/gc
     Running GC...
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -615,7 +615,7 @@ record runtime information about your tasks. Here's a basic task class:
         }
 
         @Override
-        public void execute(ImmutableMultimap<String, String> parameters, String body, PrintWriter output) throws Exception {
+        public void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
             this.database.truncate();
         }
     }
@@ -627,11 +627,29 @@ You can then add this task to your application's environment:
     environment.admin().addTask(new TruncateDatabaseTask(database));
 
 Running a task can be done by sending a ``POST`` request to ``/tasks/{task-name}`` on the admin
-port. The task will receive any query parameters and the post body as arguments. For example::
+port. The task will receive any query parameters as arguments. For example::
 
     $ curl -X POST http://dw.example.com:8081/tasks/gc
     Running GC...
     Done!
+
+You can also extend ``PostBodyTask`` to create a task which uses the body of the post request. Here's an example:
+
+.. code-block:: java
+
+    public class EchoTask extends PostBodyTask {
+        private final Database database;
+
+        public EchoTask() {
+            super("echo");
+        }
+
+        @Override
+        public void execute(ImmutableMultimap<String, String> parameters, String postBody, PrintWriter output) throws Exception {
+            output.write(postBody);
+            output.flush();
+        }
+    }
 
 .. _man-core-logging:
 
--- a/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/PostBodyTask.java
+++ b/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/PostBodyTask.java
@@ -1,0 +1,45 @@
+package io.dropwizard.servlets.tasks;
+
+import com.google.common.collect.ImmutableMultimap;
+
+import java.io.PrintWriter;
+
+/**
+ * A task which can be performed via the admin interface and provides the post body of the request.
+ *
+ * @see Task
+ * @see TaskServlet
+ */
+public abstract class PostBodyTask extends Task {
+    /**
+     * Create a new task with the given name.
+     *
+     * @param name the task's name
+     */
+    protected PostBodyTask(String name) {
+        super(name);
+    }
+
+    /**
+     * @param parameters the query string parameters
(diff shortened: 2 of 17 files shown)


scottaj:
I think I branched off of a failing build. I can rebase once I see master is green.

evnm:
Is Dropwizard 1.0 frozen yet? Could we bake this change into it to take advantage of the major version bump?

One (hacky) alternative that preserves API compatibility would be to allow `TaskExector` instances to be composed with POST bodies via a chainable method. To illustrate, your lines 99 and 100 of TaskServlet.java could become:

    taskExecutors.get(task)
        .withPostBody(getBody(req))
        .executeTask(getParams(req), output);

`withPostBody` would return a copy of the `TaskExecutor` with a `postBody` field initialized to the argument String.

I don't like having to treat the parameters and request body differently in such a way, but I think it's best to avoid breaking changes at all costs.

arteam:
Sorry for the failing build. It should pass now.

We will certainly not introduce breaking changes to 1.0.0. We plan to make a release in a short time, so only bug fixes and regressions are currently accepted. 

I will take a look at the PR implementation, but I would generally more fond to see this feature implemented in a backward-compatible way. This way we can include it in 1.1.0.

scottaj:
I rebased off the green master build, so mine should pass now.

I'm happy to change the implementation, just let me know what approach you think makes the most sense and I'll go down that road.

scottaj:
hey @arteam did you have a chance to look at this and have any thoughts on what you'd like a backwards compatible approach to look like? 

I think it would either need to be a property that is set and accessed and not passed into `execute`, or some type of custom subclass.

arteam:
Hey @scottaj.  Thanks for your persistence to contribute this feature in Dropwizard! 
I had a look at the implementation and I think it's better to go with the subclass approach. I would like to keep tasks immutable, because this makes easier for users to reason about them. They don't need to worry about the scope of a task and thread-safety concerns.

I think we could provide to users a public abstract class like that:
```java
public abstract class BodyTask extends Task {

    public BodyTask(String name) {
        super(name);
    }

    public abstract void execute(ImmutableMultimap<String, String> parameters,
                          String body, PrintWriter output) throws Exception;

    @Override
    @Deprecated
    public void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
        throw new UnsupportedOperationException("Use `execute(parameters, body, output)`");
    }
}
```
and inside `TaskServlet` invoke the correct method based on the task's type.

scottaj:
This has now been reimplemented as a backwards compatible change. 

I'm not sure how coverage dropped because I'm pretty sure I covered every line I added.

dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/PostBodyTask.java
```diff
@@ -0,0 +1,45 @@
...
+     * Deprecated, use `execute(parameters, body, output)` or inherit from Task instead.
+     *
+     * @param parameters the query string parameters
+     * @param output     a {@link PrintWriter} wrapping the output stream of the task
+     * @throws Exception
+     */
+    @Override
+    @Deprecated
+    public void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
+        throw new UnsupportedOperationException("Use `execute(parameters, body, output)`");
```
evnm: Perhaps a question for @arteam: why throw here instead of proxying to the other `execute` with the empty string? This would be the first use of `UnsupportedOperationException` in a public method of a public class within the Dropwizard codebase. It seems to me that an empty request body is a valid invocation, so is there a specific reason to throw?

scottaj: It seems like if you are explicitly extending this class, you would always want to use the post body form, otherwise you would just extend `Task`. 

Furthermore, that would complicate the TaskServlet implementation because it would have to call both execute methods, which in turn could lead to confusing implementations where a user implements both execute methods and runs conflicting code in them.

docs/source/manual/core.rst
```diff
@@ -627,12 +627,30 @@ You can then add this task to your application's environment:
...
     $ curl -X POST http://dw.example.com:8081/tasks/gc
     Running GC...
     Done!
 
+You can also extend ``PostBodyTask`` to create a task which uses the body of the post request. Here's an example:
+
+.. code-block:: java
+
+    public class EchoTask extends PostBodyTask {
+        private final Database database;
```
evnm: `database` field isn't initialized or used in this example, so please delete this line.

evnm:
LGTM, aside from documentation nit. @arteam, mind chiming in, please? Would be more comfortable merging with your signoff, since you had opinions on this change.

arteam:
Rebased and applied on master. Thanks, @scottaj for the  contribution and @evnm for the review!

nickbabcock:
The second http request in `testBuild` fails because `parameters` is is an empty map. This only happens when I run `SimpleServerFactoryTest` with `DefaultServerFactoryTest`. This is reproducible 100% for me.

evnm:
Running `SimpleServerFactoryTest` in isolation (i.e. `mvn test -Dtest=SimpleServerFactoryTest` passes, but I reliably get the failure when running all of `dropwizard-core` tests in one go.

[Printing the param map](https://gist.github.com/evnm/3d785004287edcd62af919258964f16f) within `TaskServlet#getParams` indicates that Jetty's implementation of `HttpServletRequest` isn't populating the params map with query parameters in POST requests. I initially thought that perhaps the [upgrade](https://github.com/dropwizard/dropwizard/pull/1649) from Jetty 9.3.11.v20160517 to Jetty 9.3.11.v20160721 introduced the issue, but the test failure occurs at the upgrade commit and at the commit prior.

Did some digging in Jetty's `Request` and `HttpUri` classes, but didn't turn up anything. Consider me stumped for now.

nickbabcock:
My initial thought is that there is some test in `dropwizard-core`, specifically in `DefaultServerFactoryTest`, that isn't cleaning up after itself. It's nothing but a hunch, though

evnm:
I think the issue may stem from the Jetty upgrade after all. While I was poking around earlier, 9.3.9.v20160721 was being pulled in by 1.0.1-SNAPSHOT builds of dropwizard-servlets, even when I'd checked out commits that predated the actual bump to that version of Jetty.

I pulled down the Jetty source at the `jetty-9.3.11.v20160721` tag and used a few scattered printlns to trace request state as the offending POST-with-query-params from `SimpleServerFactoryTest` passes through the server. The correct request metadata is traceable through a number of `jetty-server` classes (`HttpChannelOverHttp`, `HttpChannel`, and `Request`).

But then the query parameters aren't extracted from the URI when `TaskServlet#doPost` is invoked. Printing `req.getQueryString()` in there shows it to be `name=test_user`, as it should be. And the weirdest thing is that the extraction codepath in Jetty's `Request#extractQueryParameters` method doesn't run until _after_ the `TaskServlet` has finished.

I'm going to file [a separate issue](1699), since it doesn't seem to me that this relates to #1606

PR #1607, opened 2016-06-23
fix testing deserialization example

youngj523:
The example of Testing Deserialization won't pass unless we override the equals method in Person class


arteam:
Thanks for your efforts for improving the Dropwizard documentation!

Unfortunately, the problem is that if you override the `equals` method, you need to override `hashCode` as well. Otherwise the class doesn't conform the rule which states that equal objects mush have the same hash code. That's a lot of code, which is not directly reflects a purpose of the `Person` class . As it's presented in the documentation for the demonstration purposes, I think it's better to leave it in the current state for the sake of brevity.

PR #1610, opened 2016-06-23
Replaced my GitHub handle with my real name in contributors file

RawToast:
I noticed you only have my GitHub handle in the contributors file. So I've updated my own entry in the contributors file to include my name :smiley_cat:


nickbabcock:
I believe your name will automatically updated next time [list_contributors.rb](https://github.com/dropwizard/dropwizard/blob/master/docs/list_contributors.rb) is ran.

arteam:
Yes, we periodically update the list of contributors that with the script. But there should not be harm to amend the list manually.

Issue #1614, opened 2016-06-28
CachingAuthenticator does not block subsequent requests for same authorization


windbender:
I have a situation where multiple requests hit the server at roughly the same time, each of these request is using the same authorization.

If the CachingAuthenticator does not yet have an entry for that authorization then each of these requests will fall through to the canonical authenticator.  Coded as it is, this is less efficient than it could be.

Perhaps this should be coded with the CacheLoader pattern, such that following concurrent requests to the cache are blocked until the canonical authenticator has returned a value??

jplock:
We'll happily review a pull request if you can implement this in a backward compatible way and provide adequate tests.

nickbabcock:
#1615 closes this

PR #1615, opened 2016-06-29
Use LoadingCache in CachingAuthenticator.

pkwarren:
Update CachingAuthenticator to use a LoadingCache instead of a Cache,
which will allow multiple concurrent requests to the cache for an
uncached value to only result in one call to the authenticator. The
current behavior to not cache invalid credentials is preserved.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
@@ -4,14 +4,18 @@ import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.google.common.base.Predicate;
-import com.google.common.cache.Cache;
+import com.google.common.base.Throwables;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheBuilderSpec;
+import com.google.common.cache.CacheLoader;
 import com.google.common.cache.CacheStats;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.Sets;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 
 import java.security.Principal;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
 
 import static com.codahale.metrics.MetricRegistry.name;
 
@@ -23,8 +27,7 @@ import static com.codahale.metrics.MetricRegistry.name;
  * @param <P> the type of principals the authenticator returns
  */
 public class CachingAuthenticator<C, P extends Principal> implements Authenticator<C, P> {
-    private final Authenticator<C, P> underlying;
-    private final Cache<C, Optional<P>> cache;
+    private final LoadingCache<C, Optional<P>> cache;
     private final Meter cacheMisses;
     private final Timer gets;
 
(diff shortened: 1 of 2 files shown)


pkwarren:
Implements the feature requested in #1614.

evnm:
Very neat. LGTM.

dropwizard-auth/pom.xml
```diff
@@ -38,5 +38,10 @@
             <artifactId>jersey-test-framework-provider-grizzly2</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
arteam: You can remove this dependency, because Hamcrest is already imported to the module via JUnit.

dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
```diff
@@ -51,25 +54,37 @@ public CachingAuthenticator(final MetricRegistry metricRegistry,
...
         try {
-            Optional<P> optionalPrincipal = cache.getIfPresent(credentials);
-            if (optionalPrincipal == null) {
-                cacheMisses.mark();
-                optionalPrincipal = underlying.authenticate(credentials);
-                if (optionalPrincipal.isPresent()) {
-                    cache.put(credentials, optionalPrincipal);
-                }
+            return cache.get(credentials);
+        } catch (UncheckedExecutionException|ExecutionException e) {
```
arteam: I think it's better to handle checked and unchecked exceptions separately to indicate that we know what to do with checked exceptions and we can't do anything with unchecked. Something like that:
```java
} catch (ExecutionException e) {
    final Throwable cause = e.getCause();
    if (cause instanceof InvalidCredentialsException) {
        return Optional.empty();
    }
    Throwables.propagateIfInstanceOf(cause, AuthenticationException.class);
    throw new AuthenticationException(cause);
} catch (UncheckedExecutionException e) {
    throw Throwables.propagate(e.getCause());
}
```

dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java
```diff
@@ -51,25 +54,37 @@ public CachingAuthenticator(final MetricRegistry metricRegistry,
...
+            final Throwable cause = e.getCause();
+            if (cause instanceof InvalidCredentialsException) {
+                return Optional.empty();
             }
-            return optionalPrincipal;
+            // Attempt to re-throw as-is
+            Throwables.propagateIfPossible(cause, AuthenticationException.class);
+
+            // Rethrow other checked exceptions wrapped in an AuthenticationException (this shouldn't happen)
+            throw new AuthenticationException(Optional.ofNullable(cause).orElse(e));
```
arteam: Despite it's possible to have a null value for a cause for an exception, I think we can't get an `ExecutionException` without a cause from a `LoadingCache`, so it's safe to avoid this check.

arteam:
LGTM as well, a couple of suggestions in comments.

pkwarren:
@arteam - Thanks for the review. Updated based on your comments.

evnm:
Any reason this hasn't been merged yet? All committer feedback seems to have been addressed.

jplock:
@evnm waiting until 1.0.0 final has been released first

PR #1620, opened 2016-07-07
Commit a database transaction in the `RESP_FILTERS_START` stage

### LINKED ISSUE #1604: Hibernate insert/update exceptions not being returned to the client
**Version: 1.0.0RC3**

The fix to allow lazy loading of POJOs during serialization has introduced a key issue. We are no longer seeing SQLExceptions during an insert/update//delete being reported back to the client.

In 1.0.0RC3, the session will be closed and the transaction committed only after the serialization has happened. If there are any database exceptions during the transaction commit (column val exceed/constraints etc..), they now happen only after the serialized response has been sent back. So the client assumes things are fine, but actually the server logs show the exception.

I guess we will still need to end the transaction or commit it in the RESP_FILTERS_START event. And session can be closed in FINISHED event. 

Here's the pull that introduced this issue, https://github.com/dropwizard/dropwizard/pull/1466

arteam:
But close the session during the `FINISHED` stage. This allows to the user to catch errors during commiting transactions (constraint violations, e.g) and map them to HTTP error with
an exception mapper.

This approach also is compatible with lazy loading, because the session remain open during serializing Hibernate entities.

Fix #1604

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
@@ -64,12 +64,9 @@ class UnitOfWorkAspect {
         } catch (Exception e) {
             rollbackTransaction();
             throw e;
-        } finally {
-            session.close();
-            session = null;
-            ManagedSessionContext.unbind(sessionFactory);
         }
-
+        // We should not close the session to let the lazy loading work during serializing a response to the client.
+        // If the response successfully serialized, then the session will be closed by the `onFinish` method
     }
 
     public void onError() {
@@ -80,7 +77,16 @@ class UnitOfWorkAspect {
         try {
             rollbackTransaction();
         } finally {
-            session.close();
+            onFinish();
+        }
+    }
+
+    public void onFinish() {
+        try {
+            if (session != null) {
+                session.close();
+            }
+        } finally {
             session = null;
             ManagedSessionContext.unbind(sessionFactory);
         }
--- a/dropwizard-hibernate/pom.xml
+++ b/dropwizard-hibernate/pom.xml
@@ -41,6 +41,11 @@
             <artifactId>hibernate-core</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.dropwizard</groupId>
+            <artifactId>dropwizard-testing</artifactId>
+            <scope>test</scope>
+        </dependency>
         <!-- we need HSQL because it handles time zones, H2 totally doesn't -->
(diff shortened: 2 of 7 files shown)


dropwizard-example/src/main/java/com/example/helloworld/HelloWorldApplication.java
```diff
@@ -89,6 +89,7 @@ public void run(HelloWorldConfiguration configuration, Environment environment)
                 .setAuthorizer(new ExampleAuthorizer())
                 .setRealm("SUPER SECRET STUFF")
                 .buildAuthFilter()));
+        environment.jersey().
```
joschi: That line probably doesn't belong here. :wink:

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
```diff
@@ -80,7 +77,16 @@ public void onError() {
         try {
             rollbackTransaction();
         } finally {
-            session.close();
+            onFinish();
+        }
+    }
+
+    public void onFinish() {
```
joschi: Couldn't this lead to a session/memory leak?

dropwizard-example/src/main/java/com/example/helloworld/HelloWorldApplication.java
arteam: Yeah, a copy-paste error. A kindle reminder to myself to never use `git add .`

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
arteam: There is a possibility, but I think we can presume that Jersey will always generate a `FINISHED` event, and `onFinish` will be called and it will clean up resources.

pkwarren: According to the docs, it sounds like this is called on success or failure: 

https://github.com/jersey/jersey/blob/5f9e5fb78d48ce45431032a5867c157f191716fc/core-server/src/main/java/org/glassfish/jersey/server/monitoring/RequestEvent.java#L174-L179

pkwarren:
👍

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
joschi: ✅

PR #1622, opened 2016-07-08
Add Google Error Prone checks

jplock:
Adds Google Error Prone (http://errorprone.info) compiler checks

Also updates the plugin dependencies to their latest versions.


joschi:
LGTM. 👍

Issue #1621, opened 2016-07-07
Database exceptions can't be handled with @UnitOfWork


ogregoire:
I'm using Dropwizard 1.0.0-rc3.

A constraint violation occurs on the database level when `@UnitOfWork` tries to commit (I guess that's the moment, actually). Yet no exception is thrown when I have the hand on the code.

I would expect to be able to handle this error gracefully. [The documentation has nothing on the subject](http://www.dropwizard.io/0.9.2/docs/manual/hibernate.html).

You can reproduce with my (under development) application. Just **send twice** the following request to localhost:8080.

```
Method: POST
URL: http://localhost:8080/books
MediaType: application/json
Json: {"isbn":"978-0618640157","title":"The Lord of the rings"}
```

Relevant files:
- [`migrations.xml`](https://github.com/ogregoire/alexandria/blob/3a219efd4fe9b0d5dc887f26054eb7ee17e88721/src/main/resources/migrations.xml): puts some unique constraint on the table
- [`BookResource.java`](https://github.com/ogregoire/alexandria/blob/3a219efd4fe9b0d5dc887f26054eb7ee17e88721/src/main/java/be/ogregoire/alexandria/core/book/BookResource.java): the resource where no exception is thrown.
- [`BookDAO.java`](https://github.com/ogregoire/alexandria/blob/3a219efd4fe9b0d5dc887f26054eb7ee17e88721/src/main/java/be/ogregoire/alexandria/core/book/BookDAO.java): fairly simple AbstractDAO implementation

If I'm doing something wrong, please advise and/or redirect to the appropriate documentation page.

arteam:
I believe you hit the same issue as #1604, which was introduced in Dropwizard 1.0.0-rc3. We have a proposed fix in #1620, but it still needs to reviewed. If the fix is correct, the issue will be fixed in Dropwizard 1.0.0.

For the being, you can try downgrade to Dropwizard 1.0.0-rc2.

ogregoire:
Well, that was fast.

Indeed, downgrading to 1.0.0-rc2 provided me with a status 500 Internal Server Error, which I expect more.

**But** I still can't handle the exception instead of throwing that 500. Is there anything I should do to handle those?

arteam:
[...]

PR #1625, opened 2016-07-08
Add a way to delegate the proxy creation to the user

ogregoire:
Because of dependency injection frameworks, I need to create the proxy by myself. Using Objenesis is the way to go. This pull request lets `UnitOfWorkAwareProxyFactory` inject my javassist-created proxy with the factory's custom `MethodHandler` really easily.

Design choices:

1. It is preferrable that `UnitOfWorkAwareProxyFactory` performs the injection rather than exposing its `MethodHandler`. This way, the interface doesn't depend on javassist.
2. I think it's preferable to accept a `java.lang.Object` rather than a `javassist.util.proxy.Proxy` as the user code will be more readable.
3. If a non-proxy object is provided, an `IllegalArgumentException` is thrown rather than a `ClassCastException` to insist on the fact that the argument is incorrect and hide the implementation details.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
@@ -1,6 +1,9 @@
 package io.dropwizard.hibernate;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import com.google.common.collect.ImmutableMap;
+import javassist.util.proxy.MethodHandler;
 import javassist.util.proxy.Proxy;
 import javassist.util.proxy.ProxyFactory;
 import org.hibernate.SessionFactory;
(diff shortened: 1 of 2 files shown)


ogregoire:
Note this is directly linked with my remarks of #1621.

ogregoire:
Thanks for looking into this. But I wouldn't consider this an "improvement" per se, it's more like fixing something that has been overlooked in the context of DI, a kind of design flaw. Given this, is there any way to have this "fix" in 1.0.0 already?

The current workaround is to copy/paste the entire `UnitOfWorkAwareProxyFactory` code and adapt it.

joschi:
@ogregoire We want to keep changes to the 1.0.0 version branch minimal now, otherwise we'll never release Dropwizard 1.0.0.

This change will probably be included in Dropwizard 1.1.x.

ogregoire:
Ok, fair enough. Thanks!

ogregoire:
Do not take this PR yet. There are several other design issues. Currently, this only adds proxies. Some DI frameworks don't allow access to their proxies (as a proxy, not as a usable object).

I now think it's better to not expose javassist at all. `UnitOfWorkAspect` should be the main entry point for DI frameworks, and should probably be made visible and somehow accessible.

I'll play with this a bit more and will try to find an acceptable, more generic solution.

In the meantime, should I close this PR? Keep it open? In any case, please remove the 1.1.0 milestone

ogregoire:
I'm gonna create a new pull request with my suggested changes. I see no troubles with the new idea. I don't know how to handle properly revert and stuff like that in git / GitHub, so I hope I don't break anything by deleting my branch.

arteam:
No problem. Looking forward to your new pull request!

PR #1634, opened 2016-07-13
Update testing.rst

ipropper:
Setting out, in, and err to null messed up my unit tests in other classes that capture std out.  I felt setting these back to their original values would help other users who had similar problems.


joschi:
@ipropper Thanks!

Issue #1638, opened 2016-07-14
CachingAuthorizer?


Jimexist:
There's already a `CachingAuthenticator` which is awesome, however there's no such thing as caching proxy for authorization yet. The reason/need for this stands similar to that of `CachingAuthenticator`.

I can think of some reasons why one does not want to do that (e.g. audit trails), but having such class would be a helpful option.

So is there a plan yet? Or specific reasons not to include one?

evnm:
It seems to me that many of the concerns one could have about a `CachingAuthorizer` would apply also to `CachingAuthenticator`. For instance, concerns about auditability or credential-revocation are valid in either case. Arguably, the same logic that validates the existence of `CachingAuthenticator` supports the addition of a `CachingAuthorizer`—developers should be allowed to decide for themselves whether or not their application warrants Principal-caching.

I'm curious what the Dropwizard team thinks on the matter.

jplock:
I think it's a good idea. If someone wanted to submit a PR for an optional implementation we'll happily review it.

Jimexist:
thanks for the opinions and comments - I think that PR looks great to me!

evnm:
This issue can now be closed, since #1639 was merged.

joschi:
@evnm Thanks!

PR #1639, opened 2016-07-16
dropwizard-auth: Introduce CachingAuthorizer

evnm:
Addresses #1638

The boolean authorization state of principal+role pairs are cached individually using [`ImmutablePair`s](https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/tuple/ImmutablePair.html) internally as cache keys. This unfortunately leads to more allocation-prone `invalidate` methods than those of `CachingAuthenticator`.

I assumed that `CachingAuthorizer` wouldn't be subject to [the negative-result-caching issue](https://github.com/dropwizard/dropwizard/issues/947) that was dealt with in `CachingAuthenticator`. If this is incorrect, then we could instead cache an `ImmutableSet<String>` of role associations for each principal and always pass negative cache results through to the underlying authorizer.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
@@ -1,0 +1,182 @@
+package io.dropwizard.auth;
+
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
+import com.google.common.base.Predicate;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheBuilderSpec;
+import com.google.common.cache.CacheStats;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Sets;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+import java.security.Principal;
+
+import static com.codahale.metrics.MetricRegistry.name;
+
+/**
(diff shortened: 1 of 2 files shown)


dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
```diff
@@ -0,0 +1,172 @@
...
+ * temporarily cache principals' role associations.
+ *
+ * Cache entries include both inclusion and exclusion of a principal
+ * within a given role.
+ *
+ * @param <P> the type of principals on which the authorizer operates
+ */
+public class CachingAuthorizer<P extends Principal> implements Authorizer<P> {
+    private final Authorizer<P> underlying;
+    private final Cache<ImmutablePair<P, String>, Boolean> cache;
```
Jimexist: what about the value being an enum:
```java
private class AuthzResult {
  Allow, Deny
}
```
so it's rather clearer that the tri-state of `true`, `false`, `null` in the boolean case.

evnm: I'm all for improving clarity, but because Java enum types are defined as classes, references to them can also be `null`. For example, the following is valid code:

    enum Foo { Allow, Deny }    
    final Foo foo = null;

If it seems ambiguous to use boolean values within the context of authorization (i.e. if it's not obvious that true=authorized, false=unauthorized), then an `Allow`/`Deny` enum would be more explicit. However, given that all of this is private code internal to the class, I think it's a fair tradeoff to use booleans to avoid introducing a new class. I think an inline comment on the cache type semantics is definitely warranted.

PR #1641, opened 2016-07-19
Fix #1640: ResourceConfigurationSourceProvider - process a path to the resource in the more sophisticated way

### LINKED ISSUE #1640: ResourceConfigurationSourceProvider: process a path to the resource in the more sophisticated way
[ResourceConfigurationSourceProvider](https://github.com/dropwizard/dropwizard/blob/3d746d21c6e6656a59b26c2a7e92f19e1bb2e9b1/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ResourceConfigurationSourceProvider.java) uses `ClassLoader#getResourceAsStream()` to get a stream to the resource from the class path.

A path to the resource can't  begin from the leading `/`. It can  confuse, because, for example,  it is need to use leading `/`  while using `Class#getResourceAsStream()` for the same resourse path.

My proposal is to use a Hibernate `ClassLoaderService` approach: removing the leading slash, if an attempt to get a stream for a resource is failed: [ClassLoaderServiceImpl.java#L254](https://github.com/hibernate/hibernate-orm/blob/ec4f20a5fba79444dbd704f1305f961c339dd7d1/hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/ClassLoaderServiceImpl.java#L254).

v-ladynev:


dropwizard-configuration/src/main/java/io/dropwizard/configuration/ResourceConfigurationSourceProvider.java
```diff
@@ -6,6 +6,11 @@
 public class ResourceConfigurationSourceProvider implements ConfigurationSourceProvider {
     @Override
     public InputStream open(String path) throws IOException {
+        InputStream result = getResourceAsStream(path);
+        return result == null && path.startsWith("/") ? getResourceAsStream(path.substring(1)) : result;
```
evnm: I think an inline comment is warranted here, pointing out the convention of implicitly treating absolute resource paths as being rooted within the application's resources directory [1]. 

[1] https://docs.oracle.com/javase/8/docs/technotes/guides/lang/resources.html#res_names

v-ladynev: Sorry, it is not about 

> convention of implicitly treating absolute resource paths as being rooted within the application's resources directory

it is about (as described in [Using Methods of java.lang.ClassLoader](https://docs.oracle.com/javase/8/docs/technotes/guides/lang/resources.html#classloader))

> The methods in ClassLoader use the given String as the name of the resource without applying any absolute/relative transformation (see the methods in Class). The name should not have a leading "/".

evnm: Regardless, I think an inline comment could help. The very fact that we're having this conversation indicates that clarity is warranted.

v-ladynev: I am agree, of course. How to add that comment?

evnm: How about a Javadoc comment on the class, since there currently isn't one?

```
/**
 * A {@link ConfigurationSourceProvider} which reads the configuration
 * from a resource file.
 *
 * In order to abide by the calling conventions of
 * {ClassLoader#getResourceAsStream} [1], absolute path strings
 * (i.e. those with leading "/" characters) passed to {@link #open(String)}
 * are converted to relative paths by removing the leading "/".
 *
 * See [1] for more information on resources in Java and how they are
 * loaded at runtime.
 *
 * [1] https://docs.oracle.com/javase/8/docs/technotes/guides/lang/resources.html
 */
public class ResourceConfigurationSourceProvider implements ConfigurationSourceProvider {
...
```

v-ladynev: Looks good for me. So how to add this comment to the pull request?

evnm: You can add the Javadoc within your local `issue/1640` branch. When you commit and push the changes, the PR should update itself.

If you're feeling brave and want to keep the whole PR within a single commit, then you can [use `git rebase` to squash the two commits into one](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git).

v-ladynev: Thanks for your kind help. I will try.

v-ladynev: @evnm I have added a comment.

evnm:
Thanks, @v-ladynev!

PR #1650, opened 2016-07-27
Update request log docs

tjcutajar:
timeZone is now a member of appenders.


evnm:
I was hoping we could link to the commit which introduced this inconsistency, but the lineage is tricky to track down. `LogbackAccessRequestLogFactory` originated in `dropwizard-jetty` and its documentation has been incorrect since its inception.

This is clearly a correct documentation fix though, so LGTM.

jplock:
Thanks!

PR #1652, opened 2016-07-28
Minor Code Fixes

nickbabcock:
- Fix or comment deprecation usage in tests/docs
- Remove redundant try-catch in TaskServlet
- Prefer java8 mechanisms in CachingAuthorizer (cc @evnm)


dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
```diff
@@ -145,11 +145,7 @@ private TaskExecutor(Task task) {
         }
 
         public void executeTask(ImmutableMultimap<String, String> params, PrintWriter output) throws Exception {
-            try {
-                task.execute(params, output);
-            } catch (Exception e) {
-                throw e;
```
evnm: Good catch. Funny that this wasn't brought up as part of #473.

evnm:
LGTM. Thanks for taking care of these!

jplock:
LGTM!

PR #1656, opened 2016-07-30
Remove the test scope from `jetty-client` in the BOM

arteam:
Some Dropwizard users may want to use this library with the compile scope as an HTTP/2 client. But currently the BOM forces the test scope, so users are forced to declare the dependency with the compile scope in their projects. This especially not very convenient when the library is pulled as a transitive dependency. An example is https://github.com/arteam/dropwizard-http2-client. See this [discussion](https://github.com/arteam/dropwizard-http2-client/issues/1).


jplock:
LGTM

PR #1661, opened 2016-08-02
Make the aspect easily available to other DI frameworks

ogregoire:
The current proxy implementation makes it hard for DI frameworks to play nice with the `@UnitOfWork` annotation. It is required to have an object, its proxy and wire the two in difficult way. For instance, using Guice, there are several solutions, but none are natural.

This simple change makes `UnitOfWorkAspect` public, but only accessible from a `UnitOfWorkAwareProxyFactory` instance.

This way, DI framework users can implement their own `@UnitOfWork` behavior. This moves the responsibility to the user. Therefore a guide has been written under the form of Javadoc for the `UnitOfWorkAspect` class. It basically says how to properly use the aspect, as intended by the original developer of `UnitOfWorkAwareProxyFactory`.

For instance, my original Guice solution took 300 lines of code, made heavy use the included proxy mechanism as well as an extra one and Objenesis, and was prone to memory leaks for some of the slightest changes. This change reduce it to ~20 lines, using the standard Guice way of things.


arteam:
Looks good to me. Thanks for contributing!

Issue #1658, opened 2016-07-31
Dropwizard could not log INFO log to console and log file at very early stage of application initialization


dknc:
Dropwizard could not log INFO log to console and log file at very early stage of application initialization. 

My dropwizard application implemented a StrSubstitutor and configured it in the initialize method

``` @Override
  public void initialize(Bootstrap<ChatServiceAppConfiguration> bootstrap)
  {
    // Enable variable substitution with environment variables
    bootstrap.setConfigurationSourceProvider(
        new SubstitutingSourceProvider(
            bootstrap.getConfigurationSourceProvider(),
            new AppPropertySubstitutor(false)
        )
    );
  }
```

The AppPropertySubstitutor depends on a component which uses apache common logging.
All INFO level log is not written to the console and the log file in the depended component.

Have checked that jcl-over-slf4j.jar is successfully packaged into the application jar.

The cause is that the default constructor of the dropwizard Application calls  

```
BootstrapLogging.bootstrap()
```

The default logger initialized a ThresholdFilter with log level Level.WARN, thus INFO log is denied.

As the BootstrapLogging.bootstrap() can only be called once. Super constructor is always executed when creating sub class instance. thus it is not able to bootstrap logging with different log level other than Level.WARN.

Dropwizard should bootstrap logging with the logging configuration defined in yaml at very first stage.

Dropwizard version:  1.0.0-rc4

evnm:
You should be able to get around this by defining a default constructor in your `Application` subclass and invoking `BootstrapLogging.bootstrap(Level)` with the desired log level:

```
import ch.qos.logback.classic.Level;
import io.dropwizard.logging.BootstrapLogging;

public FooApplication extends Application<FooConfiguration> {
  protected FooApplication() {
    BootstrapLogging.bootstrap(Level.DEBUG);
  }

  ...
}
```

dknc:
Override the default constructor of the concrete dropwizard application does not work as super constructor will be always called first and call BootstrapLogging.bootstrap(Level.DEBUG) in sub class make no effect as the bootstrap method will set bootstrapped flag to true after it is called, subsequent call to this method just simply returns.

evnm:
You're right, my mistake. Had to look this one up: https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.5

[...]

PR #1662, opened 2016-08-02
Overridable log level during application bootstrap process

evnm:
Addresses #1658 

This change adds a non-final `Level` getter method within `io.dropwizard.Application` that is used to set the log level during the logging configuration bootstrap process.

Not sure how to add a test for this. Because `BootstrapLogging` doesn't set the root logger's level, AFAICT it's not possible to inspect the log level from within an `Application` subclass.


arteam:
LGTM! Thanks, @evnm for helping Dropwizard users and submitting the pull request!

PR #1664, opened 2016-08-02
Allow use of custom HostnameVerifier on clients.

### LINKED ISSUE #1663: Can't set custom HostnameVerifier on DropWizard client
While the improvements to TLS configuration of HTTP clients in 1.0.0 (maybe prior) are awesome, as part of that process the ability to set a custom HostnameVerifier easily on the HTTP client has been lost.

You used to be able to do e.g. as:

```
JerseyClientConfiguration myJerseyClientConfiguration = <some configuration>;
HostnameVerifier verifier = new MyCustomHostnameVerifier();
JerseyClientBuilder clientBuilder = new JerseyClientBuilder(env);
clientBuilder.using(myJerseyClientConfiguration).using(verifier);
Client httpClient = clientBuilder.build();
```
Same is true for `HttpClientBuilder` too.

You can still do it by creating a custom Apache `Registry<ConnectionSocketFactory>` but you need to set up socket factories for every scheme. Pretty heavyweight.

I have a pull request that I'll submit and reference this issue.

tbartley:
While the improvements to TLS configuration of HTTP clients in 1.0.0
(maybe prior) are awesome, as part of that process the ability to set a
custom HostnameVerifier easily on the HTTP client has been lost.

You used to be able to do e.g. as:

JerseyClientConfiguration myJerseyClientConfiguration = <some
configuration>;
HostnameVerifier verifier = new MyCustomHostnameVerifier();
JerseyClientBuilder clientBuilder = new JerseyClientBuilder(env);
clientBuilder.using(myJerseyClientConfiguration).using(verifier);
Client httpClient = clientBuilder.build();
Same is true for HttpClientBuilder too.

You can still do it by creating a custom Apache
Registry<ConnectionSocketFactory> but you need to set up socket
factories for every scheme.

This change restores the ability to set a custom HostnameVerifier
for clients.

[Fixes #1663]

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
@@ -19,9 +19,15 @@ import java.util.List;
 public class DropwizardSSLConnectionSocketFactory {
 
     private final TlsConfiguration configuration;
+    private final HostnameVerifier verifier;
 
     public DropwizardSSLConnectionSocketFactory(TlsConfiguration configuration) {
+        this(configuration, null);
+    }
+
+    public DropwizardSSLConnectionSocketFactory(TlsConfiguration configuration, HostnameVerifier verifier) {
         this.configuration = configuration;
+        this.verifier = verifier;
     }
 
     public SSLConnectionSocketFactory getSocketFactory() throws SSLInitializationException {
@@ -46,13 +52,11 @@ public class DropwizardSSLConnectionSocketFactory {
     }
 
     private HostnameVerifier chooseHostnameVerifier() {
-        HostnameVerifier hostnameVerifier;
         if (configuration.isVerifyHostname()) {
-            hostnameVerifier = SSLConnectionSocketFactory.getDefaultHostnameVerifier();
+            return verifier != null ? verifier : SSLConnectionSocketFactory.getDefaultHostnameVerifier();
         } else {
-            hostnameVerifier = new NoopHostnameVerifier();
+            return new NoopHostnameVerifier();
         }
-        return hostnameVerifier;
     }
 
     private SSLContext buildSslContext() throws SSLInitializationException {
--- a/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
@@ -9,9 +9,11 @@ import com.google.common.annotations.VisibleForTesting;
 import io.dropwizard.client.proxy.AuthConfiguration;
 import io.dropwizard.client.proxy.NonProxyListProxyRoutePlanner;
 import io.dropwizard.client.proxy.ProxyConfiguration;
+import io.dropwizard.client.ssl.TlsConfiguration;
 import io.dropwizard.lifecycle.Managed;
 import io.dropwizard.setup.Environment;
 import io.dropwizard.util.Duration;
+import javax.net.ssl.HostnameVerifier;
 import org.apache.http.ConnectionReuseStrategy;
 import org.apache.http.Header;
 import org.apache.http.HttpHost;
@@ -62,6 +64,7 @@ public class HttpClientBuilder {
     private Environment environment;
     private HttpClientConfiguration configuration = new HttpClientConfiguration();
     private DnsResolver resolver = new SystemDefaultDnsResolver();
+    private HostnameVerifier verifier;
     private HttpRequestRetryHandler httpRequestRetryHandler;
     private Registry<ConnectionSocketFactory> registry;
 
@@ -112,6 +115,17 @@ public class HttpClientBuilder {
      */
     public HttpClientBuilder using(DnsResolver resolver) {
         this.resolver = resolver;
+        return this;
+    }
+
+    /**
+     * Use the give (@link HostnameVerifier} instance.
+     *
+     * @param verifier a {@link HostnameVerifier} instance
+     * @return {@code this}
+     */
+    public HttpClientBuilder using(HostnameVerifier verifier) {
+        this.verifier = verifier;
         return this;
     }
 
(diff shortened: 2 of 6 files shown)


tbartley:
Also, I wasn't sure how to approach writing a unit test in HttpClientBuilderTest.java. Some guidance there would be appreciated.

dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
```diff
@@ -19,9 +19,16 @@
 public class DropwizardSSLConnectionSocketFactory {
 
     private final TlsConfiguration configuration;
+    private final HostnameVerifier verifier;
 
     public DropwizardSSLConnectionSocketFactory(TlsConfiguration configuration) {
         this.configuration = configuration;
+        this.verifier = null;
```
arteam: I think we could reduce this constructor to a call to another constructor:
```java
this(configuration, null);
```

dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
```diff
@@ -48,7 +55,12 @@ public SSLConnectionSocketFactory getSocketFactory() throws SSLInitializationExc
     private HostnameVerifier chooseHostnameVerifier() {
         HostnameVerifier hostnameVerifier;
         if (configuration.isVerifyHostname()) {
-            hostnameVerifier = SSLConnectionSocketFactory.getDefaultHostnameVerifier();
+            if (verifier == null) {
```
arteam: What do you think about reducing the size of this method by using the ternary operator and several returns?
Something like that:
```java
if (configuration.isVerifyHostname()) {
    return verifier != null ? verifier : SSLConnectionSocketFactory.getDefaultHostnameVerifier();
} else {
    return new NoopHostnameVerifier();
}
```

dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
```diff
@@ -378,9 +396,15 @@ protected InstrumentedHttpClientConnectionManager createConnectionManager(Regist
 
         final SSLConnectionSocketFactory sslConnectionSocketFactory;
         if (configuration.getTlsConfiguration() == null) {
-            sslConnectionSocketFactory = SSLConnectionSocketFactory.getSocketFactory();
+            if (verifier == null) {
```
arteam: If I understand correctly, setting a `HostnameVerifier` means using a `TlsConfiguration`. What do you think about extracting `TlsConfiguration` to a local mutable variable?
```java
TlsConfiguration tlsConfiguration = configuration.getTlsConfiguration();
if (tlsConfiguration == null && verifier != null) {
    tlsConfiguration = new TlsConfiguration();
}

final SSLConnectionSocketFactory sslConnectionSocketFactory;
if (tlsConfiguration == null) {
    sslConnectionSocketFactory = SSLConnectionSocketFactory.getSocketFactory();
} else {
    sslConnectionSocketFactory = new DropwizardSSLConnectionSocketFactory(tlsConfiguration,
        verifier).getSocketFactory();
}
```

dropwizard-client/src/test/java/io/dropwizard/client/ssl/DropwizardSSLConnectionSocketFactoryTest.java
```diff
@@ -172,4 +203,13 @@ public void shouldRejectNonSupportedProtocols() throws Exception {
             assertThat(e.getCause()).isInstanceOfAny(SocketException.class, SSLHandshakeException.class);
         }
     }
+    
+    private static class FailVerifier implements HostnameVerifier {
+
+		@Override
+		public boolean verify(String arg0, SSLSession arg1) {
```
arteam: Could you please replace tabs to spaces? A nitpick, but we prefer to stick to spaces across the project.

arteam:
> Also, I wasn't sure how to approach writing a unit test in HttpClientBuilderTest.java. Some guidance there would be appreciated.

I think you could mark the method `createConfiguredRegistry` as package protected and put the `@VisibleForTesting` annotation on it. Then you could invoke it in tests:
```java
HostnameVerifier customVerifier = (s, sslSession) -> false;
Registry<ConnectionSocketFactory> configuredRegistry = builder
    .using(customVerifier)
    .createConfiguredRegistry();
```

Then we could retrieve the SSL socket factory from the registry and extract the verifier from it via reflection. Something like that:
```java
SSLConnectionSocketFactory socketFactory = 
(SSLConnectionSocketFactory) configuredRegistry.lookup("https");
final Field hostnameVerifierField =
    FieldUtils.getField(SSLConnectionSocketFactory.class, "hostnameVerifier", true);
```

dropwizard-client/src/test/java/io/dropwizard/client/ssl/DropwizardSSLConnectionSocketFactoryTest.java
tbartley: Ach! I'd been diligent about that elsewhere. Missed it here. Thanks.

dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
tbartley: Sure

dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
tbartley: Sure

dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
tbartley: Sure

tbartley:
Thanks for the feedback @arteam I've made your suggested changes and implemented additional unit test coverage following your suggested pattern.

I ummed and ahhed over the DropwizardSSLConnectionSocketFactory constructor unit test using reflection to check private field value but I liked using a getter for the verifier field on this class even less since it would imply getter for tlsConfiguration field for consistency which exposes that to internal modification which implies need for immutable TlsConfiguration .... Also, the getter would've needed to be public rather than package protected since unit test is in a different package.

I have also squashed my commits.

arteam:

tbartley:
No problem at all. This software works so well for us. Thank you for the far greater contribution you make.

PR #1666, opened 2016-08-03
Refactored DropwizardAppRule support running different EnvironmentCommands

tburch:
I'm working on writing a [gRPC](http://www.grpc.io/) `EnvironmentCommand` for running gRPC in Dropwizard and needed the ability to be able to use a `DropwizardAppRule` for testing. 

This pull request adds constructor parameters to `DropwizardAppRule` and `DropwizardTestSupport` so you can specify a `Function` that will instantiate the `EnvironmentCommand` that's run from `DropwizardTestSupport`. `ServerCommand` was used everywhere as a default to maintain backwards compatibility.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
@@ -6,6 +6,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import io.dropwizard.Application;
 import io.dropwizard.Configuration;
+import io.dropwizard.cli.Command;
 import io.dropwizard.cli.ServerCommand;
 import io.dropwizard.configuration.YamlConfigurationFactory;
 import io.dropwizard.lifecycle.Managed;
@@ -21,6 +22,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Throwables.propagate;
@@ -40,6 +42,7 @@ public class DropwizardTestSupport<C extends Configuration> {
     protected final String configPath;
     protected final Set<ConfigOverride> configOverrides;
     protected final Optional<String> customPropertyPrefix;
+    protected final Function<Application<C>, Command> commandInstantiator;
 
     /**
      * Flag that indicates whether instance was constructed with an explicit
@@ -62,11 +65,19 @@ public class DropwizardTestSupport<C extends Configuration> {
 
     public DropwizardTestSupport(Class<? extends Application<C>> applicationClass, String configPath,
                                  Optional<String> customPropertyPrefix, ConfigOverride... configOverrides) {
+        this(applicationClass, configPath, customPropertyPrefix, ServerCommand::new, configOverrides);
+    }
+
+    public DropwizardTestSupport(Class<? extends Application<C>> applicationClass, String configPath,
+                                 Optional<String> customPropertyPrefix,
+                                 Function<Application<C>, Command> commandInstantiator,
+                                 ConfigOverride... configOverrides) {
         this.applicationClass = applicationClass;
         this.configPath = configPath;
         this.configOverrides = ImmutableSet.copyOf(firstNonNull(configOverrides, new ConfigOverride[0]));
         this.customPropertyPrefix = customPropertyPrefix;
         explicitConfig = false;
+        this.commandInstantiator = commandInstantiator;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
```diff
@@ -81,7 +92,21 @@ public DropwizardTestSupport(Class<? extends Application<C>> applicationClass, S
...
+    /**
+     * Alternate constructor that allows specifying the command the Dropwizard application is started with.
+     * @since 1.1.0
+     * @param applicationClass Type of Application to create
+     * @param configuration Pre-constructed configuration object caller provides; will not
+     * @param envCommandInistaniator The {@link Function} used to instantiate the {@link EnvironmentCommand} used to
+     *   start the Application
+     */
+    public DropwizardTestSupport(Class<? extends Application<C>> applicationClass,
+                                 C configuration, Function<Application<C>, EnvironmentCommand<C>> envCommandInistaniator) {
```
evnm: *Instantiator

dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
```diff
@@ -81,7 +92,21 @@ public DropwizardTestSupport(Class<? extends Application<C>> applicationClass, S
...
+                                 C configuration) {
+        this(applicationClass, configuration, ServerCommand::new);
+    }
+
+
+    /**
+     * Alternate constructor that allows specifying the command the Dropwizard application is started with.
+     * @since 1.1.0
+     * @param applicationClass Type of Application to create
+     * @param configuration Pre-constructed configuration object caller provides; will not
```
evnm: "Pre-constructed configuration object" would suffice

dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
tburch: Thanks! fixed...

dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
tburch: Are you thinking a `DropwizardTestSupportConfig` POJO for all this instead of constructor overloading?

evnm:
LGTM

Does your application override `Application#addDefaultCommands(Bootstrap)` to swap out the default `ServerCommand`? This brings up an interesting question: what is the recommended path for running a non-Jetty/HTTP Dropwizard server?

Another option would be augment `ServerCommand` to support server types beyond Jetty. This will be relevant to a future Dropwizard release which includes HTTP/2 support.

dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
evnm: Nope, I was just pointing out that the `configuration` parameter doc is either clipped or needs rewording.

tburch: Ahh, IntelliJ didn't copy it fully for me. Fixed that...

tburch:
@evnm it does modify `Application#addDefaultCommands`. I first went down the path of refactoring `ServerCommand`, but then decided that what I wanted to do was different enough to warrant another command. I also noticed that `dropwizard-http2` took this approach, so decided it was ok. I'll opensource it at some point (or I would happy to add it to Dropwizard)...

It looks like `DropwizardTestSupport` is pretty tightly coupled to Jetty, so it might make sense to abstract all of this a little (I saw #846 that looked like what may be needed).

evnm:
Gotcha, thanks for clarifying. I admittedly haven't looked into dropwizard-http2 yet.

gRPC support would be a great candidate for a [3rd-party module](http://modules.dropwizard.io/thirdparty/). To date, Dropwizard has been staunchly focused on HTTP services, so maintaining official support for other protocols would warrant a wider discussion.

Ferada:
I need something similar for testing a separate command for an application, so this comes in handy. Now, could this also use `ConfiguredCommand` instead of `EnvironmentCommand`?

tburch:
@Ferada I don't see why not. I've updated the pull request.

evnm:
Why not go all the way and have the function produce `Command`s?

dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
evnm: Looks like this didn't get updated in your more recent commits. Mind fixing the `configuration` param documentation so I can merge, please?

tburch: @evnm It's on the next line (to match prior formatting)...

tburch:
@jplock what are the chances of this being in a point release instead of `1.1.0`? It's the same functionality as previous versions with an additional constructor to support additional functionality.

dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
evnm: Derp. My bad.

evnm:
If we're going to be strict, given that the new constructor exposes additional functionality, this change would warrant a minor version bump, according to the [precepts of semantic versioning](http://semver.org/).

I think that the best we can offer right now is to clarify a roadmap for the next few releases. From this, we can all rally around a specific task list for changes that are to be included in the next release.

pparth:
Anything new regarding Dropwizard + GRPC integration? Is there, at least, a way to swap the Jetty server to another one?

tburch:
@pparth checkout https://github.com/tburch/dropwizard-grpc

pparth:
@tburch Looks good! Nice! thanks!

PR #1667, opened 2016-08-04
dropwizard-servlets: Combine unnecessarily nested if blocks in TaskServlet

evnm:
These two nested if statements within `io.dropwizard.servlets.tasks.TaskServlet` were [flagged by Code Climate](https://codeclimate.com/github/dropwizard/dropwizard/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java) and can be flattened.


arteam:
Nice catch!

Issue #1633, opened 2016-07-12
Caching Http Client


gurrie:
Hi All, 

I would like to implement the ability to create a caching http client (https://hc.apache.org/httpcomponents-client-ga/tutorial/html/caching.html) when using the JerseyClientBuilder. 

Is there any interest in this feature? Or is this ability already present and I am just missing something? 

The rationale for the request is that on my project we have a number of collaborating microservices, and it would be desirable to reduce the number of network requests that are made by specifying the appropriate HTTP cache headers on each response and having the clients cache the data for the specified period. I have previously posted to both the dev and user mailing lists but haven't received any responses.

Any feedback or suggestions would be greatly appreciated.

Kind Regards 
Brian

arteam:
Hi, @gurrie. Sorry for the lack of feedback in the mailing lists, usually we try to answer to such questions.

I think your proposal is an interesting feature and I can see other users who can be interested in it. Dropwizard is quite often used in projects consisting of many microservices, so the benefits of client-side caching can be quite extensive. Considering we can provide an external cache configuration for the application administrators, it can be a powerful way to optimize existed Dropwizard application which make a lot of HTTP requests. 

But I am a little bit skeptic about including such functionality in the project's core. We will need to pull external library `httpclient-cache` and make caching transparent to the user. Despite it looks like that the Apache Client makes it rather easy (`CachingHttpClientBuilder` extends `HttpClientBuilder`), there could be some hidden issues. Also people tend to use complex caching solutions like EhCache, Memcached, Redis, Hazelcast, etc.., so naturally they would want to have them supported as in the core as well. I don't want to open a Pandora's box.

What do you think about starting to work on this feature as an external Dropwizard module? In the core we could expose some internals of `io.dropwizard.client.HttpClientBuilder` for overriding, if you want to reuse some of its functionality.

gurrie:
Hi,

Many thanks for the feedback. My thinking was not to replace the existing http client with the caching client but rather to provide an option on the JerseyClientBuilder to create a caching client or a standard client. That way the developer would be free to chose the implementation based on their needs. 

As you mentioned this would require pulling in the httpclient-cache module as a dependancy. If you are still uncomfortable with putting it in the core, I would be happy to create it as an external module, but as you mentioned it would require a few changes to the existing code to support the change. 

I am happy to describe what I would require in that case if you see this as being beneficial? It would also potentially be beneficial to instrument the cache so we could provide metrics on the cache usage etc.

Regards
[...]

PR #1668, opened 2016-08-04
Allow HttpClientBuilder subclasses to customize the underlying builder

evnm:
This change adds an overridable method which will allow `io.dropwizard.client.HttpClientBuilder` subclasses to customize the underlying `org.apache.http.impl.client.HttpClientBuilder` that is used. For example, this should allow the injection of [HTTP client caching](https://hc.apache.org/httpcomponents-client-ga/tutorial/html/caching.html), as proposed in #1633.


arteam:
Looks good to me, thanks for contributing!

PR #1594, opened 2016-06-12
JUnit rule for testing database interactions

rbarbey:
In several projects, I wrote a test code which allows testing of database interactions independently of `@UnitOfWork` annotations only present on resource methods. So I thought this might be useful to others as well. I rewrote it to be similar to other existing JUnit rule that are part of the `dropwizard-testing` module. I hope I followed all the coding conventions.

I added Hibernate and H2 to the dependencies of the testing module using scopes which hopefully don't pollute the classpath of users who are not interested in database testing (`provided` and `test`, respectively).

I also added a small section to the "Testing Dropwizard" part of the manual explaining how to use this rule.

Let me know if you have questions are if anything is missing.

PR #1669, opened 2016-08-04
JUnit rule for testing database interactions (redux)

arteam:
This pull requests is based on the great work of @rbarbey in #1594. There were too many changes I wanted to point out, so I decided to open a new pull request. This is basically the same JUnit rule, but with several implementation improvements (comments, a more apt API , additional tests, conformity with `HibernateBundle`, etc..).


dropwizard-testing/src/test/java/io/dropwizard/testing/junit/DAOTestRuleTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        // when we prepare an update of that entity
+        testEntity.setDescription("newDescription");
+        try {
+            // ... but cause a constraint violation during the actual update
+            daoTestRule.inTransaction(() -> {
+                persist(testEntity);
+                persist(new TestEntity(null));
+            });
+            Assert.fail("Expected a constraint violation");
```
jplock: Would you want to use `failBecauseExceptionWasNotThrown(ConstraintViolationException.class)` here instead?

jplock:
One minor comment in the tests otherwise this looks great!

dropwizard-testing/src/test/java/io/dropwizard/testing/junit/DAOTestRuleTest.java
arteam: Yes, thanks for pointing this out!

rishair:
I got to this PR a couple days late but it seems to assume usage of the Hibernate ORM. We have a similar rule and set of classes that provide support for JDBI based DAOs that I'm hoping to contribute back eventually. Would it make sense to package the changes in this PR under a more specific namespace and perhaps extract out common interfaces to leave room for other DB libraries to fill in? 

Although, after going through this diff I'm finding it difficult to find ways to converge on a single interface because usages of Hibernate ORM vs JDBI are very different. Support for multiple DB access methods is just something to keep in mind as the story for testing continues to unfold.

arteam:
Hey, @rishair! For JDBI there is (shameless plug) [jdit|https://github.com/arteam/jdit], which is implemented as a separate library. It uses another approach (with a custom test runner), but provides a rather simple testing approach for the developer (similar to Mockito). I used it successfully in several production projects and should be fairy stable.

Issue #1671, opened 2016-08-05
JerseyClient ExecutorService not Shutdown gracefully


saumitrabhave:
Hi:

I am looking at https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardExecutorProvider.java

which provides an executor service to Jersey Client for Async Execution. I think that, because of NO-OP implementation of dispose method, any async requests waiting in queue are just killed like that.

It would be good to have an implementation for dispose which calls shutdown and awaitTermination(With Configurable wait time maybe). This will make sure that any async() WebTargets submitted already are given time to complete in case DW Service is being closed/terminated.

evnm:
I think this is a valid concern, but I don't think it's as simple as invoking `ExecutorService#shutdown` within the `dispose` method. Doing so is the right thing to do _as long as the `ExecutorService` isn't shared_ (see third paragraph in [ExecutorServiceProvider Javadoc](https://jersey.java.net/apidocs/2.22/jersey/org/glassfish/jersey/spi/ExecutorServiceProvider.html)).

In this case, we can't guarantee that the service isn't shared. For example, the `ExecutorService` used in `JerseyClientBuilder` may be passed in via the appropriate `using` method, in which case that executor could be used in other parts of an application.

All of this is to say that a fix for this issue is still possible, it would just need to be nuanced. I can put a change together and we can take further discussion to a PR.

saumitrabhave:
ohh, I missed that use-case,
 
one alternate place can be https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java#L349 

here we can shutdown the executorService explicitly after client.close(), as it is the environment shutdown() callback everything has to go away anyways i guess.

or some mix of both of these approaches, to handle shutdown in dispose() when custom execService is not provided via using() and do it in lifecycle shutdown if custom execService is provided to builder.

I think you will have better idea of whats needed, please link this issue to PR so I can add my 2 cents.

arteam:
Closed via #1676

PR #1676, opened 2016-08-05
Gracefully shut down default ExecutorService used by JerseyClientBuilder

evnm:
Addresses #1671 & #1679

This change introduces a package-private `ExecutorService` decorator which is used to signal to `DropwizardExecutorProvider` that the executor should be shut down upon its disposal by the Jersey runtime. This enables graceful shutdown of the default `ExecutorService`s created within `JerseyClientBuilder` if the builder isn't configured with one.

Included tests for this graceful shutdown behavior, as well as its absence in the case where a "non-disposable" `ExecutorService` is used.

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardExecutorProvider.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardExecutorProvider.java
@@ -1,10 +1,38 @@
 package io.dropwizard.client;
 
+import com.google.common.util.concurrent.ForwardingExecutorService;
 import org.glassfish.jersey.spi.ExecutorServiceProvider;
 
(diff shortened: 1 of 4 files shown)


evnm:
Thanks for the feedback, @saumitrabhave. Good suggestion. I updated the change accordingly.

Since a shutdown grace period has to be threaded through `JerseyClientBuilder` via a public method, this change will now be slated for the 1.1.0 release.

I also took the liberty of tacking on the `@ClientAsyncExecutor` annotation proposed in #1679, since I'm already in that file.

saumitrabhave:
Thanks, @evnm .

arteam:
LGTM!

arteam:
Thanks @saumitrabhave for reporting these issues and helping to fix them!

PR #1677, opened 2016-08-06
Set FAIL_ON_UNKNOWN_PROPERTIES in DefaultConfigurationFactoryFactory

cakofony:
The logic is moved to DefaultConfigurationFactoryFactory from
YamlConfigurationFactory, making it easier to modify with a
custom ConfigurationFactoryFactory implementation.

There are some (uncommon) use cases where it's benefitial to
update an application without modifying the configuration, which
may fail if the application configuration has removed a configuration
property.

### DIFF
--- a/dropwizard-configuration/src/main/java/io/dropwizard/configuration/DefaultConfigurationFactoryFactory.java
+++ b/dropwizard-configuration/src/main/java/io/dropwizard/configuration/DefaultConfigurationFactoryFactory.java
@@ -1,5 +1,6 @@
 package io.dropwizard.configuration;
 
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import javax.validation.Validator;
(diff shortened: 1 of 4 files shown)


evnm:
I agree that this makes sense as something that the library should offer a means of overriding.

Instead of simply moving the location of the hardcoded `ObjectMapper#enable` call, what if we added a `protected ObjectMapper configureObjectMapper(ObjectMapper)` method to `DefaultConfigurationFactoryFactory`? That way, consumers could enable `ObjectMapper` functionality without having to touch the `create` method.

cakofony:
Thanks for the quick feedback!  I implemented your suggestion and added testing.
I'm not sure how to run the style tests locally, I might have to re-push in a couple minutes

dropwizard-configuration/src/main/java/io/dropwizard/configuration/DefaultConfigurationFactoryFactory.java
```diff
@@ -11,6 +12,23 @@
...
     }
+
+    /**
+     * Provides additional configuration for the {@link ObjectMapper} used to read
+     * the configuration. By default {@link DeserializationFeature#FAIL_ON_UNKNOWN_PROPERTIES}
+     * is enabled to protect against misconfiguration.
+     *
+     * @param objectMapper template to be configured
+     * @return configured object mapper
+     */
```
evnm: A+ Javadoc, would read again

Issue #1672, opened 2016-08-05
requestLog formats with %reqAttribute{someAttibuteName} work intermittently


tjcutajar:
When setting a request attribute in a resource and configuring it to be logged via a custom logFormat with %reqAttribute{someAttibuteName} the expected value only gets logged from time to time.

I've done a little digging and it seems [`prepareForDeferredProcessing()`](https://github.com/qos-ch/logback/blob/master/logback-access/src/main/java/ch/qos/logback/access/spi/AccessEvent.java#L576) gets called twice, once from [`io.dropwizard.request.logging.async.AsyncAccessEventAppenderFactory`](https://github.com/dropwizard/dropwizard/blob/3d746d21c6e6656a59b26c2a7e92f19e1bb2e9b1/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/async/AsyncAccessEventAppenderFactory.java#L22) and from [`ch.qos.logback.core.OutputStreamAppender`](https://github.com/qos-ch/logback/blob/658963544cc08059fc3cec48f5d35f650de5f6df/logback-core/src/main/java/ch/qos/logback/core/OutputStreamAppender.java#L205) where there's a race to set the attributeMap.

DW 1.0

evnm:
Good catch. Pity this Logback codepath isn't thread-safe. Perhaps they're wanting to avoid synchronization because logging tends to be such a hot path. The `DeferredProcessingAware` interface could use a Javadoc.

If I'm understanding the issue correctly, then a fix would be to simply remove the `prepareForDeferredProcessing` invocation in Dropwizard: https://github.com/dropwizard/dropwizard/pull/1678

arteam:
Closed via #1678

tjcutajar:
@evnm @arteam On review this looks a bit trickier than just removing the `prepareForDeferredProcessing()` call. If you run the hello world application in dropwizard-example for instance and hit the /hello-world resource the logging of the requests themselves become inconsistent.

tjcutajar:
[...]

PR #1678, opened 2016-08-07
Fix request attribute-related race condition in Logback request logging

evnm:
Addresses #1672 

`i.d.request.logging.async.AsyncAccessEventAppenderFactory` invokes `AccessEvent. prepareForDeferredProcessing`, which triggers a race with an identical invocation [within Logback itself](https://github.com/qos-ch/logback/blob/658963544cc08059fc3cec48f5d35f650de5f6df/logback-core/src/main/java/ch/qos/logback/core/OutputStreamAppender.java#L205). The reported issue indicates that the race manifests in inconsistent logging behavior when request attributes are set within a resource class.

This change removes the `prepareForDeferredProcessing` call from dropwizard-request-logging. As far as I know, all `IAccessEvent`s encountered by this class will be of concrete type `AccessEvent` and all non-async appenders extend `OutputStreamAppender`, so the `prepareForDeferredProcessing` method should always be invoked exactly once.

### DIFF
--- a/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/async/AsyncAccessEventAppenderFactory.java
+++ b/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/async/AsyncAccessEventAppenderFactory.java
@@ -10,17 +10,11 @@ import io.dropwizard.logging.async.AsyncAppenderFactory;
 public class AsyncAccessEventAppenderFactory implements AsyncAppenderFactory<IAccessEvent> {
 
     /**
-     * Creates an {@link AsyncAppenderFactory} of type {@link IAccessEvent} that prepares events
-     * for deferred processing
-     * @return the {@link AsyncAppenderFactory}
+     * Creates an {@link AsyncAppenderBase} of type {@link IAccessEvent}.
+     * @return the {@link AsyncAppenderBase}
      */
     @Override
     public AsyncAppenderBase<IAccessEvent> build() {
-        return new AsyncAppenderBase<IAccessEvent>() {
-            @Override
-            protected void preprocess(IAccessEvent event) {
-                event.prepareForDeferredProcessing();
-            }
-        };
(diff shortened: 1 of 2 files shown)


arteam:
It's very strange to step on those thread-safety issues with Logback (or the SLF4J bridge). Let's see if the fix helps, I even don't known how to write a unit to reproduce it.

dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/async/AsyncAccessEventAppenderFactory.java
```diff
@@ -16,11 +16,6 @@
      */
     @Override
     public AsyncAppenderBase<IAccessEvent> build() {
-        return new AsyncAppenderBase<IAccessEvent>() {
-            @Override
-            protected void preprocess(IAccessEvent event) {
-                event.prepareForDeferredProcessing();
-            }
-        };
+        return new AsyncAppenderBase<IAccessEvent>();
```
arteam: Could you please also update the Javadoc for this method? In looks a lit bit wrong for me.

* It should link to `AsyncAppenderBase` rather than to `AsyncAppenderFactory` (not-related issue to this change)
* I think we should remove a claim about deferred processing.

evnm: Definitely, thanks for mentioning. It would be nice to collapse this and `io.dropwizard.logging.async.AsyncLoggingEventAppenderFactory` into a single class parameterized on the type of event, since they're now almost identical. This would be a breaking change, though.

I took the liberty of updating `AsyncLoggingEventAppenderFactory`'s Javadoc too, since it also incorrectly links to `AsyncAppenderFactory`.

arteam:
Thanks, @evnm!

arteam:
Cherry-picked to the 1.0,x branch.

PR #1682, opened 2016-08-10
Update maven archetype version in getting started

danielbcorreia:


joschi:
@danielbcorreia Thanks!

PR #1683, opened 2016-08-10
Add ResourceCollection support ServletEnvironment

afrin216:
PR to support adding a collection of static resource folders in `ServletEnvironment`. 

Basically this was done to negate the need for consumers to have a dependency on `jetty-util` library, as they would need to create a `ResourceCollection` if they have multiple resources for serving static content.

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/setup/ServletEnvironment.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/setup/ServletEnvironment.java
@@ -148,6 +148,17 @@ public class ServletEnvironment {
     }
 
     /**
+     * Sets the base resources for this context.
+     *
+     * @param resources A list of strings representing the base resources to serve static
+     *                  content for the context. Any string accepted by Resource.newResource(String)
+     *                  may be passed and the call is equivalent to {@link #setBaseResource(Resource...)}}
+     */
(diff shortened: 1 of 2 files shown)


dropwizard-jetty/src/main/java/io/dropwizard/jetty/setup/ServletEnvironment.java
```diff
@@ -130,26 +131,45 @@ public void setProtectedTargets(String... targets) {
...
     }
 
     /**
+     * Sets the base resources for this context.
+     *
+     * @param resources A list of strings representing the base resources to serve static
+     *                  content for the context. Any string accepted by Resource.newResource(String)
+     *                  may be passed and the call is equivalent to {@link #setBaseResource(Resource...)}}
+     */
+    public void setResourceBase(String... resources) {
```
mattnelson: The method name should match the wrapped name `setBaseResource `

afrin216: Addressed in https://github.com/dropwizard/dropwizard/pull/1683/commits/a6b55d1c1dc48646370290a0d1d6c1fa0a04f633

arteam:
Looks good! Thanks for the contribution!

PR #1685, opened 2016-08-12
Make GZipHandler sync-flush configurable #1673

### LINKED ISSUE #1673: It's too hard to turn off Gzip selectively
I wanted to use [Jersey's SSE](https://jersey.java.net/documentation/latest/sse.html) support with Dropwizard 1.0 however just as described in [JERSEY-3000](https://java.net/jira/browse/JERSEY-3000) Dropwizard by _default_ has GZip-Compression. I somehow didn't want to go the easy way (_just turn off Gzip_), so I tried _a lot_:

* modifying response headers in JAX-RS by `@Context`ing all kind of ServletContexts
* intercepting in Jersey with a [Pre-Matching filter](https://jersey.java.net/documentation/latest/filters-and-interceptors.html#d0e9678) as well as a name-bound interceptor
* servlet filters

Digging deeper I finally hit the underlying [Jetty `GzipHandler`](http://www.eclipse.org/jetty/documentation/9.4.x/gzip-filter.html) which happens to have the perfect `excludedPath` (exclude a path from Gzip treatment) *and* [`excludedMimeTypes`](http://download.eclipse.org/jetty/9.3.9.v20160517/apidocs/org/eclipse/jetty/server/handler/gzip/GzipHandler.html) (exclude a MIME type) methods that would not require me to re-create the _positive_ whitelist of stuff that shall be Gzipped (supported in configuration)...

I guess this should be the result of `getApplicationContext().getGzipHandler()` but this one is `null` no matter where in the lifecycle. I ended up learning something about Jetty and came up with this (which works nicely):

```java
//env.getApplicationContext().getGzipHandler(); // <-- this is always null? why?!?
env.lifecycle().addServerLifecycleListener(new ServerLifecycleListener() {
	@Override
	public void serverStarted(Server server) {
		Handler handler = server.getHandler();
		LOG.debug("Disable GZip for SSE: Starting with handler {}", handler);
		while (handler instanceof HandlerWrapper) {
			handler = ((HandlerWrapper) handler).getHandler();
			LOG.debug("Disable GZip for SSE: Found handler {}", handler);
			if (handler instanceof BiDiGzipHandler) {
				LOG.info("Excluding mime-type {} from gzip compression handler", SseFeature.SERVER_SENT_EVENTS);
				((BiDiGzipHandler) handler).addExcludedMimeTypes(SseFeature.SERVER_SENT_EVENTS);
			}
		}
	}
});
```

I really think either MIME type `text/event-stream` should be excluded by default **or** flushing fixed somehow **or** at the blacklist be exposed for configuration **or** some special annotation to whitelist... No idea what is the most _idiomatic_ way...

LeDominik:
This would resolve #1673 by making sync-flush behaviour of GZipHandler configurable

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -95,7 +95,9 @@ GZip
 +---------------------------+---------------------+------------------------------------------------------------------------------------------------------+
 | gzipCompatibleInflation   | true                | If true, then ZLIB inflation(decompression) will be performed in the GZIP-compatible mode.           |
 +---------------------------+---------------------+------------------------------------------------------------------------------------------------------+
-
+| syncFlush                 | false               | The flush mode. Set to true if the application wishes to stream (e.g. SSE) the data,                 |
+|                           |                     | but this may hurt compression performance (as all pending output is flushed).                        |
++---------------------------+---------------------+------------------------------------------------------------------------------------------------------+
 
 .. _man-configuration-requestLog:
 
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/GzipHandlerFactory.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/GzipHandlerFactory.java
@@ -92,6 +92,8 @@ public class GzipHandlerFactory {
 
     private boolean gzipCompatibleInflation = true;
 
+    private boolean syncFlush = false;
+    
     @JsonProperty
     public boolean isEnabled() {
         return enabled;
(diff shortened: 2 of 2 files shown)


evnm:
Change looks great. Thanks for the contribution!

I guess we can punt on adding similar configurability for path/MIME-type inclusion/exclusion as a separate change, if there's still any interest in doing so.

LeDominik:
I think this solution is great, it's about streaming and not about MIME types so for this purpose the change just feels right 👍

victornoel:
@evnm is this present in 1.1.x only?

arteam:
Yes, but we can backport it to 1.0.* if it's needed.

victornoel:
@arteam for now I enable it by myself with this workaround:
```java
        environment.lifecycle().addServerLifecycleListener(server -> {
            Handler handler = server.getHandler();
            while (handler instanceof HandlerWrapper) {
                handler = ((HandlerWrapper) handler).getHandler();
                if (handler instanceof BiDiGzipHandler) {
                    LOG.info("Setting sync flush on gzip compression handler");
                    ((BiDiGzipHandler) handler).setSyncFlush(true);
                }
            }
        });
```

So don't bother with it except if it's very easy and you plan to release many 1.0.x before 1.1.x is released :)
Thanks!

victornoel:
@arteam by the way, I think you forgot to add it to the changelog of the released 1.1.0 :)

arteam:
The improvement has been added to the changelog. Thanks for letting us know!

PR #1688, opened 2016-08-13
Use a List instead an array in YamlConfigurationFactory

arteam:
A minor improvement to the escaping code in `YamlConfigurationFactory`.

A good practice is to avoid manipulating arrays in favour of Java collections, because arrays are relatively dangerous to operate and provide a less convenient API.

See "Effective Java Item 25: Prefer lists to arrays".


jplock:
LGTM!

PR #1690, opened 2016-08-15
Allow overriding of configuration class

jhaber:
I'm opening this PR to get your thoughts on removing the final modifier from `Application#getConfigurationClass`. My use-case is creating a reusable subclass of `Application` so that we have less boilerplate and duplication in each of our REST APIs (of which we have ~400). I put our subclass in a gist [here](https://gist.github.com/jhaber/136d46c6c795117d5a5c56d035cac434#file-restapplication-java), a usage would look like:
```java
public class Main {
  
  public static void main(String... args) throws Exception {
    RestApplication.newBuilder().modules(new MyGuiceModule()).build().run(args);
  }
}
```
This works, but only as long as everyone is using the same configuration class. We would like to allow users to subclass our default configuration, which we could do by modifying our subclass to look like [this](https://gist.github.com/jhaber/136d46c6c795117d5a5c56d035cac434#file-restapplication2-java) (note the [override](https://gist.github.com/jhaber/136d46c6c795117d5a5c56d035cac434#file-restapplication2-java-L32-L35)). The only issue we're running into is that this method is final.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/Application.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/Application.java
@@ -40,7 +40,7 @@ public abstract class Application<T extends Configuration> {
      * @return the configuration class
      * @see Generics#getTypeParameter(Class, Class)
      */
-    public final Class<T> getConfigurationClass() {
+    public Class<T> getConfigurationClass() {
         return Generics.getTypeParameter(getClass(), Configuration.class);
     }
 


evnm:
This is a great use case.

`Application#getConfigurationClass` has been `final` since its inception within `com.yammer.dropwizard.Service` in 2011, so I think it's safe to say that this was done as a precaution rather than to prevent a regression that we should be mindful of. Thankfully, [removing final from a method doesn't break ABI compatibility](http://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.4.17), so I think this would be a fine addition to 1.1.0.

For flavor, I noticed that [the equivalent method on `io.dropwizard.cli.ConfiguredCommand`](https://github.com/dropwizard/dropwizard/blob/v1.0.0/dropwizard-core/src/main/java/io/dropwizard/cli/ConfiguredCommand.java#L41) isn't declared `final`, so this change arguably improves consistency across the codebase, albeit in a marginal way.

I'm curious if any other maintainers will have different opinions on this change, so I'm going to let the PR bake for a day or two before merging.

arteam:
The change looks harmless to me. It doesn't break the compatibility and
opens the Application class to extension, which could be useful as shown in
the gist.
Am 16.08.2016 06:13 schrieb "Evan Meagher" <notifications@github.com>:

evnm:
Merged. @jhaber, thanks for the contribution,  and thanks @arteam for chiming in!

jhaber:
Thanks for the quick turnaround!

PR #1693, opened 2016-08-16
Update location of dropwizard-scala repository.

bretthoerner:
The `dropwizard-scala` docs currently point at my outdated repo, this PR points them to a modern one and I verified this was OK with the owner here: https://github.com/datasift/dropwizard-scala/issues/9

Alternatively this could be dropped from the docs entirely, up to y'all. :)


joschi:
@bretthoerner Thanks!

Issue #1680, opened 2016-08-08
Ability to return non-zero status codes from Commands


tburch:
I'd like to have the ability to return a non-zero exit code from a command. My current use case is that I run migrations prior to deploying my application, but would like to fail the application deployment if the migrations are not successfully run based on the status code returned from the migration command.

Would a change with this functionality added be welcomed?

evnm:
I think [it's important](https://monkey.org/~marius/unix-tools-hints.html) for processes to return sensible exit codes. Currently, doing so should be as simple as invoking `System.exit(exitCode)` in your `Command` implementation.

One way to improve upon this would be to modify `Command#run` to return an `int` which could be fed to `System.exit()` by the framework rather than force the user to invoke it manually. This unfortunately would be a breaking change.

Another way that I'm not a huge fan of would be to introduce an `Exception` type which encodes an exit status to be fed to `System.exit()` by Dropwizard's [CLI mechanism](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java). Maybe others disagree, but it seems wrong to rely on exception-handling to deal with exit status.

Happy to entertain any ideas you have, so fire away!

tburch:
I don't know what the "right" solution is, however I lean towards making `Command#run` return an `Optional<Integer>` used in ` System#exit` if present because it seems like a clean API design. I don't like that it's a breaking change.

I could also make an argument for creating something like a `CommandException` (extending `Exception`) that contains the the `Optional<Integer>` used in ` System#exit` if present since the primary use case for this is when we hit an exception case in the command and needed to exit appropriately. 

I'm happy to implement either solution. Any other input?

evnm:
I think the exception-as-nonzero-status-code-container idea is the best near-term option. I think the best _overall_ solution is to change `Command#run` to return an int, but we should punt on that until plans for a 2.x release train start whispering on the wind.

tburch:
I have the basic foundation comitted at https://github.com/tburch/dropwizard/tree/CommandExitCodes. What are your thoughts on trapping any exceptions from `ConfiguredCommand#run(Bootstrap<T>, Namespace, T)` and then re-throwing a `CommandException` with the trapped `Exception` and a non-zero exit code? The current behavior is that the `Application`/jvm just exits.

evnm:
[...]

PR #1695, opened 2016-08-17
Catch Throwable and return false if any command errors out

tburch:
Dropwizard should exit gracefully with a non-zero (`1`) status code in the event that a command throws an `Exception` instead of just exiting the `Application`/JVM. This PR is the result of https://github.com/dropwizard/dropwizard/issues/1680#issuecomment-240132023

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
@@ -1,6 +1,5 @@
 package io.dropwizard.cli;
 
-import io.dropwizard.configuration.ConfigurationException;
 import io.dropwizard.setup.Bootstrap;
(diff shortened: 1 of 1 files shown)


evnm:
Since the last two `catch`s are identical, let's collapse them. You can do this by simply replacing `ConfigurationException` with `Throwable` (and removing the import). A comment to the effect of "unexpected exceptions result in non-zero exit status of the process" is probably a good idea, too.

tburch:
@evnm updated the commit.

evnm:
New commit doesn't seem to have been pushed.

tburch:
@evnm my bad. Pushed the wrong branch after rebasing.

evnm:
Thanks for taking this one home!

PR #1698, opened 2016-08-18
Log Jetty initialized SSLContext not the Default

nickbabcock:
Instead of listing supported cipher suites, protocols, etc of the default
SSLContext algorithm, log what Jetty's will actually use, which as of this
commit is a "TLS" SSLContext. This will also future proof the logging if
this property ever becomes configurable in Dropwizard.


evnm:
LGTM

Issue #1597, opened 2016-06-15
Default showDetails in JsonProcessingExceptionMapper to true


jeremygehrshg:
After using Dropwizard for 5 months at work now, I discovered this option recently and it's saved hours of debugging and digging through stack traces since then.  It doesn't look like anything is gained by _not_ providing the error field and what the validation issue is, so I think showDetails should be set to true by default.  

Here is the class:
https://github.com/dropwizard/dropwizard/blob/ae9844a80a1ab273b7ccdb35d86fa624ee915d7a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java

The same theme might also might apply to LoggingExceptionMapper, but I'm less sure that exposing e.getMessage() or something similar to the client would be the desired behavior.

evnm:
Looking at the history of the file in question, the `showDetails` guard [was introduced in 2014](https://github.com/dropwizard/dropwizard/commit/4fec8a95cc4de3a8f2f87ea4e3437616f19a0d41). @joschi, do you remember what the motivator was for this?

If I were to guess, I'd assume it was to prevent cases where a pathological client could induce large response payloads if a developer was longwinded when writing exception messages. Debatable whether this is a realistic concern, but this file's history shows a similar tack being taken with regard to logging: https://github.com/dropwizard/dropwizard/commit/9f2df976b2b1fda7dfb19fbdfd0c670914791f8a

joschi:
@evnm The idea was to prevent information leakage to malicious clients.

jeremygehrshg:
@joschi How about default to showing the error message without the classpath and have the showDetails flag append the offending class?  I don't think an error message like this wouldn't be useful for anything malicious.

```
{
  "code": 400,
  "message": "Unable to process JSON",
  "details": "Unrecognized field \"foo\" not marked as ignorable"
}
```

tmack8001:
I agree with @jeremygehrshg that providing some information regarding the failure wouldn't be "leaking potentially insecure implementation details" to a malicious actor. Having this available to be configured would greatly ease my integration efforts internally (e.g. having showDetail=true in a development and integration environment while leaving off in a staging and production environment). Instead I'm currently just re-registering this JsonExceptionMapper with jersey a second time and ignoring the warning. 

I'm not sure if defaulting to "true" is really the correct approach, but definitely exposing a way for downstream users to set it up in their configuration files would be ideal.

arteam:
[...]

PR #1705, opened 2016-08-23
Config override for showDetails in JsonProcessingExceptionMapper

tmack8001:
From a discussion in https://github.com/dropwizard/dropwizard/issues/1597 there is some want to have the ability to override the showDetails flag when registering the `JsonProcessingExceptionMapper` with jersey. If I just override the existing registered component in my application I receive the following warning log event at application startup which is less than ideal, though also not anything to be concerned about as I am intentionally re-registering this component.

```
{
  "@timestamp":"2016-08-23T03:07:17.054+00:00",
  "@version": 1,
  "message":"The following warnings have been detected: WARNING: Cannot create new registration for component type class io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper: Existing previous registration found for the type.\n",
  "logger_name":"org.glassfish.jersey.internal.Errors",
  "thread_name":"main",
  "level":"WARN",
  "level_value":30000
}
```

The approach I took was to:

1. Add config property to AbstractServerFactory to expose setting of this boolean property
2. Default to existing behavior `showDetails=false` (the default constructor in `JsonProcessingExceptionMapper`).
3. Added tests for both default and override cases (which required me to expose the value of `showDetails` within the `JsonProcessingExceptionMapper` via a getter).

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java
@@ -250,6 +250,8 @@ public abstract class AbstractServerFactory implements ServerFactory {
 
     private Boolean registerDefaultExceptionMappers = Boolean.TRUE;
 
+    private Boolean detailedJsonProcessingExceptionMapper = Boolean.FALSE;
+
     private Duration shutdownGracePeriod = Duration.seconds(30);
 
     @NotNull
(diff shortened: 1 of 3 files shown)


arteam:
I think it makes sense to provide this feature, because it could be useful in staging/development environments to catch integration errors between the client and server side early and it could be changed in the config file depending on the environment.

arteam:
@tmack8001, Thanks for the contribution!

tmack8001:
yeah no problem @arteam I've been using this setup for several months and definitely helps to reduce the number of integration questions I get from other developers

PR #1715, opened 2016-09-01
Allow for passing a filter as a class

### LINKED ISSUE #1630: @Context or @Session injection not working in ContainerRequestFilter
Hi,

According to Jersey people, this (at least with `@Context`) should be supported since Jersey 2.4 (see [JERSEY-1960](https://java.net/jira/browse/JERSEY-1960) but I'm trying with dropwizard 0.9.3 and and the context simply does not get injected. I also tried with the dropwizard's `@Session` annotation but it's the same. 

Is that normal or is there a bug somewhere?

Note: my filter is registered via `AuthDynamicFeature` (because it's an alternative authentication filter based on the content session, maybe it's not the best way to do that?). 

Thanks!

victornoel:
Jersey does not behave exactly the same when registering
object and class during a DynamicFeature configuration.

This commit allows for both behaviour to be exploited in
AuthDynamicFeature.

This fixes #1630.

@evnm @psamsotha please review and tell me if we need more doc that what I added :)

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
@@ -21,13 +21,26 @@ import java.lang.annotation.Annotation;
  * <p>If authorization is not a concern, then {@link RolesAllowedDynamicFeature}
  * could be omitted. But to enable authentication, the {@link PermitAll} annotation
  * should be placed on the corresponding resource methods.</p>
+ * <p>Note that registration of the filter will follow the semantics of
+ * {@link FeatureContext#register(Class)} and {@link FeatureContext#register(Object)}:
+ * passing the filter as a {@link Class} to the {@link #AuthDynamicFeature(Class)}
+ * constructor will result in dependency injection, while objects passed to
+ * the {@link #AuthDynamicFeature(ContainerRequestFilter)} will be used directly.</p>
  */
 public class AuthDynamicFeature implements DynamicFeature {
 
     private final ContainerRequestFilter authFilter;
 
+    private final Class<? extends ContainerRequestFilter> authFilterClass;
+
     public AuthDynamicFeature(ContainerRequestFilter authFilter) {
         this.authFilter = authFilter;
+        this.authFilterClass = null;
+    }
+
+    public AuthDynamicFeature(Class<? extends ContainerRequestFilter> authFilterClass) {
+        this.authFilter = null;
+        this.authFilterClass = authFilterClass;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
```diff
@@ -21,13 +21,26 @@
  * <p>If authorization is not a concern, then {@link RolesAllowedDynamicFeature}
  * could be omitted. But to enable authentication, the {@link PermitAll} annotation
  * should be placed on the corresponding resource methods.</p>
+ * <p>Note that registration of the filter will be done following the semantics of
```
evnm: Nit: "Note that registration of the filter will follow..."

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
```diff
@@ -21,13 +21,26 @@
...
+ * {@link FeatureContext#register(Class)} and {@link FeatureContext#register(Object)}:
+ * passing the filter as a {@link Class} to the {@link #AuthDynamicFeature(Class)}
+ * constructor will do dependency injection while passing the filter as an object to
```
evnm: Nit: "...will inject the filter as an object to the..."

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
```diff
@@ -21,13 +21,26 @@
...
+ * the {@link #AuthDynamicFeature(ContainerRequestFilter)} constructor won't.</p>
```
evnm: Nit: "passing the filter as a {@link Class} to the {@link #AuthDynamicFeature(Class)} constructor will result in dependency injection, while objects passed to {@link #AuthDynamicFeature(ContainerRequestFilter)} will be used directly."

evnm:
Change looks good! Just a couple documentation nit-picks.

victornoel:
Should be alright now :)

evnm:
Thanks for taking care of this, @victornoel.

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
meilleureVie: how to make it work.
I need to register a custom intance of requestFilter from dynamic feature and want to get all injected objects with @Context or @inject injected.
In my case, I need a custom instance of requestFilter because i override RolesAllowedDynamicFeature to support i18n response output. so instead to instanciate your RolesAllowedRequestFilter within AuthorizationDynamicFeature i need to inject a messageInterpolator that i have created and binded

meilleureVie: please replace instead by despite the fact of. Sorry for my bad english

victornoel: @meilleureVie please open a ticket or ask on the google group for help, this is a merged PR.

meilleureVie:
Ok

Le lun. 16 sept. 2019 à 12:23, Victor Noël <notifications@github.com> a
écrit :

Issue #1716, opened 2016-09-02
When logging sub-resource that is polymorphic Object, DropwizardResourceConfig causes NPE


apatrida:
When loging sub-resources that are polymorphic and are methods returning a base class or Object that is not identifiable as a resource, the call at line 164 (branch 1.x) of 

```
populate(basePath, klass, isLocator, Resource.from(klass), endpointLogLines);
```

causes the populate call to NPE at same file line 173 because the `Resource.from(kclass)` returns NULL and there is no null checking.  

Therefore this completely breaks using polymorphic sub resources just due to logging bug.

apatrida:
this originated from: http://stackoverflow.com/a/39295594/3679676

would be good to update that SO when this is resolved.

evnm:
Hello Jayson. This is an interesting issue.

Subresources are supported by Dropwizard, and I think it'd be worthwhile to [add a test case to `DropwizardResourceConfigTest` that illustrates this](https://github.com/dropwizard/dropwizard/pull/1717). As mentioned by the SO questioner, the subresource function works if its return type is a concrete resource class. As shown by the test in #1717, subresource locators can point to subclasses, so a good approach could be to bound the expected polymorphic resource class with an interface.

At the risk of seeming heretical, I suspect that the problem stems from the usage of the `Any` type rather than a particular deficiency in Dropwizard or Jersey. In Scala (which I'm far more familiar with than Kotlin), the analogous `Any` type is the type system's wildcard lower bound used as an alternative to `j.l.Object` to encompass primitive types as well as objects. Thus it's arguably more of a type system formalism than a type that should be used in practice. If the case is the same in Kotlin, then I don't see how using `Object` would be any worse than using `Any`, not to mention that using `Object` wouldn't result in this NPE.

apatrida:
@evnm Any in Kotlin will compile the method signature to ByteCode as:

From source:
```
 fun foo(): Any 
```

actual bytecode
```
public final foo()Ljava/lang/Object;
  @Lorg/jetbrains/annotations/NotNull;() // invisible
```

As you can see from a perspective of Dropwizard and anything Java, it sees Object.  In fact the user actually tried using `java.lang.Object` and hit the same error.  

And it is intended that `Any` functions as both a type system bound and also as an alias for `java.lang.Object`.

So this error should be easily reproducible in Java with `java.lang.Object` return type.

apatrida:
[...]

PR #1717, opened 2016-09-02
Illustrate subresource method usage in DropwizardResourceConfigTest

evnm:
This adds a test case to `DropwizardResourceConfigTest` that ensure that subresource methods are supported. This functionality was called into question in #1716.


nickbabcock:
Actually, I was able to reproduce the error, I'll have a PR up in a couple mins.

PR #1718, opened 2016-09-02
Fix NPE of non-resource sub-resource methods

### LINKED ISSUE #1716: When logging sub-resource that is polymorphic Object, DropwizardResourceConfig causes NPE

nickbabcock:
Closes #1716
Supersedes #1717

Whether it is a good idea to have a method return `Object` can be left up to discussion, but the code should not throw a NPE.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -186,7 +186,11 @@ public class DropwizardResourceConfig extends ResourceConfig {
                         final Class<?> erasedType = !responseType.getTypeBindings().isEmpty() ?
                                 responseType.getTypeBindings().getBoundType(0).getErasedType() :
                                 responseType.getErasedType();
-                        populate(path, erasedType, true, endpointLogLines);
+                        if (Resource.from(erasedType) == null) {
+                            endpointLogLines.add(new EndpointLogLine(method.getHttpMethod(), path, erasedType));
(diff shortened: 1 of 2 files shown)


evnm:
Ah, I see. The issue occurs when the resource method lacks an HTTP verb annotation. When I change the interface-returning resource method in #1717 to instead return `Object`, the tests still pass because I'd added `@GET`.

Thanks for taking care of this, Nick. LGTM

apatrida:
@nickbabcock 
> Whether it is a good idea to have a method return Object can be left up to discussion, but the code should not throw a NPE.

This judgement was already made by Jersey as promoted not just acceptable.  As you see in https://jersey.java.net/documentation/latest/jaxrs-resources.html#d0e2496 there is a section talking about polymorphism that shows using `Object` (or in our case `Any`) as a valid model.  So it is recommended by Jersey and yet crashed Dropwizard.  (all about sub-resources here: https://jersey.java.net/documentation/latest/jaxrs-resources.html#d0e2592)

nickbabcock:
Ah sorry, my statement was misleading. Everything that Jersey allows, Dropwizard should allow. I meant to convey that the return type should be as specific as possible 😝

PR #1719, opened 2016-09-02
Fix occasional null comparison introduced in #1718

nickbabcock:
😊

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -6,6 +6,7 @@ import com.fasterxml.classmate.ResolvedType;
 import com.fasterxml.classmate.TypeResolver;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ComparisonChain;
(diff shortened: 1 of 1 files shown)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -232,7 +233,7 @@ public String toString() {
         public int compare(EndpointLogLine endpointA, EndpointLogLine endpointB) {
             return ComparisonChain.start()
                 .compare(endpointA.basePath, endpointB.basePath)
-                .compare(endpointA.httpMethod, endpointB.httpMethod)
+                .compare(endpointA.httpMethod, endpointB.httpMethod, Ordering.natural().nullsLast())
```
evnm: According to [the Javadoc](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Ordering.html#nullsLast()), we should use `Comparator.nullsLast(Ordering.natural())` here.

nickbabcock:
Ok updated, let me know if I got it right. Third time is the charm :wink:

evnm:
Thanks, Nick! Nice job catching this one.

PR #1721, opened 2016-09-07
JerseyClientBuilder can create rx-capable client

nickbabcock:
Jersey allows for clients to have a reactive api if created appropriately:
https://jersey.java.net/documentation/latest/rx-client.html

This functionality for creating an rx-capable client is not currently in
dropwizard, which may lead users down the wrong path. They'll try wrapping
the dropwizard configured client like so (Java8 example):

```java
RxCompletionStage.from(client);
```

This is very bad because this will not use the executor carefully crafted
by dropwizard, but instead use `ForkJoinPool#commonPool`, which is, by
default, created to only have as many threads as there are numbers of CPUs - 1.
The end result is that many requests will end up blocking, the exact
opposite of what a user may think by using this API.

This commit introduces an rx-capable client that wraps the dropwizard
client with the dropwizard created executor so that both `async` and `rx`
methods can take advantage of the same thread pool, something that is more
intuitive than the current behavior.

Jersey supports four different reactive libraries and this commit works
with any of them and all future versions.

An alternative implementation would be to expose the executor that
dropwizard cofigures, but I find the approach taken in this commit better.

### DIFF
--- a/docs/source/manual/client.rst
+++ b/docs/source/manual/client.rst
@@ -34,7 +34,7 @@ To create a :ref:`managed <man-core-managed>`, instrumented ``HttpClient`` insta
         public HttpClientConfiguration getHttpClientConfiguration() {
             return httpClient;
         }
-        
+
         @JsonProperty("httpClient")
         public void setHttpClientConfiguration(HttpClientConfiguration httpClient) {
             this.httpClient = httpClient;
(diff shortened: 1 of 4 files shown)


jplock:
This looks good to me. Should we update the release notes and provide an example in the docs for this?

nickbabcock:
Ask and you shall receive 😉 

I'll update the release notes out-of-band once this PR is merged

nickbabcock:
Updated release 1.1 notes with this: 78dff9b

arteam:
Awesome, thanks!

PR #1723, opened 2016-09-08
validation: Pull parameter names from QueryParam, FormParam, and PathParam

cmicali:
Currently jersey resource paramere validation returns arg0, arg1, etc... for parameter names.  This PR adds a ParameterNameProvider that attempts to pull parameter names from QueryParam, FormParam, and PathParam annotations before falling back to the old behavior.


nickbabcock:
What version of dropwizard are you using, as I believe that this was resolved in 1.0?

The implementation foundation is potentially better as it hooks into `parameterNameProvider`

cmicali:
I'm using 1.0 - without this change I was still seeing arg0, arg1, etc.

nickbabcock:
Can you post an example or add a test case?

nickbabcock:
@cmicali, there needs to be something that's reproducible or this PR may be closed.

cmicali:
@nickbabcock re-submitted as #1734

PR #1729, opened 2016-09-16
Add `characterEncoding` and `useUnicode` to JDBC properties.

sam-berry:
I am proposing the addition of these two properties as part of the example because they required in the DW 1.0 app I am currently working on. It is worth noting that the DB is MySQL, so this is potentially only an issue with MySQL implementations. Omitting `characterEncoding` and `useUnicode` does not cause anything to fail but it prevents the app from properly enforcing UTF-8. A simple example includes persisting and retrieving the ™ symbol.

### DIFF
--- a/docs/source/manual/hibernate.rst
+++ b/docs/source/manual/hibernate.rst
@@ -77,6 +77,8 @@ Your application's configuration file will then look like this:
       # any properties specific to your JDBC driver:
       properties:
         charSet: UTF-8
+        characterEncoding: UTF-8
+        useUnicode: true
         hibernate.dialect: org.hibernate.dialect.PostgreSQLDialect
 
       # the maximum amount of time to wait on an empty pool before throwing an exception


jplock:
@sam-berry Do you know if all JDBC implementations support these options or just MySQL?

sam-berry:
@jplock I believe that I've confirmed that this is a MySQL specific issue. According to the Hibernate documentation ([see the second to last paragraph about arbitrary connection properties](https://docs.jboss.org/hibernate/orm/5.0/manual/en-US/html/ch03.html#configuration-hibernatejdbc)), and the MySQL documentation ([see "`useUnicode` and `characterEncoding`"](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-charsets.html)) these properties appear to be MySQL configuration that Hibernate just passes through.

So, this probably isn't appropriate for the DW docs since the example is using Postgres. I've added a [StackOverflow Q&A](http://stackoverflow.com/q/39523080/1756430) for this that will hopefully find others who are in the same position.

nickbabcock:
Thanks for digging into this, Sam. Since these properties don't exist and are not needed for Postgres, modifying the docs to reference these properties in a Postgres example will only confuse others, I'm afraid. This PR and stackoverflow should be enough documentation for mysql users. Googling "mysql java unicode" let's users know they have to specify it as a property 😉 

Let me know if you have any questions/concerns about closing this PR.

Issue #1727, opened 2016-09-15
Should OptionalMessageBodyWriter return 404 when the response is Optional.empty()


qinfchen:
Instead of returning NotFoundException for Optional.empty(), should it return 204 instead 404 based on the definition of the status codes below?

*404* client error
>The server has not found anything matching the Request-URI. No indication is given of whether the condition is temporary or permanent. The 410 (Gone) status code SHOULD be used if the server knows, through some internally configurable mechanism, that an old resource is permanently unavailable and has no forwarding address. This status code is commonly used when the server does not wish to reveal exactly why the request has been refused, or when no other response is applicable.

*204* success
> The server has fulfilled the request but does not need to return an entity-body, and might want to return updated metainformation. The response MAY include new or updated metainformation in the form of entity-headers, which if present SHOULD be associated with the requested variant.

>If the client is a user agent, it SHOULD NOT change its document view from that which caused the request to be sent. This response is primarily intended to allow input for actions to take place without causing a change to the user agent's active document view, although any new or updated metainformation SHOULD be applied to the document currently in the user agent's active view.

>The 204 response MUST NOT include a message-body, and thus is always terminated by the first empty line after the header fields.

>Wikipedia
The server successfully processed the request, but is not returning any content.

 >Status when wrapped responses (e.g. JSEND) are not used and nothing is in the body (e.g. DELETE).

http://www.restapitutorial.com/httpstatuscodes.html

nickbabcock:
This makes sense, but I'm always wary of changing status codes, so CC: @dropwizard/committers 

I think treating this as a bug could be the appropriate response.

evnm:
I think the short answer is "no". JAX-RS specifically intends for `NotFoundException` to trigger 404s.

See [the `javas.ws.rs` package summary](http://docs.oracle.com/javaee/7/api/javax/ws/rs/package-summary.html):

> NotFoundException—A runtime exception indicating a resource requested by a client was [not found](http://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Response.Status.html#NOT_FOUND) on the server.

nickbabcock:
I think @qinfchen was arguing that the writer shouldn't throw `NotFoundException` -- not that the behavior of `NotFoundException` should change.

evnm:
D'oh. My bad.

[...]

PR #1732, opened 2016-09-19
OptionalMessageBodyWriter to return `NoContentException` instead of `NotFoundException`

qinfchen:
fixes https://github.com/dropwizard/dropwizard/issues/1727

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/OptionalLongMessageBodyWriter.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/OptionalLongMessageBodyWriter.java
@@ -38,7 +38,7 @@ public class OptionalLongMessageBodyWriter implements MessageBodyWriter<Optional
                         MultivaluedMap<String, Object> httpHeaders,
                         OutputStream entityStream) throws IOException {
         if (!entity.isPresent()) {
-        	throw new WebApplicationException(Status.NO_CONTENT);
+            throw new WebApplicationException(Status.NO_CONTENT);
         }
 
         entityStream.write(Long.toString(entity.getAsLong()).getBytes(StandardCharsets.US_ASCII));
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/OptionalLongMessageBodyWriter.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/OptionalLongMessageBodyWriter.java
@@ -1,17 +1,19 @@
 package io.dropwizard.jersey.optional;
 
-import javax.ws.rs.NotFoundException;
-import javax.ws.rs.Produces;
-import javax.ws.rs.core.MediaType;
-import javax.ws.rs.core.MultivaluedMap;
-import javax.ws.rs.ext.MessageBodyWriter;
-import javax.ws.rs.ext.Provider;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 import java.nio.charset.StandardCharsets;
 import java.util.OptionalLong;
+
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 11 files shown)


qinfchen:
constructor taking the exception type? another option is provide a not found body writer for users to configure, but mark it as deprecated

nickbabcock:
Hmm, that wasn't my initial thought. I was thinking about something like a boolean `notFoundOnAbsent` for those that relied on the behavior.

dropwizard-core/src/test/java/io/dropwizard/server/DefaultServerFactoryTest.java
```diff
@@ -4,11 +4,14 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.io.CharStreams;
 import com.google.common.io.Resources;
+import com.sun.research.ws.wadl.Option;
```
joschi: Wrong import?

qinfchen: yep, fixing the style issues.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/OptionalLongMessageBodyWriter.java
```diff
@@ -36,7 +37,7 @@ public void writeTo(OptionalLong entity,
                         MultivaluedMap<String, Object> httpHeaders,
                         OutputStream entityStream) throws IOException {
         if (!entity.isPresent()) {
-            throw new NotFoundException();
+            throw new NoContentException("Empty content.");
```
nickbabcock: Not sure if I'm a fan of using this exception, as it looks like it forces the use of another exception mapper.

Have you tried doing just a `WebApplicationException` with the `Response.noContent().build()`?

qinfchen: good point. fixed.

nickbabcock: Actually, [looking at the code](https://github.com/dropwizard/dropwizard/blob/a0c5e2ab97a91200e09c91397fab0a0c359b8483/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/LoggingExceptionMapper.java#L22-L32), I believe even if you supplied a `WebApplicationException` with a fully built response, it will respond with a _json payload_ and a no content status 😢 

I'm not sure if this is the intended purpose of `WebApplicationException`. Let me ponder this.

nickbabcock:
Random hodgepodge of thoughts:

1. Current implementation of PR does work. 204's are returned with an empty body. I expressed concerns that [`LoggingExceptionMapper`](https://github.com/dropwizard/dropwizard/blob/89ab17b88345e76f5b35eb1fc526679235e2dc6b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/LoggingExceptionMapper.java) would interpret the `WebApplicationException` and return a 204 but contain a body. I'm unsure how this is possible, but tests should assert that the body is empty to ensure the behavior. Sorry for any doubt 😌 

2. Since this can be an opinionated issue, for Dropwizard 1.x, PR should keep previous behavior and allow the new behavior with code customization. Re-evaluate defaults for 2.0. Where to embed this customization is up for discussion. Though since #1768, new opportunities opened up. It used to be bittersweet when adding a new default exception mapper, but now that they can be easily overridden, it is no longer an issue. Spit-balling here but here's a possible game plan:

    1. Create `EmptyOptionalException` that is thrown at current location.
    2. Create default mapper to handle this exception
    3. If default exception is not good enough, register a different exceptionmapper

3. What happens / should happen when a `null` `Optional` is returned? EDIT: [looks like 204](http://stackoverflow.com/questions/14102797/how-to-throw-http-204-status-code-using-jersey-framework-in-restful-web-service), but need to confirm

jplock:
I'm going to close this as we just merged in #1784

PR #1734, opened 2016-09-21
Jersey Validation Improvements

cmicali:
This PR is a few changes we have been using for a while to improve the built-in jersey parameter validation error messages provided by Dropwizard.

**Better Enum Validation Messages**
Currently dropwizard resources return a 404 if a request is made with an invalid value for a parameter of an `Enum` type. This PR includes `ValidatingEnumParamConverterProvider` which provides converters for all enums used in method parameters. The converter then provides better error messages like _"query param choice must be one of [OptionA, OptionB, OptionC]"_ when invalid values are sent.

**Parameter Names in Parse Failure Messages**
Currently when an invalid value is passed into a *Param parameter dropwizard returns an error message like _"Parameter is not a number."_.  This PR includes `AbstractParamConverterProvider` 
along with a slight enhancement to the *Param classes that returns messages that include the name of the parameter that failed validation, like _"query param user_id is not a number."_

**Hibernate ParameterNameProvider**
Small change to move most of parameter naming logic from `ConstraintMessage` to `JerseyParameterNameProvider`. This implements the standard `ParameterNameProvider` interface that Hibernate uses to resolve parameter names during validation.

**New DurationParam**
Include a new `DurationParam` that can parse String input in the format supported by dropwizard's existing `Duration` class.


dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/DurationParam.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2016. Sense Labs, Inc. All Rights Reserved
```
nickbabcock: Please remove copyright

nickbabcock:
Awesome PR, did a quick glance and everything looks top notch.

I'm going to let this PR sit for a little bit as I mull over some of the implementation (like better enum validation messages).

dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapperTest.java
```diff
@@ -548,4 +549,60 @@ public void testInvalidFieldQueryParam() {
...
+            .request()
+            .get();
+        assertThat(response.getStatus()).isEqualTo(400);
+        assertThat(response.readEntity(String.class))
+            .containsOnlyOnce("query param length may not be null");
+        response = target("/valid/paramValidation")
+            .queryParam("length", 50)
+            .request()
+            .get();
+        assertThat(response.getStatus()).isEqualTo(400);
```
nickbabcock: Do you mind splitting up each of these assertions into a separate test case with a descriptive name like the other tests found in the file?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/DurationParam.java
cmicali: D'oh, sorry - auto-added by intellij.

dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapperTest.java
cmicali: Yeah no problem - will do now.

cmicali:
Addressed comments on comments & tests and also fixed test failures related to `NonEmptyStringParam`.  The new converter for *Param types was not handling `NonEmptyStringParam` properly. I added support for it and removed `NonEmptyStringParamFeature` which was no longer needed.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/DurationParam.java
```diff
@@ -0,0 +1,20 @@
+package io.dropwizard.jersey.params;
+
+import io.dropwizard.util.Duration;
+
+public class DurationParam extends AbstractParam<Duration> {
```
cmicali: This isn't directly related to the PR, but we use this often and seems like something Dropwizard should include by default.

nickbabcock: Ideally, this may have been split into another PR, but that's ok here.

Can you write a test for this?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParamConverterProvider.java
```diff
@@ -0,0 +1,60 @@
...
+            final String parameterName = JerseyParameterNameProvider.getParameterNameFromAnnotations(annotations).orElse("Parameter");
+            final Constructor<T> constructor;
+            try {
+                constructor = rawType.getConstructor(String.class, String.class);
+            } catch (NoSuchMethodException ignored) {
+                // The Param class did not have a (String, String) constructor. We return null,
+                // leaving Jersey to handle these parameters as it normally would.
+                return null;
+            }
+            return new ParamConverter<T>() {
```
nickbabcock: Should this param converter be more like Jersey's default [`AbstractStringReader`](https://github.com/jersey/jersey/blob/927b4d0605aeb119d15e9623dab65c8aec2c8e20/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverters.java#L77-L112)?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/JerseyParameterNameProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class JerseyParameterNameProvider extends ReflectionParameterNameProvider {
+
+    @Override
+    public List<String> getParameterNames(Constructor<?> constructor) {
```
nickbabcock: Does this method need to be included? Looks like you're just delegating to `super`.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ValidatingEnumParamConverterProvider.java
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright (c) 2016. Sense Labs, Inc. All Rights Reserved
```
nickbabcock: Oh no, another copyright 😛

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/DurationParam.java
cmicali: Done - updated description and added test.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ValidatingEnumParamConverterProvider.java
cmicali: Removed!

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/JerseyParameterNameProvider.java
cmicali: No, will remove.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParamConverterProvider.java
cmicali: Possibly - the one thing we might not want to change is the null handling. Jerey's provider will throw an `IllegalArgumentException` if the value is null.  My change does not, as this would be a breaking change with previous versions of dropwizard.

Otherwise, we could change the exception handling to match abstract string provider or we could inherit from AbstractStringReader and just modify the behavior we don't like. Thoughts?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParamConverterProvider.java
```diff
@@ -0,0 +1,60 @@
...
+
+import javax.ws.rs.InternalServerErrorException;
+import javax.ws.rs.ext.ParamConverter;
+import javax.ws.rs.ext.ParamConverterProvider;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Type;
+
+public class AbstractParamConverterProvider implements ParamConverterProvider {
```
evnm: Could you add a brief Javadoc to this class explaining what it does, please?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/JerseyParameterNameProvider.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.ws.rs.core.Context;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class JerseyParameterNameProvider extends ReflectionParameterNameProvider {
+
```
evnm: Rm extra newline

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/JerseyParameterNameProvider.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.ws.rs.MatrixParam;
+import javax.ws.rs.core.Context;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class JerseyParameterNameProvider extends ReflectionParameterNameProvider {
```
evnm: Ditto brief Javadoc

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ValidatingEnumParamConverterProvider.java
```diff
@@ -0,0 +1,146 @@
...
+import javax.ws.rs.ext.Provider;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.EnumSet;
+import java.util.Set;
+
+@Provider
+public class ValidatingEnumParamConverterProvider<C extends Enum<C>> implements ParamConverterProvider {
```
evnm: Ditto brief Javadoc

cmicali: Thanks for the feedback - newline removed and javadoc added to those 3 classes.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParamConverterProvider.java
nickbabcock: Do you know how it will break backwards compatibility because doesn't the current classes get instantiated through [`StringConstructor`](https://github.com/jersey/jersey/blob/927b4d0605aeb119d15e9623dab65c8aec2c8e20/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverters.java#L118-L137), which calls [`AbstractStringReader`](https://github.com/jersey/jersey/blob/927b4d0605aeb119d15e9623dab65c8aec2c8e20/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverters.java#L77-L112)

cmicali: I think you're right, I'm getting confused. Just committed an update that makes it basically match `AbstractStringReader` and added a test for not provide vs empty parameter. All tests pass so I think this is OK.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ValidatingEnumParamConverterProvider.java
nickbabcock: Looks like the copyright came back 😉

cmicali: Removed, again.  Intellij: amazing 99.999% of the time.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ValidatingEnumParamConverterProvider.java
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Provides converters to jersey for enums used as resource parameters.
+ *
+ * <p>By default jersey will return a 404 if a resource parameter of an enum type cannot be converted. This class
+ * provides converters for all enum types used as resource parameters that provide better error handling. If an
+ * invalid value is provided for the parameter a {@code 400 Bad Request} is returned and the error message will
+ * include the parameter name and a list of valid values.</p>
+ */
+@Provider
+public class ValidatingEnumParamConverterProvider<C extends Enum<C>> implements ParamConverterProvider {
```
nickbabcock: [I was able to cut the implementation in half](https://gist.github.com/nickbabcock/d34c360e7af44be4f8ed142c14e143cc) and make it somewhat compatible with the [`FuzzyEnumModule`](https://github.com/dropwizard/dropwizard/blob/9e01f7ec99e47f24f26698dbde7379e51b4e3aad/dropwizard-jackson/src/main/java/io/dropwizard/jackson/FuzzyEnumModule.java) that we use for Jackson enum serialization. I think it could be a good idea if params and body enums were handled the same way:

If there isn't anything fundamentally wrong with my suggested foundation, I'd like to see it integrated and improved upon 😄

cmicali: 😮 this is much better. Facepalm for the scanning work when all you need is

```java
 if (!rawType.isEnum()) {
    return null;
}
```

I will update this PR with this - should I include 

```java
 final String text = CharMatcher.WHITESPACE
                    .removeFrom(jp.getText())
                    .replace('-', '_')
                    .replace('.', '_');
```

from [`FuzzyEnumModule`](https://github.com/dropwizard/dropwizard/blob/9e01f7ec99e47f24f26698dbde7379e51b4e3aad/dropwizard-jackson/src/main/java/io/dropwizard/jackson/FuzzyEnumModule.java) to make it behave exactly the same way?

cmicali: Two other questions:

1. Should I rename the class FuzzyEnumParamConverterProvider to make it match [`FuzzyEnumModule`](https://github.com/dropwizard/dropwizard/blob/9e01f7ec99e47f24f26698dbde7379e51b4e3aad/dropwizard-jackson/src/main/java/io/dropwizard/jackson/FuzzyEnumModule.java) ?
2. Should we move the 'fuzzy' part out into `dropwizard-util`, maybe to an `Enums` class?

nickbabcock: No problem -- the power of open source 🎉 

To answer your questions:

- Yes it is probably prudent to make the implementations exactly the same.
- So extracting the shared code into an `Enums` class in `dropwizard-util` would be a fine idea
- And yes a rename to `FuzzyEnumParamConverterProvider`, I'm assuming, will help discover-ability

cmicali: Changes made and committed.

nickbabcock: Looks like there was a change that made `readsEnumsUsingToStringWithDeserializationFeatureOff` test case fail

cmicali: Forgot to add that change to the previous commit - commited/pushed.

dropwizard-util/src/main/java/io/dropwizard/util/Enums.java
```diff
@@ -0,0 +1,46 @@
...
+            .removeFrom(value)
+            .replace('-', '_')
+            .replace('.', '_');
+        for (Enum<?> constant : constants) {
+            if (constant.name().equalsIgnoreCase(text)) {
+                return constant;
+            }
+        }
+
+        //In some cases there are certain enums that don't follow the same patter across an enterprise.  So this
```
evnm: * Single space after slashes
* s/patter/pattern/

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
```diff
@@ -0,0 +1,69 @@
...
+                if (Strings.isNullOrEmpty(value)) {
+                    return null;
+                }
+
+                Enum<?> constant = Enums.fromStringFuzzy(value, constants);
+                if (constant != null) {
+                    return (T) constant;
+                }
+
+                throw new WebApplicationException(getErrorResponse(String.format("%s must be one of [%s]", parameterName, Joiner.on(", ").join(constants))));
```
evnm: This line's pretty long. Let's break it up and move the `Joiner` to a static field:

```
final String errorMsg = String.format(
    "%s must be one of [%s]", parameterName, joiner.join(constants));
throw new WebApplicationException(getErrorResponse(errorMsg));
```

dropwizard-util/src/main/java/io/dropwizard/util/Enums.java
cmicali: 👍

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
cmicali: Good idea - done 👍

evnm:
Thanks, @cmicali!

nickbabcock:
Updated release notes in b1c563e2

Thanks for the huge effort, @cmicali 😄

cmicali:
Happy to contribute back, not huge at all. Thanks for all the great detailed review, the result is significantly better for it.

Issue #1699, opened 2016-08-19
Broken query parameter extraction in POST requests


evnm:
See post-closure comments on #1606.

nickbabcock:
What I can't figure out is that the `testBuild` runs flawlessly by itself (and on Travis), but fails when run with other tests 😕

evnm:
Spent a bunch more time digging on this today, but still no solution. It seems to be related to how `o.e.jetty.server.Request` objects are recycled between requests.

To summarize, `Request` has a `recycle` method which nulls out the object's state (including the `_queryParameters` field) and is presumably meant to be run between individual HTTP requests. For some reason, `TaskServlet#doPost` is invoked prior to `Request#recycle`, which results in the request's query parameters not being extracted and the default empty map being used. 

Patching `o.e.jetty.server.Request` to remove [the `if (_queryParameters == null)` guard](https://github.com/eclipse/jetty.project/blob/9fd6d4354e7a6d5ea0b0a8fff47351ba37b9babd/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java#L370) around the `extractQueryParameters()` method results in query parameters universally being extracted and the `SimpleServerFactoryTest#testBuild` passing.

It's unclear to me if this situation is a Jetty bug or the result of mismanagement of HTTP request lifecycles in `SimpleServerFactoryTest`. I fiddled with `httpRequest` helper method to ensure that all input/output buffers associated with requests are closed, but the test still fails.

nickbabcock:
A few more hours of debugging as only let to the following insight: 

Changing `testBuild` to have three requests will fail every time when ran independently. eg:

```java
assertThat(httpRequest("GET", "http://localhost:" + port + "/service/test"))
        .isEqualTo("{\"hello\": \"World\"}");
assertThat(httpRequest("POST", "http://localhost:" + port + "/secret/tasks/hello?name=test_user"))
        .isEqualTo("Hello, test_user!");
assertThat(httpRequest("POST", "http://localhost:" + port + "/secret/tasks/hello?name=test_user"))
        .isEqualTo("Hello, test_user!");
```

And strangely, putting the GET request as the last request will make the test pass :confused:

nickbabcock:
I extracted almost all the admin task servlet into a [self contained gist](https://gist.github.com/nickbabcock/318d233c293444cb4fc22a8a80947f1f) that doesn't import dropwizard and I'm unable to replicate the errors we're seeing.

[...]

PR #1739, opened 2016-09-23
Add Windows CI via AppVeyor

nickbabcock:
We need a windows ci runner, as many of my commits deal with fixing broken windows builds. These, ideally, should be fixed before a PR is merged. Since I'm most familiar with AppVeyor, I decided to go with them -- it was pretty simple, add `appveyor.yml` file to invoke maven.

Some notes:

- Something was going on with H2 where the same table existed when on setup, so I modified the tests to create the table if not already existing.
- I enabled the CI on my fork, and [this CI run](https://ci.appveyor.com/project/nickbabcock/dropwizard/build/1.0.5) would have caught the #1699 issue before #1678 was merged.
- To test, I reverted mentioned pull request and the [CI does pass](https://ci.appveyor.com/project/nickbabcock/dropwizard/build/1.0.4)

We'll have to add AppVeyor as a trusted integration, like we do with Travis. I haven't checked if I have permission to do so.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,3 @@
+test_script:
+  - mvn package
+build: off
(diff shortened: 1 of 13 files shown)


jplock:
Looks good to me.  It doesn't look like Github natively supports AppVeyor (it doesn't show up in the integrations drop down anyway), but maybe we could trigger it via a web hook?

nickbabcock:
Actually, it looks like I just had to add it in the AppVeyor UI, it may not require any permissions at all from us. I'm going to go ahead and merge this so we can see. It should pass

nickbabcock:
While appveyor built on the code on master, looks like it may have not added it as a check to pull requests:

Like @jplock said, some additional configurations may be needed for the dropwizard account. For my account and projects, I have:

image

image

It's definitely possible to add AppVeyor as a check. Take a look at any of the [Json.Net pull requests](https://github.com/JamesNK/Newtonsoft.Json/pulls)

Issue #1742, opened 2016-09-26
Discoverable bug for RequestLogFactory


jmoney8080:
Pretty sure that `META-INF.services` -> `META-INF/services` [here](https://github.com/dropwizard/dropwizard/tree/master/dropwizard-request-logging/src/main/resources/META-INF.services).  This is causing custom RequestLogFactories from not being discovered.

Workaround, in your project add a `io.dropwizard.jackson.Discoverable` file in `META-INF/services` with contents of `io.dropwizard.request.logging.RequestLogFactory` and everything is hunky-dory again!  

I *think* the only reason this works in 1.0 is because `RequestLogFactory` has a default impl class in the jackson annotation.

nickbabcock:
Nice, did you want to send a PR to fix this?

Is it possible to write a test to ensure that this doesn't regress in the future?

jmoney8080:
I have a PR for this: https://github.com/dropwizard/dropwizard/pull/1743
I'm unsure how to write a test case as Discoverable is kind of opt-in.  How would you recommend one?  Hard code a check for `META-INF.services`?

nickbabcock:
This may be enough of a unit test (it'll fail without your PR) 💭 

```java
public class RequestLogFactoryTest {

    @Test
    public void isDiscoverable() throws Exception {
        assertThat(new DiscoverableSubtypeResolver().getDiscoveredSubtypes())
            .contains(LogbackAccessRequestLogFactory.class);
    }
    // ...
}
```

There are other tests worth looking at in dropwizard-logging

jmoney8080:
Added test, ci is running…

jmoney8080:
ci ran and github checks passed

PR #1743, opened 2016-09-27
Rename META-INF.services -> META-INF/services for dropwizard-request-logging

jmoney8080:
Rename META-INF.services -> META-INF/services for dropwizard-request-logging.  Closes https://github.com/dropwizard/dropwizard/issues/1742


evnm:
LGTM. Good find! Thanks for taking care of this and thanks, @nickbabcock for the test case.

PR #1749, opened 2016-09-28
Jackson 2.8 Redux

nickbabcock:
Based on joschi's work in #1647

- Rebased on current master
- Bump to latest 2.8 (2.8.3)
- Maintain compatibility with how we treat errors resulting from JSON input being the wrong type. They should be treated as client errors, not server errors. The current heuristics were not sufficient enough for Jackson 2.8, so another heuristic was added via a straightforward regex.

Can squash/rebase on request.

Can update release notes on merge.


jplock:
This looks good to me.

jplock:
Updated release notes in https://github.com/dropwizard/dropwizard/commit/d619dede86b7f0d3888539e846fa152c638d876e

PR #1762, opened 2016-10-06
Also point to metrics-scala module.

erikvanoosten:


jplock:
Thanks!

joschi:
@erikvanoosten Wouldn't this be better fitting in the https://github.com/dropwizard/metrics project?

erikvanoosten:
Yes, that is the better place. In fact the project _is_ listed there.

_But why allow only one place?_

This PR was created after some confusion from a user. The other referred project, 'dropwizard-scala' apparently also has metrics integration. By including this reference we can at least indicate a choice.

BTW, I tried to get more information from the dropwizard-scala maintainers about their plans and goals so that it would be easier to recommend which project combination to use. Unfortunately I did not get a response yet.

PR #1763, opened 2016-10-07
Edit trustSelfSignedCertificates description

ipropper:
The old description was misleading. One could think `false` meant no self-signed certs could be trusted.


nickbabcock:
Looks good, thanks for the clarification.

PR #1765, opened 2016-10-07
Avoid needlessly binding 8080 and 8081 in tests

nickbabcock:
This way I can have a default dropwizard app I'm testing and still run all of dropwizard's tests.


arteam:
LGTM

PR #1768, opened 2016-10-07
Bind default exception mappers

nickbabcock:
Currently when a user wants to override one of the default exception mappers they must specify `registerDefaultExceptionMappers: false` in their configuration, register their exception mapper, and then re-register all the other default exception mappers they want. This is not user friendly. A user should be able to register their own exception mapper, override the default exception mapper while preserving other default exception mappers.

It should be possible to achieve this behavior as alluded to in [JERSEY-2722](https://java.net/jira/browse/JERSEY-2722)

> your own ExceptionMapper<ValidationException> should always have a higher priority than anything provided by Jersey

and alluded in [Jersey's Bean Validation code](https://github.com/jersey/jersey/blob/1a7919181d8a1f550c7fc20898db67fe9da5d596/ext/bean-validation/src/main/java/org/glassfish/jersey/server/validation/internal/ValidationBinder.java#L102)

```java
// Custom Exception Mapper and Writer - registering in binder to make possible for users register their own providers.
bind(ValidationExceptionMapper.class).to(ExceptionMapper.class).in(Singleton.class);
bind(ValidationErrorMessageBodyWriter.class).to(MessageBodyWriter.class).in(Singleton.class);
```

This pull request takes a crack at introducing this to Dropwizard, while remaining 100% backwards compatible. I've tried creating this behavior before in #1144, but it didn't work out, so I'd appreciate some help testing this :smile:. I believe it is working as I have an app that has:

```java
@Override
public void run(final DropwizardSampleConfiguration configuration,
                final Environment environment) {
    environment.jersey().register(new Resource());
    environment.jersey().register(new MyJerseyException());
}


public class MyJerseyException implements ExceptionMapper<JerseyViolationException> {
    @Override
    public Response toResponse(JerseyViolationException e) {
        return Response.ok("Oh yeah").build();
    }
}
```

and I receive the "Oh yeah" response with every restart and request (100 requests across 10 restarts).

We can even keep around the `registerDefaultExceptionMappers` for users who want no default exception mappers registered, but I'm assuming most users will not use this option and opt for overriding specific exception mappers.

The major benefit to this approach is the ability to add new exception mappers without having the relatively large cost of forcing the user to know them all 😋 

TODO:
- [x] Add tests
- [x] Update docs
- [x] To use the provider annotation or not?

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -1219,14 +1219,23 @@ and then registering the exception mapper:
 Overriding Default Exception Mappers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you want more control, you can disable the exception mappers Dropwizard provides by default. This is done
-by setting ``server.registerDefaultExceptionMappers`` to ``false``. Since this disables all default exception
-mappers make sure to re-enable exception mappers that are wanted. The default exception mappers are:
+To override a specific exception mapper, register your own class that implements the same
+``ExceptionMapper<T>`` as one of the default. For instance, we can customize responses caused by
+Jackson exceptions:
 
-- ``LoggingExceptionMapper<Throwable>``
-- ``JerseyViolationExceptionMapper``
-- ``JsonProcessingExceptionMapper``
-- ``EarlyEofExceptionMapper``
+.. code-block:: java
+
+    public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProcessingException> {
+        @Override
+        public Response toResponse(JsonProcessingException exception) {
+            // create the response
+        }
+    }
+
+With this method, one doesn't need to know what the default exception mappers are, as they are
+overridden if the user supplies a conflicting mapper. While not preferential, one can also disable
+all default exception mappers, by setting ``server.registerDefaultExceptionMappers`` to ``false``.
+See the class ``ExceptionMapperBinder`` for a list of the default exception mappers.
 
 .. _man-core-resources-uris:
 
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/ExceptionMapperBinder.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/ExceptionMapperBinder.java
@@ -1,0 +1,30 @@
+package io.dropwizard.setup;
+
+import io.dropwizard.jersey.errors.EarlyEofExceptionMapper;
+import io.dropwizard.jersey.errors.LoggingExceptionMapper;
+import io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper;
+import io.dropwizard.jersey.validation.JerseyViolationExceptionMapper;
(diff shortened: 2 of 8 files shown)


nickbabcock:
Added the test case. Writing it made me kinda wish that we had a dedicated spot for integration tests, and I didn't want to pile on additional tests in dropwizard-example 😕

jplock:
Should we leave the note in the docs about this http://www.dropwizard.io/1.0.2/docs/manual/core.html#overriding-default-exception-mappers ?

dropwizard-core/src/main/java/io/dropwizard/setup/ExceptionMapperBinder.java
```diff
@@ -0,0 +1,30 @@
...
+import io.dropwizard.jersey.errors.EarlyEofExceptionMapper;
+import io.dropwizard.jersey.errors.LoggingExceptionMapper;
+import io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper;
+import io.dropwizard.jersey.validation.JerseyViolationExceptionMapper;
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
+
+import javax.ws.rs.ext.ExceptionMapper;
+
+public class ExceptionMapperBinder extends AbstractBinder {
+    private boolean showDetails;
```
jplock: nitpick: this could be `final`

nickbabcock:
Yeah, imo I should update the docs in this PR too

docs/source/manual/core.rst
```diff
@@ -1219,14 +1219,23 @@ and then registering the exception mapper:
...
-If you want more control, you can disable the exception mappers Dropwizard provides by default. This is done
-by setting ``server.registerDefaultExceptionMappers`` to ``false``. Since this disables all default exception
-mappers make sure to re-enable exception mappers that are wanted. The default exception mappers are:
-
-- ``LoggingExceptionMapper<Throwable>``
-- ``JerseyViolationExceptionMapper``
-- ``JsonProcessingExceptionMapper``
-- ``EarlyEofExceptionMapper``
+To override a specific exception mapper, register your own class that implements the same
+``ExceptionMapper<T>`` as one of the default. For instance, we can customize responses caused by
```
nickbabcock: Yeah, couple sentences later

> See the class ``ExceptionMapperBinder`` for a list of the default exception mappers.

Which is similar to the other snippets in the documentation.

davidmankin:
Thank you.  I've wanted this feature several times!

nickbabcock:
Looks like even Jersey code doesn't standardize on whether to annotate exception mappers with `@Provider`, so as far as I'm concerned this is ready for further review/merge.

jplock:
Looks good to me!

jplock:
@nickbabcock mind updating the release notes?

nickbabcock:
@jplock done, 89ab17b88345e76f5b35eb1fc526679235e2dc6b 😄

PR #1770, opened 2016-10-08
correct sample code on building apache http client

tmack8001:
This contract with [HttpClientBuilder$build](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java#L220-L226) is the following:

```
public CloseableHttpClient build(String name);
```


jplock:
Thanks!

PR #1771, opened 2016-10-11
Update JerseyClientConfiguration link to dropwizard 1.0.1

rwwarren:


dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientConfiguration.java
```diff
@@ -12,7 +12,7 @@
  * {@link HttpClientConfiguration}.
  *
  * @see HttpClientConfiguration
- * @see <a href="http://dropwizard.io/0.9.1/docs/manual/configuration.html#jerseyclient">Jersey Client Configuration</a>
+ * @see <a href="http://dropwizard.io/1.0.1/docs/manual/configuration.html#jerseyclient">Jersey Client Configuration</a>
```
jplock: It'd be nice to link this to a place that's not version specific, but if we are going to update it, would you mind updating it to the latest release `1.0.2`?

rwwarren: Just updated it to 1.0.2 and I agree there should be a non version specific link

jplock:
Thanks!

PR #1775, opened 2016-10-14
Upgrade to Jackson 2.8.4

joschi:
This PR also includes the (unofficial) Jackson BOM so that users can include Jackson dependencies in the correct version into their projects.


jplock:
@joschi is the bom something @cowtowncoder may maintain himself or maybe we should move it under the dropwizard github org first?  Unless you want to maintain it in your personal repo.

cowtowncoder:
Yes, this is something that would make sense to add under FasterXML repo I think.
We haven't gone through the details (not that there should be much) yet; I think it'd be used from 2.9.0 onwards. Until then, for older versions, I think outside location would probably work best.

joschi:
@jplock I can maintain the Jackson BOM until it's being moved to the FasterXML organization with Jackson 2.9.x.

jplock:
Sounds good. As soon as Travis is done, I'll merge it if no one else has already.

PR #1778, opened 2016-10-20
Add web test container agnostic client target

nickbabcock:
Depending on the web test container passed to
`ResourceTestRule#setTestContainerFactory`, one would either invoke
`client()` or `getJerseyTest()`, which is inconvenient. The fix is to
standardize on a new method that can be used regardless of web test
container that also doesn't infringe on the law of Demeter (as the other
solution would be to rewrite all usage as using `getJerseyTest()`)


nickbabcock:
Updated release notes in 9a41974

Issue #1496, opened 2016-03-28
ResourceTestRule does not support testing file upload


svenwb:
I tried to write a unit test for testing a file upload by using the ResourceTestRule as described in the manual under "Testing Resources" http://www.dropwizard.io/0.9.2/docs/manual/testing.html

But doing something like this, will result in an exception when running the test case.
```
CsvUpload response = resource.client().target("/csvupload")
                .request().post(Entity.entity(formDataMultiPart, formDataMultiPart.getMediaType()), CsvUpload.class);
```

```
org.glassfish.jersey.message.internal.MessageBodyProviderNotFoundException: MessageBodyWriter not found for media type=multipart/form-data, type=class org.glassfish.jersey.media.multipart.FormDataMultiPart, genericType=class org.glassfish.jersey.media.multipart.FormDataMultiPart.

	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.aroundWriteTo(WriterInterceptorExecutor.java:247)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor.proceed(WriterInterceptorExecutor.java:162)
	at org.glassfish.jersey.message.internal.MessageBodyFactory.writeTo(MessageBodyFactory.java:1130)
	at org.glassfish.jersey.client.ClientRequest.writeEntity(ClientRequest.java:502)
	at org.glassfish.jersey.test.inmemory.InMemoryConnector.apply(InMemoryConnector.java:239)
	at org.glassfish.jersey.client.ClientRuntime.invoke(ClientRuntime.java:255)
	at org.glassfish.jersey.client.JerseyInvocation$2.call(JerseyInvocation.java:700)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:315)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:297)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:228)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:444)
	at org.glassfish.jersey.client.JerseyInvocation.invoke(JerseyInvocation.java:696)
	at org.glassfish.jersey.client.JerseyInvocation$Builder.method(JerseyInvocation.java:448)
	at org.glassfish.jersey.client.JerseyInvocation$Builder.post(JerseyInvocation.java:349)
	at CsvUploadResourceTest.testCsvUpload(CsvUploadResourceTest.java:71)
[...]

PR #1780, opened 2016-10-20
Add example of multipart forms test

nickbabcock:
Takes @psamsotha [comment](https://github.com/dropwizard/dropwizard/issues/1496#issuecomment-202352516) and adds it to the documentation as it failed my litmus test of a dropwizard feature that has testing documented 😋 

Docs will slightly change once #1778 is merged (use `target` instead of `client`), so feel free to merge in any order and I can make the appropriate change.


jplock:
I merged in #1778 so let's update this to reflect the new preferred usage 👍

jplock:
Looks good to me!

PR #1781, opened 2016-10-21
Remove dead code in the jackson provider

nickbabcock:
The previous workaround rendered useless by the changes in Jackson 2.8.3. 🎉


jplock:
@nickbabcock I'm guessing we had tests for this that will continue to pass?

nickbabcock:
Yup, sure thing. I ran the test suite and everything passed, and `catch` code was never executed. Since the removed lines of code only had special handling with IOExceptions, which was fixed in 2.8.3 the test case `returnsA400ForMalformedInputCausingIoException` is kinda a misnomer now (it's a wrapped IOException).

Also no need to override the method anymore because it is the exact same as the parent.

PR #1784, opened 2016-10-23
Empty optional exception mapper

### LINKED ISSUE #1727: Should OptionalMessageBodyWriter return 404 when the response is Optional.empty()

nickbabcock:
Another implementation of #1732, but using exceptionmappers, which now have a lower code tax due to #1768. By default, backwards compatibility is preserved with the ability to easily remap it to a 204.

The shape of the custom exception (`EmptyOptionalException`) inspired by netty's [ReadTimeoutException](https://github.com/netty/netty/blob/cfa5b85075f9f3a74e7eb46e344f44775dbcea57/handler/src/main/java/io/netty/handler/timeout/ReadTimeoutException.java) to be a lightweight exception, as 204's may be a normal state and shouldn't be treated exceptionally.

Maybe controversial, but I've decided to add another submodule, dropwizard-e2e (end-to-end). It's like dropwizard-example except that it is meant to hold multiple `Applications`, as it's possible that a single example application can't represent all options especially if some are conflicting. It also presents another avenue for tests that ensure unit + integration tests work. If others want to verify it looks correct, it'd be much appreciated.

Closes #1727

cc original author: @qinfchen 

I can update the release notes once this is merged.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/ExceptionMapperBinder.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/ExceptionMapperBinder.java
@@ -3,6 +3,7 @@ package io.dropwizard.setup;
 import io.dropwizard.jersey.errors.EarlyEofExceptionMapper;
 import io.dropwizard.jersey.errors.LoggingExceptionMapper;
 import io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper;
+import io.dropwizard.jersey.optional.EmptyOptionalExceptionMapper;
 import io.dropwizard.jersey.validation.JerseyViolationExceptionMapper;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
 
@@ -26,6 +27,7 @@ public class ExceptionMapperBinder extends AbstractBinder {
         bind(new JerseyViolationExceptionMapper()).to(ExceptionMapper.class);
         bind(new JsonProcessingExceptionMapper(isShowDetails())).to(ExceptionMapper.class);
         bind(new EarlyEofExceptionMapper()).to(ExceptionMapper.class);
+        bind(new EmptyOptionalExceptionMapper()).to(ExceptionMapper.class);
     }
 
     public boolean isShowDetails() {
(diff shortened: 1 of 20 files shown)


jplock:
I like the idea of the `dropwizard-e2e` module.  I'm wondering if we should standardize on some naming conventions though instead of a generic `App1`?  Are there any other tests floating around that could be integrated into this new module as well (or anything that we haven't been able to test thoroughly in the past)?

nickbabcock:
I know for #1768, `dropwizard-e2e` would have been the preferred test implementation because it is a much more realistic integration test.

I'm fine if we can find a theme for each application. Might have to ponder possible themes.

jplock:
I think we can merge this for now and refactor the tests later.

nickbabcock:
Updated release notes in 00cc42f62

PR #1788, opened 2016-10-27
Allow user to override configured JsonProvider

### LINKED ISSUE #1005: Make registration of JacksonMessageBodyProvider optional
see AbstractServerFactory.createAppServlet:
jersey.register(new JacksonMessageBodyProvider(objectMapper, validator))

Problem description: (I noticed this during migration from 0.7.1 to 0.8.1)
Given:
- an entity class without a default constructor
- a method @POST + @Consumes("application/json) which takes an instance of the entity class.
- A custom MessageBodyReader for deserializing the class.
  When you post json to the given resource method you can get errors because JacksonMessageBodyProvider is consulted before my own messagebodyreader.

Notes:
- In my environment this manifests itself as a heisenbug: when I debug an integration test exposing this problem I often end up in my own MessageBodyReader (depending on where  I installed breakpoints). Presumably it's related to timing, but I have yet to pin it down.
- One could create a custom ServerFactory which extends DefaultServerFactory and overrides AbstractServerFactory.createAppServlet, but that method calls private methods and I would have to end up copying almost everything.
- IIUC I may be able to work around this using @JsonIgnoreType
- There are other providers being registered at AbstractServerFactory.createAppServlet, one might want to make these optional as well.

nickbabcock:
I wondered if I could create a custom JsonProvider where I could read and write a special payload at the top the body (don't ask why, it's just a thought experiment):

``` json
/* I'm a special header */
{
    "check": "mate"
}
```

This PR allows a user configured `JacksonJaxbJsonProvider` or another appropriate class to take priority over the dropwizard configured class (`JacksonMessageBodyProvider`).

I believe this PR closes #1005

@jplock I used App1 in dropwizard-e2e, not sure if it is too early to call it a theme, but so far it is about overriding defaults 😋

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java
@@ -6,7 +6,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import io.dropwizard.jersey.gzip.ConfiguredGZipEncoder;
 import io.dropwizard.jersey.gzip.GZipDecoder;
-import io.dropwizard.jersey.jackson.JacksonMessageBodyProvider;
+import io.dropwizard.jersey.jackson.JacksonBinder;
 import io.dropwizard.jersey.validation.HibernateValidationFeature;
 import io.dropwizard.jersey.validation.Validators;
 import io.dropwizard.lifecycle.Managed;
@@ -403,7 +403,7 @@ public class JerseyClientBuilder {
             config.register(provider);
         }
 
-        config.register(new JacksonMessageBodyProvider(objectMapper));
+        config.register(new JacksonBinder(objectMapper));
         config.register(new HibernateValidationFeature(validator));
 
         for (Map.Entry<String, Object> property : this.properties.entrySet()) {
--- a/dropwizard-client/src/test/java/io/dropwizard/client/JerseyClientBuilderTest.java
+++ b/dropwizard-client/src/test/java/io/dropwizard/client/JerseyClientBuilderTest.java
@@ -8,7 +8,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import io.dropwizard.jersey.gzip.ConfiguredGZipEncoder;
 import io.dropwizard.jersey.gzip.GZipDecoder;
-import io.dropwizard.jersey.jackson.JacksonMessageBodyProvider;
 import io.dropwizard.jersey.validation.Validators;
 import io.dropwizard.lifecycle.setup.ExecutorServiceBuilder;
 import io.dropwizard.lifecycle.setup.LifecycleEnvironment;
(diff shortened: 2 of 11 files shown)


nickbabcock:
Pushed a commit to fix the style issues and added explicit test case that'll close #1005

jplock:
@nickbabcock sorry, I forgot about this one.  Good to merge?

nickbabcock:
Rebased PR on master, this should be good to go once Travis gives the green light!

jplock:
Travis failed :( Seems like its something unrelated

nickbabcock:
Yeah maven is having some problems with their certificates

nickbabcock:
Looks like there is now a test failure with the view integration, I'll have to look into and report back.

nickbabcock:
Ok fixed previous issue by specifying that the view produces html.

I've also changed the exception mapper that looks for the view render exception. cc @acwwat and @dren-dk to get their insight and see if they prefer this approach (and if so, it'll need to be documented).

acwwat:
@nickbabcock I have some questions/comments about the exception mapper change.

1. I personally don't have a preference on `ExtendedExceptionMapper` over `ExceptionMapper` and it seems to fit the use case better anyway, so the change is OK. But just wondering - is the usage of `ExtendedExceptionMapper` common in JAX-RS development?

2. For the example in App1, given that the custom exception mapper is specifically for handling `MustacheNotFoundException` to return a 404 error, it should be checked for instead of `ViewRenderException`.

acwwat:
@nickbabcock Oh and also, if we decide to use `ExtendedExceptionMapper`, maybe we should also update `ViewRenderExceptionMapper` that I added earlier to be more consistent.

nickbabcock:
I'm not sure how common `ExtendedExceptionMapper` is, but it's common enough to warrant a mention in the [jersey docs](https://jersey.java.net/documentation/latest/representations.html#d0e6754) (emphasis mine).

> [`ExtendedExceptionMapper`] can for example check the exception parameters and based on them return false and let *other provider* to be chosen for the exception mapping.

By defining an WebApplicationException mapper in dropwizard-views, it would override the WebApplicationException logic found in [`LoggingExceptionMapper`](https://github.com/dropwizard/dropwizard/blob/1cce996bab38f6b4ae0643f7cc0e9ec468239e3a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/LoggingExceptionMapper.java). Which I can see as undesirable for those who were previously relying on that behavior.

It should be preferential to have exception mappers that target specific exceptions (in `App1` case, `MustacheNotFoundException`), but since the exception thrown is a  `MustacheNotFoundException` wrapped in a `WebApplicationException` I used `ExtendedExceptionMapper`. I realize that it is not 100% necessary, but since others may look to `dropwizard-e2e` or `dropwizard-example` for code examples I'd rather it keep unexpected behavior (eg. overriding `LoggingExceptionMapper` web exception behavior) to a minimum.

So yes, I think `ViewRenderExceptionMapper` should be updated as well.

Thank you for the code review, I've updated the code to just remap `MustacheNotFoundException` to 404.

jplock:
This looks good to me.  Any other comments before we merge?

nickbabcock:
I've moved the discussion/issue about `ViewRenderExceptionMapper` to #1837, as to not impede this PR.

nickbabcock:
Updated release notes in 361218c570d

PR #1790, opened 2016-11-01
Bump jetty

nickbabcock:
The [release notes are pretty small](https://github.com/eclipse/jetty.project/blob/206122756f8988cf9e5abf4a4b4da83a3a562e07/VERSION.txt#L3-L22), but contain an important snippet:

> Support certificates hot reload

Which'll set the ground for better integration with Let's Encrypt (for instance), for those eschewing tls termination proxies.

Just wanted to point it out, as, in the future, I may be tinkering with it a bit 😄


evnm:
Thanks, @nickbabcock!

PR #1791, opened 2016-11-01
Log WebApplicationException redirects

jroitgrund:
Example use case: log OAuth2 errors.


nickbabcock:
Hey @jroitgrund thanks for your interest. Not sure I understand the applicability of logging an error on a redirect. If this is a requirement for the site you're working on, prefer creating a custom exception mapper.

jroitgrund:
My rationale was that if you wanted a simple redirect you wouldn't wrap it in a WebApplicationException. What is the difference in semantics between a plain redirect response and a WebApplicationException if not logging?

nickbabcock:
If one's endpoints don't return `Response`, the way to communicate a redirect is through a `WebApplicationException`

Issue #1794, opened 2016-11-02
Make HttpConfiguration#blockingTimeout configurable


CodingFabian:
Right now `org.eclipse.jetty.server.HttpConfiguration.setBlockingTimeout(long)`is not configurable via `io.dropwizard.jetty.HttpConnectorFactory`but it probably should be. -1 as default meaning block forever is probably not a good value.

arteam:
Hi @CodingFabian!

Thanks for pointing on this omission. I've submitted a pull request to add this parameter to the configuration. Please have a look at #1795.

CodingFabian:
Fixed via https://github.com/dropwizard/dropwizard/pull/1795 thanks

Issue #1674, opened 2016-08-05
Dropwizard Request timeout


darrenhaken:
I'm trying to ensure that requests timeout after a suitable period. 
Looking through the docs the idleTimeout config param should define that (default 30 seconds).

I tried running a request with a Thread.sleep for > 30 seconds and it didn't time out to the client.
Any idea why this might be?

evnm:
I don't know Jetty's internals super well, but I think that it comes down to the semantics of what the [idle timeouts on `Connector`s](http://download.eclipse.org/jetty/9.3.9.v20160517/apidocs/org/eclipse/jetty/server/AbstractConnector.html#setIdleTimeout-long-) mean (emphasis mine):

>The max idle time is applied:
>
> * When waiting for a new message to be received on a connection
> * When waiting for a new message to be sent on a connection
>
> This value is interpreted as the maximum time between some progress being made _on the connection_.

I think that in the case you're describing, where you stick a `Thread.sleep(...)` within a resource method, the server is pausing during request-handling rather than while Jetty is waiting for bytes to be transferred. That is to say, a connection is _idle_ if it's hung in the transfer state rather than the request-handling state.

Somebody please correct me if I'm wrong. This is what "idle timeouts" mean in Netty/Finagle parlance, at least. These other server libraries provide more granular types of timeouts (ex: [Finagle](https://twitter.github.io/finagle/guide/Clients.html#timeouts-expiration)), which unfortunately Jetty doesn't seem to provide.

nickbabcock:
[Someone asked a similar question several years ago](http://jetty.4.x6.nabble.com/how-to-set-connection-request-timeout-for-jetty-sever-td43743.html). I'm under the impression that nothing changed under this situation. The answer then:

> Jetty has no timeout that limits how long a handler can handle for. 
>
> You could simply write a handler wrapper than imposed such a timeout 
> and interrupted the thread if it took longer than your 2s etc. and 
> wrote a response in the finally block. 

Nowadays, one could use [Guava's SimpleTimeLimiter](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/util/concurrent/SimpleTimeLimiter.html) or [Jersey's async server timeout handler](https://jersey.java.net/documentation/latest/async.html).

nickbabcock:
Closing the issue as the solution presented should be satisfactory. If not, please respond with additional details.

BenRomberg:
[...]

PR #1795, opened 2016-11-02
Add support for configuring the blocking timeout for Jetty connectors

arteam:
The issue reported in #1794.

Jetty 9.3.2.v20150730 introduced a new option for the configuration of HTTP connection factories. The option is called `blockingTimeout` and allows to configure the timeout which applies to all blocking operations in addition to `idleTimeout`. The rationale behind it is to allow to set a timeout
for blocking operations which perform I/O activity, but with a slow rate (for example serving a big static file). Such tasks won't exhaust the pool of worker threads which should improve the QoS for a Jetty server.

See:
https://dev.eclipse.org/mhonarc/lists/jetty-announce/msg00085.html
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472621

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -303,6 +303,9 @@ idleTimeout              30 seconds          The maximum idle time for a connect
                                              or when waiting for a new message to be sent on a connection.
                                              This value is interpreted as the maximum time between some progress being made on the
                                              connection. So if a single byte is read or written, then the timeout is reset.
+blockingTimeout          (none)              The timeout applied to blocking operations. This timeout is in addition to
+                                             the `idleTimeout`, and applies to the total operation (as opposed to the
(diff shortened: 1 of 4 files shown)


CodingFabian:
@arteam thanks should be sufficient - i tried to come up with a pr on my own, but I was unable to decide how to deal with the -1 setting. Looks like you decided that one would not want to manually set it?

arteam:
Yes, I think making the field nullable is a better way to represent its absence than using -1. We use the `Duration` class for time durations, so we would need to specify the unit of time along with the number. I find quite strange from an outsiders perspective to specify `-1ms` as an indicator of the  default value.

CodingFabian:
agreed. in my opinion this is good to be merged.

nickbabcock:
Would this be a better solution for #1674, or are these two different problems?

CodingFabian:
In my case it is @nickbabcock

Issue #1803, opened 2016-11-07
dw-jersey: Better endpoint logging to include manually created Resources


vemilyus:
Currently the `EndpointLogger` in [DropwizardResourceConfig.java](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java) only shows resources which were added using their `Class` objects:
```java
jersey().addResource(DummyResource.class);
```
These show up when starting dw and initializing everything.

However, resources that are created programmatically and added dynamically using `jersey().getResourceConfig().registerResources(...)` don't show up in the logging output which is very irritating.

What I am proposing would be to expand the `EndpointLogger` to also look through `ResourceConfig.getResources()` to catch all that were not found using `DropwizardResourceConfig.allClasses()`.

Specifically it logs the child resources that were grouped using a manually created Resource object to share a common prefix, for example.

PR #1804, opened 2016-11-07
Enhanced the logging mechanism of registered endpoints in DropwizardR…

vemilyus:
…esourceConfig.

With the enclosed changes, the logger also processes programatically added Resources which group other resources together.
This differs from the previous implementation where only the directly added Resource classes were logged.

This pull request accompanies this issue: [#1803](https://github.com/dropwizard/dropwizard/issues/1803)

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -26,16 +26,18 @@ import javax.ws.rs.Path;
 import javax.ws.rs.ext.Provider;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
-import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.regex.Pattern;
 
 public class DropwizardResourceConfig extends ResourceConfig {
     private static final Logger LOGGER = LoggerFactory.getLogger(DropwizardResourceConfig.class);
     private static final String NEWLINE = String.format("%n");
     private static final TypeResolver TYPE_RESOLVER = new TypeResolver();
+    
+    private static final Pattern PATH_DIRTY_SLASHES = Pattern.compile("\\s*/\\s*/+\\s*");
 
     private String urlPattern = "/*";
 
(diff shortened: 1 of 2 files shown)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -122,16 +123,30 @@ public String getEndpointsInfo() {
...
         }
 
-        for (Class<?> klass : allResources) {
+        for (Class<?> klass : allResourcesClasses) {
             new EndpointLogger(urlPattern, klass).populate(endpointLogLines);
         }
+        
+        final Set<Resource> allResources = this.getResources();
+        for (Resource res : allResources) {
+            if (containsAny(allResourcesClasses, res.getHandlerClasses())) {
```
arteam: I think we could avoid creating a helper method and use a declarative approach with Guava's  `Iterables`.
```
Iterables.any(res.getHandlerClasses(), allResourcesClasses::contains)
```

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -122,16 +123,30 @@ public String getEndpointsInfo() {
...
+                continue;
```
arteam: Could you add a test for this use case (the `continue` branch when resource handlers are already processed)?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -143,6 +158,24 @@ public String getEndpointsInfo() {
...
+            if (collection.contains(value)) {
+                return true;
+            }
+        }
+        
+        return false;
+    }
+    
+    private String cleanUpPath(String path) {
+        return path.replaceAll("//+", "/");
```
arteam: I think we could extract the regular expression to a constant as `Pattern` to avoid compiling it every time we invoke `replaceAll`.

arteam: Also, could you add a unit test which verifies that path cleaning works correctly?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
arteam: Java 8's `collection.stream().anyMatch()` would also work.

arteam:
@vemilyus Thanks for your contribution!

PR #1805, opened 2016-11-08
Remove FindBugs from the Dropwizard build process

arteam:
The FindBugs project is not actively maintained, and I don't remember last time it reproduced an error in our build.  As see it, FindBugs only slow down our builds in Travis. We already use Google's `error-prone` which,  in my opinion,a better tool for catching programming errors:

 * Is actively maintained and targets Java 8
 * Much faster than FindBugs
 * Works on the compiler level
 * Produces fewer false positive warnings


Sounie:
Interesting timing, as the developer of FindBugs has recently announced that he will be coming back to work on it.

https://news.ycombinator.com/item?id=12886665

I can't vouch for how FindBugs compares with error-prone or how long it will be before it does get some updates, but just thought that it was worth mentioning.

joschi:
@arteam Is there a specific problem with using FindBugs right now?

Errorprone (compile-time checks) and FindBugs (static byte-code analysis) target different bug patterns, so they're not really interchangeable.

arteam:
> Is there a specific problem with using FindBugs right now

No problem with it, except it slow downs our builds.

> Errorprone (compile-time checks) and FindBugs (static byte-code analysis) target different bug patterns, so they're not really interchangeable.

That's true. But from the practical point of view, I find that Errorprone is much useful as a tool. It's reports more issues with code and has faster turnaround time.

jplock:
@arteam @joschi should we keep findbugs for now?  Personally, I don't have a preference.

arteam:
I too, let's put it on hold.

Issue #1812, opened 2016-11-10
Custom appender for requestLog not working in 1.0.3


matthewjsummers:
We have a custom appender (it uses the logstash-encoder) to output logs in json.  This is working fine for:

logging:
  appenders:
    - type: console
    - type: logstash-file
      threshold: TRACE
      customFields: '{"service-name":"ms-iam"}'
      currentLogFilename: log/application-json.log
      archivedLogFilenamePattern: log/application-json-%d{yyyy-MM-dd}-%i.log.gz
      archivedFileCount: 7
      timeZone: Europe/Oslo
      maxFileSize: 10MB

However, when we try and add it to requestLog, no logging appears in the request log file.  Are custom appenders not supported for the requestLog?

server:
    requestLog:
    appenders:
    - type: console
    - type: logstash-file
      threshold: TRACE
      currentLogFilename: log/access-json.log
      archivedLogFilenamePattern: log/access-json-%d{yyyy-MM-dd}-%i.log.gz
      archivedFileCount: 7
      maxFileSize: 10MB

arteam:
Hi @matthewjsummers,

Please check if your custom appender correctly works with logging events of the `IAccessEvent` type.
Now in Dropwizard 1.0.* normal and request logs are produced by different libraries (`logback-classic` and `logback-access` respectively). Ideally, your appender should be able to work both with `ILoggingEvent` and `IAccessEvent`. But not all do so (for example, the syslog appender #1801).

matthewjsummers:
Hi @arteam 

The custom appender is simply an extension of FileAppender, that uses a LogstashLayout:

https://gist.github.com/matthewjsummers/65feb02d4e6a4822919e88f2b06e93f2

LogstashLayout should support both ILoggingEvents and IAccessEvents.  Is there anything that needs to be changed in the extended FileAppender, shown in the gist above?

matthewjsummers:
Hi @arteam 
My bad - the underlying framework does support both, but you need to choose a LogstashAccessLayout. I think this just means we need two logappenders - one logging for ILoggingEvents and one for IAccessEvents.

Thanks for your help, closing this down.

arteam:
Yes, I think it's a right approach. You need to create an additional appender with `LogstashAccessLayout` and make it discoverable. But we are thinking to return back old request logging (via `logback-classic` and custom logger) in the form a custom `RequestLogFactory`, because it looks like many people can't easily transit to `logback-access`. 

[...]

Issue #1801, opened 2016-11-04
Request logging through syslog broken in DW1.0?


wilbertpol:
We have been using the following snippet to log requests through syslog:
```
# Listeners
server:
  applicationConnectors:
    - type: http
      port: 6720
  adminConnectors:
    - type: http
      port: 6721
  requestLog:
    appenders:
      - type: syslog
        threshold: ALL
        logFormat: "app-ms %msg"
```

This was still working fine in DW0.9.2. After upgrading to DW1.0.2 the requests are no longer getting logged through syslog; the regular application logging through syslog is still working fine.

Do we have to make some configuration changes?

arteam:
Hi, @wilbertpol,

Thanks for reporting this issue. We migrated request logging in 1.0 from a native Jetty solution to `logback-access`. Unfortunately, it doesn't support syslog appenders, see [LOGBACK-317](http://jira.qos.ch/browse/LOGBACK-317). 

I think it can be considered a regression for Dropwizard, and we need to figure out a way to workaround this limitaion.

mattnelson:
The logs could be configured to the console appender, then you can update your process command to redirect console to syslog.

arteam:
Hi @wilbertpol,

Could you please try to update to Dropwizard 1.0.4? We brought back old-style request logging with `logback-classic` and it should work with the syslog appender. You need explicitly set the type of request logging factory, though:
```yaml
  requestLog:
    type: classic
    appenders:
      - type: syslog
        threshold: ALL
        logFormat: "app-ms %msg"
```

wilbertpol:
I did a quick check and request logging is working again. Thanks!

Issue #1737, opened 2016-09-23
No request logs in dropwizard 1.0.1


m-at-nu:
Hello,
since I've updated dropwizard to version 1.0.1 I have no request logs anymore, using dropwizard version 1.0.0 everything works. Is this a known bug or was anything changed regarding the configuration? According to the [release logs](https://github.com/dropwizard/dropwizard/blob/master/docs/source/about/release-notes.rst#v101-sep-21-2016), issue #1678 may be related to this? Even so, having only the requestLog's appenders without logging appenders doesn't work either with version 1.0.1.

I've attached my server configuration _server.yml_ file below:
```
server:
  applicationContextPath: /
  applicationConnectors:
    - type: http
      port: 8080
  requestLog:
    appenders:
      - type: console
logging:
  appenders:
    - type: console
````
I've also attached a small minimal working example. Please change the dropwizard version to 1.0.0 and 1.0.1 respectively to test the issue.

Thanks in advance!
[test.zip](https://github.com/dropwizard/dropwizard/files/489626/test.zip)

**Comment by nickbabcock** [comment]
I can confirm that this is a bug affecting v1.0.1

@m-at-nu correctly identified where the bug was introduced.

I can also confirm that checking out the 1.0.x branch, the following command fixed the issue

```
git revert -m 1 e7c16c97
```

arteam:
Yes, I can also confirm this regression. Which is strange, because it's exactly the problem which had to fixed by #1678. I would like to get some feedback from @evnm and @tjcutajar before reverting the commit.

arteam:
Oh, now I see @tjcutajar pointed out to this problem in #1672, we ignored his comment. How inconsiderate of us.

arteam:
Digging into internals I see that logback doesn't log messages because of a `NullPointerException` in `org.eclipse.jetty.server.Request#extractQueryParameters` method because the `metadata` object is null.

arteam:
It looks like there is a race between calling the `org.eclipse.jetty.server.Request#recycle` method by Jetty and extracting information from the request by Logback.

evnm:
Those are the same Logback codepaths that I've been spelunking in while debugging #1699. There seems to be a growing number of fingers pointing to Dropwizard's Logback-enabled request logging.

tjcutajar:
This is a duplicate of #1686 

[...]

PR #1813, opened 2016-11-10
Bring back request logging with logback-classic

arteam:
Several users reported that they have issues with the logback-access implementation.
See #1801, #1812, #1737. It would be great to allow these users to continue to use `logback-classic`,
while they can't move their appenders to `logback-access`.

### DIFF
--- a/dropwizard-request-logging/src/main/resources/META-INF.services/io.dropwizard.request.logging.RequestLogFactory
+++ b/dropwizard-request-logging/src/main/resources/META-INF.services/io.dropwizard.request.logging.RequestLogFactory
@@ -1,1 +1,2 @@
 io.dropwizard.request.logging.LogbackAccessRequestLogFactory
+io.dropwizard.request.logging.old.LogbackClassicRequestLogFactory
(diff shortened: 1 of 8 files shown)


arteam:
> Should we update the logging documentation and release notes too?

Yes, thanks. Updates the docs and 1.0.4 release notes.

jplock:
@arteam we should probably merge this into `master` too

jplock:
Should we try and merge `release/1.0.x` into `master`?  There seems to be a few commits not present in `master`

arteam:
I've cherry-picked this feature to the master branch f4f8258 and 867a101.

dropwizard-request-logging/src/main/resources/META-INF.services/io.dropwizard.request.logging.RequestLogFactory
```diff
@@ -1 +1,2 @@
 io.dropwizard.request.logging.LogbackAccessRequestLogFactory
+io.dropwizard.request.logging.old.LogbackClassicRequestLogFactory
```
mattnelson: Should this file be located at `/META-INF/services/` instead of `/META-INF.services/`?

arteam:
You're right, I found this mistake too. I fixed it with a follow-up commit
4c57844. It's strange, that this pull request passed all the tests. The
1.0.4 version should correctly provide the request log discovery.
Am 15.11.2016 18:10 schrieb "Matt Nelson" <notifications@github.com>:

Issue #1289, opened 2015-10-20
Configuration Options for dropwizard-views-mustache


lefloh:
The `MustacheViewRenderer` is caching templates forever so there's no chance for hot-deployment during development. Would it make sense to provide an option to disable caching during development or to define a cache-lifetime?

This should be no big change for the `MustacheViewRenderer` but as far as I've seen there's currently no way to pass any configuration to a ViewRenderer. `ViewBundle#getViewConfiguration` returns an empty Map. Am I missing something here?

fexbraun:
Hi,

Return of empty map was introduced to simplify usage of mustache views module: 
"It's awkward to demand configuration from apps, even when they don't have any need to override the defaults."

"You can pass configuration through to view renderers by overriding getViewConfiguration:"
See:  https://dropwizard.github.io/dropwizard/manual/views.html

And +1, I like the idea of enabling hot-deployment during development.

lefloh:
How could I miss that? Thanks!

Seems like then it could easily be solved by using `CacheBuilder#expireAfterWrite`?

jplock:
Is this still an issue?  Do we need to clarify the documentation or is there a code change required to support this functionality?

lefloh:
Yes there is a code change required. I would suggest to add a cache property to the view configuration:

    views:
      .mustache:
        cache: false

The `ViewRenderer` implementations would need to bypass the `LoadingCache` in this case. Shall I provide a pull request?

lefloh:
Tu sum up again: If you want to recompile mustache views during local development you should [create a new MustacheFactory on each request](https://github.com/spullara/mustache.java/issues/57). There's no big code change necessary. An example [can be found here](https://github.com/dropwizard/dropwizard/compare/master...lefloh:1289).

acwwat:
Hello, any update on this issue? This feature would certainly help with faster local development. Even better, this can be combined with support for loading templates from the file system which I have implemented for my own project. Can open new issue and submit PR for it if appropriate.

lefloh:
I already [proposed a solution](https://github.com/dropwizard/dropwizard/compare/master...lefloh:1289) 7 month ago. It would be nice to get some feedback or otherwise just close this issue.

arteam:
Hi @lefloh,

[...]

PR #1814, opened 2016-11-10
#1289: Enabling configuration option for caching of Mustache Views

lefloh:
Solves [1289](https://github.com/dropwizard/dropwizard/issues/1289).


arteam:
Thanks for your contribution!

Issue #1811, opened 2016-11-09
Need finer-grain control of exception behaviour in view renderers


acwwat:
For background details, please refer to the following user group thread: https://groups.google.com/forum/#!topic/dropwizard-user/wy8ksKjz-HA

This issue is a feature request to enable finer-grain control over exceptions thrown when rendering views, so that one can use custom `ExceptionHandler`s to handle certain exception flows.

For example, when using Mustache a `MustacheNotFoundException` is thrown by the Mustache Java API when the template is not found. It is then wrapped/re-thrown as a `RuntimeException` by `MustacheViewRenderer`, which finally gets caught and re-thrown as a `WebApplicationException` with a generic "template error" response. The exception handler logic would ultimately catch this `WebApplicationException` will simply return the enclosed `Response` instead of filtering further. What I would like to be able to do is to somehow use an `ExceptionHandler` to catch a wrapped exception and inspect the cause which is the view implementation-specific exceptions (Mustache, Fremarker, etc.) In my case, I want to create a 404 error response with custom page instead of the generic 500 error response.

It achieve this, one can perhaps add a config option which `ViewMessageBodyWriter` uses to determine if the re-thrown `WebApplicationException` should include a `Response` or the caught exception in the cause. Then I have a choice to drill down into the exception causes and decide what to do with it in my custom `ExceptionHandler`. (But of course, this will not be view renderer agnostic by my own choice, as I am handling renderer-specific exceptions manually.)

Please let me know if more info is needed, otherwise I appreciate that you consider this as a feature request.

acwwat:
Submitted PR #1820 with a possible implementation.

arteam:
Closed via #1820

PR #1820, opened 2016-11-15
Finer-grain control of exception behaviour in view renderers

acwwat:
This PR is one possible implementation of issue #1811. In short, I am standardizing the exception thrown by each `ViewRenderer` implementation as `ViewRenderException`, which is then caught and rethrown as `WebApplicationException` by `ViewMessageBodyWriter`.

To preserve previous behavior of returning an HTML error response on render exceptions, I have added a new `ViewRenderExceptionMapper` class which should be registered. Consequently I am still changing the original behavior by adding this registration requirement. Let me know if this is OK from backward-compatibility standpoint - if needed we can register the mapper in `ViewBundle`, although I don't know how I can register another exception mapper afterwards if I want custom behaviour. This also requires change to the documentation.

Otherwise, one can choose to provide their own `ExceptionMapper` to catch `WebApplicationException` and check that the cause is `ViewRenderException`, then do whatever is needed with its cause in the form of an exception thrown by the rendering code (Viewmarker, Mustache, etc.)

I tried updating the tests to verify both cases of with and without `ViewRenderExceptionMapper`, but I can't seem to get registration in each test case working like this:

` target("/test/bad").regsiter(new ViewRenderExceptionMapper()).request().get(String.class);`

I had to register the mapper in `configure()`. Please let me know if you have any solution to this so I can increase the test coverage.

Thanks.

### DIFF
--- a/dropwizard-views/src/main/java/io/dropwizard/views/ViewMessageBodyWriter.java
+++ b/dropwizard-views/src/main/java/io/dropwizard/views/ViewMessageBodyWriter.java
@@ -3,8 +3,6 @@ package io.dropwizard.views;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.Produces;
 import javax.ws.rs.WebApplicationException;
@@ -27,7 +25,6 @@ import static com.codahale.metrics.MetricRegistry.name;
 @Provider
 @Produces({ MediaType.TEXT_HTML, MediaType.APPLICATION_XHTML_XML })
 public class ViewMessageBodyWriter implements MessageBodyWriter<View> {
-    private static final Logger LOGGER = LoggerFactory.getLogger(MessageBodyWriter.class);
 
     @Context
     private HttpHeaders headers;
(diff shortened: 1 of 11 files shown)


nickbabcock:
I think it would be worthwhile to have a test in `dropwizard-e2e/src/main/java/com/example/app1` showing that the behavior can be changed.

acwwat:
@nickbabcock I added a test/example to e2e on overriding the standard 503 response on missing mustache view with a 404 response.

dropwizard-views/src/main/java/io/dropwizard/views/ViewMessageBodyWriter.java
```diff
@@ -82,12 +76,9 @@ public void writeTo(View t,
                 }
             }
             throw new ViewRenderException("Unable to find a renderer for " + t.getTemplateName());
-        } catch (Exception e) {
+        } catch (ViewRenderException e) {
             LOGGER.error("Template Error", e);
```
dren-dk: My preference would be to format the log message in my own ExceptionMapper, so could this go into the ExceptionMapper as well?

acwwat: That makes sense. I've removed the logging from `ViewMessageBodyWriter` and added the logging code to `ViewRenderExceptionMapper` to preserve the original exception logging behavior.

arteam:
The pull request looks good to me. I don't think the code coverage is so important, what matters you provided relevant tests. And thank you very much for a very detailed documentation for `ViewRenderException`!

PR #1826, opened 2016-11-16
Backport of httpCompliance to Release/1.0.x

dren-dk:
Hi, what is the ETA for 1.1.0?

I'd very much like to avoid having to drag around a local fork, so how do you guys feel about having the httpCompliance setting backported to 1.0.x?


arteam:
We usually avoid to backport changes, except bugfixes (And we released 1.0.4 only a couple of days ago). We probably could make a release candidate for 1.1.0, because there is already a lot of changes in the master branch since 1.0.0. Let us discuss that among maintainers.

dren-dk:
I understand, if the 1.0.x branch isn't going to live for ages, then I'll be happy to wait for 1.1.0

joschi:
@dren-dk With Dropwizard 1.1.0 around the corner (now for real), we won't release another version of Dropwizard 1.0.x.

Issue #1827, opened 2016-11-16
It loses logs when the request log format is customized


Wstunes:
My dropwizard version is 1.0.4
Now my request log is like this in the base.yml
<pre><code>requestLog:
    appenders:
      - type: file
        currentLogFilename: ./logs/runtimeService-requests.log
        archive: true
        logFormat: "%h|%reqParameter{something}|%t|%s|%b|%U|%i{Referer}|%i{User-Agent}"
</pre></code>
The log format is customized because I want to get some parameters out in the url.
But the problem is that if I have a sequence of rapid requests to the server like 1000 requests, it only logs about 50 messages in the log file. I doubt it's because the log formatting costs much.

And if I try the classic request log configuration which generates logs in common log format, which does not meet my requirement. Like the following configuration.
<pre><code>requestLog:
    type: classic
</pre></code>

And send 1000 requests to the server. The 1000 request can be all logged into the log file. Nothing missing.

Why does it lose logs when I use custom request log format?
Now the documentation about request log has too many TODOs. Could someone please add some custom information to it?
And will it be same situation in the logging configuration in addition to the requestLog configuration?
Please help.

Wstunes:
@arteam  Could you please have a look at this issue?

arteam:
Hi @Wstunes,

This issue has already been reported in #1672 and #1678. For some reason using `reqParameter` in the request logging pattern leads to race conditions and inconsistent logging. We tried to fix this issue on the Dropwizard side, but unfortunately without success. We need some collaboration with the logback maintainers to figure out how to correctly fix it.

Wstunes:
@arteam
ah Got it.
  And how long do you estimate it will take to contact the logback team and fix this issue? We encoutered this critical problem in prodution.
  Is there a workaround for now?what about I write a log appender on my own and get the request parameter in the url programmingly? Are there examples or documentations for the  customized log appender?

arteam:
I tracked this issue to a buggy Logback converter (`RequestParameterConverter`) which don't work correctly with async appenders. I am going to create a PR which will replace it with a custom one. Let's see if it will help.

arteam:
I sent #1828, it would be great if you could test it.

Wstunes:
@arteam  I want to test it now. But I add dropwizard from the pom.xml. How can I add your latest change to my project and test it?

arteam:
On a unix machine with Maven, you can do something like that:

[...]

Issue #1686, opened 2016-08-12
Inconsistent request logging


tjcutajar:
I've created this on the back of #1672 where my late comments may have been missed.

The problem looks a bit trickier than just removing the prepareForDeferredProcessing() call. If you run the hello world application in dropwizard-example for instance and hit the /hello-world resource the logging of the requests themselves become inconsistent.

There seems to be another race to call [`appendLoopOnAppenders`](https://github.com/qos-ch/logback/blob/658963544cc08059fc3cec48f5d35f650de5f6df/logback-core/src/main/java/ch/qos/logback/core/spi/AppenderAttachableImpl.java#L45) in [`RequestLogImpl`](https://github.com/qos-ch/logback/blob/658963544cc08059fc3cec48f5d35f650de5f6df/logback-access/src/main/java/ch/qos/logback/access/jetty/RequestLogImpl.java#L138) and [`AsyncAppenderBase`](https://github.com/qos-ch/logback/blob/658963544cc08059fc3cec48f5d35f650de5f6df/logback-core/src/main/java/ch/qos/logback/core/AsyncAppenderBase.java#L265) where they both end up calling `doAppend` in  [`UnsynchronizedAppenderBase`](https://github.com/qos-ch/logback/blob/658963544cc08059fc3cec48f5d35f650de5f6df/logback-core/src/main/java/ch/qos/logback/core/UnsynchronizedAppenderBase.java#L60).

tjcutajar:
Yeah a reproducer would be nice. I will report back if I manage to write a failing test.

evnm:
I too saw your comment on #1672, but hadn't had a chance yet to reply.

I agree that a test case would be the best way to move forward. If possible, we should avoid further shots in the dark like #1678. Given that this latest race is seemingly within Logback itself, perhaps we should reach out to its maintainers. I'm curious if these issues have arisen before and if the Logback team will have any guidance.

nickbabcock:
#1678 was merged.

nickbabcock:
oops, I think I prematurely closed the issue, sorry

mattnelson:
I'm not sure if I'm hitting this exact issue, but I've noticed that access logging looks to be broken with DW-1.0.1-SNAPSHOT. I created a repo with shaded jars against 1.0.0 and 1.0.1-SNAPSHOT. I haven't dug into it a whole lot yet, but I did try managing the logback/jetty versions back in sync with 1.0.0 and that did not resolve the issue.

[...]

PR #1828, opened 2016-11-16
Fix request logs with request parameter in layout pattern

arteam:
Add a safe version of `RequestParameterConverter` which works with async appenders.
It loads request parameters from a cached map rather than trying to load request data
from the original request, which may be closed during the time of a call.  The map is
cached by the  `prepareForDefferedProcessing` method, which invokes before invoking
appenders, so in theory the call to the map should be safe.

As a result, Dropwizard users can use `%reqParameter{something}` in their request
logging patterns.

References #1827, #1672, #1686.

### DIFF
--- a/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/layout/LogbackAccessRequestLayout.java
+++ b/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/layout/LogbackAccessRequestLayout.java
@@ -15,6 +15,12 @@ import java.util.TimeZone;
  */
 public class LogbackAccessRequestLayout extends PatternLayout {
 
+    static  {
+        // Replace the buggy default converter which don't work async appenders
+        defaultConverterMap.put("requestParameter", SafeRequestParameterConverter.class.getName());
+        defaultConverterMap.put("reqParameter", SafeRequestParameterConverter.class.getName());
+    }
+
     public LogbackAccessRequestLayout(Context context, TimeZone timeZone) {
         setOutputPatternAsHeader(false);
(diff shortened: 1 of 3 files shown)


Wstunes:
@coveralls   @arteam   @jplock 
I've tested the new fix_reqAttribute_request_logs branch.
No logs are missing now. But another critical problem appears.
I just used the same log format in the base.yml like this
<pre><code>
requestLog:
    appenders:
      - type: file
        currentLogFilename: ./logs/runtimeService-requests.log
        archive: true
        logFormat: "%h|%reqParameter{something}|%reqParameter{somethingelse}
</pre></code>
But in the request log file it writes
%PARSER_ERROR[something]|%PARSER_ERROR[somethingelse]
The logback request parameter parser is not working fine.


I've done a deep test in your branch and found out the problem.
If the log format configuration in the base.yml contains a patternLayout like reqParameter and reqAttribute, the corresponding "RequestAttributeConverter.class" and "RequestParameterConverter.class" will invoke their start() method to register when the application is started.
But now you've overrided the original RequestParameterConverter to SafeRequestParameterConverter in the defaultConverterMap. I find when the application is on, the start() method in "RequestAttributeConverter.class" is invoked but nothing is invoked in the new "SafeRequestParameterConverter.class". This results in the PASERERROR.
And Could you leave your contact information with me because it's a urgent issue for me . My email is wangshuogr@qq.com. Thanks

arteam:
@Wstunes reviewed and confirmed the pull request works in #1827. I am going to go ahead and merge it. If anyone has concerns with it, please says so.

PR #1830, opened 2016-11-16
Fix configPath in DropwizardTestSupport could only be file

### LINKED ISSUE #1709: DropwizardTestSupport not compatible with Failsafe 2.19
Failsafe 2.19 now runs tests out of JAR files: https://issues.apache.org/jira/browse/SUREFIRE-855

DropwizardTestSupport assumes that `configPath` can be a String (the file path) but this isn't possible for config files located inside JAR files.

The only way I can think of fixing this is to replace `String configPath` with a URI, then open an `InputStream` to read the config file whether it resides in target/classes or inside a JAR file.

However, you just released version 1.0.0 and this sounds like a breaking change ;)

aeolus:
trying to fix #1709


arteam:
Wouldn't it to be better to expose `configurationSourceProvider` as a field for `DropwizardTestSupport`?
Users who load resources from URLs will be able to set `UrlConfigurationSourceProvider` explicitly in a `DropwizardTestSupport` constructor.

PR #1832, opened 2016-11-17
Fix guava deprecations with 20.0 update

nickbabcock:
Fixes the deprecations introduced in the Guava 20.0 update.

Most of the `propagate` can be converted to a `throw new RuntimeException`. The one instance where `throwIfUnchecked` functionality is kept is in `DAOTestRule` due mainly to a test that is written an older junit style that only looks at the shallowest exception and not the root cause.

### DIFF
--- a/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
+++ b/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
@@ -1,6 +1,5 @@
 package io.dropwizard.lifecycle.setup;
 
-import com.google.common.base.Throwables;
 import io.dropwizard.util.Duration;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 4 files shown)


jplock:
Do we want to convert everything into a `RuntimeException`?  I'm not sure

nickbabcock:
So I pushed a changeset (and it will break travis, so don't be alarmed). The issue is that it may break clients who rely on catching `ConstraintViolationException`. If the `ConstraintViolationException` (or another unchecked exception) is wrapped, it becomes harder to catch, as one has to catch `RuntimeException` and look at the cause.

Let me know if you think I'm missing something. If not, I'll fixup the deprecations but keep the same behavior for those who may be relying on unwrapped unchecked exceptions.

jplock:
I think that makes sense.   Looks like most of the impact with this change is in the tests anyway.

arteam:
I think we need to add `Throwables.throwIfUnchecked(e)` before `throw new RuntimeException(e)` to preserve runtime exceptions. I think the Guava documentation suggests to do that.

nickbabcock:
Old behavior reinstated. Should be ok to go 👌

arteam:
Thanks!

PR #1835, opened 2016-11-20
Enable option to load Mustache templates from file system

acwwat:
This PR is for adding the ability to load Mustache templates from the file system instead of from the classpath.

To use this feature, add the new, Mustache-specific config option called `fileRoot` pointing to the directory in the file system where Mustache templates should be loaded. When specified, a `FileSystemResolver` will be used instead of `PerClassSystemResolver`. When combined with the new cache options from #1289 , you can update Mustache templates in the file system and Dropwizard will pick up the changes w/o restarting the application.

From implementation point of view, please see whether there is a better way to handle the tests. AFAIK each `JerseyTest` class can only configure the server once normally, so I cheated and added an instance variable to store the `MustacheViewRenderer` object, such that I can configure it. For each relevant test, I then added a "load from file system" version of the test by configuring the renderer with the `fileRoot` and `cache` option.

Thanks.

### DIFF
--- a/dropwizard-views-mustache/src/main/java/io/dropwizard/views/mustache/MustacheViewRenderer.java
+++ b/dropwizard-views-mustache/src/main/java/io/dropwizard/views/mustache/MustacheViewRenderer.java
@@ -28,16 +28,15 @@ import java.util.Optional;
 public class MustacheViewRenderer implements ViewRenderer {
     private final LoadingCache<Class<? extends View>, MustacheFactory> factories;
     private boolean useCache = true;
-    private File fileRoot;
+    private Optional<File> fileRoot = Optional.empty();
 
     public MustacheViewRenderer() {
-        this.factories = CacheBuilder.newBuilder()
-                                     .build(new CacheLoader<Class<? extends View>, MustacheFactory>() {
-                                         @Override
-                                         public MustacheFactory load(Class<? extends View> key) throws Exception {
-                                             return createNewMustacheFactory(key);
-                                         }
-                                     });
+        this.factories = CacheBuilder.newBuilder().build(new CacheLoader<Class<? extends View>, MustacheFactory>() {
+            @Override
+            public MustacheFactory load(Class<? extends View> key) throws Exception {
+                return createNewMustacheFactory(key);
+            }
+        });
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


dropwizard-views-mustache/src/main/java/io/dropwizard/views/mustache/MustacheViewRenderer.java
```diff
@@ -26,6 +28,7 @@
 public class MustacheViewRenderer implements ViewRenderer {
     private final LoadingCache<Class<? extends View>, MustacheFactory> factories;
     private boolean useCache = true;
+    private File fileRoot;
```
arteam: I think it would be better to declare this field as `Optional` and initialize with `Optional.empty`. Doing so, we explicitly state that field's value can be absent.

dropwizard-views-mustache/src/main/java/io/dropwizard/views/mustache/MustacheViewRenderer.java
```diff
@@ -62,6 +65,9 @@ public void configure(Map<String, String> options) {
         useCache = Optional.ofNullable(options.get("cache"))
             .map(Boolean::parseBoolean)
             .orElse(true);
+        if (options.containsKey("fileRoot")) {
+            fileRoot = new File(options.get("fileRoot"));
```
arteam: If we declare `fileRoot` as `Optional`, we can initialize it with one-liner:
```java
Optional.ofNullable(options.get("fileRoot")).map(File::new)
```

dropwizard-views-mustache/src/main/java/io/dropwizard/views/mustache/MustacheViewRenderer.java
```diff
@@ -75,7 +81,8 @@ public String getSuffix() {
     }
 
     private MustacheFactory createNewMustacheFactory(Class<? extends View> key) {
-        return new DefaultMustacheFactory(new PerClassMustacheResolver(key));
+        return (fileRoot != null) ? new DefaultMustacheFactory(new FileSystemResolver(fileRoot))
```
arteam: I think we could make this initialization more terse if we initialize `MustacheResolver` with the ternary operator and pass it to `DefaultMustacheFactory`.
```java
new DefaultMustacheFactory(fileRoot.isPresent() ? new FileSystemResolver(fileRoot.get()) :
            new PerClassMustacheResolver(key))
```

dropwizard-views-mustache/src/test/java/io/dropwizard/views/mustache/MustacheViewRendererTest.java
```diff
@@ -73,12 +80,32 @@ public void rendersViewsWithAbsoluteTemplatePaths() throws Exception {
     }
 
     @Test
+    public void rendersViewsWithAbsoluteTemplatePathsFromFileSystem() throws Exception {
```
arteam: I think it would be better to create an additional independent test class (which starts a new server, application, etc...) that verifies loading templates from the filesystem, rather than reusing the same renderer across different tests.

arteam:
Thanks for your contribution!

PR #1838, opened 2016-11-23
Update to Jackson 2.8.5

jplock:
This also simplifies the `dropwizard-bom` by just relying on the `jackson-bom` and `jersey-bom`.  All of the tests still pass and the example application is properly packaged without having to maintain all of the individual dependencies in the bom pom.xml.

(the preferred `jackson-bom` version to use is `2.8.5.1` per https://twitter.com/fasterxml/status/801240492000231424)


arteam:
Wow, `jackson-bom` is a great thing.

PR #1842, opened 2016-11-27
General Code Improvements

nickbabcock:
These should be pretty uncontroversial code improvements, but let me know if otherwise. Most of these improvements were automatically done through intellij with me reviewing the changes.

- Fixup mockito deprecations
- Rest of Guava's propagate replaced
- Remove unneeded throws declaration
- Simplify stream calls to collection foreach
- Remove caught and rethrown exception
- Replace anonymous types with lambdas
- Replace isPresent check with ifPresent function
- Replace asList with singleton list with one element
- Use diamond operator
- Prefer contains to indexOf() != -1
- Directly sort list instead of Collections.sort

Feel free to squash this into a single commit


nickbabcock:
Today I learned that Jersey doesn't accept lambda `ExceptionMapper`s 😞 (that's why travis failed earlier, for the curious few)

PR #1851, opened 2016-12-09
Mockito should be a test dependency

umcodemonkey:
Mockito is being exposed as a compile dependency to dependents of dropwizard, which may force them to upgrade to Mockito 2 before they are ready.

### DIFF
--- a/dropwizard-example/pom.xml
+++ b/dropwizard-example/pom.xml
@@ -92,6 +92,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.glassfish.jersey.test-framework.providers</groupId>
             <artifactId>jersey-test-framework-provider-inmemory</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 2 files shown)


evnm:
Thanks for taking care of this. This change is fine, but I question how much of a problem the issue in reality.

Neither `dropwizard-testing` nor `dropwizard-example` should be imported within an application's `compile` scope. Dropwizard's root pom and `dropwizard-bom/pom.xml` both put Mockito in the test scope, so applications depending on `dropwizard-core` shouldn't be affected.

joschi:
@umcodemonkey @evnm I vote for reverting this change.

In a correct Dropwizard project, `dropwizard-testing` will only be imported in `test` scope and so will Mockito.

With the applied change of this PR, you now will always have to explicitly add a dependency to Mockito in all projects using `dropwizard-test`, which was pulled in automatically before.

joschi:
After checking the actual code, this PR seems correct. Sorry for the noise. 😉

umcodemonkey:
To clarify for anyone coming later (thanks @joschi for validating), prior to this change, if you import the `dropwizard-bom`, and depend on both `dropwizard-testing` and `mockito` in test scope, the dependency management in the bom would force your dependency of `mocktio` to Mockito 2. IMO, the `bom` should *not* be managing this dependency, as the point of the `bom` is to manage dependencies of end users, not for internal usage. Also, dropwizard-test does not need to have a `compile` dependency on `mockito`, as it's only used for it's internal tests, and is not required to use any of the classes from `dropwizard-test`.

@joschi is correct that after this change users will no longer get a transitive dependency on `mockito`, but since it a best practice with Maven to explicitly depend on artifacts that you use, this is a net positive to impacted users in any case.

Issue #1853, opened 2016-12-10
HTTPS Connector doc does not accurately reflect Jetty 9.3 default value for excludedCipherSuites


rondagostino:
As per #1700 only TLS 1.2 is supported out-of-the-box with Dropwizard 1.0+ due to a code change in Jetty 9.3 (see https://github.com/eclipse/jetty.project/commit/0a1b0b2bc69ea7e7f5f44992f47a84f926cdeebb) that sets a very restrictive default `excludedCipherSuites` value, but the HTTPS Connector documentation shows the default value for `excludedCipherSuites` as being `(none)`. The documentation should accurately reflect the default, which is `^.*_RSA_.*_(MD5|SHA|SHA1)$`. (I guess regular expressions are supported, which the doc should also indicate.)

rondagostino:
Duplicate of #1702

rondagostino:
Actually, #1702 documents how to workaround the default value in Jetty 9.3 but does not actually fix the configuration guide, which still incorrectly states `(none)` instead of the Jetty 9.3 default of `^.*_RSA_.*_(MD5|SHA|SHA1)$`

evnm:
Good catch. I took a stab at updating the documention in #1854.

arteam:
#1854 has been merged to master. Thanks @rondagostino, for pointing out to this documentation mismatch.

PR #1854, opened 2016-12-10
Document SSL protocol/cipher suite defaults inherited from the JVM and Jetty

evnm:
Addresses #1853

This change updates the user manual and the Javadoc of `io.dropwizard.jetty.HttpsConnectorFactory` to call out the fact that the lists of supported and excluded protocols and cipher suites are inherited from the JVM and Jetty. Specifically, the supported lists fall back to JVM defaults and the excluded lists to defaults [hardcoded in Jetty](https://github.com/eclipse/jetty.project/blob/14611d2f76be9ebd955bff6cdd38fde4a008a4c3/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java#L202-L203).

Note that this change replaces the values of `supportedProtocols` and `supportedCipherSuites` documented in the user manual. Based on my reading of the Jetty code linked above (at the version currently used in Dropwizard, 9.3.14.v20161028), SSLv3 is explicitly excluded, so it doesn't make sense that it would be a supported protocol. Also if the default list of supported cipher suites is [pulled from the JVM](https://github.com/eclipse/jetty.project/blob/14611d2f76be9ebd955bff6cdd38fde4a008a4c3/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java#L319), then I'm unsure how we arrived at the specified default of TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256. Please advise here.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -401,10 +401,10 @@ Extends the attributes that are available to the :ref:`HTTP connector <man-confi
           jceProvider: (none)
           validateCerts: true
           validatePeers: true
-          supportedProtocols: [SSLv3]
-          excludedProtocols: (none)
-          supportedCipherSuites: [TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256]
-          excludedCipherSuites: (none)
+          supportedProtocols: (JVM default)
+          excludedProtocols: [SSL, SSLv2, SSLv2Hello, SSLv3] # (Jetty's default)
+          supportedCipherSuites: (JVM default)
+          excludedCipherSuites: [.*_(MD5|SHA|SHA1)$] # (Jetty's default)
           allowRenegotiation: true
           endpointIdentificationAlgorithm: (none)
 
@@ -494,6 +494,7 @@ This connector extends the attributes that are available to the :ref:`HTTPS conn
           keyStorePassword: changeit
           trustStorePath: /path/to/file # required
           trustStorePassword: changeit
+          supportedCipherSuites: TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256
 
 
 ========================  ========  ===================================================================================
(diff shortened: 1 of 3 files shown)


docs/source/manual/configuration.rst
```diff
@@ -401,10 +401,10 @@ Extends the attributes that are available to the :ref:`HTTP connector <man-confi
           jceProvider: (none)
           validateCerts: true
           validatePeers: true
-          supportedProtocols: [SSLv3]
-          excludedProtocols: (none)
-          supportedCipherSuites: [TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256]
-          excludedCipherSuites: (none)
+          supportedProtocols: (JVM default)
+          excludedProtocols: [SSL, SSLv2, SSLv2Hello, SSLv3] (Jetty's default)
```
nickbabcock: I would feel better if "(Jetty's default)" were in yaml comments, to aid the copy + paste use case 😄 

```yaml
excludedProtocols: [SSL, SSLv2, SSLv2Hello, SSLv3] # (Jetty's default)
```

docs/source/manual/configuration.rst
```diff
@@ -401,10 +401,10 @@ Extends the attributes that are available to the :ref:`HTTP connector <man-confi
...
+          supportedCipherSuites: (JVM default)
+          excludedCipherSuites: [.*_(MD5|SHA|SHA1)$] (Jetty's default)
```
nickbabcock: Comment out the jetty default to convey specifying a regex as an excluded cipher suite is supported (not many people may know this).

docs/source/manual/configuration.rst
evnm: Re: Http2, the manual already states as such:

```
Note that your JVM also must provide ``TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`` cipher.
```

I will add a `supportedCipherSuites` override to the provided h2 connector example configuration.

evnm:
Thanks for the review, @nickbabcock. I figured you'd be a good person to add as a reviewer here, since you've touched the SSL cert code recently.

arteam:
Looks good to me. Sorry for the noise, I tried resolve merge conflicts with the GitHub UI and messed up, but it should be fine now. The change is applied to upstream and 1.0.x as a6cfcc8.

PR #1855, opened 2016-12-11
Switch cert and peer validation to false by default

nickbabcock:
If any either of these two options are enabled, a HTTPs enabled Dropwizard
app will most likely fail to start with cryptic error messages. Since
Jetty has these options disabled by default, we should follow their lead
and disable these options as well.

Future work includes adding in additional certificate features as
Certificate Revocation List (CRL),  CRL Distribution Points Protocol
(CRLDP), On-Line Certificate Status Protocol (OCSP)

This may be considered a bugfix and is a candidate for either 1.0.x or 1.1

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -399,8 +399,8 @@ Extends the attributes that are available to the :ref:`HTTP connector <man-confi
           maxCertPathLength: (unlimited)
           ocspResponderUrl: (none)
           jceProvider: (none)
-          validateCerts: true
-          validatePeers: true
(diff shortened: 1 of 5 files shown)


evnm:
Makes sense to me. We should be noisy when releasing this in the next stable versions, given the possibility that it changes behavior out from under folks.

arteam:
Makes sense to me as well. Merged to upstream and applied to the 1.0.x branch as 7be01126.

nickbabcock:
Updated release notes in 31068e5b and 94fd90f41

PR #1863, opened 2016-12-20
ensuring that server-side JsonProcessingExceptions return JSON, consistent with other 500 errors

ibwhite:
Currently, most 500s from Dropwizard internal exceptions return as application/json, with a nice logging ID, courtesy of `LoggingExceptionMapper`. This is a pleasant default.

However, `JsonProcessingException`s that have been determined to be server-side issues are still returning as the default Jersey HTML format, which is less pleasant.

This PR causes server-side JsonProcessingException 500s to return as application/json. It re-uses the `LoggingExceptionMapper` code and returns in the exact same format as other internal exceptions.

Feedback welcomed.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -45,8 +45,7 @@ public class JsonProcessingExceptionMapper extends LoggingExceptionMapper<JsonPr
          * If the error is in the JSON generation, it's a server error.
          */
         if (exception instanceof JsonGenerationException) {
-            LOGGER.warn("Error generating JSON", exception);
-            return Response.serverError().build();
+            return super.toResponse(exception); // LoggingExceptionMapper will log exception
         }
 
         final String message = exception.getOriginalMessage();
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -13,12 +13,11 @@ import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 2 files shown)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
```diff
@@ -17,7 +18,7 @@
 import java.util.regex.Pattern;
 
 @Provider
-public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProcessingException> {
+public class JsonProcessingExceptionMapper extends LoggingExceptionMapper<JsonProcessingException> implements ExceptionMapper<JsonProcessingException> {
```
evnm: `LoggingExceptionMapper` itself implements `ExceptionMapper`, so the implements clause here isn't necessary.

ibwhite: Thanks, great point. I've pushed a commit to address this as you suggested.

evnm:
I slotted this change for 1.1.0, but technically the change in response payload format would be a breaking change (since folks might be relying on the HTML output to detect 5xx's, for example). I'm inclined to let it slip, since this strikes me as somewhat of a bugfix.

ibwhite:
Thanks for the review and merge @evnm!

That leaves the other place where I've noticed a non-JSON response on errors -- the `DefaultUnauthorizedHandler` for 401s, which returns text/plain.

That one doesn't bug me as much because it's so trivial to override when you set up auth (I have a little mini-lib I use to do this -- https://github.com/ibwhite/starwizard) -- but maybe if we're fixing this error response for 1.1.0, we should fix that default as well in the same version?

Let me know your thoughts, happy to submit a PR for that as well if you'd like.

jplock:
@ibwhite I'm the maintainer of the dropwizard-swagger module.  I like the AuthParamFilter you have in your starwizard project.  Would you be interested in having that incorporated into the main module?

ibwhite:
Hey @jplock -- sure, I think that would be great. I got that code from @pablo-meier FTR. The dropwizard-swagger module is awesome btw, thanks for your work on it.

PR #1864, opened 2016-12-22
Upgrade tomcat-jdbc to 8.5.9

arteam:
Unfortunately, the tomcat-jdbc project doesn't provide own release notes, but
you can find the changelog for the Tomcat project here: (you can see the `tomcat-jdbc` section):
https://tomcat.apache.org/tomcat-8.0-doc/changelog.html. The list of commits here:
https://github.com/apache/tomcat/commits/trunk/modules/jdbc-pool

The biggest feature which  is an additional statistic for the connection pool plus
some bugfixes for handling various timeouts.


joschi:
Thanks!

PR #1865, opened 2016-12-24
Dependency updates

joschi:
Assorted dependency updates.


arteam:
The change looks good to me, but it looks like that the latest Jersey update broke some unit tests for the Jersey module. It looks like the build at Travis was killed because a test took too much time to complete.

joschi:
@arteam Looks like Jetty and Errorprone currently cause some problems. I've removed these commits from this PR to take a look at them separately.

arteam:
Crap, looks like the build failed after the merge. I've restarted it several times, but without success. Some Jersey test never completes in a timely manner and Travis kills a test suite.

joschi:
@arteam The build has passed now. No idea why Travis CI keeps failing from time to time.

Maybe we should take a look at alternatives if this keeps happening.

mattnelson:
Discovered the non-passive changes in logback-1.1.8 today, when we were upgrading all of our transitive dependencies. Glad to see it has already been addressed in dropwizard. Are there plans to either patch back the logback changes to 1.0.X or release 1.1.0?

joschi:
@mattnelson We're aiming to publish the first release candidate of Dropwizard 1.1.0 within January.

Backporting this change to Dropwizard 1.0.x will most likely not happen.

PR #1869, opened 2017-01-04
Avoid NPE while processing NPE

kjetilv:
Some exceptions, most notably NullPointerException, have null messages.
Masking them with a new NullPointerException is unhelpful.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -65,10 +65,11 @@ public class JsonProcessingExceptionMapper extends LoggingExceptionMapper<JsonPr
             // Until completely foolproof mechanism can be worked out in coordination
             // with Jackson on how to communicate client vs server fault, compare
             // start of message with known server faults.
-            final boolean beanError = cause.getMessage().startsWith("No suitable constructor found") ||
-                cause.getMessage().startsWith("No serializer found for class") ||
-                (cause.getMessage().startsWith("Can not construct instance") &&
-                    !WRONG_TYPE_REGEX.matcher(cause.getMessage()).find());
+            final boolean beanError = cause.getMessage() != null &&
+                (cause.getMessage().startsWith("No suitable constructor found") ||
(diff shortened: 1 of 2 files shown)


nickbabcock:
Is there a test case for this so we don't have a regression in the future?

kjetilv:
Not yet, at least. It's just a null check before dereferencing.

nickbabcock:
Ok, just curious if this bug was encountered in the wild or more theoretical.

kjetilv:
I got an actual NPE in the wild, yes. It happens :-)

nickbabcock:
In that case, I would like to see a test case 😄 

That exception mapper is going to see an update when Jackson 2.9 comes out and I just want to make sure that there would be no regressions.

nickbabcock:
Ah sorry, I wasn't clear. The test case should send a payload to a resource endpoint so that the exception mapper would be triggered as it naturally would (see sibling test cases)

kjetilv:
In that case, I need to get back to this later and figure out how to provoke an NPE within the test case.

All I really want is an error handler that doesn't mask my errors with its own. This is a pretty obvious fix, since NPE's do happen in the wild, and they will be masked with a new NPE by this code.

nickbabcock:
Yup, I agree with everything you say. The fix is obvious. What's not obvious is when we bump to jackson 2.9 and refactor the class if the bug will reappear, hence I ask for a test case using Jersey endpoints and JSON payload 😄

kjetilv:
OK, that probably won't happen, sorry. I won't get the time. I hope you're not feeling obligated to let me do the honors or anything. I don't need commit credits, I just need a runtime that handles exceptions :-)

This code assumes that the message of any given Throwable is non-null. It is well-known and documented since the dawn of time as a property that may be null. It's in the Javadoc and has been there since 1.0. Not checking it isn't a bug, it's a trivial mistake, almost a typo. Easily remedied by checking it. These things creep into the code base from time to time, and when we spot them we fix them.

Hoping to keep using dropwizard in the future,

Yours truly,

Kjetil

nickbabcock:
Here's the plan

- I'll apply the null check to the code
- The test case, while exercising the theoretical bug does seem a bit contrived. For now, I won't apply it
- I have a hunch, which I may explore, that a custom deserializer could cause said behavior to be exhibited.

Thanks for the code review 😄

Issue #523, opened 2014-04-03
admin interface hangs on 24-core machine


jbyler:
_Primary symptom_: on a machine with 22 or more cores and the default server configuration, the admin interface accepts TCP connections and then never processes the requests, causing a browser to hang forever.  This can happen on any machine given the wrong config parameters.

Details:
- Server thread configuration must adhere to the following invariants, determined through debugging and trial and error:
  - `maxThreads` > ∑ (`acceptorThreads` + `selectorThreads`) over all `applicationConnectors`
  - `adminMaxThreads` > ∑ (`acceptorThreads` + `selectorThreads`) over all `adminConnectors`
- Presumably what this means is that `maxThreads` _includes_ all the `acceptorThreads` and `selectorThreads`, and what's left over is used for handling requests. If there's nothing left over, the requests queue up and never get handled.
- If this invariant is not satisfied, either the `applicationConnectors` or the `adminConnectors` (respectively) will accept TCP connections but then never handle them
- DropWizard's defaults for `maxThreads` (1024) and `adminMaxThreads` (64) are fixed, while the defaults for `acceptorThreads` (#CPUs/2) and `selectorThreads` (#CPUs) vary on different machines.
- So on a machine with 22 or more cores and an admin interface with 2 connectors (one for HTTP, one for HTTPS), the invariant doesn't hold for the admin interface.  With 342 or more cores and 2 connectors, it won't hold for the application interface.

There are potentially 3 parts to this bug:
- DropWizard defaults should probably be made so they work on all modern hardware.
- DropWizard should probably validate the configuration and fail to start if an invariant isn't satisfied, rather than silently hanging.
- Perhaps the invariant should be documented.
- Alternatively, the meaning of the parameters could be changed so that they are independent.  Using `maxThreads` only for processing requests (and not for selector threads and acceptor threads) might be more intuitive.

Tested with version `v0.7.0.rc3`.

How to reproduce: use the `dropwizard-example` application on a 24-core machine, or use the following server config on any machine:

```
server:
  minThreads: 2
  maxThreads: 2
  applicationConnectors:
    - type: http
      port: 8080
      acceptorThreads: 1
      selectorThreads: 1
  adminConnectors:
    - type: http
      port: 8081
      acceptorThreads: 1
      selectorThreads: 1
```

App starts successfully, but requests to either the application or the admin interface hang.

vishvananda:
This seems to be fixed for maxThreads but not adminMaxThreads.

[...]

PR #1875, opened 2017-01-10
Upgrade Jetty to 9.4.0.v20161208

### LINKED ISSUE #1874: jetty 9.4 support
I couldn't find any activity about jetty 9.4 and it seems 1.1.0 release will also be 9.3 based.

Any plans about it?

I'm asking mainly for unix domain socket support, which will probably require some dropwizard changes in order to support it with yaml. Or we can also live with our own ConnectorFactory until the changes are there I suppose.

arteam:
Hopefully, it's just a version bump.

Resolves #1874.

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
@@ -25,6 +25,7 @@ import org.eclipse.jetty.util.thread.ThreadPool;
 
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 import static com.codahale.metrics.MetricRegistry.name;
@@ -141,13 +142,15 @@ import static com.codahale.metrics.MetricRegistry.name;
  *     </tr>
  *     <tr>
  *         <td>{@code acceptorThreads}</td>
- *         <td>half the # of CPUs</td>
- *         <td>The number of worker threads dedicated to accepting connections.</td>
+ *         <td>(Jetty's default)</td>
+ *         <td>The number of worker threads dedicated to accepting connections.
+ *         By default is <i>max</i>(1, <i>min</i>(4, #CPUs/8)).</td>
  *     </tr>
  *     <tr>
  *         <td>{@code selectorThreads}</td>
- *         <td>the # of CPUs</td>
- *         <td>The number of worker threads dedicated to sending and receiving data.</td>
+ *         <td>(Jetty's default)</td>
+ *         <td>The number of worker threads dedicated to sending and receiving data.
+ *         By default is <i>max</i>(1, <i>min</i>(4, #CPUs/2)).</td>
  *     </tr>
  *     <tr>
  *         <td>{@code acceptQueueSize}</td>
@@ -259,10 +262,10 @@ public class HttpConnectorFactory implements ConnectorFactory {
     private Size maxBufferPoolSize = Size.kilobytes(64);
 
     @Min(1)
-    private int acceptorThreads = Math.max(1, Runtime.getRuntime().availableProcessors() / 2);
+    private Optional<Integer> acceptorThreads = Optional.empty();
 
     @Min(1)
-    private int selectorThreads = Runtime.getRuntime().availableProcessors();
+    private Optional<Integer> selectorThreads = Optional.empty();
 
     @Min(0)
     private Integer acceptQueueSize;
@@ -405,22 +408,22 @@ public class HttpConnectorFactory implements ConnectorFactory {
     }
 
     @JsonProperty
-    public int getAcceptorThreads() {
+    public Optional<Integer> getAcceptorThreads() {
         return acceptorThreads;
     }
 
     @JsonProperty
-    public void setAcceptorThreads(int acceptorThreads) {
+    public void setAcceptorThreads(Optional<Integer> acceptorThreads) {
         this.acceptorThreads = acceptorThreads;
     }
 
     @JsonProperty
-    public int getSelectorThreads() {
+    public Optional<Integer> getSelectorThreads() {
         return selectorThreads;
     }
 
     @JsonProperty
-    public void setSelectorThreads(int selectorThreads) {
+    public void setSelectorThreads(Optional<Integer> selectorThreads) {
         this.selectorThreads = selectorThreads;
     }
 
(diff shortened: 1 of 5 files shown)


joschi:
@arteam It's failing on Travis CI. That's why I didn't create a PR from https://github.com/dropwizard/dropwizard/tree/jetty-9.4.0. 😉

arteam:
Sorry, didn't see that branch. The build in fact fails on Travis CI. Will try then to run it several times there to figure out a reason of failing. *Unfortunately*, the build passes locally, so we can't debug it in a simple way.

nickbabcock:
Heads up, I ran this branch on [appveyor (windows ci) and it passed](https://ci.appveyor.com/project/nickbabcock/dropwizard/build/1.0.92)

By the way, Dropwizard has an `appveyor.yml` in the repo if one of the owners want to allow permissions, so it'll build all commits 😋

arteam:
I've set up an integration with AppVeyour [here](https://ci.appveyor.com/project/arteam/dropwizard), so now it should build all pull requests.

arteam:
I've set up the build on another CI sever [CircleCI](https://circleci.com/gh/dropwizard/dropwizard). Let see whether the build fails there or not.

arteam:
Looks like the build fails on Circle CI at the same test as on Travis CI.
The test suite just hangs up (I guess because the relevant test doesn't
have timeouts).

jplock:
Would closing these - https://github.com/dropwizard/dropwizard/blob/master/dropwizard-testing/src/test/java/io/dropwizard/testing/DropwizardTestSupportTest.java#L154 help at all?

jplock:
This might be dumb, but what if we changed all the localhost's to 127.0.0.1? That would remove any reliance on the DNS resolver being set up properly (as localhost is really just an /etc/resolv.conf) thing.

jplock:
It looks like Jetty's session handling was a major change in 9.4 - http://www.eclipse.org/jetty/documentation/9.4.0.v20161208/upgrading-jetty.html#_upgrading_from_jetty_9_3_x_to_jetty_9_4_0

Would that impact anything?

arteam:
I believe the issue lies somewhere in handling administrative tasks, but still can figure it out. For instances, many tests in `dropwizard-testing` took ~30 seconds, because the thread pool for admin tasks doesn't shut down gracefully. Another possibility is that Jetty 9.4.* somehow creates too many TCP sockets that are not closed correctly, so we hit limit on Linux CI boxes.

arteam:
Looks like I boiled down the problem to `DefaultServerFactory` and admin thread pools. Will try to poke around there.

jplock:
@arteam this may be unrelated, but have you ever seen the `DefaultServerFactory` register the handlers twice:

```
INFO  [2017-01-16 18:18:33,623] org.eclipse.jetty.util.log: Logging initialized @1206ms
INFO  [2017-01-16 18:18:33,682] io.dropwizard.server.DefaultServerFactory: Registering jersey handler with root path prefix: /
INFO  [2017-01-16 18:18:33,682] io.dropwizard.server.DefaultServerFactory: Registering admin handler with root path prefix: /
INFO  [2017-01-16 18:18:34,371] com.smoketurner.uploader.handler.UploadInitializer: SSL Provider: OpenSSL
INFO  [2017-01-16 18:18:34,388] com.smoketurner.uploader.handler.UploadInitializer: SSL ENABLED (certificate: 'scripts/server.crt', key: 'scripts/server_pkcs8.key', trust store: 'scripts/ca.crt')
INFO  [2017-01-16 18:18:34,392] com.smoketurner.uploader.handler.AccessControlListFilter: Loaded 1 IP filter(s) (1 accept, 0 reject)
INFO  [2017-01-16 18:18:34,394] com.smoketurner.uploader.config.NettyConfiguration: Event Loop: NIO
INFO  [2017-01-16 18:18:39,487] io.dropwizard.server.DefaultServerFactory: Registering jersey handler with root path prefix: /
INFO  [2017-01-16 18:18:39,487] io.dropwizard.server.DefaultServerFactory: Registering admin handler with root path prefix: /
INFO  [2017-01-16 18:18:39,494] io.dropwizard.server.ServerFactory: Starting uploader
```

That's from a service I've been messing around with which includes a managed-Netty instance, but I don't know why the "Registering XXX handler with root path prefix" messages are showing up twice.

arteam:
Hm, I have not. Thanks for the pointer, though.

arteam:
Looks like I found the culprit. I believe, the issue is reported in #523. The OS on CI Linux machines returns astonishingly big amount of CPUs (128). Jetty changed their algorithm for calculating the maximum amount of selector and acceptor threads, see https://github.com/eclipse/jetty.project/commit/2d52280179a7d9f8b2449c49bcb0847e02115082. But, in Dropwizard we still set the amount of acceptors as `#CPUs/2` and selectors `#CPU`. Looks like that's too much for Jetty, and it can't handle such a big amount of threads. I propose not to guess the right amount of threads and delegate it to Jetty.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
```diff
@@ -258,11 +258,11 @@ public static ConnectorFactory admin() {
...
 
-    @Min(1)
-    private int acceptorThreads = Math.max(1, Runtime.getRuntime().availableProcessors() / 2);
+    @Min(-1)
+    private int acceptorThreads = -1;
 
-    @Min(1)
-    private int selectorThreads = Runtime.getRuntime().availableProcessors();
+    @Min(0)
+    private int selectorThreads = 0;
```
jplock: zero in this case means use jetty's calculation (not -1)?

arteam: Hm, the Jetty docs say:
> selectors - the number of selector threads, or <=0 for a default value.

I guess we could use -1 for consistency with the `acceptorThreads` property. Another option is to switch to `Optional`.

jplock: I like the explicitness of `Optional` instead of people needing to know that `-1` has a special meaning.

jplock:
Thanks for all your hard work @arteam 👍 I think this is good to merge now.

jplock:
With the optional change, users can still provide overrides in a yaml file and they'll be picked up correct?  I've internally stored things in an Optional, but I haven't used the setters with `@JsonProperty` before to know whether that works or not?

arteam:
Yes, they can. The [HttpConnectorFactoryTest#testParseFullConfiguration](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jetty/src/test/java/io/dropwizard/jetty/HttpConnectorFactoryTest.java#L83) test loads a custom YAML [config](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jetty/src/test/resources/yaml/http-connector.yml) and verifies that options are applied correctly. I verified it on a sample application and properties were picked up correctly. I think even the `@Min` annotation worked.

arteam:
Yay! @jplock, thank you for merging these changes and your suggestions fox fixing the issue.

PR #1876, opened 2017-01-10
Protect against NPE from rogue deserializers

nickbabcock:
If someone defines a custom jackson deserializer, throws a custom `JacksonMappingException`, and overrides `getMessage()` to return `null`, `JsonProcessingExceptionMapper` will have a NPE as reported in #1869. If someone creates such a devious deserializer, it should be treated as a server error, hence a 500 status code is returned.

Closes #1869

thanks @kjetilv
cc @cowtowncoder so that he is aware of this edge case. Not 100% what the behavior of 2.9 will/should be in this situation, but I figured I'd give you a heads up in case this changes anything.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -65,10 +65,11 @@ public class JsonProcessingExceptionMapper extends LoggingExceptionMapper<JsonPr
             // Until completely foolproof mechanism can be worked out in coordination
             // with Jackson on how to communicate client vs server fault, compare
             // start of message with known server faults.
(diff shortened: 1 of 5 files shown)


kjetilv:
Thanks for taking it seriously 👍 

Though I'd argue it is not contrived, and that you won't need a "devious" deserializer to provoke it, just a badly written one. You refer to a rogue JacksonMappingException, but if you look at the code, you'll see it's the CAUSE of the JME that is queried. 

All it takes is for some code to handle a null badly and end up with an NPE, which gets wrapped in the JME. NPE's have null messages, as can any exception. It is perfectly legal according to the Javadoc for Throwable. 

(I agree the test case was contrived though. I may have been subconsciously trying to make a point. Thanks again.)

cowtowncoder:
@nickbabcock thanks for the ping. Since code here is for dropwizard-only, I don't think that directly affects Jackson itself. However, if NPE problems are found from within Jackson components (possible), I would be interested in learning about those to add bit defensive handling.

PR #1877, opened 2017-01-12
Add configuration hooks to UnitOfWorkAwareProxyFactory

scottaj:
Ran into a situation where I want to be able to do some configuration sessions that are set up by the `UnitOfWorkAwareProxyFactory`. Specifically, in our app I wanted to set up a bunch of default hibernate filters that should be turned on for every session. 

In our resource methods that use `UnitOfWork`, we do that by implementing Jersey event listeners with lower priority than the UnitOfWork listener. I wanted to do something similar and poke some configuration in between the session being set up, and my code actually running.

With the current state of things (using 1.0.5) that required us to rewrite our own version of `UnitOfWorkAwareProxyFactory`. Additionally, because `UnitOfWorkAspect` is still package private in the latest release, we also had to copy and create our own version of that.

I spiked out what this change might look like in this PR, it simply takes in two lambdas for a before and after configuration and provides the session factory to both. I'm open to feedback on other approaches to solve this, but I figured the discussion is easier if there is some code.

Let me know any feed back on if this change in general seems reasonable, and on the specific implementation I added.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
@@ -75,7 +75,7 @@ public class UnitOfWorkAwareProxyFactory {
                     factory.create(constructorParamTypes, constructorArguments));
             proxy.setHandler((self, overridden, proceed, args) -> {
                 final UnitOfWork unitOfWork = overridden.getAnnotation(UnitOfWork.class);
-                final UnitOfWorkAspect unitOfWorkAspect = newAspect();
+                final UnitOfWorkAspect unitOfWorkAspect = newAspect(sessionFactories);
                 try {
                     unitOfWorkAspect.beforeStart(unitOfWork);
                     Object result = proceed.invoke(self, args);
@@ -99,9 +99,17 @@ public class UnitOfWorkAwareProxyFactory {
     }
 
     /**
-     * @return a new
+     * @return a new aspect
      */
     public UnitOfWorkAspect newAspect() {
+        return new UnitOfWorkAspect(sessionFactories);
+    }
+
+    /**
+     * @return a new aspect
+     * @param sessionFactories
+     */
+    public UnitOfWorkAspect newAspect(ImmutableMap<String, SessionFactory> sessionFactories) {
         return new UnitOfWorkAspect(sessionFactories);
     }
 }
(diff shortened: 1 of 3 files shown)


scottaj:
I can overload the existing create method with all the new variations. 

#1661 doesn't really solve this issue. It makes it easier to roll your own proxy factory, because at least you don't have to recreate the aspect, but you still have to roll your own because you can't configure `UnitOfWorkAwareProxyFactory`

arteam:
I agree with @ogregoire. I think a better way to to provide such functionality is allow users to implement a custom `UnitOfWorkAspect. What do you think if we:

* Make the `UnitOfWorkAspect` constructor public;
* Make the `configureSession` method in `UnitOfWorkAspect` protected;
* Expose `session` and `sessionFactory` via protected getters;
* Change the `newAspect` method in `UnitOfWorkAwareProxyFactory` to accept `sessionFactories`

With these change, the user could implement custom aspects like this:
```java
public class CustomAspect extends UnitOfWorkAspect {

    public CustomAspect(Map<String, SessionFactory> sessionFactories) {
        super(sessionFactories);
    }

    @Override
    protected void configureSession() {
        super.configureSession();
        Filter filter = getSession().enableFilter("customFilter");
    }
}

new UnitOfWorkAwareProxyFactory("default", sessionFactory) {
            @Override
            public UnitOfWorkAspect newAspect(Map<String, SessionFactory> sessionFactories) {
                return new CustomAspect(sessionFactories);
            }
        }
```

scottaj:
That looks like a much better solution. I can rework this PR to incorporate those changes.

We've found `UnitOfWorkAwareProxyFactory` very useful, we have a number of tasks and some middleware that need database access and that makes it very easy to do that.

scottaj:
I made the changes for the implementation suggested by @arteam and added a quick test.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
```diff
@@ -100,8 +100,9 @@ public UnitOfWorkAwareProxyFactory(HibernateBundle<?>... bundles) {
 
     /**
      * @return a new
+     * @param sessionFactories
      */
-    public UnitOfWorkAspect newAspect() {
-        return new UnitOfWorkAspect(sessionFactories);
+    public UnitOfWorkAspect newAspect(ImmutableMap<String, SessionFactory> sessionFactories) {
+        return new UnitOfWorkAspect(this.sessionFactories);
```
ogregoire: Shouldn't it be `sessionFactories` instead of `this.sessionFactories`? If yes, would it be possible to overload `newAspect` instead of changing its signature?

scottaj: Yes, that was incorrect. Fixed and also made it an overload so we could keep the no args version.

arteam:
@scottaj, thank you for your contribution and @ogregoire for the review!

PR #1883, opened 2017-01-13
Evaluate CircleCI

arteam:
Add support for running tests on CircleCI. The idea is too have an additional CI sever for running tests on the Linux environment, because in recent times we have had some issues with Travis CI.

The reason for that is PR #1875. Currently we have a CircleCI integration set up for the Dropwizard repository, but without a `circle.yml` config file all other pull requests fail. So I would like to merge it to master, so #1875 doesn't block other pull requests, but still allows to run tests in TravisCI and CircleCI.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,18 @@
+machine:
+  timezone: UTC
+  java:
+    version: oraclejdk8
+  environment:
+    _JAVA_OPTIONS: '-Xms512m -Xmx2g'
+dependencies:
+  override:
+    - mvn dependency:go-offline --fail-never
+compile:
+  override:
+    - mvn compile
+test:
+  override:
+    - mvn test
+  post:
+    - mkdir -p $CIRCLE_TEST_REPORTS/junit/
+    - find . -type f -regex ".*/target/surefire-reports/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
--- a/dropwizard-e2e/src/test/java/com/example/forms/FormsAppTest.java
+++ b/dropwizard-e2e/src/test/java/com/example/forms/FormsAppTest.java
@@ -5,11 +5,13 @@ import io.dropwizard.client.JerseyClientBuilder;
 import io.dropwizard.client.JerseyClientConfiguration;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import io.dropwizard.testing.junit.DropwizardAppRule;
+import io.dropwizard.util.Duration;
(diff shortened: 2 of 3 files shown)


circle.yml
```diff
@@ -0,0 +1,23 @@
+machine:
+  timezone: UTC
+  java:
+    version: oraclejdk8
+  environment:
+    _JAVA_OPTIONS: "-Xms512m -Xmx1024m"
```
joschi: I think using `2g` should be fine here. Travis CI and Circle CI allow to use 4 GB of memory per build job.

circle.yml
```diff
@@ -0,0 +1,23 @@
...
+dependencies:
+  override:
+    - mvn dependency:go-offline --fail-never
```
joschi: I think this override could be removed once #1882 has been merged.

circle.yml
```diff
@@ -0,0 +1,23 @@
...
+  override:
+    - mvn test
+  post:
+    - mkdir -p $CIRCLE_TEST_REPORTS/junit/
+    - find . -type f -regex ".*/target/surefire-reports/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
+deployment:
+  staging:
+    branch: master
+    commands:
+      - ./deploy_staging.sh
```
joschi: Do we really want to deploy artifacts from Travis CI *and* Circle CI?

circle.yml
arteam: Makes sense, will change.

circle.yml
arteam: Got it. Will remove the section and rebase the branch against master after #1882 merged.

circle.yml
arteam: Yes, this section shouldn't exist. It's a leftover from experiments on CircleCI. Will remove.

joschi:
We also might want to disable the running default services (such as MongoDB or PostgreSQL) which we don't need: https://circleci.com/docs/build-image-trusty/#databases

circle.yml
arteam: Hm, looks like `mvn dependency:go-offline` still fails on `Dropwizard End-to-end Tests` 
https://circleci.com/api/v1.1/project/github/dropwizard/dropwizard/25/output/11/0?file=true

arteam:
> We also might want to disable the running default services (such as MongoDB or PostgreSQL) which we don't need: https://circleci.com/docs/build-image-trusty/#databases

I tried to do this, but could not figure out how do disable default services. I couldn't pass an empty list to the`services` section, so I guess we need to run _some_ service?

PR #1890, opened 2017-01-17
Add timeouts for the all tests in the `dropwizard-jersey`module

arteam:
Recently, the tests for Jersey resources were rather flaky and failed on Travis CI. Unfortunately, we can't get any debug information about it, because they the test suite is killed by the CI environment because it becomes unresponsive. It would be great to have timeouts for each tests, so even in case we have errors, the whole test suite won't fail.

This is achieved by introducing a new class `AbstractJerseyTest` which contains common features of the all tests for Jersey resources. Now the tests are inherited from it and perform HTTP requests with timeouts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
-sudo: false
+sudo: true
 language: java
 jdk:
   - oraclejdk8
+before_install:
+  - ulimit -c unlimited -S
 after_success:
   - bash .travis_after_success.sh
 cache:
--- a/dropwizard-http2/pom.xml
+++ b/dropwizard-http2/pom.xml
@@ -375,5 +375,17 @@
                 <alpn-boot.version>8.1.10.v20161026</alpn-boot.version>
             </properties>
         </profile>
+        <profile>
+            <id>jdk-1.8.0_121</id>
+            <activation>
+                <property>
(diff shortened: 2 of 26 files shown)


arteam:
Looking at Travis logs, it looks like I got I wanted: now tests fail in a timely manner. Now the question is, _why_ the fail :) The main suspect is `ConstraintViolationExceptionMapperTest` which has a lot of test methods comparing to the rest.

arteam:
Now the build fails during starting HTTP/2 tests on Travis... That's sorcery.

arteam:
Okay, looks like the reason for the failed HTTP/2 tests was #1899 and the fact that we run them in parallel (doesn't make sense anyway, because the time to start a new JVM eats up any performance improvement on such a small set). Because of that we couldn't see an error message (the JVM just hanged up).

Also, I had to set up an unlimited amount of open files, because otherwise the Jersey test suite hangs up. I presume Travis machines have some default limit which we hit during peak test loads (other possibility is that we leak file descriptors in our tests).

arteam:
Yes! Finally the build has passed.

arteam:
I'm going to go ahead and merge this to restore the build on master. If anyone has feedback for this PR, please feel free to express it.

PR #1891, opened 2017-01-17
Add support for HTTP client managed by DropwizardAppRule

arteam:
A common use case for the users of `DropwizardAppRule` is creating an HTTP client which performs requests against the tested server. I think we could make it little bit easier by providing a method for `DropwizardAppRule`which creates a new Jersey HTTP client client managed by the rule. The user
doesn't need to care about creating own client and shutting it down. The created is also have default sane timeouts in contrary to the default Jersey client. In the end, this allows to write tests like that:

```java
RULE.client().target("http://localhost:" + RULE.getLocalPort() + "/hello-world")
    .queryParam("name", name.get())
    .request()
    .get(Saying.class);
```

It makes testing easy for users who don't use the `dropwizard-client` module or find it excessive for testing.

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -338,6 +338,10 @@ running and stop it again when they've completed (roughly equivalent to having u
 ``DropwizardAppRule`` also exposes the app's ``Configuration``,
 ``Environment`` and the app object itself so that these can be queried by the tests.
 
+If you don't want to use the ``dropwizard-client`` module or find it excessive for testing, you can get access to
+a Jersey HTTP client by calling the `client` method on the rule. The returned client is managed by the rule
+and can be reused across tests.
+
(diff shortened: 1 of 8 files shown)


jplock:
I like this a lot.  Would it make sense to expose the builder somehow in case other providers need to be registered on the client?

arteam:
I the builder is exposed via the protected `clientBuilder` method, so the user can override it. Something like that:
```java
new DropwizardAppRule<TestConfiguration>(TestApplication.class, 
resourceFilePath("test-config.yaml")) {
        @Override
        protected JerseyClientBuilder clientBuilder() {
            return super.clientBuilder()
                .withConfig(new ClientConfig()
                    .connectorProvider(new ApacheConnectorProvider()));
        }
```

PR #1897, opened 2017-01-19
added JsonConfigurationFactory to support all JSON files

### LINKED ISSUE #1896: JSON configuration not working
Heyo,
the manual states

> If your configuration file doesn’t end in .yml or .yaml, Dropwizard tries to parse it as a JSON file.

This seems to be no longer the case. In order to actually use a JSON configuration file I had to write a custom ConfigurationFactory.
```java
public class JSONConfigurationFactory<T> extends YamlConfigurationFactory<T> {

	private ObjectMapper mapper;
	private JsonFactory factory;

	public JSONConfigurationFactory(Class<T> klass, Validator validator, ObjectMapper objectMapper, String propertyPrefix) {
		super(klass, validator, objectMapper, propertyPrefix);
		
        // Sub-classes may choose to omit data-binding; if so, null ObjectMapper passed:
        if (objectMapper == null) { // sub-class has no need for mapper
            mapper = null;
            factory = null;
        } else {
            mapper = objectMapper.copy()
                    .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
            factory = new JsonFactory();
        }
	}

	@Override
    public T build(ConfigurationSourceProvider provider, String path) throws IOException, ConfigurationException {
        try (InputStream input = provider.open(requireNonNull(path))) {
            final JsonNode node = mapper.readTree(factory.createParser(input));

            if (node == null) {
                throw new RuntimeException("Configuration at " + path + " must not be empty");
            }

            return build(node, path);
        }
    }
}
```

Maybe I am just misunderstanding something here and  it is simply my fault but I think this is either a bug or an outdated note in the documentation. I would be willing to fix the JSON support and submit a pull request if this is indeed a missing/bugged feature.

manuel-hegner:
changed YAMLConfigurationFactory to support JSON files with different whitespaces

added JSON version of ConfigurationFactoryTest

fixes #1896


arteam:
Thank you for your interest in contributing to Dropwizard! We would be interested in the first-class support for JSON configuration files. But, I believe such support should be provided as a separate configuration factory rather than as a fix for `YamlConfigurationFactory`. Dropwizard is opinionated about its config format and I would prefer maintain `YamlConfigurationFactory` as a parser strictly for YAML files. Despite JSON is a subset of YAML and, in theory, should be parsed it, we can't guarantee such compatibility. I think your original solution with a JSON configuration factory is a right way to go with this feature. At least, in this format it can be accepted.

manuel-hegner:
Okay, I did this as a new ConfigurationFactory now. The remaining codeclimate issues should be all false positives now.

dropwizard-configuration/src/main/java/io/dropwizard/configuration/JsonConfigurationFactory.java
```diff
@@ -0,0 +1,232 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A factory class for loading JSON configuration files, binding them to configuration objects, and
+ * validating their constraints. Allows for overriding configuration parameters from system properties.
+ *
+ * @param <T> the type of the configuration objects to produce
+ */
+public class JsonConfigurationFactory<T> implements ConfigurationFactory<T> {
```
arteam: Very good! But I think we could avoid duplication by extending `JsonConfigurationFactory` from `YamlConfigurationFactory`. We could expose the `getParser` method in `YamlConfigurationFactory`. The method will create a parser from an input stream. `JsonConfigurationFactory` can override this method to produce a `JsonParser` instead of a `YAMLParser`. I had something like that in mind:
```java
public class JsonConfigurationFactory<T> extends YamlConfigurationFactory<T> {

    private final JsonFactory jsonFactory = new JsonFactory();

    public JsonConfigurationFactory(Class<T> klass,
                                    Validator validator,
                                    ObjectMapper objectMapper,
                                    String propertyPrefix) {
        super(klass, validator, objectMapper, propertyPrefix);
    }

    @Override
    protected JsonParser getParser(InputStream input) throws IOException {
        return jsonFactory.createParser(input);
    }

    @Override
    public T build(ConfigurationSourceProvider provider, String path) throws IOException, ConfigurationException {
        try {
            return super.build(provider, path);
        } catch (JsonParseException e) {
            throw ConfigurationParsingException
                .builder("Malformed JSON")
                .setCause(e)
                .setLocation(e.getLocation())
                .setDetail(e.getMessage())
                .build(path);
        }
    }
}
```

YamlConfigurationFactory
```java
protected JsonParser getParser(InputStream input) throws IOException {
        return yamlFactory.createParser(input);
    }

    @Override
    public T build(ConfigurationSourceProvider provider, String path) throws IOException, ConfigurationException {
        try (InputStream input = provider.open(requireNonNull(path))) {
            final JsonNode node = mapper.readTree(getParser(input));
        //...
}
```

dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationFactoryJSONTest.java
```diff
@@ -0,0 +1,44 @@
...
+            factory.build(malformedFile);
+            failBecauseExceptionWasNotThrown(ConfigurationParsingException.class);
+        } catch (ConfigurationParsingException e) {
+            assertThat(e.getMessage())
+                    .containsOnlyOnce("* Malformed JSON at line:");
+        }
+    }
+
+    @Override
+    public void printsDetailedInformationOnMalformedYaml() throws Exception {
```
arteam: I guess we could name the test `printsDetailedInformationOnMalformedJson`.

dropwizard-configuration/src/test/resources/factory-test-invalid.json
```diff
@@ -0,0 +1,3 @@
+{
+	"name": "Boop"
+}
```
arteam: Just a nitpick, could you please add a newline symbol to the end of JSON files? That way they're displayed correctly in terminal windows.

dropwizard-configuration/src/test/resources/factory-test-typo.json
```diff
@@ -0,0 +1,14 @@
+{
+	"name": "Migty Wizard",
+	"type": [
+		"wizard"
+	],
+	"propertis": {
+		"admin": true
+	},
+	"servers": [
```
arteam: Very nice that you use tabs to verify that we can parse any JSON file.

dropwizard-configuration/src/main/java/io/dropwizard/configuration/JsonConfigurationFactory.java
mattnelson: It feels wrong to have a json extend yaml. What about creating a base factory that both yaml and json extend?

arteam: > It feels wrong to have a json extend yaml
I also thought about that, should be other way around. The base factory approach looks even better to me.

dropwizard-configuration/src/test/resources/factory-test-invalid.json
manuel-hegner: I think I did that now but I am not sure if my GIT client removed them again. The github view is confusing for these cases.

arteam:
@manuel-hegner Thank you very much for the contribution and your persistence in addressing feedback!

manuel-hegner:
No problem. It is a pleasure to work for an open source project with a contribution process that works so well.

PR #1899, opened 2017-01-23
Add a jetty-alpn version for the JDK 1.8.0u121

arteam:
Should fix the build on the Travis CI machines, which apparently have been already updated to the new JDK.


arteam:
Superseded and incorporated into #1890. Without it the suite fails hangs up on Jersey tests.

PR #1908, opened 2017-02-01
Register Jackson parameter names module

### LINKED ISSUE #1852: Should jackson ParameterNamesModule be supplied?
As part of the uplift to Java 8 for 1.x should the jackson parameter names module have been supplied by default?

https://github.com/FasterXML/jackson-modules-java8/tree/master/parameter-names

nickbabcock:
Closes #1852

Addresses @arteam's concern about users who don't compile with `-parameters`, as only the `dropwizard-jackson` test module is compiled with `-parameters` and the other modules don't fail.

### DIFF
--- a/dropwizard-jackson/pom.xml
+++ b/dropwizard-jackson/pom.xml
@@ -58,6 +58,10 @@
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-parameter-names</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-afterburner</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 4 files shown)


dropwizard-jackson/src/test/java/io/dropwizard/jackson/ParanamerModuleTest.java
```diff
@@ -0,0 +1,38 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.ObjectWriter;
+import com.fasterxml.jackson.module.paramnames.ParameterNamesModule;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+
+import static org.assertj.core.api.Java6Assertions.assertThat;
```
joschi: Wrong assertion?

nickbabcock:
oops 😊 

Good catch, fixed.

dropwizard-jackson/pom.xml
```diff
@@ -73,4 +77,25 @@
...
+                <artifactId>maven-compiler-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-testCompile</id>
+                        <phase>test-compile</phase>
+                        <goals>
+                            <goal>testCompile</goal>
+                        </goals>
+                        <configuration>
+                            <compilerArgument>-parameters</compilerArgument>
```
arteam: I believe Maven does that by default. But I'm not sure, will try look it up.

nickbabcock: I commented out that line and the tests fails (locally), so it doesn't look like a default.

joschi: Same for me here. I think this is the correct way to enable retrieving method parameters via reflection.

Reference: https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html

nickbabcock:
Updated release notes in 95c0468

Issue #1760, opened 2016-10-06
Unable to package dropwizard-example


jrwiegand:
I tried packaging the dropwizard-example on my machine (MacBook Pro Mid-2014 running macOS v 10.12 with Apache Maven 3.3.9 (installed via Homebrew 1.0.6)) which was unsuccessful. 

I first ran the `mvn package` with a clean local maven repository (deleted ~/.m2/) with the following message output.

``` bash
mvn package
[INFO] Scanning for projects...
Downloading: http://oss.sonatype.org/content/repositories/snapshots/io/dropwizard/dropwizard-bom/1.0.1-SNAPSHOT/maven-metadata.xml
Downloading: http://oss.sonatype.org/content/repositories/snapshots/io/dropwizard/dropwizard-bom/1.0.1-SNAPSHOT/dropwizard-bom-1.0.1-SNAPSHOT.pom
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Non-resolvable import POM: Could not find artifact io.dropwizard:dropwizard-bom:pom:1.0.1-SNAPSHOT in sonatype-nexus-snapshots (http://oss.sonatype.org/content/repositories/snapshots) @ line 38, column 25
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-core:jar is missing. @ line 49, column 21
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-auth:jar is missing. @ line 53, column 21
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-assets:jar is missing. @ line 57, column 21
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-http2:jar is missing. @ line 61, column 21
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-hibernate:jar is missing. @ line 65, column 21
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-migrations:jar is missing. @ line 69, column 21
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-views-freemarker:jar is missing. @ line 73, column 21
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-views-mustache:jar is missing. @ line 77, column 21
[ERROR] 'dependencies.dependency.version' for io.dropwizard:dropwizard-metrics-graphite:jar is missing. @ line 81, column 21
[...]

Issue #1510, opened 2016-04-05
Error building Dropwizard-example


moezubair:
When I try to run
```
mvn package
```
I get the following errors:
```
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-core:jar is missing. @ line 49, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-auth:jar is missing. @ line 53, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-assets:jar is missing. @ line 57, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-http2:jar is missing. @ line 61, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-hibernate:jar is missing. @ line 65, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-migrations:jar is missing. @ line 69, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-views-freemarker:jar is missing. @ line 73, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-views-mustache:jar is missing. @ line 77, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-metrics-graphite:jar is missing. @ line 81, column 21
[ERROR]     'dependencies.dependency.version' for com.h2database:h2:jar is missing. @ line 85, column 21
[ERROR]     'dependencies.dependency.version' for io.dropwizard:dropwizard-testing:jar is missing. @ line 89, column 21
[ERROR]     'dependencies.dependency.version' for org.glassfish.jersey.test-framework.providers:jersey-test-framework-provider-inmemory:jar is missing. @ line 94, column 21
[ERROR]     'dependencies.dependency.version' for org.glassfish.jersey.test-framework.providers:jersey-test-framework-provider-grizzly2:jar is missing. @ line 109, column 21
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
[...]

PR #1913, opened 2017-02-02
added run from root directory for maven command in example readme

zalberico:
This small addition should reduce the confusion that caused the following issues: https://github.com/dropwizard/dropwizard/issues/1760, https://github.com/dropwizard/dropwizard/issues/1510


jplock:
Thanks!

PR #1917, opened 2017-02-06
Add support for neverBlock on AsyncAppenders

### LINKED ISSUE #1916: Expose neverBlock option for async appender
logback offers an option, `neverBlock`, to make it so that an async appender will drop events rather than block when the queue fills. The default for this option is `false`, meaning the async appender will, in fact, block. I'd like to guard against this scenario in my application and prevent threads from backing up waiting on the appender by setting the option to `true`.

I'd propose Dropwizard config expose this option in the configuration block.

markelliot:
fixes #1916

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -924,6 +924,25 @@ single ``Logger``:
 
 .. _man-core-testing-applications:
 
+Common Logging Configuration
+----------------------------
+
+Some settings are available to all loggers.
+
+.. code-block:: yaml
+
+    logging:
+      level: DEBUG
+      appenders:
+        - type: console
+          threshold: WARN
+          target: stderr
+          # all loggers are wrapped to append asynchronously, set neverBlock
+          # to true to ensure that the application will not block if the async
+          # append queue is full (default: false)
+          neverBlock: true
+
+
 Testing Applications
 ====================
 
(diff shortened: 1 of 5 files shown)


docs/source/manual/core.rst
```diff
@@ -924,6 +924,25 @@ single ``Logger``:
 
 .. _man-core-testing-applications:
 
+Common Logging Configuration
```
markelliot: not really sure where to put common settings in docs, LMK if you'd like me to move this somewhere else.

nickbabcock: How do you feel about the [logging section on the configuration page](http://www.dropwizard.io/1.0.6/docs/manual/configuration.html#logging)?

markelliot: cool, moved there

jplock:
@markelliot @nickbabcock should we provide a sane default and set the default to `true` (to never block)?  I'm not sure what the more explicit approach would be or what users would expect the default behavior to be.  /cc @dropwizard/committers

markelliot:
The logback default is false, so I stuck with the existing behavior as the
default -- this seems like it will be the least surprising. Happy to flip
the default if there's a preference to do so.

jplock:
@markelliot good call, that's probably best for now.

PR #1922, opened 2017-02-10
Convencience AbstractDAO method for Query<T>

vvondra:
I've been migrating our codebase to 1.1 and this popped up to be one missing thing from my previous PR:

Since the old Query object is now deprecated, this helper eases migration to the new structure.

Quick follow up to #1871 

Code before:
```java
Query<E> query = currentSession().createQuery(hql, getEntityClass());
```

Code after:
```java
Query<E> query = query(hibernateQuery);
```

btw for our pretty large app the migration has been fairly smooth, `HashLoginService` changed in Jetty, the Hibernate upgrade needs a few things here and there, and Liquibase now forces a default schema, otherwise it seems much much easier than 0.9 -> 1.0


arteam:
LGTM

PR #1931, opened 2017-02-17
dropwizard-auth: Add support for optional resource protection

evnm:
The current support for optionally-protected resources isn't great. We require users to implement a custom auth filter to inject a dummy `SecurityContext` for unauth'd requests.

I had occasion to write such an endpoint at work and decided to take a stab at supporting `Optional` principals with the `@Auth` annotation. With this patch, I'm able to do things like this:

```
@GET
public String getGreeting(@Auth Optional<User> userOpt) {
  if (userOpt.isPresent()) {
    return "Hello, " + userOpt.get().getName() + "!";
  } else {
    return "Greetings, anonymous visitor!"
  }
}
```
This change breaks down into two steps:

1. A new `ContainerRequestFilter` decorator is used to silence exceptions thrown by failed `AuthFilter` invocations. This results in every request succeeding and principals being injected into the `SecurityContext` of auth'd requests.
2. An `OptionalPrincipalContainerRequestValueFactory` class is used to return the optional principal rather than throwing on its absence.

These two new classes are wired into `AuthDynamicFeature` and `AuthValueFactoryProvider` and their polymorphic equivalents. Tests are added to verify the behavior of each combination and the auth section of the used manual is updated to reflect this new capability.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
@@ -11,7 +11,6 @@ import javax.ws.rs.container.DynamicFeature;
 import javax.ws.rs.container.ResourceInfo;
 import javax.ws.rs.core.FeatureContext;
 import java.lang.annotation.Annotation;
-import java.lang.reflect.Type;
 import java.util.Optional;
 
 /**
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
@@ -11,6 +11,8 @@ import javax.ws.rs.container.DynamicFeature;
 import javax.ws.rs.container.ResourceInfo;
 import javax.ws.rs.core.FeatureContext;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Optional;
 
 /**
  * A {@link DynamicFeature} that registers the provided auth filter
@@ -47,7 +49,24 @@ public class AuthDynamicFeature implements DynamicFeature {
     public void configure(ResourceInfo resourceInfo, FeatureContext context) {
         final AnnotatedMethod am = new AnnotatedMethod(resourceInfo.getResourceMethod());
         final Annotation[][] parameterAnnotations = am.getParameterAnnotations();
-        //@DenyAll shouldn't be attached to classes
+        final Class<?>[] parameterTypes = am.getParameterTypes();
+
+        // First, check for any @Auth annotations on the method.
+        for (int i = 0; i < parameterAnnotations.length; i++) {
+            for (final Annotation annotation : parameterAnnotations[i]) {
+                // Optional auth requires that a concrete AuthFilter be provided.
+                if (annotation instanceof Auth && parameterTypes[i].equals(Optional.class) && authFilter != null) {
+                    context.register(new WebApplicationExceptionCatchingFilter(authFilter));
+                    return;
+                } else if (annotation instanceof Auth) {
+                    registerAuthFilter(context);
+                    return;
+                }
+            }
+        }
+
+        // Second, check for any authorization annotations on the class or method.
+        // Note that @DenyAll shouldn't be attached to classes.
         final boolean annotationOnClass = (resourceInfo.getResourceClass().getAnnotation(RolesAllowed.class) != null) ||
             (resourceInfo.getResourceClass().getAnnotation(PermitAll.class) != null);
         final boolean annotationOnMethod = am.isAnnotationPresent(RolesAllowed.class) || am.isAnnotationPresent(DenyAll.class) ||
@@ -55,15 +74,6 @@ public class AuthDynamicFeature implements DynamicFeature {
 
         if (annotationOnClass || annotationOnMethod) {
             registerAuthFilter(context);
-        } else {
-            for (Annotation[] annotations : parameterAnnotations) {
-                for (Annotation annotation : annotations) {
-                    if (annotation instanceof Auth) {
-                        registerAuthFilter(context);
-                        return;
-                    }
-                }
-            }
         }
     }
 
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
@@ -54,13 +54,15 @@ public class AuthDynamicFeature implements DynamicFeature {
         // First, check for any @Auth annotations on the method.
         for (int i = 0; i < parameterAnnotations.length; i++) {
             for (final Annotation annotation : parameterAnnotations[i]) {
-                // Optional auth requires that a concrete AuthFilter be provided.
-                if (annotation instanceof Auth && parameterTypes[i].equals(Optional.class) && authFilter != null) {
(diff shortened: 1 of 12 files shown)


dropwizard-auth/src/main/java/io/dropwizard/auth/OptionalPrincipalContainerRequestValueFactory.java
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * @return {@link Optional}{@code <}{@link Principal}{@code >}
+     *         stored on the request, or {@code Optional.empty()} if
+     *         no object was found.
+     */
+    @Override
+    public Optional<Principal> provide() {
+        final Principal principal = getContainerRequest().getSecurityContext().getUserPrincipal();
+
+        if (principal == null) {
```
jplock: could this be simplified to `return Optional.ofNullable(principal);`?

dropwizard-auth/src/main/java/io/dropwizard/auth/PolymorphicAuthValueFactoryProvider.java
```diff
@@ -57,8 +59,15 @@ public PolymorphicAuthValueFactoryProvider(
      */
     @Override
     public AbstractContainerRequestValueFactory<?> createValueFactory(Parameter parameter) {
-        if (!parameter.isAnnotationPresent(Auth.class) || !principalClassSet.contains(parameter.getRawType())) {
+        final boolean isOptionalPrincipal = parameter.getRawType().equals(Optional.class)
```
arteam: I think we could implement this in a more concise form as well:
```java
if (!parameter.isAnnotationPresent(Auth.class)) {
    return null;
} else if (principalClassSet.contains(parameter.getRawType())) {
    return new PrincipalContainerRequestValueFactory();
} else {
    final boolean isOptionalPrincipal = parameter.getRawType() == Optional.class
        && ParameterizedType.class.isAssignableFrom(parameter.getType().getClass())
        && principalClassSet.contains(((ParameterizedType) parameter.getType()).getActualTypeArguments()[0]);
    return isOptionalPrincipal ? new OptionalPrincipalContainerRequestValueFactory() : null;
}
```

dropwizard-auth/src/main/java/io/dropwizard/auth/WebApplicationExceptionCatchingFilter.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * A {@link ContainerRequestFilter} decorator which catches any {@link
+ * WebApplicationException WebApplicationExceptions} thrown by an
+ * underlying {@code ContextRequestFilter}.
+ */
+class WebApplicationExceptionCatchingFilter implements ContainerRequestFilter {
+    private final ContainerRequestFilter underlying;
+
+    public WebApplicationExceptionCatchingFilter(ContainerRequestFilter underlying) {
+        this.underlying = underlying;
```
arteam: I think we could add a precondition check that `underlying` should be not null

dropwizard-auth/src/test/java/io/dropwizard/auth/AuthResource.java
```diff
@@ -28,6 +28,13 @@ public String showForEveryUser(@Auth Principal principal) {
         return "'" + principal.getName() + "' has user privileges";
     }
 
+    @PermitAll
+    @GET
+    @Path("optional")
+    public String checkOptionalAuth(@Auth Optional<Principal> principalOpt) {
+        return "principal was " + ((principalOpt.isPresent()) ? "" : "not ") + "present";
```
arteam: I guess we could omit the parentheses around `principalOpt.isPresent`

dropwizard-auth/src/main/java/io/dropwizard/auth/OptionalPrincipalContainerRequestValueFactory.java
evnm: Derp, yes.

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java
```diff
@@ -47,23 +49,31 @@ public AuthDynamicFeature(Class<? extends ContainerRequestFilter> authFilterClas
...
         final AnnotatedMethod am = new AnnotatedMethod(resourceInfo.getResourceMethod());
         final Annotation[][] parameterAnnotations = am.getParameterAnnotations();
-        //@DenyAll shouldn't be attached to classes
+        final Class<?>[] parameterTypes = am.getParameterTypes();
+
+        // First, check for any @Auth annotations on the method.
+        for (int i = 0; i < parameterAnnotations.length; i++) {
+            for (final Annotation annotation : parameterAnnotations[i]) {
+                // Optional auth requires that a concrete AuthFilter be provided.
+                if (annotation instanceof Auth && parameterTypes[i].equals(Optional.class) && authFilter != null) {
```
evnm: I typically prefer flatter conditional tree, but in this case it seems reasonable. 👍

dropwizard-auth/src/main/java/io/dropwizard/auth/PolymorphicAuthDynamicFeature.java
```diff
@@ -30,12 +33,24 @@ public void configure(ResourceInfo resourceInfo, FeatureContext context) {
...
-                    context.register(authFilterMap.get(parameterTypes[i]));
-                    return;
+                // If the parameter type is an Optional, extract its type
+                // parameter. Otherwise, use the parameter type itself.
+                final Type paramType = (parameterTypes[i].equals(Optional.class))
+                    ? ((ParameterizedType) parameterGenericTypes[i]).getActualTypeArguments()[0]
+                    : parameterTypes[i];
+
+                if (annotation instanceof Auth && authFilterMap.containsKey(paramType)) {
+                    if (parameterTypes[i].equals(Optional.class)) {
```
evnm: Not sure why I opted for `.equals` all over the place 😑

ibwhite:
@evnm Is there any chance that this feature could make it into an earlier milestone, perhaps the 1.1.x release?

I really need optional resource protection for something I'm working on now, and was actually about to go and write a PR for the exact same approach myself, so I was really delighted to see this written and merged. I was slightly less delighted to see it's two whole minor versions away.

I can rely on 1.2.0-SNAPSHOT for a bit while developing, but I don't want to roll something into prod with a snapshot dependency.

Since Dropwizard made a breaking change to remove the optional flag from `@Auth` several releases back, addressing this is somewhat of a bugfix as well as a feature. Understand if 1.1.x is feature-locked, but figured I'd ask!

evnm:
@ibwhite I also would like to be able to make use of this change in production, so I share your sense of haste.

I'm going to defer to other committers, though. It seems to me that it's somewhat debatable whether this change represents an introduction of new functionality or, as you say, a bugfix-esque improvement. @jplock, @arteam, what do you guys think?

jplock:
I wouldn't mind seeing this in 1.1.0 as well. +1 from me.

evnm:
Set the milestone to 1.1.0. Perhaps this can be cherry-picked into a second 1.1.0 release candidate.

jplock:
I can cherry-pick it into `release/1.1.x` later today

jplock:
Cherry picked into `release/1.1.x` in https://github.com/dropwizard/dropwizard/commit/9c90c512ea3e3b4203d9a321a929e50b71c5417d

ibwhite:
👍  Thank you so much @evnm and @jplock! I really appreciate moving this up. My production codebases thank you!

PR #1936, opened 2017-02-24
Upgrade to Dropwizard Metrics 3.2.0

arteam:
http://metrics.dropwizard.io/3.2.0/about/release-notes.html

### DIFF
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -27,7 +27,7 @@
         <jackson.version>2.8.6</jackson.version>
         <jetty.version>9.4.1.v20170120</jetty.version>
         <servlet.version>3.0.0.v201112011016</servlet.version>
-        <metrics3.version>3.1.2</metrics3.version>
+        <metrics3.version>3.2.0</metrics3.version>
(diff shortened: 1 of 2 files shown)


jplock:
@dropwizard/committers should we cherry-pick this into `release/1.1.x`, or update 1.1.0 with metrics `3.1.3`?  Any thoughts?  Personally I'd probably prefer `3.2.0` as it has been awhile since a new metrics release.

evnm:
I vote for including this change in 1.1.0, provided we can get it into a release candidate or two in order to kick the tires.

arteam:
One of nice features of the 3.2.0 release is the ability to disable some metric attributes from being reported. We can provide the YAML configuration of such attributes the same way we filter metric names.

jplock:
Cherry picked into `release/1.1.x` in https://github.com/dropwizard/dropwizard/commit/676c9ebbbd7785ce57e98be7e685ae8517804407 and added release notes in https://github.com/dropwizard/dropwizard/commit/a61abef87097acc68733444b32f9af4e04fdfde4

jplock:
We should probably create an rc2 soonish :)

PR #1938, opened 2017-02-25
Use the new setDefault() option to set the default metric registry

jplock:
Related to https://github.com/dropwizard/metrics/pull/801


arteam:
Looks good to me. The fact that the `SharedMetricRegistries.setDefault` and `SharedMetricRegistries.getDefault` methods are not synchronized is a little bit scary, but let's try to deal with that in the metrics project.

PR #1939, opened 2017-02-26
Implement TaskServlet::doGet to list tasks lexically.

aaanders:
Tasks are logged on startup but aren't easily accessible otherwise. Implement `TaskServlet::doGet` to print tasks lexically by name.

Example:
```
$ curl http://localhost:8081/tasks
echo
gc
log-level
```
TaskServlet is mapped on "/tasks/*" and any GET request not matching root 
will yield 404 or 405.

404: Task name is unknown.  
405: Task name is registered but must be invoked with POST.

### DIFF
--- a/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
+++ b/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
@@ -7,9 +7,13 @@ import com.codahale.metrics.annotation.ExceptionMetered;
 import com.codahale.metrics.annotation.Metered;
 import com.codahale.metrics.annotation.Timed;
 import com.google.common.base.Charsets;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.io.CharStreams;
 import com.google.common.net.MediaType;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
```diff
@@ -91,6 +95,25 @@ public void add(Task task) {
     }
 
     @Override
+    protected void doGet(HttpServletRequest req,
+                         HttpServletResponse resp) throws ServletException, IOException {
+        if (Strings.isNullOrEmpty(req.getPathInfo())) {
+            try (final PrintWriter output = resp.getWriter()) {
+                resp.setContentType(MediaType.PLAIN_TEXT_UTF_8.toString());
+                final ArrayList<Task> tasks = new ArrayList<>(getTasks());
```
joschi: This (and the following lines) could be replaced by:
```java
getTasks().stream()
    .map(Task::getName)
    .sorted()
    .forEach(output::println);
```

dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
```diff
@@ -7,9 +7,13 @@
 import com.codahale.metrics.annotation.Metered;
 import com.codahale.metrics.annotation.Timed;
 import com.google.common.base.Charsets;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.io.CharStreams;
 import com.google.common.net.MediaType;
+import java.util.ArrayList;
```
joschi: Please remove the unused imports. Then we're good to go.

dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
aaanders: Nice, thanks @joschi, nice and clean. (Grepping for `stream()` made me think it was a bad idea to use it :D )

dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
aaanders: Fixed

aaanders:
Thanks for reviewing @joschi

Issue #1941, opened 2017-02-27
Logging disabled after uncaught exception


Sijmen:
If an uncaught exception is thrown in the `run` method of the `Application` class, this cannot be logged by the `main` method which started the application. Logging in the shutdown hook and ucaught exception handler is also disabled. This is due to the fact that in the event of a shutdown Dropwizard [runs a cleanup](https://github.com/dropwizard/dropwizard/blob/7295c5faf42840325c161cc82ad4ed1db2bf8419/dropwizard-core/src/main/java/io/dropwizard/cli/ConfiguredCommand.java#L88) method in `ConfiguredCommand`, which stops the logging. In the event of an unexpected exception logging is required in order to debug the program.  This issue is similar to #1119.

I've got two workarounds:

1.  Catch all exceptions in the `run` method of the `Application`. However the method signature suggests that exceptions should be handled and can be thrown. The disadvantage is that logging in the shutdown hook and the uncaught exception handler will not work. 
2. Create a new command `start` which extends `ServerCommand` and overrides the `cleanup` method with a no-op. This keeps the logging alive till the last moment of shutdown. 

I would like configuration or a command option in order to prevent the logging from stopping instead of the current workarounds. 

Class to reproduce the situation. Ran it in eclipse with console logging. I'm using version 1.0.2.

```java
import java.lang.Thread.UncaughtExceptionHandler;

import org.eclipse.jetty.util.component.LifeCycle;
import org.eclipse.jetty.util.component.LifeCycle.Listener;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import io.dropwizard.Application;
import io.dropwizard.Configuration;
import io.dropwizard.setup.Environment;

public class DropwizardExample extends Application<Configuration> {
    private static final Logger LOG = LoggerFactory.getLogger(DropwizardExample.class);

    public static void main(String[] args) throws Throwable {
        LOG.info("Uninitialized logger");
        enableShutdownLogging();
        enableUncaughtExceptionLogging();
        try {
            new DropwizardExample().run(new String[] { "server" });
        } catch (Throwable e) {
            LOG.info("Cant see me, now can you?");
            LOG.error("O no...", e);
            System.out.println("You see me, but no timestamp and no appenders");
            throw e;
        }
        LOG.info("Finished running");
    }

    @Override
    public void run(Configuration configuration, Environment environment) throws Exception {
        LOG.info("You can see me...");
[...]

Issue #1119, opened 2015-06-19
Unable to log after RULE.getApplication().run("db", "migrate", CONFIG_PATH);


joaocenoura:
I'm unable see any logs after calling `RULE.getApplication().run("db", "migrate", CONFIG_PATH);`. Take the following example:  
If I add some loggers to `com.example.helloworld.IntegrationTest.migrateDb()`:

```
    @BeforeClass
    public static void migrateDb() throws Exception {
        LoggerFactory.getLogger(IntegrationTest.class).warn("now you see me");
        RULE.getApplication().run("db", "migrate", CONFIG_PATH);
        LoggerFactory.getLogger(IntegrationTest.class).warn("now you don't");
        System.out.println("+++++++++ just a control line +++++++++");
    }
```

I'll get the following output:

```
WARN  [2015-06-19 09:31:53,676] com.example.helloworld.IntegrationTest: now you see me
INFO  [2015-06-19 09:31:54,767] liquibase: Successfully acquired change log lock
INFO  [2015-06-19 09:31:55,964] liquibase: Creating database history table with name: PUBLIC.DATABASECHANGELOG
INFO  [2015-06-19 09:31:55,969] liquibase: Reading from PUBLIC.DATABASECHANGELOG
INFO  [2015-06-19 09:31:55,983] liquibase: migrations.xml: 1::codahale: Table people created
INFO  [2015-06-19 09:31:55,983] liquibase: migrations.xml: 1::codahale: ChangeSet migrations.xml::1::codahale ran successfully in 6ms
INFO  [2015-06-19 09:31:56,114] liquibase: Successfully released change log lock
+++++++++ just a control line +++++++++
```

After that I can't see any logs from slf4j (from the test class, applications class etc). I've tried to change logging settings in `test-example.yml` but without success.  
I'm using maven 3.0.5 and openjdk 1.7.0_79.

joaocenoura:
Sorry, just had the opportunity to dig deeper just now.  
The problem seems to be in `io.dropwizard.logging.DefaultLoggingFactory`, particularly in:
```
private Logger configureLevels() {
        final Logger root = loggerContext.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
        loggerContext.reset();
        ...
}
```
and also in `io.dropwizard.cli.ConfiguredCommand`
```
protected void cleanup() {
    if (configuration != null) {
        configuration.getLoggingFactory().stop();
    }
}
```
[...]

PR #1947, opened 2017-03-03
Flush loggers on cleanup, don't destroy

nickbabcock:
This PR fixes the issue where dropwizard would close all logging appenders in tests and in code when it's reasonable for there to be more log statements (eg. in case of an exception)! The fix is instead of closing all appenders, wait for them to flush all their log messages (which, as far as I can tell, was the original impetus for closing the `LoggerContext` on application shutdown, anyways)

In tests, the logging is reset to a sane default (I've chosen the default console appender as a default), so that users may still see the output of logging statements in their tests code (outside application code).

1. It's probably prudent to not merge this for 1.1.0 as logging is finicky, even though I've included a decently thorough test case (there's an API change to `LoggingFactory` 😞 to communicate resetting for tests).

Closes https://github.com/dropwizard/dropwizard/issues/1119
Closes https://github.com/dropwizard/dropwizard/issues/1941

CC'ing original authors: @Sijmen, @joaocenoura

EDIT: and of course, any comments/reviews/suggestions are sincerely welcome 😄

### DIFF
--- a/dropwizard-e2e/src/main/java/com/example/badlog/BadLogApp.java
+++ b/dropwizard-e2e/src/main/java/com/example/badlog/BadLogApp.java
@@ -1,0 +1,25 @@
+package com.example.badlog;
+
+import io.dropwizard.Application;
+import io.dropwizard.Configuration;
+import io.dropwizard.setup.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BadLogApp extends Application<Configuration> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(BadLogApp.class);
+
+    @Override
+    protected void onFatalError() {
+        LOGGER.warn("Mayday we're going down");
+    }
+
+    public static void runMe(String[] args) throws Exception {
+        new BadLogApp().run(args);
+    }
+
+    @Override
+    public void run(Configuration configuration, Environment environment) throws Exception {
+        throw new RuntimeException("I'm a bad app");
+    }
+}
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java
@@ -1,11 +1,14 @@
 package io.dropwizard.logging;
 
+import ch.qos.logback.classic.AsyncAppender;
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.classic.jmx.JMXConfigurator;
 import ch.qos.logback.classic.jul.LevelChangePropagator;
 import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.Appender;
+import ch.qos.logback.core.ConsoleAppender;
 import ch.qos.logback.core.util.StatusPrinter;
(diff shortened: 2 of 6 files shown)


dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java
```diff
@@ -167,12 +172,59 @@ public void stop() {
...
+            // giving up that the appender will be completely flushed.
+            final Logger logger = loggerContext.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
+            final ArrayList<Appender<ILoggingEvent>> appenders = Lists.newArrayList(logger.iteratorForAppenders());
+            for (Appender<ILoggingEvent> appender : appenders) {
+                if (appender instanceof AsyncAppender) {
+                    flushAppender((AsyncAppender) appender);
+                }
+            }
+        } catch (InterruptedException ignored) {
+            // If the thread waiting for the logs to be flushed is aborted then
```
jplock: is there any benefit to putting a `Thread.currentThread().interrupt();` in here?

arteam: I think it always a good practice, because we generally don't control the thread which loads `DefaultServerFactory`. There is a lot of people who use Dropwizard within app servers.

nickbabcock: Ah yes, great catch. Updated 👍

jplock:
@nickbabcock is this good to merge?

dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java
```diff
@@ -167,12 +172,60 @@ public void stop() {
...
+            Thread.sleep(100);
+            timeWaiting += 100;
+        }
+
+        if (appender.getNumberOfElementsInQueue() > 0) {
+            // It may seem odd to log when we're trying to flush
+            // a logger that isn't flushing, but the same warning
+            // is issued inside appender.stop() if the appender
+            // isn't able to flush.
+            appender.addWarn("Possibly " +
```
evnm: Mild preference to word this "n records will possibly be discarded".

nickbabcock: Compromise with "n events may be discarded"? I realized that "will possibly" means "maybe" 😊

dropwizard-e2e/src/main/java/com/example/badlog/BadLogApp.java
```diff
@@ -0,0 +1,25 @@
+package com.example.badlog;
+
+import io.dropwizard.Application;
+import io.dropwizard.Configuration;
+import io.dropwizard.setup.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BadLogApp extends Application<Configuration> {
```
nickbabcock: I agree that public is not needed, but dropwizard-e2e isn't deployed to maven and it's not imported into other dropwizard modules. The other applications in e2e suffer the same problem that the access is too wide. Maybe another PR is needed to clean up all the unnecessary wide access.

evnm: Ah, right. sgtm.

PR #1948, opened 2017-03-05
Add some commonly used and useful jersey request/response filters

jplock:
* `X-Request-Id` filter is based upon https://devcenter.heroku.com/articles/http-request-id
* `X-Runtime` filter is based upon https://github.com/rack/rack/blob/master/lib/rack/runtime.rb

These are not registered by default.  They are merely provide for convenience if people want to opt-in to using them.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/CharsetFilter.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/CharsetFilter.java
@@ -14,7 +14,7 @@ import javax.ws.rs.ext.Provider;
 
 /**
  * This class ensures that any HTTP response that includes a Content-Type
- * response header, that it also includes the UTF-8 character set.
+ * response header, will also include the UTF-8 character set.
  */
 @Provider
 @Priority(Priorities.HEADER_DECORATOR)
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/CharsetFilter.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/CharsetFilter.java
@@ -1,0 +1,35 @@
+package io.dropwizard.jersey.filter;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Locale;
+import javax.annotation.Priority;
+import javax.ws.rs.Priorities;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * This class ensures that any HTTP response that includes a Content-Type
+ * response header, that it also includes the UTF-8 character set.
+ */
+@Provider
+@Priority(Priorities.HEADER_DECORATOR)
+public class CharsetFilter implements ContainerResponseFilter {
+
+    private static final String UTF_8 = StandardCharsets.UTF_8.displayName(Locale.ENGLISH);
+
+    @Override
+    public void filter(final ContainerRequestContext request,
+            final ContainerResponseContext response) throws IOException {
+
+        final MediaType type = response.getMediaType();
+        if (type != null && !type.getParameters().containsKey(MediaType.CHARSET_PARAMETER)) {
+            final MediaType typeWithCharset = type.withCharset(UTF_8);
+            response.getHeaders().putSingle(HttpHeaders.CONTENT_TYPE, typeWithCharset);
+        }
+    }
+}
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java
@@ -14,7 +14,7 @@ import org.slf4j.LoggerFactory;
 /**
  * This class adds a "X-Request-Id" HTTP response header and logs the following
  * information: request method, request path, request ID, response status,
- * response bytes (or -1 if not known).
+ * response length (or -1 if not known).
  *
  * @see https://devcenter.heroku.com/articles/http-request-id
  */
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java
@@ -1,0 +1,36 @@
+package io.dropwizard.jersey.filter;
+
+import java.io.IOException;
+import java.util.UUID;
+import javax.annotation.Priority;
+import javax.ws.rs.Priorities;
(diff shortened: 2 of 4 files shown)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/CharsetFilter.java
```diff
@@ -0,0 +1,35 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * This class ensures that any HTTP response that includes a Content-Type
+ * response header, that it also includes the UTF-8 character set.
```
evnm: s/, that it/ will/
s/includes/include/

dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RequestIdFilter.class);
+    private static final String REQUEST_ID = "X-Request-Id";
+
+    @Override
+    public void filter(final ContainerRequestContext request,
+            final ContainerResponseContext response) throws IOException {
+
+        final UUID id = UUID.randomUUID();
+        LOGGER.info("method={} path={} request_id={} status={} bytes={}",
```
evnm: Maybe `length` instead of `bytes`?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.ext.Provider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class adds a "X-Request-Id" HTTP response header and logs the following
+ * information: request method, request path, request ID, response status,
+ * response bytes (or -1 if not known)
+ */
```
evnm: Let's add a `@see` pointing to e.g. the Heroku documentation.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java
joschi: Shouldn't that be logged on DEBUG or TRACE instead of INFO where it potentially spams the application logs?

dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RequestIdFilter.class);
+    private static final String REQUEST_ID = "X-Request-Id";
+
+    @Override
+    public void filter(final ContainerRequestContext request,
+            final ContainerResponseContext response) throws IOException {
+
+        final UUID id = UUID.randomUUID();
+        LOGGER.info("method={} path={} request_id={} status={} length={}",
```
joschi: Shouldn't that be logged on DEBUG or TRACE instead of INFO where it potentially spams the application logs?

jplock: Good point.  I'll lower it to TRACE and also see if I can extract `X-Request-Id` from the request and use that if it exists to maintain any flows.

mattnelson:
I'm going to play devils advocate on two of these new filters. No issues with the charset filter

RequestIdFilter could be modeled of a more opinionated stance on request tracking similar to this bundle. It doesn't handle requests that already have an id in the request header. It isn't utilizing MDC which enables rich correlation of logs events from the entire request. Though the MDC does require the user to property format their log patterns.
https://github.com/service-enabled/dropwizard-request-tracker

RuntimeFilter seems to be duplicating the functionality/purpose of the access logs and will likely report a different time since it will be based on when the filter executed vs when jetty logs the access event.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/CharsetFilter.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * This class ensures that any HTTP response that includes a Content-Type
+ * response header, will also include the UTF-8 character set.
+ */
+@Provider
+@Priority(Priorities.HEADER_DECORATOR)
+public class CharsetFilter implements ContainerResponseFilter {
+
+    private static final String UTF_8 = StandardCharsets.UTF_8.displayName(Locale.ENGLISH);
```
mattnelson: Should this defer to the default?

http://docs.oracle.com/javase/8/docs/api/java/nio/charset/Charset.html#defaultCharset--

jplock:
@mattnelson thanks for the feedback.  I think I can make the `X-Request-Id` better by looking for an existing `X-Request-Id` header in the request and using that value instead of generating a new UUID.  I'll take a look at the request-tracker link to see if there's other functionality which may be missing.  The `X-Runtime` filter was just to provide an easy mechanism for knowing how long an API response took.  If you think this is confusing, I can remove it, but I've found it useful in some of the API's I've built.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/CharsetFilter.java
jplock: I've specifically used this in the past to enforce adding a `;charset=utf-8` into JSON responses, so I'm not sure I'd want to default this to whatever `defaultCharset` returns.

mattnelson: Should the filter be changed to only decorate json responses?

jplock: I can do that

mattnelson:
The request filter should probably be opt-in since UUID performance is highly dependent on the concurrent load.

[1] http://www.cowtowncoder.com/blog/archives/2010/10/entry_429.html
[2] https://github.com/Netflix/netflix-commons/blob/v0.3.0/netflix-commons-util/src/main/java/com/netflix/util/concurrent/ConcurrentUUIDFactory.java

jplock:
@mattnelson none of these are required.  They are just provided in the code so someone can optionally use them.  They aren't being tied in by default.

mattnelson:
> They aren't being enabled by default.

That alleviates most of my concerns. I was thinking that these filters would be auto registered based on the annotations, but it does not appear that this package is added to the package scanning list.

jplock:
@mattnelson I pushed up some updates to address your feedback

mattnelson:
> I pushed up some updates to address your feedback

Changes look good to me.

arteam:
Thanks @jplock for the contribution, @joschi, @evnm, and @mattnelson for the review!

PR #1957, opened 2017-03-12
Support disabled and enabled attributes for metrics

arteam:
Metrics 3.2.* allows to configure the set of attributes which can be disabled for a reporter. It would be nice to provide the ability to set them via the YAML config the same way we provide that for
metric names.


jplock:
@arteam mind updating the release notes too?

joschi:
@arteam Please update the documentation as well: https://github.com/dropwizard/dropwizard/blob/release/1.1.x/docs/source/manual/configuration.rst#metrics

arteam:
Will do.

arteam:
Thanks for fixing the markup!

Issue #1806, opened 2016-11-08
UnitOfWorkApplicationListener doesn't seem to work with sub-resources


victornoel:
Hi,

I have this kind of resource definition:
```java
@Singleton
@Path("/workspaces")
public class WorkspacesResource {

    @POST
    @Consumes(MediaType.APPLICATION_JSON)
    @Produces(MediaType.APPLICATION_JSON)
    @UnitOfWork
    public Workspace create(NewWorkspace ws, @Pac4JProfile CockpitProfile profile) {
        WorkspacesDAO.DbWorkspace w = workspaces.create(ws.name, profile.getUser());

        return new Workspace(Long.toString(w.getId()), w.getName(),
                w.getUsers().stream().map(UsersDAO::username).toArray(String[]::new));
    }

    @Path("/{wsId}")
    public Class<WorkspaceResource> workspaceResource() {
        return WorkspaceResource.class;
    }

    @Singleton
    public static class WorkspaceResource {

        @GET
        @Produces(MediaType.APPLICATION_JSON)
        @UnitOfWork(readOnly = true, transactional = false)
        public WorkspaceTree get(@PathParam("wsId") long wsId, @Pac4JProfile CockpitProfile profile) {
            DbWorkspace w = getWorkspace(wsId, profile);

            return new WorkspaceTree(Long.toString(w.getId()), w.getName());

        }

        @Path("/buses")
        public Class<BusesResource> getBuses() {
            return BusesResource.class;
        }
    }
}
```

And while the `UnitOfWork` annotations on `WorkspacesResource`'s methods initialise the `Session` as needed, the annotations on `WorkspaceResource` do not!
I checked in debug what happens inside `UnitOfWorkApplicationListener.registerUnitOfWorkAnnotations(ResourceMethod)` and it does seem that the method is registered as it should for the sub-resources!

In case it is a bug, would you have a workaround for this until it is fixed? Thanks :)

arteam:
Hi @victornoel,

It looks like a bug. For some reason `UnitOfWorkApplicationListener` doesn't intercept requests to sub-resources.  As a workaround you could try to use `UnitOfWorkAwareProxyFactory` to create sub-resources:
```java
@Path("/{wsId}")
public WorkspaceResource workspaceResource() {
    return new UnitOfWorkAwareProxyFactory(hibernateBundle).create(WorkspaceResource.class);
}
```

victornoel:
[...]

PR #1959, opened 2017-03-13
Support `@UnitOfWork` in sub-resources

arteam:
Currently Dropwizard doesn't open transactions in sub-resources as reported in #1806. The problem is that Dropwizard scans resource methods for `@UnitOfWork` during resource initialization and sub-resource methods are not resolved at that time and can't be registered by Dropwizard. The fix is
to defer the lookup of the `@UnitOfWork` annotation on a method until it's invoked and than cache it. One downside of this approache is that `ConcurrentMap` doesn't support null values, so we have to use `Optional` from `UnitOfWork` which can add an additional overhead, but it should be negligible.

### DIFF
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -40,6 +40,7 @@ v1.1.0: Unreleased
 * Register Jackson parameter-names modules `#1908 <https://github.com/dropwizard/dropwizard/pull/1908>`_
 * Native Jackson deserialization of enums when Jackson annotations are present `#1909 <https://github.com/dropwizard/dropwizard/pull/1909>`_
 * Add `JsonConfigurationFactory` for first-class support of the JSON configuration `#1897 <https://github.com/dropwizard/dropwizard/pull/1897>`_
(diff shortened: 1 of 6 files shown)


jeppe-style:
I have an alternative fix for this issue that adds the sub-resource locator methods in the `UnitOfWorkApplicationListener` (it was done as an university assignment). Please see the [code in my fork](https://github.com/jeppe-style/dropwizard/commit/410c6b9e033414c7c2356b672e93c7b6082498b5#diff-4dd0b0c461524faa7c0b0ba7a996054dR106). I also created a test for it based on the `JerseyIntegrationTest`.

It has one caveat that when an

arteam:
Thanks! I will take a look at your fix tomorrow.

arteam:
I had a look and your proposed fix and it works too. Nevertheless, I believe it's better to go with the pull request proposed by me. The motivation is too try to switch from the static lookup to dynamic to avoid guessing completely and delegate it to Jersey. We can use your fix as a backup plan in case the original fix causes any problems.

Issue #1546, opened 2016-05-14
setContextPath doesn't seem to work.


TomCools:
Maybe it's just me doing something wrong, but in my run(con,env) method, i do the following:
`environment.getApplicationContext().setContextPath("/rest");
 environment.jersey().register(HelloWorldEndpoint.class);`

The HelloWorldEndpoint has the **@Path("/hello")** annotation.

When i launch dropwizard however (without config, just using the **server** command), my endpoint is available on _localhost:8080/hello_, not on _localhost:8080/rest/hello_ wich i would expect.

I have tried this in multiple versions, but it doesn't seem to work. Versions tried:

- 1.0.0-rc2
- 0.9.2
- 0.8.5

Logs for 1.0.0-rc2 list: _Registering jersey handler with root path prefix: /_

nickbabcock:
Is modifying the `rootPath` in the configuration not working as well (it does a `setUrlPattern` behind the scenes)?

TomCools:
rootPath works, but i wonder what the setContextPath is used for then?
I like to set it in code for a simple integration test, for which i just want to invoke the run()-method.

nickbabcock:
Ok I looked into a little more and the reason why you're seeing this issue is that the server is not configured until after the Application's `run`. During server configuration it overwrites whatever context path may have been previously set (`SimpleServerFactory` and `DefaultServerFactory`). To get a custom context path in tests use `ConfigOverride` or a different configuration file to set the `applicationContextPath`

PR #1965, opened 2017-03-15
Enable setting context path programmatically 

ghenkes:
As mentioned in #1546, setting the context path programmatically in an application's `run` method is being ignored.
`environment.getApplicationContext().setContextPath(somePath);`

This is due to DefaultServerFactory's `build` method (which executes after the application's `run`) reapplying the default (`/`). However, since #1504, the context path information has been getting set prior to the application's run method being invoked, so the configuration being performed in `build` is redundant. Here's the execution flow:

[EnvironmentCommand.run](https://github.com/dropwizard/dropwizard/blob/v1.0.6/dropwizard-core/src/main/java/io/dropwizard/cli/EnvironmentCommand.java#L40-L44)
1. calls `configure` which sets the context paths
2. calls the application's `run` method (where I'm attempting to set context path)
3. calls `run` which ends up executing ServerCommand's `run` method

[ServerCommand's `run` method]( https://github.com/dropwizard/dropwizard/blob/v1.0.6/dropwizard-core/src/main/java/io/dropwizard/cli/ServerCommand.java#L49)
1. calls ServerFactory's `build` method which calls `configure` again, wiping out the what the application set.

I'm proposing removing the call to `configure` in DefaultServerFactory's `build` method, as it is redundant and overwrites attempts to set the context path.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/server/DefaultServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/server/DefaultServerFactory.java
@@ -152,9 +152,6 @@ public class DefaultServerFactory extends AbstractServerFactory {
 
     @Override
     public Server build(Environment environment) {
-        // ensures that the environment is configured before the server is built
-        configure(environment);
-
         printBanner(environment.getName());
         final ThreadPool threadPool = createThreadPool(environment.metrics());
         final Server server = buildServer(environment.lifecycle(), threadPool);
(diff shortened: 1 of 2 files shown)


arteam:
At the first sight it looks like a positive and harmless change, but let's me think a little bit more about an impact of it and whether it could break anything.

jplock:
Any more thoughts on this @arteam?

arteam:
Sorry, I didn't have much time to dig into it, but I can't say anything​ against this change. I would merge it.

PR #1967, opened 2017-03-19
Make AllowedMethodsFilter immutable

arteam:
It looks like that the set of allowed methods in `AllowedMethodsFilter` is not modified after it was created. We could embrace it and make immutable during the creation time. This also should provide some performance benefits, because immutable hash tables could be optimized during construction times (e.g. they can use linear probing instead chaining)


jplock:
Looks good to me.  Any idea why appveyor has been failing recently?

nickbabcock:
The reason is that there is already a `test` table in some of the tests and so the `CREATE TABLE test` will fail. Whether it is because h2 isn't shutting down cleaning or tests are being executed concurrently, I'm not sure. The temporary fix would be to transform all `CREATE TABLE test` to `CREATE TABLE test IF NOT EXISTS`

arteam:
I guess the pull requests with new Liquebase tests could be a culprit. It creates shared H2 databases to keep the state between different migrations.

arteam:
I think I fixed the issue with the JDBI errors. I added shutdown code in the JDBI tests to make sure the tests don't reuse the same database.

PR #1972, opened 2017-03-21
docs: Fix `IllegalArgumentException` misspelling

evnm:
Just noticed this misspelling in the user manual.


joschi:
@evnm Thanks!

PR #1973, opened 2017-03-23
Set the console logging context after a reset

arteam:
When resetting logging, we should set a context for the console appender. Otherwise, we get an annoying message: `LOGBACK: No context given for ch.qos.logback.core.ConsoleAppender[null]` in the tests which use `DropwizardAppRule`.

This is related to #1947.


shawnsmith:
FYI: I'm getting a new Logback status warning from any attempt to use this `ConsoleAppender` because its `start()` method is never called.  Hooking up the context allows the context to capture the following warning:
```
WARN in ch.qos.logback.core.ConsoleAppender[null] - Attempted to append to non started appender [null].
```

PR #1976, opened 2017-03-23
Remove javax.el from jersey-bean-validation

vrolijken:
Fixes #1974

PR #1977, opened 2017-03-23
Exclude duplicated JTA 1.1 from dropwizard-hibernate dependencies

arteam:
`jackson-datatype-hibernate5` pulls a `javax.transaction.jta` artifact and Jadira pulls `geronimo-jta_1.1_spec`. They are outdated, because Hibernate already pulls JTA1.2.

This exclude affords to avoid warnings about duplicated classes from the `maven-shade-plugin` during assembling a fat jar.

### DIFF
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -114,6 +114,10 @@
                         <groupId>org.joda</groupId>
                         <artifactId>joda-money</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>org.apache.geronimo.specs</groupId>
+                        <artifactId>geronimo-jta_1.1_spec</artifactId>
+                    </exclusion>
                 </exclusions>
             </dependency>
             <dependency>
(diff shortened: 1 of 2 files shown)


arteam:
I think this PR should be applied to the 1.1 branch along with #1976

dropwizard-testing/pom.xml
```diff
@@ -63,6 +63,12 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>javax.transaction</groupId>
+            <artifactId>jta</artifactId>
+            <version>1.1</version>
```
joschi: Shouldn't we use `javax.transaction:javax.transaction-api:1.2` here?

arteam: Oh, right! I forgot about the differences between JTA 1.1 and JTA 1.2. I guess we need to remove JTA 1.1 from `jackson-datatype-hibernate5` then.

arteam:
Ping, I guess this should be good to merge.

PR #1981, opened 2017-03-24
Add setting system/default timezone in logging configuration

### LINKED ISSUE #1980: Allow system/default timezone in logging configuration
It is currently required to set the Timezone for a logging appender. Either it defaults to UTC or an explicit value has to be set. It is not possible to set it to the system/vmdefault timezone. 

A suggested edit is to change the AbstractAppenderFactory where the the setTimeZone allows a String instead of a TimeZone object. This way a special value could be introduced (i.e. 'default' or null) , another bonus is that unknown ZoneIds can be logged. A potential disadvantage is that it may affect existing configuration which uses the special value, which now defaults to 'GMT' as per Javadoc. 

Because code says more than a thousand words, an example test-case:

```java
import java.io.IOException;
import java.util.TimeZone;

import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.core.JsonParseException;
import com.fasterxml.jackson.databind.JsonMappingException;
import com.google.common.collect.Lists;

import org.junit.Assert;
import org.junit.Test;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class TimezoneTest {

    @Test
    public void testTimezoneDefault() throws JsonParseException, JsonMappingException, IOException {
        ObjectMapper mapper = new ObjectMapper();
        TestTimezone readValue = mapper.readValue("{\"zone\":\"default\"}", TestTimezone.class);
        Assert.assertEquals(TimeZone.getDefault(), readValue.getZone());
    }

    @Test
    public void testTimezoneUTC() throws JsonParseException, JsonMappingException, IOException {
        ObjectMapper mapper = new ObjectMapper();
        TestTimezone readValue = mapper.readValue("{\"zone\":\"UTC\"}", TestTimezone.class);
        Assert.assertEquals(TimeZone.getTimeZone("UTC"), readValue.getZone());
    }

    public static class TestTimezone {
        private TimeZone zone;
        private static final Logger LOG = LoggerFactory.getLogger(TimezoneTest.TestTimezone.class);

        public TestTimezone() {

        }

        @JsonProperty
        public void setZone(String zone) {
            if (zone.equals("default")) {
                this.zone = TimeZone.getDefault();
            } else {
                if (!Lists.newArrayList(TimeZone.getAvailableIDs()).contains(zone)) {
                    LOG.info("Unknown timezone {} provided, using GMT as per Java documentation", zone);
                }
                this.zone = TimeZone.getTimeZone(zone);
            }
        }

        @JsonProperty
        public TimeZone getZone() {
            return zone;
        }
    }
}
```

arteam:
Users should have the ability to use the system timezone in the logging configuration. It allows administrators to take advantage of the operation system/JVM time zone management tools and use the same configuration for Dropwizard applications in different time zones.

Resolves #1980.

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
@@ -21,6 +21,8 @@ import javax.validation.constraints.NotNull;
 import java.util.List;
 import java.util.TimeZone;
 
+import static com.google.common.base.Strings.nullToEmpty;
+
 /**
  * A base implementation of {@link AppenderFactory}.
  * <p/>
@@ -100,7 +102,7 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
     private boolean includeCallerData = false;
 
     private ImmutableList<FilterFactory<E>> filterFactories = ImmutableList.of();
-    
+
     private boolean neverBlock = false;
 
     @JsonProperty
@@ -149,6 +151,12 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
     }
 
     @JsonProperty
+    public void setTimeZone(String zoneId) {
+        this.timeZone = nullToEmpty(zoneId).equalsIgnoreCase("system") ? TimeZone.getDefault() :
+            TimeZone.getTimeZone(zoneId);
+    }
+
+    @JsonProperty
     public void setTimeZone(TimeZone timeZone) {
         this.timeZone = timeZone;
     }
@@ -172,7 +180,7 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
     public void setFilterFactories(List<FilterFactory<E>> appenders) {
         this.filterFactories = ImmutableList.copyOf(appenders);
     }
-    
+
     @JsonProperty
     public void setNeverBlock(boolean neverBlock) {
         this.neverBlock = neverBlock;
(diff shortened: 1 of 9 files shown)


arteam:
@Sijmen, Could you review?

dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
```diff
@@ -149,6 +151,19 @@ public TimeZone getTimeZone() {
...
 
     @JsonProperty
+    public void setTimeZone(String zoneId) {
+        if (zoneId.equalsIgnoreCase("system")) {
+            this.timeZone = TimeZone.getDefault();
+        } else {
+            if (Arrays.asList(TimeZone.getAvailableIDs()).contains(zoneId)) {
+                this.timeZone = TimeZone.getTimeZone(zoneId);
+            } else {
+                LoggerFactory.getLogger(getClass()).warn("Unknown timezone '{}' provided, using UTC", zoneId);
```
arteam: Right, it breaks with a time zone in the custom format (like `GMT-02:00`). Let's then remove the check for available time zones and just use `TimeZone.getTimeZone` as Jackson if the zone id is not `system`.

dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
```diff
@@ -149,6 +149,12 @@ public TimeZone getTimeZone() {
     }
 
     @JsonProperty
+    public void setTimeZone(String zoneId) {
+        this.timeZone = zoneId.equalsIgnoreCase("system") ? TimeZone.getDefault() :
```
jplock: In case zoneId is null, maybe put "system".equalsIgnoreCase(zoneId)

arteam: The method should be invoked in that case, but that a good idea in the spirit of defensive programming. I'm not a big fan of  reversed conditions (they don't look natural for me), so I will just use `Strings.nullOrEmpty`.

PR #1986, opened 2017-03-28
Fixed typo in javax.el exclusion

isaki-x:
The 1.1.0 release introduced a maven dependency conflict which has already been corrected in the 1.1.x branch. However, the fix itself has a typo. This fixes that typo. This will need to be merged into master as well. I have worked around this by forcing the javax.el and javax.el-api versions to 3.0.0 in my own pom for the time being.


isaki-x:
Actually, the web is correct for the legacy version. Just ignore me.

Issue #1985, opened 2017-03-27
Enum `fromStringFuzzy` should trim() and replace inner whitespace with '_'


stefan-zh:
I think that the `Enums` class in `io.dropwizard.util` should `trim()` whitespace around the String and then convert inner whitespace to `_`. This way you can use space characters as delimiters too.
```java
final String text = value.trim()
            .replaceAll("\\s+", "_") // <- space can be used as a delimiter too
            .replace('-', '_')
            .replace('.', '_');
```

Somewhere before version `1.0` or `1.1` of Dropwizard it was possible to use the `fromString()` method on Enums to convert Strings to the right Enum, but not they all go through `Enums.fromStringFuzzy()`, which eliminates any custom conversion logic.

That would be a different issue, but wouldn't it be helpful to check whether the custom `Enum` class actually implements a `fromString()` and only in case it doesn't, fall back to `final Enum<?> constant = Enums.fromStringFuzzy(value, constants);` I guess this logic would go somewhere in `FuzzyEnumParamConverterProvider`

stefan-zh:
Proposing this solution: https://github.com/dropwizard/dropwizard/pull/1987

BorisNaguet:
Hello,

We have the same issue here, as we have an enum with a custom fromString() method. It's useful when the strings contains chars that are not allowed in java enum name for example.

It used to be managed by 
`org.glassfish.jersey.server.internal.inject.ParamConverters$TypeFromStringEnum`
that is included in 
`org.glassfish.jersey.server.internal.inject.ParamConverters$AggregatedProvider`

AggregatedProvider is still present but is shadowed by FuzzyEnumParamConverterProvider because `@Custom` providers have a higher priority.

BorisNaguet:
I also noticed that all ConverterProviders are `@Singleton` except `FuzzyEnumParamConverterProvider`.

I don't think there's a reason for that?

arteam:
I think it's a good idea. I can imagine a situation when we want to map enums to values with spaces in the middle (e.g. country names).

arteam:
Closed via #2161

PR #1987, opened 2017-03-28
Adding support for fromString() for Enums in the FuzzyEnumParamConverter

stefan-zh:
Issue: https://github.com/dropwizard/dropwizard/issues/1985

Summary: Somewhere before versions `1.0` or `1.1` `@PathParam` objects that were `Enum` were able to be mapped by Jackson to their types. That is also described in the `@PathParam` documentation: `Have a static method named valueOf or fromString that accepts a single String argument`. However, that rule is being overriden by the logic in `FuzzyEnumParamConverterProvider` part of `dropwizard-jersey` which does not look for `fromString` method but directly tries to parse an `Enum` by performing some operations such as removing whitespace, mapping `.` or `-` characters to `_`

I think that this change (or a similar change) would allow for the original behavior to be restored.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
@@ -5,6 +5,10 @@ import com.google.common.base.Strings;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import io.dropwizard.util.Enums;
 
+import java.lang.reflect.Method;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.stream.Stream;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -39,6 +43,10 @@ public class FuzzyEnumParamConverterProvider implements ParamConverterProvider {
         final Class<Enum<?>> type = (Class<Enum<?>>) rawType;
         final Enum<?>[] constants = type.getEnumConstants();
         final String parameterName = getParameterNameFromAnnotations(annotations).orElse("Parameter");
+        final Method fromStringM = Stream.of(type.getMethods())
+            .filter(m -> m.getName().equals("fromString") && Arrays.equals(m.getParameterTypes(), new Class[]{String.class}))
+            .findFirst()
+            .orElse(null);
 
         return new ParamConverter<T>() {
             @Override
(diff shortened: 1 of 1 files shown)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
```diff
@@ -39,6 +42,10 @@
         final Class<Enum<?>> type = (Class<Enum<?>>) rawType;
         final Enum<?>[] constants = type.getEnumConstants();
         final String parameterName = getParameterNameFromAnnotations(annotations).orElse("Parameter");
+        final Method fromStringM = Stream.of(type.getMethods())
+            .filter(m -> m.getName().equals("fromString"))
+            .findFirst()
+            .orElse(null);
```
stefan-zh: I was looking for a way to get this safely `type.getMethod("fromString", String.class)` without throwing a `NoSuchMethodException`

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
```diff
@@ -39,6 +43,10 @@
         final Class<Enum<?>> type = (Class<Enum<?>>) rawType;
         final Enum<?>[] constants = type.getEnumConstants();
         final String parameterName = getParameterNameFromAnnotations(annotations).orElse("Parameter");
+        final Method fromStringM = Stream.of(type.getMethods())
+            .filter(m -> m.getName().equals("fromString") && Arrays.equals(m.getParameterTypes(), new Class[]{String.class}))
+            .findFirst()
+            .orElse(null);
```
stefan-zh: I was looking for an `Exception`-free equivalent of `type.getMethod("fromString", String.class)`

BorisNaguet: We should also test if method is static
`Modifier.isStatic(method.getModifiers())`
and the return type is the enum Type
`method.getReturnType() == type`

You can also directly use 
`Method fromStringMethod = AccessController.doPrivileged(ReflectionHelper.getFromStringStringMethodPA(rawType));`
that also manage privileges

arteam: I like the last proposed solution with delegating the lookup.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
```diff
@@ -47,7 +55,15 @@ public T fromString(String value) {
                     return null;
                 }
 
-                final Enum<?> constant = Enums.fromStringFuzzy(value, constants);
+                Object constant;
+                try {
```
arteam: I think it would be better to make a bill check instead of catching a NPE.

arteam: bill -> null

stefan-zh:
If https://github.com/dropwizard/dropwizard/pull/2161 goes in, I will close this PR

arteam:
Closed via #2161. Thanks for submitting the original version of the bugfix!

Issue #1988, opened 2017-03-29
Consider making DropwizardSlf4jRequestLog.java public?


qinfchen:
We are doing some customizations on the top of the LogbackClassicRequestLogFactory class. Is it possible to make DropwizardSlf4jRequestLog public instead of package private?

joschi:
@qinfchen Out of curiosity, what kind of customizations are necessary for your code?

qinfchen:
We are extending LogbackClassicRequestLogFactory to have different types of request logs based on sensitivity level.

qinfchen:
ping?

PR #1989, opened 2017-03-29
making DropwizardSl4jRequestLogger public instead of package private

qinfchen:
fixes https://github.com/dropwizard/dropwizard/issues/1988


arteam:
I think that's a reasonable improvement. Thank you!

Issue #1990, opened 2017-03-30
Application start fails due to "java.lang.ClassCastException: io.dropwizard.util.Size cannot be cast to java.util.Optional"


maksymgendin:
Hi!

Has anyone seen this error already? I am using Dropwizard 1.1.0.
It happens while validating some fields (e.g. `inputBufferSize` field) in `io.dropwizard.jetty.HttpConnectorFactory` even if I do not specify a `server` block in my configuration YAML, so the default values are used.

```
java.lang.ClassCastException: io.dropwizard.util.Size cannot be cast to java.util.Optional
	at org.hibernate.validator.internal.engine.valuehandling.OptionalValueUnwrapper.handleValidatedValue(OptionalValueUnwrapper.java:20)
	at org.hibernate.validator.internal.engine.ValueContext.getCurrentValidatedValue(ValueContext.java:121)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:110)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:87)
	at org.hibernate.validator.internal.metadata.core.MetaConstraint.validateConstraint(MetaConstraint.java:73)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateMetaConstraint(ValidatorImpl.java:620)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraint(ValidatorImpl.java:583)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForSingleDefaultGroupElement(ValidatorImpl.java:527)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForDefaultGroup(ValidatorImpl.java:495)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForCurrentGroup(ValidatorImpl.java:460)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateInContext(ValidatorImpl.java:410)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateCascadedConstraint(ValidatorImpl.java:761)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateCascadedConstraints(ValidatorImpl.java:684)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateInContext(ValidatorImpl.java:419)
[...]

PR #1993, opened 2017-03-31
Add missing @UnwrapValidatedValue annotations

joschi:
Some classes/fields were missing the `@UnwrapValidatedValue` annotations in combination with javax.validation annotations.


arteam:
Is this this bug causes #1990?

evnm:
Change LGTM.

@arteam I wouldn't think so. The fields claimed to trigger the issue in #1990 aren't optional, so the `@UnwrapValidatedValue` shouldn't be relevant to them.

joschi:
> Does this this bug cause #1990?

No, but I discovered the missing annotations while looking into that issue.

PR #2001, opened 2017-04-10
Add support for specifying specific client certificate

### LINKED ISSUE #2000: Ability to specify specific client certificate alias in dropwizard-client TLS configuration
Sometimes it's convenient to use a key store that has more than one private key and certificate in it when authenticating to a server with a certificate.

The dropwizard-client does not currently provide a way to define a `PrivateKeyStrategy` when setting up the`SSLContext` 

A simple, configuration based approach is simply to add a `certAlias` parameter to the TLS configuration of the HTTP client and, when set, define a PrivateKeyStrategy that unconditionally returns this configured alias.

I have a pull request to go along with this which I'll submit for 1.2.x.

Thanks,

Tim

tbartley:
Adds certAlias configuration parameter to TLS configuration of HTTP
clients. This allows a specific certificate in a keystore to be
specified as the one to use when authenticating to servers.

Fixes #2000


dropwizard-client/src/main/java/io/dropwizard/client/ssl/TlsConfiguration.java
```diff
@@ -150,7 +156,18 @@ public void setSupportedProtocols(@Nullable List<String> supportedProtocols) {
         this.supportedProtocols = supportedProtocols;
     }
 
-    @ValidationMethod(message = "keyStorePassword should not be null or empty if keyStorePath not null")
+    @Nullable
+    @JsonProperty
+    public String getCertAlias() {
+		return certAlias;
+	}
```
jplock: spacing seems off here

dropwizard-client/src/main/java/io/dropwizard/client/ssl/TlsConfiguration.java
```diff
@@ -38,6 +40,10 @@
     @Nullable
     private List<String> supportedCiphers = null;
 
+    @Nullable
+    @Size(min=1)
```
jplock: what is the `@Size` doing here?

tbartley: Ensuring that if a value for certAlias is provided it's min length is 1. Maybe this is unnecessary.

dropwizard-client/src/main/java/io/dropwizard/client/ssl/TlsConfiguration.java
tbartley: Yeah - the codeclimate checker pulled me up on that too. Am working through everything it reported.

dropwizard-client/src/main/java/io/dropwizard/client/ssl/TlsConfiguration.java
jplock: How about making this:

```
@UnwrapValidatedValue
public Optional<String> certAlias = Optional.empty();
```

tbartley:
Only concern is that doesn’t match the style of other optional parameters for that class e.g. trustStorePath, keyStorePath, ….

Cheers,

Tim

On Apr 10, 2017, at 2:44 PM, Justin Plock <notifications@github.com<mailto:notifications@github.com>> wrote:


@jplock requested changes on this pull request.

Let's make private String certAlias be optional

jplock:
@tbartley gotcha, maybe just remove the @Size annotation then.  I think that's throwing me off as we usually don't use that to check for emptiness or length of a string.

tbartley:
Thanks @jplock - I've removed the @Size constraint.

jplock:
Thanks!

PR #2005, opened 2017-04-11
Improve unhandled exception logging in Cli.run

### LINKED ISSUE #2003: Unhandled exceptions from Application run methods are not well logged.
The only place that logs an unhandled exception from an Application's run method is in Cli.run where it just does:

```java
// Unexpected exceptions should result in non-zero exit status of the process
stdErr.println(t.getMessage());
return false;
 ```

This is fine, but an unhandled NPE for instance ends up just logging the String `null`. Since NPE's don't have a message.

This is not particularly helpful in diagnosing the situation.

I think this should log the stack trace as well. This affected me when debugging something recently.

Will submit a PR.

Cheers,

Tim

tbartley:
Only log the exception message if there is one and
always print the exceptiopn stack trace when Cli.run
receives an unhandled exception when invoking Command.run.

Fixes #2003

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/cli/Cli.java
@@ -84,7 +84,7 @@ public class Cli {
             return false;
         } catch (Throwable t) {
             // Unexpected exceptions should result in non-zero exit status of the process
-            stdErr.println(t.getMessage());
+            t.printStackTrace(stdErr);
             return false;
         }
     }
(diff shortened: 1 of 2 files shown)


evnm:
Thanks for reporting and tackling this, @tbartley!

The stack trace [should include a `Throwable`'s message](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#printStackTrace--), if present, so I think we can remove the `stdErr.println(t.getMessage());` entirely.

tbartley:
Thanks @evnm - true. I've tidied that up.

evnm:
Thanks, @tbartley!

PR #2014, opened 2017-04-18
Add flag to make ThreadNameFilter optional

### LINKED ISSUE #2013: Make ThreadNameFilter optional
I'd like to sometimes disable ThreadNameFilter because I don't want to share actual request information outside of request logs, and I would like to, generally, include thread name in my service logs.

I'm happy to make a PR to add an option to disable the ThreadNameFilter in AbstractServerFactory, if this seems acceptable.

markelliot:
Include a flag enableThreadNameFilter in AbstractServerFactory to allow
disabling of the ThreadNameFilter's installation. Existing default behavior
is preserved.

Fixes #2013.


arteam:

PR #2018, opened 2017-04-21
Add allowCoreThreadTimeOut option to ExecutorServiceBuilder.

natnan:
[allowCoreThreadTimeOut](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html#allowCoreThreadTimeOut(boolean))

Today it's only possible to do this by casting:
```
ThreadPoolExecutor executor = (ThreadPoolExecutor) environment.lifecycle().executorService("test").....build();
executor .allowCoreThreadTimeOut(true);
```

### DIFF
--- a/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
+++ b/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
@@ -21,6 +21,7 @@ public class ExecutorServiceBuilder {
     private final String nameFormat;
     private int corePoolSize;
     private int maximumPoolSize;
+    private boolean allowCoreThreadTimeOut;
     private Duration keepAliveTime;
     private Duration shutdownTime;
     private BlockingQueue<Runnable> workQueue;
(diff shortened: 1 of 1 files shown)


nickbabcock:
I think this is a fine change. For my curiosity did you have a use case where `allowCoreThreadTimeOut` mattered and the `keepAliveTime` wasn't sufficient?

natnan:
http://stackoverflow.com/questions/1800317/impossible-to-make-a-cached-thread-pool-with-a-size-limit
This sums it up I believe.

Dropwizard's default ExecutorService comes with `new LinkedBlockingQueue<>();` which is an infinite queue. That means, if you set up a pool with `minThread:1` (and set maxThread to int-max if you want), you'll never get more than one thread. Which we realized this once on production some time ago :) and we changed the queue to `SynchronousQueue` with `maxThread(Integer.MAX)`. For that use case, that was enough.

But recently, we required better fine-tuning with the ability of starting many threads (say 30) right away when needed, and increase it to a certain point (say 100 threads) on high load. You can accomplish that with minThread, limited queue, and maxThread.
However, without `allowCoreThreadTimeOut`, that means 30 threads will always be alive. And that's not what we want. 

`<rant>`
It's actually quite surprising to me how limited the ExecutorService options are in Java. You can't do much smart tuning. For example, I would have liked another queue after `maxThread` (which was my assumption in the first place). `RejectedExecutionHandler`s are hardly sufficient.
Or have `minThread` where the threads are always alive (`allowCoreThreadTimeOut: false` today), and then threads that are produced as soon as `minThread` is not enough. 
`</rant>`

nickbabcock:
Thank you for such an explanation. Looks good to me.

Issue #1998, opened 2017-04-09
Update to Hibernate 5.2.10


yateam:
Hello, I just switched to DW 1.1.0 (from 1.0.6) and immediately found a problem with the new Hibernate (5.2.8).
It is described pretty good here: https://hibernate.atlassian.net/browse/HHH-11496

It seems, I am experiencing the same thing.
I have two entities: Club and Tenant
Club:
```
@Entity
public class Club {
  @Id
  @GeneratedValue
  private Integer id;

  @Column(nullable = false)
  private String name;

  @OneToOne
  @JoinColumn(name = "tenant_id")
  private Tenant tenant;
...
}
```
and Tenant:
```
@Entity
public class Tenant {
  @Id
  @GeneratedValue
  private Integer id;

  @OneToOne(mappedBy = "tenant", cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.LAZY)
  private Club club;
...
}
```

Well, basically, constructing these entities I followed the official Hibernate User Guide.
Everything worked fine in DW 1.0.6 (and HIbernate 5.1.0 respectively) but after switching I noticed exceptions reporting about ConstraintViolations. I checked the logs, debugged and saw that Hibernate tries to insert a `Club` first, however, it should  insert it only after `Tenant`.

Writing a test for this would be pretty time consuming task. If you really need this - I might try.

Have you experiences anything like this?

yateam:
Seems like it had been fixed here: https://github.com/hibernate/hibernate-orm/commit/f90845c30c2a6d5e14eeafd32a4c9d321d3a55ef

And the corresponding Hibernate issue is this one: https://hibernate.atlassian.net/browse/HHH-11585

According to this issue the fix got only into 5.2.10. Sad.

arteam:
Thanks for the investigation! I guess we will wait until Hibernate 5.2.10 released and then update its dependency in Dropwizard.

yateam:
Maybe rename the issue to Update to Hibernate 5.2.10 then?

arteam:
Done

arteam:
Should be resolved as part of #2021.

PR #2021, opened 2017-04-22
Dependency updates

joschi:


nickbabcock:
Considering Hibernate 5.2.10, SLF4J 1.7.25, and Logback 1.2.3 are bug fixes that dropwizard users are running into, are they candidates for a 1.1.x?

arteam:
I think it's a good idea to make a 1.1 release. We have #1998, which affects Hibernate users.

Issue #2024, opened 2017-04-24
Use latest version of alpn-boot by default


t-tsutsumi:
Currently, dropwizard-http2 using the alpn-boot 8.1.3.v20150130 by default. Therefore, building with a future Java version that is not defined in the profile always fails.

Since it takes a while for alpn-boot to be released after the latest version of Java is released, I suggest that dropwizard-http2 use the latest alpn-boot by default.

Note: Probably, Travis CI currently uses Java 8u131, because Dropwizard uses oracle-java 8-installer.

arteam:
Good idea. I think we could upgrade alpn-boot to 8.1.11.v20170118, because it should be compatible with the latest JDKs. Would you interested in submitting a pull request?

t-tsutsumi:
Yes. I will submit a pull request.

FYR: If alpn-boot version is old, Http2IntegrationTest ends with thread death. When testing with Java 8u131, the following log is recorded.
```
INFO  [2017-04-24 17:50:59,036] org.eclipse.jetty.server.Server: Started @4081ms
WARN  [2017-04-24 17:50:59,148] org.eclipse.jetty.util.thread.QueuedThreadPool: 
! java.lang.NoSuchMethodError: sun.security.ssl.SignatureAndHashAlgorithm.getSupportedAlgorithms(Ljava/util/Collection;)Ljava/util/Collection;
! at sun.security.ssl.ServerHandshaker.clientHello(ServerHandshaker.java:698)
! at sun.security.ssl.ServerHandshaker.processMessage(ServerHandshaker.java:217)
! at sun.security.ssl.Handshaker.processLoop(Handshaker.java:1026)
! at sun.security.ssl.Handshaker$1.run(Handshaker.java:966)
! at sun.security.ssl.Handshaker$1.run(Handshaker.java:963)
! at java.security.AccessController.doPrivileged(Native Method)
! at sun.security.ssl.Handshaker$DelegatedTask.run(Handshaker.java:1416)
! at org.eclipse.jetty.io.ssl.SslConnection$DecryptedEndPoint.fill(SslConnection.java:736)
! at org.eclipse.jetty.server.NegotiatingServerConnection.fill(NegotiatingServerConnection.java:153)
! at org.eclipse.jetty.server.NegotiatingServerConnection.onFillable(NegotiatingServerConnection.java:98)
! at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:279)
[...]

PR #2025, opened 2017-04-24
Use latest version of alpn-boot by default

t-tsutsumi:
Solves: #2024


arteam:

PR #2026, opened 2017-04-25
[ENHANCEMENT] SubstringMatchingStrategy for metrics

jelmerterwal:
The reason I came across this `problem` (maybe it is meant to be used like the way it is now) is because I use datadog and use a thin library on top of dropwizard (https://raw.githubusercontent.com/coursera/metrics-datadog). Which gave the following description how includes work;

---

#### Filtering

If you want to filter only a few metrics, you can use the `includes` or 
`excludes` key to create a set of metrics to include or exclude respectively.

~~~yaml
metrics:
  frequency: 1 minute                       # Default is 1 second.
  reporters:
    - type: datadog
      host: <host>
      includes:
        - jvm.
        - ch.
~~~

The check is very simplistic so be as specific as possible. For example, if 
you have "jvm.", the filter will check if the includes has that value in any 
part of the metric name (not just the beginning).

---
After trying a lot of different configurations I checked the code to see if this was really the case. For now I used a regular expression to fix my problem, but I would like to improve the code to adhere to the description above.

Dropwizard's own documentation is very brief about the usage (should it be the complete string or not?);
```
includes	(all)	Metrics to include in reports, by name. When defined, only these metrics will be reported.
```

### DIFF
--- a/dropwizard-metrics/src/test/java/io/dropwizard/metrics/BaseReporterFactoryTest.java
+++ b/dropwizard-metrics/src/test/java/io/dropwizard/metrics/BaseReporterFactoryTest.java
@@ -17,10 +17,10 @@ import static org.mockito.Mockito.mock;
 @RunWith(Parameterized.class)
 public class BaseReporterFactoryTest {
     private static final ImmutableSet<String> INCLUDES = ImmutableSet.of("inc", "both", "inc.+");
-    private static final ImmutableSet<String> EXCLUDES = ImmutableSet.of("both", "exc", "exc.+");
+    private static final ImmutableSet<String> EXCLUDES = ImmutableSet.of("exc", "both", "exc.+");
     private static final ImmutableSet<String> EMPTY = ImmutableSet.of();
 
-    @Parameterized.Parameters(name = "{index} {4} {2}={3}")
+    @Parameterized.Parameters(name = "{index} {6} {2}")
     public static List<Object[]> data() {
 
         return ImmutableList.of(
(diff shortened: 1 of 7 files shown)


dropwizard-metrics/src/test/java/io/dropwizard/metrics/BaseReporterFactoryTest.java
```diff
@@ -37,19 +34,21 @@
...
+                /*
                  * case2: If include list is NOT empty and exclude list is empty, only the ones
                  * specified in the include list should be included.
                  */
                 new Object[]{INCLUDES, EMPTY, "inc", true, true, "case2"},
                 new Object[]{INCLUDES, EMPTY, "both", true, true, "case2"},
                 new Object[]{INCLUDES, EMPTY, "exc", false, false, "case2"},
                 new Object[]{INCLUDES, EMPTY, "any", false, false, "case2"},
-                new Object[]{INCLUDES, EMPTY, "incWithSuffix", false, true, "case2"},
+                new Object[]{INCLUDES, EMPTY, "incWithSuffix", true, true, "case2"},
```
jelmerterwal: 'inc' is part of 'incWithSuffix'

dropwizard-metrics/src/test/java/io/dropwizard/metrics/BaseReporterFactoryTest.java
```diff
@@ -58,18 +57,20 @@
                 new Object[]{EMPTY, EXCLUDES, "exc", false, false, "case3"},
                 new Object[]{EMPTY, EXCLUDES, "any", true, true, "case3"},
                 new Object[]{EMPTY, EXCLUDES, "incWithSuffix", true, true, "case3"},
-                new Object[]{EMPTY, EXCLUDES, "excWithSuffix", true, false, "case3"},
+                new Object[]{EMPTY, EXCLUDES, "excWithSuffix", false, false, "case3"},
```
jelmerterwal: 'exc' (one of the excludes) is part of 'excWithSuffix'

dropwizard-metrics/src/test/java/io/dropwizard/metrics/BaseReporterFactoryTest.java
```diff
@@ -58,18 +57,20 @@
...
                  * case4: If include list is NOT empty and exclude list is NOT empty, only things not excluded
                  * and specifically included should show up. Excludes takes precedence.
                  */
                 new Object[]{INCLUDES, EXCLUDES, "inc", true, true, "case4"},
                 new Object[]{INCLUDES, EXCLUDES, "both", false, false, "case4"},
                 new Object[]{INCLUDES, EXCLUDES, "exc", false, false, "case4"},
                 new Object[]{INCLUDES, EXCLUDES, "any", false, false, "case4"},
-                new Object[]{INCLUDES, EXCLUDES, "incWithSuffix", false, true, "case4"},
-                new Object[]{INCLUDES, EXCLUDES, "excWithSuffix", false, false, "case4"}
+                new Object[]{INCLUDES, EXCLUDES, "incWithSuffix", true, true, "case4"},
```
jelmerterwal: none of the excludes match 'incWithSuffix'

arteam:
I believe it's originally were meant to be a complete metric name and then regular expressions were added on top. I think it's an interesting idea to improve the default matching strategy, but the biggest concern to me that such a change would be backward incompatible. Users may find new metrics pop up in dashboards after an update, because the new strategy would more lenient. Maybe we could improve the documentation to make more clear how matching works.

jelmerterwal:
I agree about the backward compatibility, maybe an additional matching strategy which will behave this way (which must be explicitly set, like the way it is for regex now)? Or, change this behaviour with a major version? The way includes and excludes work combined with the default matching strategy makes no sense in my opinion.

arteam:
I think a new strategy will be the best option. Doing so we reduce the risk to broke existing applications and can provide this matching in the 1.2 version. We could add a new option `useSubstringMatching` or something like that (if you have ideas for a better name, go ahead). It could be a middle ground between full string matching and regular expressions.

jelmerterwal:
I think the name is in line with `useRegexFilters`. If you have any further feedback let me know.

arteam:
Thank you very much for the contribution!

PR #2028, opened 2017-04-26
Add new criteriaQuery method to avoid hibernate criteria deprecations.

lightswitch05:
Hibernate deprecated `session.createCriteria(<E>)`.

I added a `criteriaQuery()` method as a migration path away from `criteria()`. I considered `createCriteriaUpdate` and `createCriteriaDelete` methods as well, but wasn't sure if they would be wanted. Please let me know your thoughts.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java
@@ -12,7 +12,6 @@ import org.hibernate.query.internal.AbstractProducedQuery;
 import java.io.Serializable;
 import java.util.List;
 
-import javax.persistence.criteria.CriteriaBuilder;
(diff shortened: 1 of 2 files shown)


dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java
```diff
@@ -55,6 +56,16 @@ protected Criteria criteria() {
     }
 
     /**
+     * Creates a new {@link CriteriaQuery} for {@code <E>}.
+     *
+     * @return a new {@link CriteriaQuery} query
+     */
+    protected CriteriaQuery<E> criteriaQuery() {
+        CriteriaBuilder builder = this.currentSession().getCriteriaBuilder();
```
arteam: I think the builder variable could be omitted because it's used only once.

arteam:
Thanks for the contribution!

Issue #2031, opened 2017-04-27
sessionDestroyed is not being called in latest DW version


orangecoding:
We've upgraded the dw version and noticed that the sessionDestroyed in our internal event handler is not being called anymore.

I've reported this to jetty in the first place, they've tracked it down and finally noticed you are implementing a wrong Interface.
Please see the issue I'Ve reported in the jetty repo here: https://github.com/eclipse/jetty.project/issues/1432#issuecomment-297665805

> So all dropwizard needs to do to make this work is to change RoutingHandler to respect the HandlerContainer interface.

I've also a testcase for you.
http://www.orange-coding.net/dw_test.zip

build it with `mvn clean install -DskipTests`, afterwards go to ./target start it with `java -jar Test-1.2.0-SNAPSHOT.jar server` then point your browser to localhost:8080.

In the System.out, you'll see a CREATED (Session has been created) but no DESTROY. (Should appear after 10 seconds)

janbartel:
If the dropwizard folks need any further help, the jetty team will be happy to answer any questions.

orangecoding:
👍 Thanks!

t-tsutsumi:
I also saw the same thing. Probably, the `SessionHandler` will work with [this](https://github.com/dropwizard/dropwizard/compare/master...t-tsutsumi:pr/implement-HandlerContainer). However, there is concern about maintaining backward compatibility. I think more tests will be required.

Also, If you run the test case of `@orangecoding`, I recommend adding the following code.
```diff
diff --git a/src/main/java/com/example/test/TestSessionHandler.java b/src/main/java/com/example/test/TestSessionHandler.java
index 75726cd..5d7924b 100644
--- a/src/main/java/com/example/test/TestSessionHandler.java
+++ b/src/main/java/com/example/test/TestSessionHandler.java
@@ -3,6 +3,10 @@ package com.example.test;
 import javax.servlet.http.HttpSessionEvent;
 import javax.servlet.http.HttpSessionListener;

+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.SessionIdManager;
+import org.eclipse.jetty.server.session.DefaultSessionIdManager;
+import org.eclipse.jetty.server.session.HouseKeeper;
 import org.eclipse.jetty.server.session.SessionHandler;

 public class TestSessionHandler extends SessionHandler{
@@ -24,4 +28,18 @@ public class TestSessionHandler extends SessionHandler{
     });

   }
+
[...]

PR #2032, opened 2017-04-27
Fix HttpSessionListener.sessionDestroyed is not being called #2031

t-tsutsumi:
Solves: #2031

Also, in order to make it possible to detect degradation caused by Jetty's version upgrade, I added several tests.


CodingFabian:
is it possible to merge this to 1.1.x? as its actually a 1.1.x regression due to jetty upgrade.

We would like to be able to use this feature again without the need to wait for 1.2

jplock:
I cherry-picked these commits in https://github.com/dropwizard/dropwizard/commit/bfc370b5fb16a8d7b93ea9cd2520c2921a13d95e and https://github.com/dropwizard/dropwizard/commit/5f85ffab2f09e6ff43f88f8763ecbef147e2754e into the `release/1.1.x` branch

PR #2035, opened 2017-04-28
More docs for project structure

### LINKED ISSUE #1979: Improve "Organizing Your Project" section of Dropwizard Core manual
In the `Organizing Your Project` section of the Dropwizard Core manual, I think the following could be improved:

> In general, we recommend you separate your projects into three Maven modules: project-api, project-client, and project-application.

> project-api should contain your Representations; project-client should use those classes and an HTTP client to implement a full-fledged client for your application, and project-application should provide the actual application implementation, including Resources.

It would be nice to provide a little bit more context on **why** splitting a Dropwizard project into those three modules is such a good idea. And if that's really the case, it would also be nice to update the [dropwizard-example](https://github.com/dropwizard/dropwizard/tree/master/dropwizard-example) to reflect this recommendation.

nickbabcock:
The organizing your code section has been confusing, as reported in #1979. This PR attempts to fix this 👼 

CC @fabriziocucci

Closes #1979

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -26,27 +26,43 @@ Dropwizard consists mostly of glue code to automatically connect and configure t
 Organizing Your Project
 =======================
 
-In general, we recommend you separate your projects into three Maven modules: ``project-api``,
-``project-client``, and ``project-application``.
+If you plan on developing a client library for other developers to access your service, we recommend
+you separate your projects into three Maven modules: ``project-api``, ``project-client``, and
+``project-application``.
 
 ``project-api`` should contain your :ref:`man-core-representations`; ``project-client`` should use
 those classes and an :ref:`HTTP client <man-client>` to implement a full-fledged client for your
 application, and ``project-application`` should provide the actual application implementation, including
 :ref:`man-core-resources`.
 
-Our applications tend to look like this:
+To give a concrete example of this project structure, let's say we wanted to create a Stripe_-like
+API where clients can issue charges and the server would echo the charge back to the client.
+``stripe-api`` project would hold our ``Charge`` object as both the server and client want to work
+with the charge and to promote code reuse, ``Charge`` objects are stored in a shared module.
+``stripe-app`` is the Dropwizard application. ``stripe-client`` abstracts away the raw HTTP
+interactions and deserialization logic. Instead of using a HTTP client, users of ``stripe-client``
+would just pass in a ``Charge`` object to a function and behind the scenes, ``stripe-client`` will
+call the HTTP endpoint. The client library may also take care of connection pooling, and may
+provide a more friendly way of interpreting error messages. Basically, distributing a client library
+for your app will help other developers integrate more quickly with the service.
+
+If you are not planning on distributing a client library for developers, one
+can combine ``project-api`` and ``project-application`` into a single project,
+which tends to look like this:
 
 * ``com.example.myapplication``:
 
-  * ``api``: :ref:`man-core-representations`.
+  * ``api``: :ref:`man-core-representations`. Request and response bodies.
   * ``cli``: :ref:`man-core-commands`
-  * ``client``: :ref:`Client <man-client>` implementation for your application
-  * ``core``: Domain implementation
+  * ``client``: :ref:`Client <man-client>` code that accesses external HTTP services.
+  * ``core``: Domain implementation; where objects not used in the API such as POJOs, validations, crypto, etc, reside.
   * ``jdbi``: :ref:`Database <man-jdbi>` access classes
   * ``health``: :ref:`man-core-healthchecks`
   * ``resources``: :ref:`man-core-resources`
   * ``MyApplication``: The :ref:`application <man-core-application>` class
   * ``MyApplicationConfiguration``: :ref:`configuration <man-core-configuration>` class
+
+.. _Stripe: https://stripe.com/docs/api/java
 
 .. _man-core-application:
 


docs/source/manual/core.rst
```diff
@@ -26,28 +26,43 @@ Dropwizard consists mostly of glue code to automatically connect and configure t
...
 
-Our applications tend to look like this:
+To give a concrete example of this project structure, let's say we wanted to create a Stripe_-like
+API where clients can issue charges and the server would echo the charge back to the client.
+``stripe-api`` project would hold our ``Charge`` object as both the server and client want to work
+with the charge and to promote code reuse, ``Charge`` objects are stored in a shared module.
+``stripe-app`` is the Dropwizard application. ``stripe-client`` abstracts away the raw HTTP
+interactions and deserialization logic.  Instead of using a HTTP client, users of ``stripe-client``
+would just pass in a ``Charge`` object to a function and behind the scenes, ``stripe-client`` will
+call the HTTP endpoint.  The client library may also take care of connection pooling, and may
```
fabriziocucci: Two white spaces at line 45 after the dot and before "The client library..."

docs/source/manual/core.rst
```diff
@@ -26,28 +26,43 @@ Dropwizard consists mostly of glue code to automatically connect and configure t
...
+provide a more friendly way of interpreting error messages. Basically, distributing a client library
+for your app will help other developers integrate more quickly with the service.
+
+If you are not planning on distributing a client library for developers, one can combine the
```
fabriziocucci: "...one can combine the ``project-api`` and ``project-application`` modules into a single project..."

Does it sound more accurate?

docs/source/manual/core.rst
```diff
@@ -26,28 +26,43 @@ Dropwizard consists mostly of glue code to automatically connect and configure t
...
 application, and ``project-application`` should provide the actual application implementation, including
 :ref:`man-core-resources`.
 
-Our applications tend to look like this:
+To give a concrete example of this project structure, let's say we wanted to create a Stripe_-like
+API where clients can issue charges and the server would echo the charge back to the client.
+``stripe-api`` project would hold our ``Charge`` object as both the server and client want to work
+with the charge and to promote code reuse, ``Charge`` objects are stored in a shared module.
+``stripe-app`` is the Dropwizard application. ``stripe-client`` abstracts away the raw HTTP
+interactions and deserialization logic.  Instead of using a HTTP client, users of ``stripe-client``
```
fabriziocucci: "...and deserialization logic.  Instead of using a HTTP client..."

Double white space after the dot, not sure if it's intentional.

docs/source/manual/core.rst
fabriziocucci: "...call the HTTP endpoint.  The client library ..."

Double white space after the dot, not sure if it's intentional.

docs/source/manual/core.rst
```diff
@@ -26,28 +26,43 @@ Dropwizard consists mostly of glue code to automatically connect and configure t
...
+If you are not planning on distributing a client library for developers, one can combine the
+projects into a single project, which tend to look like this:
 
 * ``com.example.myapplication``:
 
   * ``api``: :ref:`man-core-representations`.
   * ``cli``: :ref:`man-core-commands`
-  * ``client``: :ref:`Client <man-client>` implementation for your application
+  * ``client``: :ref:`Client <man-client>` code used to access other HTTP endpoints -- not your endpoints.
   * ``core``: Domain implementation
```
evnm: I think we should clarify what the difference is between the `api` and `core` packages.

nickbabcock:
Updated PR with the suggestions 💪

docs/source/manual/core.rst
```diff
@@ -26,28 +26,44 @@ Dropwizard consists mostly of glue code to automatically connect and configure t
...
+which tends to look like this:
 
 * ``com.example.myapplication``:
 
   * ``api``: :ref:`man-core-representations`.
   * ``cli``: :ref:`man-core-commands`
-  * ``client``: :ref:`Client <man-client>` implementation for your application
-  * ``core``: Domain implementation
+  * ``client``: :ref:`Client <man-client>` code used to access other HTTP endpoints -- not your endpoints.
+  * ``core``: Domain implementation. Whereas you should try and keep objects that are serialized and deserialized to and from clients in the ``api`` package, the ``core`` package contains objects not used in the API, validations, crypto, monitoring, exception handling, etc, and other business logic at the heart of the service.
```
evnm: Consider putting the "validations, crypto, ..." list in parentheses.

docs/source/manual/core.rst
```diff
@@ -26,28 +26,44 @@ Dropwizard consists mostly of glue code to automatically connect and configure t
...
+can combine ``project-api`` and ``project-application`` into a single project,
+which tends to look like this:
 
 * ``com.example.myapplication``:
 
   * ``api``: :ref:`man-core-representations`.
   * ``cli``: :ref:`man-core-commands`
-  * ``client``: :ref:`Client <man-client>` implementation for your application
-  * ``core``: Domain implementation
+  * ``client``: :ref:`Client <man-client>` code used to access other HTTP endpoints -- not your endpoints.
```
fabriziocucci: I'm going to throw this as it come up to my mind right now so feel free to ignore! :smile: 

Naming the module ``project-client`` (in the "client library" scenario) and the package ``client`` (in the "NO client library" scenario) _can_ still cause a little bit of misunderstanding (even though the new comments help a lot). 

Have you guys thought about renaming the ``client`` package into something like ``remote``, ``service`` or something like that?

To be more concrete, I'll write down an example based on the conventions I've adopted in one of my projects. I have two microservices, aggregation-service and recommendation-service. The aggregation-service depends on the recommendation-service in that it requires an invocation of the recommendation-service REST API. So, in the aggregation-service I have my ``remote`` package where I've created a JAX-RS annotated interface representing the recommendation-services and called ``RemoteRecommendationService`` (and ultimately instantiated using jersey-proxy-client). Again, this is just a (fairly opinionated) convention, but makes it clear what are the "remote" dependencies of a service and could potentially avoid the "client" naming clash...

Ultimately, that package (i.e. ``client``) should contain [_client_]s for [_external_]/[_remote_] [_service_]s (i.e. any word surrounded by square brackets is, kind of, representative of that package).

nickbabcock: I can get on board the `remote` train.

nickbabcock:
🎬 2

@evnm went with a slightly different approach and tried to make the description for clear and concise.

Let me know if this was a step in the right direction.

docs/source/manual/core.rst
```diff
@@ -26,28 +26,44 @@ Dropwizard consists mostly of glue code to automatically connect and configure t
...
+which tends to look like this:
 
 * ``com.example.myapplication``:
 
-  * ``api``: :ref:`man-core-representations`.
+  * ``api``: :ref:`man-core-representations`. Request and response bodies.
   * ``cli``: :ref:`man-core-commands`
-  * ``client``: :ref:`Client <man-client>` implementation for your application
-  * ``core``: Domain implementation
+  * ``remote``: :ref:`Client <man-client>` code that accesses external HTTP services.
```
evnm: Just wondering, primarily: Is this terminology that's widely used already or are we innovating here?

fabriziocucci: @evnm I can honestly tell you that I usually take Dropwizard, Spring Boot and JHipster as reference for widely used conventions but I don't remember seeing this anywhere.

Anyway, I see your point: introducing a brand new terminology may cause even more confusion than keeping the `project-client` module and the `client` package as they are.

I guess it's all a matter of deciding whether this explicit disambiguation is worth it or not, also bearing in mind that, ultimately, this is just a recommendation.

nickbabcock: Spring boot makes references to consuming web services with a REST client 🤔 

If we take the case where a dropwizard app will have a client library and use REST clients (using Stripe here to keep things straightforward), the project structure would look like

- com.stripe.stripe-client
  - bank-api-client
  - etc
- com.stripe.stripe-app
  - client
  - etc

I'd think that having `com.stripe.stripe-client.*` and `com.stripe.stripe-app.client` could cause confusion even though the terminology is correct.

I'm even open to switching to a structure like:

- com.stripe.stripe-library (stripe's client actual artifact name is stripe-java, so something to consider)
- com.stripe.stripe-app
  - client
- com.stripe.stripe-core 

To me, core sounds like a module used in other modules, library sounds like something that is included as part of a larger whole, and app is something that is deployed.

Why does naming things have to be so hard 😖

fabriziocucci: > I'd think that having `com.stripe.stripe-client.*` and `com.stripe.stripe-app.client` could cause confusion even though the terminology is correct.

My point too, this is why I ended up using `remote` as package in the application module.

But, as you mentioned, the terminology `client` for that package is indeed correct, at least from a technical prospective. It is also generic enough such that the package can (potentially) host _any_ client for services the application depends upon (i.e. not only REST over HTTP).

> To me, core sounds like a module used in other modules, library sounds like something that is included as part of a larger whole, and app is something that is deployed.

I feel, and I may be wrong, that `core` is kind of more suitable (or at least more common) in the context of frameworks rather than applications. While `library` sounds quite generic (maybe too much?).

Naming is a tough business indeed! :sweat:

nickbabcock: Since naming is hard, I've gone back to calling it `client` instead of `remote`. We can discuss what they should be named another time, and keep this PR focused on elaborating their descriptions.

jplock:
This ready to merge?

nickbabcock:
Yes

PR #2037, opened 2017-04-29
Add a WeldBundle to Dropwizard

jelmerterwal:
This feature makes it easy to get started with CDI using the Weld implementation.


joschi:
@jelmerterwal Thanks for your contribution!

We try to keep the core Dropwizard project as small and lean as possible, so please create a separate project for your Weld bundle and add it to the [Dropwizard Modules Directory](http://modules.dropwizard.io/thirdparty/) via https://github.com/dropwizard/modules.dropwizard.io/pulls

joschi:
@jelmerterwal As a side note, it would be great if you could integrate Weld with the HK2 injector used by Jersey, e. g. via https://jersey.java.net/documentation/latest/cdi.support.html#cdi.support.weld.se

jelmerterwal:
@joschi Just thought this might be something a lot of people might need. However, I do understand your point.

I believe that is exactly the approach taken :)

PR #2039, opened 2017-05-01
Add JDBI support for Period data types

jplock:


evnm:
This is great. Couple question:

1. Why not support `java.time.Duration` too?
2. Why not use the SQL interval type instead of strings?

joschi:
@jplock Do you think it would be possible to contribute support for `java.time.Period` and `java.time.Duration` to the upstream [jdbi project](https://github.com/jdbi/jdbi) directly?
They already support other `java.time` classes, so adding support for `Duration` and `Period` should be possible to merge quickly.

jplock:
Good call, let me submit a PR to JDBI.

jplock:
Looks like v3 already supports this - https://github.com/jdbi/jdbi/blob/master/postgres/src/main/java/org/jdbi/v3/postgres/PeriodArgumentFactory.java.  I'll close this PR.

PR #2046, opened 2017-05-04
fixed error of not using given factory in JsonConfigurationFactory

manuel-hegner:
This was an oversight on my part in the JsonConfigurationFactory. I created a new JsonParser like the YamlConfigurationFactory does. But it is better to use the one given in the ObjectMapper because the user might have changed the settings of this one in the bootstrapping phase.


jplock:
@manuel-hegner do you have a simple test case that we could add to this?

arteam:
Looks like the pull requests forces to use `MappingJsonFactory` which carries `ObjectMapper`  configured by Dropwizard. A unit test would be great, but I think in the current form it's justifiable to be merged.

arteam:
Applied to the 1.1.x branch as a5ecf60a8f26145a08f8654fddec8d03c5911472

PR #2052, opened 2017-05-10
adding test case for issue2049

anujkumar04:
Added the test case for the below scenario 
if a resource method parameter is annotated with @Valid @Validated(abc.class) and @NotNull, it throws a 500 instead of a 422


nickbabcock:
Hey the fix is to do:

```java
    @POST
    @Path("fooValidatedAndNotNull")
    public ValidRepresentation blahValidatedAndNotNull(
        @NotNull(groups = Partial1.class) 
        @Valid @Validated(Partial1.class) 
            ValidRepresentation representation) {
        return new ValidRepresentation();
    }
```

Notice the `NotNull` has `groups` specified. This is more of a docs issue, I believe!

Thank you, by the way, for the pull request. Makes replicating errors a lot easier 😄

anujkumar04:
thank you that works

nickbabcock:
I'm going to keep this PR open as a reminder that I need to update the docs 😄

PR #2057, opened 2017-05-11
Update to Jackson 2.9

nickbabcock:
Jackson 2.9 greatly simplifies exception handling with the new exception,
`InvalidDefinitionException`.

This PR technically breaks behavior for custom deserializers, but I
believe it is now the correct behavior. If a deserializer fails to to deserialize
input then a 400 should be returned as it denotes unexpected input. A 500
should only be returned if a deserializer encountered an error unrelated
to the input (if it read from a non-existant file or db...(not saying it's
a good design)). Added a test case to assert this.

Do not merge this PR as Jackson 2.9 is not actually released. I'm not sure when it will
be released, but I figured it would be a good idea to know what to expect.

### DIFF
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -24,7 +24,7 @@
         <dropwizard.version>${project.version}</dropwizard.version>
         <guava.version>22.0</guava.version>
         <jersey.version>2.25.1</jersey.version>
-        <jackson.version>2.8.9</jackson.version>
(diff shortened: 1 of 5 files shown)


jplock:
2.9 was released - https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9

nickbabcock:
Awesome, let me update the PR to rebase the commit and remove "WIP" 😄

PR #2059, opened 2017-05-16
Add support for HTML error responses and other non-JSON content types

### LINKED ISSUE #2038: HTML error responses
It would be nice to have proper support for HTML error responses (and other non-JSON content types too). Writing your own `ExceptionMapper` gets most of the way but is a bit ugly, and it doesn't seem to work at all for `AbstractParam` validation.

I've put together a tentative solution consisting of an `ErrorResponseProvider` interface that can be hooked up to your app's error views.

```java
public interface ErrorResponseProvider {

    Response getResponse(ErrorMessage errorMessage);

    Response getValidationResponse(ValidationErrorMessage errors);

    boolean isAcceptable();
}

```

It works with `LoggingExceptionMapper` and `JerseyViolationExceptionMapper` and I've managed to integrate with `AbstractParam` too although that was a bit more tricky.

Here's an outline of the solution:
https://gist.github.com/BenScholl/95236ce186f05d0f5bdd46f1e5dc93ad

Any comments would be much appreciated.

Ben

BenScholl:
Fixes #2038 .

This adds `ErrorEntityWriter`, a generic class that works for HTML error responses but also for other content types. Unit test + docs included.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/ErrorEntityWriter.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/ErrorEntityWriter.java
@@ -58,7 +58,7 @@ public abstract class ErrorEntityWriter<T, U> implements MessageBodyWriter<T> {
         responseHeaders.putSingle(HttpHeaders.CONTENT_TYPE, contentType);
 
         writer.writeTo(getRepresentation(entity), representation, representation, annotations,
-            MediaType.TEXT_HTML_TYPE, responseHeaders, entityStream);
+            contentType, responseHeaders, entityStream);
     }
 
     protected abstract U getRepresentation(T entity);
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/ErrorEntityWriter.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/ErrorEntityWriter.java
@@ -20,7 +20,7 @@ import java.lang.reflect.Type;
  * @param <T> The entity type to handle
  * @param <U> The response type to produce
  */
-public abstract class ErrorEntityWriter<T,U> implements MessageBodyWriter<T> {
+public abstract class ErrorEntityWriter<T, U> implements MessageBodyWriter<T> {
 
     /**
      * @param contentType Content type the writer will produce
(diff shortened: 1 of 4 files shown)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/ErrorEntityWriter.java
```diff
@@ -0,0 +1,74 @@
...
+                        MediaType mediaType,
+                        MultivaluedMap<String, Object> responseHeaders,
+                        OutputStream entityStream)
+        throws IOException, WebApplicationException {
+
+        final MessageBodyWriter<U> writer = mbw.get().getMessageBodyWriter(representation,
+            representation, annotations, contentType);
+
+        // Fix the headers, because Dropwizard error mappers always set the content type to APPLICATION_JSON
+        responseHeaders.putSingle(HttpHeaders.CONTENT_TYPE, contentType);
```
nickbabcock: Thinking out loud here; it seems like a little bit of a hack to override the content-type. I was thinking that it may be possible to remove the hard coded application json content type from the exceptionmappers, but it looks like it is since [we moved to Jersey 2.0](#706)

dropwizard-jersey/src/test/java/io/dropwizard/jersey/errors/ErrorEntityWriterTest.java
```diff
@@ -0,0 +1,71 @@
...
+        public ErrorEntityWriterTestResourceConfig() {
+            super(true, new MetricRegistry());
+
+            register(DefaultLoggingExceptionMapper.class);
+            register(DefaultJacksonMessageBodyProvider.class);
+            register(ExceptionResource.class);
+            register(new ErrorEntityWriter<ErrorMessage, String>(MediaType.TEXT_HTML_TYPE, String.class) {
+                @Override
+                protected String getRepresentation(ErrorMessage entity) {
+                    return "Uh oh: " + entity.getMessage();
```
nickbabcock: Is a plain string valid HTML?

BenScholl: Fair point - that is a bit misleading for anyone reading the test. I'll add some tags to make clear that it's HTML.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/ErrorEntityWriter.java
BenScholl: Totally agree. That's the same reason I'm injecting `HttpHeaders` here instead of just using the `mediaType` from the method parameters.

It would be nice to get rid of the hard coded application/json references at some point, but of course that affects all the parameter classes as well as the exception mappers ... and with my limited DW knowledge I have no idea what it might break...

arteam:
Thanks for the contribution!

PR #2062, opened 2017-05-18
Updating all jersey URL's to new Jersey home

sankate:
java.net site has closed. Updating all jersey URL's to new Jersey home.


nickbabcock:
Awesome, thank you so much!

PR #2063, opened 2017-05-18
Update the URL's for grizzly and jax-rs-spec

sankate:
`java.net` has closed. 
Update the URL's for `grizzly` and `jax-rs-spec`
Unfortunately, `jax-rs-spec` API Docs are no longer published on their site.


joschi:
@sankate Thanks! I've merged it into master with a small change.

Issue #2056, opened 2017-05-11
JDBI: Extending DBI/DBIFactory


isaki-x:
Greetings,

I am experimenting with extending DBIFactory to inject custom argument and column mapper factories into a custom IDBI object (which currently extends DBI, but I could just as easily use a decorator pattern). I've run into two issues (and I only need to solve one of them).

1) DBIFactory returns DBI instead of IDBI
2) DBIFactory invokes the DBI constructor with no way to intercept and inject my own version.

I can create one or more of the following patches to address this:

**Change DBIFactory to return IDBI (Problem 1)**

This will break existing code for users of dropwizard-jdbi. As such, I don't really like this idea. Also, it would remove the visibility of the registration methods for column mappers and arguments, which isn't ideal. As a result, I think this idea is a non-starter. I put it out there as it could solve my issue in a non-optimal way.

**Allow custom DBI instance creation into DBIFactory (Problem 2)**

There are two ways we can do this. If we want to be traditional, we could set up a protected method (createDBI) that could be overridden by child classes and defaults to "new DBI()".

If we want to go with a more Java 8-esque approach, I could add a new constructor to DBIFactory that takes a Supplier<DBI> for generating new DBI instances. The current constructor would be updated to provide a default supplier. Then, the build method could just invoke "get" on the supplier to create new DBI instances and then operate on them as it currently does.

Please let me know your thoughts on this and I will create a patch and submit a pull request if any of this works for you. I'm open to alternatives as well; worst case I end up writing my own factory that decorates DBIFactory and returns my own IDBI implementation and updating the handful of places I reference DBIFactory to use my own object instead. I don't necessarily like this approach as I prefer to subclass DBI instead of decorating it.

Thank you!

nickbabcock:
Related:

- We have [Jackson class](https://github.com/dropwizard/dropwizard/blob/c0d7379530df147c629b00798430bd73c8d7e7ab/dropwizard-jackson/src/main/java/io/dropwizard/jackson/Jackson.java) for configuring ObjectMapper
[...]

PR #2067, opened 2017-06-01
Refactor of DBIFactory (Issue #2056)

isaki-x:
Refactor of DBIFactory (Issue #2056)
    
- Split DBI construction and configuration into dedicated protected methods.
- Added unit tests to verify default configuration.

Please also merge this into master.

Thank you!

### DIFF
--- a/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
+++ b/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
@@ -52,7 +52,12 @@ import java.util.TimeZone;
 import static com.codahale.metrics.MetricRegistry.name;
 
 public class DBIFactory {
+
+    /**
+     * An SLF4J created instance of a LogBack logger for use with created DBI instances.
+     */
     private static final Logger LOGGER = (Logger) LoggerFactory.getLogger(DBI.class);
+    
     private static final String RAW_SQL = name(DBI.class, "raw-sql");
 
     private static class SanerNamingStrategy extends DelegatingStatementNameStrategy {
@@ -83,6 +88,21 @@ public class DBIFactory {
         return Optional.empty();
     }
 
+    
+    /**
+     * Build a fully configured DBI instance managed by the DropWizard lifecycle
+     * with the configured health check; this method should not be overridden
+     * (instead, override {@link #configure(DBI, String)} and/or
+     * {@link #newInstance(ManagedDataSource)})
+     * 
+     * @param environment
+     * @param configuration
+     * @param name
+     * @return A fully configured {@link DBI} object using a managed data source
+     *         based on the specified environment and configuration
+     * @see #build(Environment, PooledDataSourceFactory, ManagedDataSource,
+     *      String)
+     */
     public DBI build(Environment environment,
                      PooledDataSourceFactory configuration,
                      String name) {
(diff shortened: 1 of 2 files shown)


dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java
```diff
@@ -83,33 +88,104 @@ private SanerNamingStrategy() {
...
+     * if this is overridden it is strongly recommend that
+     * {@code super.configure(dbi, configuration)} is invoked before any other
+     * changes are made if you intend to use the default as a base so that the
+     * customized argument settings will supersede the defaults
+     * </p>
+     * 
+     * @param dbi
+     * @param configuration
+     */
+    protected void configure(final DBI dbi, final PooledDataSourceFactory configuration) {
```
nickbabcock: Should the argument be `String driverClazz` instead of `PooledDataSourceFactory`? It doesn't look like any other properties from the configuration are useful for JDBI. And it would make it easier to extract the creation and configuring of a `DBI` instance that isn't tied to a `DBIFactory`

nickbabcock: Just realized `databaseTimeZone` needs to be accounted for too, so might not be as straightforward as I initially thought 🤔

isaki-x: I originally had configure taking a second parameter as a `String` (called driverClazz), however the configuration class also happens to be the factory. Should anyone extend `PooledDataSourceFactory` to have extra options in their YAML config, they may wish to have that extra data available to their own `configure` method. That was my thinking anyway. I can very easily swap this back to using just a plain `String`.

As for `databaseTimeZone`, that is taken care of by a method so configure doesn't access it directly.

In retrospect, it might be ideal to take that as a parameter to configure instead of the method override approach; however I don't want to break any existing code in the field that might be using an override of the database time zone method.

nickbabcock: Good point, I probably should stop trying to kill two birds with one stone, and continue to do what I am currently doing. I maintain a [`JdbiUtils`](https://gist.github.com/nickbabcock/5c598065b0fbb80ec3ba9aea047fcc5a) in my code, so that I can do the following in my tests:

```java
final DBI jdbi = JdbiUtils.configureDbi(new DBI("url"), "", Optional.empty());
```

I don't think this should hold up this PR. :+1:

isaki-x: There could be another approach. We could create a minimally defined interface that is parent of PooledDataSourceFactory that has only the method(s) we need for setting up the default argument settings. Then, the configure method could take this interface as a second argument. If this interface has only one method, say `getDriverClass()`, then for easy testing you could do something like this:

```Java
configure(new DBI("url"), () -> "io.dropwizard.driver.Driver");
```

nickbabcock: That would be interesting to see 😄 

But if you think that would detract from the PR, we can continue in another issue/pr.

isaki-x: I'll look into it; I want to get the API right now so when I subclass this, I don't have to change it later if I don't have to. It might be worth moving to another issue, but it doesn't hurt to think about for a bit.

isaki-x: After looking into this, I feel it is probably worth pushing to another issue/PR and I'll update my downstream code when the time comes.

I have pushed one last change to remove some unused imports from the unit test.

nickbabcock: Ok sounds good to me 👍

isaki-x:
For some reason the Travis CI changes completed successfully but the pull request shows them as pending.

isaki-x:
Github support has asked me to close and re-open this to try and fix the hung Travis status.

isaki-x:
Re-opening as per GitHub support.

arteam:
Thank you very much for the contribution and a very thorough unit test!

arteam:
Applied to master as b6097f2ee2b3714f11d7c43318b2a44f51397cc7

isaki-x:
Thank you!

PR #2069, opened 2017-06-06
Add time zone to Java 8 datetime mappers

scottaj:
Unless I'm missing something this seems like a bug. I ran into it today while playing with JDBI.

Is there a good reason to keep around the default constructor? It looks like it is only used in tests right now. Those could easily pass an empty optional and get the same result, and that would have made this impossible to miss.


arteam:
Looks like so. Thanks for spotting and submitting the pull request.

arteam:
Applied to 1.1.x as c360323582e6c5a1817a06fa47f43216990f110b

PR #2010, opened 2017-04-13
use SizeAndTimeBasedRollingPolicy instead of deprecated SizeAndTimeBasedFNATP

qinfchen:

PR #2071, opened 2017-06-06
Fix size and time rolling appender

### LINKED ISSUE #2068: dropziward-example doesn't write any file log
Hi all,
last versions of dropwizard (1.1.0 1.2.0) doesn't write any file log, neither in my application based on dropwizard-example neither in dropwizard-example without any modifications

nickbabcock:
Without this fix, file logging configurations that use time and size based rotations will not work (ie, no log file or logging statements). Added test case to guard against regressions.

cc @charbonnier666 who reported the issue and also reported this PR fixed the issue

Closes #2068

### DIFF
--- a/dropwizard-example/src/test/java/com/example/helloworld/IntegrationTest.java
+++ b/dropwizard-example/src/test/java/com/example/helloworld/IntegrationTest.java
@@ -13,8 +13,12 @@ import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.MediaType;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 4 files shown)


arteam:
LGTM! It seems that the bug was introduced only in the 1.2 branch by #2010.

charbonnier666:
hi @arteam , I don't want to say stupid things :) , but I believe in 1.1.0 doesn't work too (seem to remember having tried it)

arteam:
Okay, I will check.

nickbabcock:
@charbonnier666, I adapted the test on the v1.1.1 tag and the unit test passed, so my initial thought is that this should already work for v1.1.1

charbonnier666:
@nickbabcock thanks nick :)

PR #2080, opened 2017-06-19
Test for #2046

manuel-hegner:
Sorry for the late response. I had a lot to do the last few weeks. Here is a test for the problem in #2046 . I added an example json config that contains a comment and can only be parsed if the object mapper is configured accordingly.


arteam:

PR #2081, opened 2017-06-19
Test for Generics class

manuel-hegner:
I create a test class for the Generics class in dropwizard-util.


jplock:
Would it be possible to convert this to assertj to be inline with the rest of the test suite?

manuel-hegner:
Oh sure, I didn't realize this library was used.

jplock:
Thanks!

Issue #2085, opened 2017-06-26
Update Jackson to address Deserializer security vulnerability


msymons:
A Jackson Deserializer security vulnerability is detailed by:

FasterXML/jackson-databind#1599
https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.8.9

Fixed by 
•	2.7.9.1 (for dropwizard 1.0.x branch)
•	2.8.9 (for dropwizard 1.1.x branch)
•	2.9.0.pr3

evnm:
Thanks for submitting these, @msymons!

BenScholl:
I **think** this won't affect most Dropwizard apps, unless they specifically set `objectMapper.enableDefaultTyping()` or do funky things with e.g. `@JsonTypeInfo`.

At least that's my understanding after reading the excellent paper at https://github.com/mbechler/marshalsec

For reference: https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization

arteam:
I think so too. Nevertheless, I've pushed bugfix releases 1.1.2 and 1.0.8 which update Dropwizard to the Jackson versions which are not affected by this vulnerability.

PR #2088, opened 2017-06-26
Update jackson to 2.8.9

msymons:
Update jackson version from 2.8.8 to 2.8.9 to address Deserializer security vulnerability, per #2085
vulnerability is also fixed in jackson 2.9.0.pr3.    (see #2057)


evnm:
Jackson 2.8.9 [includes](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.8.9) a handful of patches in addition to the `Deserializer` fix, but none of them sound worrisome. I'm inclined to merge this, since we should address the vulnerability ASAP.

Issue #2084, opened 2017-06-23
NTLM Proxy Auth for the HTTP client


stringy05:
I'm doing some work at BigCorp who use Windows and NTLM proxies everywhere,  so I need my client to be able to authenticate against a proxy using NTLM.

I've written a patch for it in dropwizard-client, basic outline is:
- Change the AuthConfiguration, adding NTLM requrired fields such as 
  - realm
  - domain
  - hostname
  - a credentialType parameter, which determines which implementation of the Credential interface to use

The alter the proxy creation part of the createClient method to provide a NTLM aware 

``` 
    if (auth != null) {
        if (credentialsProvider == null) {
            credentialsProvider = new BasicCredentialsProvider();
        }
        // set the AuthScope
        AuthScope authScope = new AuthScope(httpHost, auth.getRealm(), auth.getAuthScheme());

        // set the credentials type
        Credentials credentials = null;
        if (null != auth.getCredentialType() && auth.getCredentialType().equalsIgnoreCase(AuthConfiguration.NT_AUTH_TYPE)) {
            credentials = new NTCredentials(auth.getUsername(), auth.getPassword(), auth.getHostname(), auth.getDomain());
        } else {
            credentials = new UsernamePasswordCredentials(auth.getUsername(), auth.getPassword());
        }

        credentialsProvider.setCredentials(authScope, credentials);
    }
} 
```

Works on my machine :)

arteam:
This looks like a useful addition! I would accept the patch.

arteam:
Closed via #2091

PR #2091, opened 2017-06-28
Client ntlm auth

stringy05:
patch for #2084 to allow a NTCredentials instance to be used for proxy auth. I've also updated the doco, however my ServiceUnavailableRetryStrategy is somewhat unreliable at my work, sometimes it works, sometimes it doesnt


nickbabcock:
Thank you for an awesome pull request 🎉 

Something that isn't obvious to me is how `ServiceUnavailableRetryStrategy` is related to NTLM authentication. If not required, it may be prudent to break that into another PR so those changes can be more easily digested and reviewed.

stringy05:
It's not directly related so I will put it into a second PR. Basically I
was trying to use the Retry Strategy to inject a NTLM token when a 407 is
received from the Proxy. I'd added it to the docs, however took it out as
It works but is pretty unreliable

arteam:
Looks good to me! Thank you very much for this contribution!

PR #2096, opened 2017-07-05
Use LoadingCache in CachingAuthorizer

FredDeschenes:
As done in #1615 for CachingAuthenticator. Since `Authorizer::authorize` doesn't have a `throws` clause, I've only handled `UncheckedExecutionException`s.

Result when `null` is returned by the authorizer will differ from previously, where a `NullPointerException` would have been thrown by the cache's `put` method, whereas now a `InvalidCacheLoadException` will be thrown by the cache itself when the loader returns null. There might be a way to have the same result as previously, let me know if you want me to look into it (but personally I don't think it really matters as most users won't be catching those exceptions anyway, and it might even be clearer as to what happened).

EDIT: Looking at this again, `Authorizer::authorize` returns a `boolean` and not a `Boolean` so null returns shouldn't happen anyway.

### DIFF
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/CachingAuthorizerTest.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/CachingAuthorizerTest.java
@@ -10,6 +10,7 @@ import org.mockito.InOrder;
 import java.security.Principal;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNullPointerException;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.anyString;
 import static org.mockito.Mockito.inOrder;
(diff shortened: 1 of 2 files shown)


FredDeschenes:
Coverage went down because [this line](https://coveralls.io/builds/12260797/source?filename=dropwizard-auth%2Fsrc%2Fmain%2Fjava%2Fio%2Fdropwizard%2Fauth%2FCachingAuthorizer.java#L96) is unreachable but needed for compilation (since Throwables.propagate is now deprecated). Not much I can do I think.

dropwizard-auth/src/test/java/io/dropwizard/auth/CachingAuthorizerTest.java
```diff
@@ -119,8 +125,17 @@ public void calculatesTheSizeOfTheCache() throws Exception {
...
+        cached.authorize(principal, role);
+        assertThat(cached.stats().loadCount()).isEqualTo(1);
         assertThat(cached.size()).isEqualTo(1);
     }
+
+    @Test
+    public void shouldPropagateRuntimeException() throws AuthenticationException {
+        final RuntimeException e = new NullPointerException();
+        when(underlying.authorize(principal, role)).thenThrow(e);
+        expected.expect(CoreMatchers.sameInstance(e));
```
arteam: I think we could use AssertJ's exception assertion, which looks concise and provide a more wide range of assertions than the `ExpectedException` rule. Something like that:
```java
assertThatNullPointerException()
            .isThrownBy(() -> cached.authorize(principal, role))
            .isSameAs(e);
```

FredDeschenes: Oh that's nice, I didn't even know it existed! I'll change this and update the PR.

FredDeschenes:
No problem, I was looking at the code for CachedAuthorizer and CachedAuthenticator and noticed the authorizer hadn't been updated.

I've push a new commit with the modification to the unit test.

arteam:
Thanks for your contribution!

FredDeschenes:
You're welcome!

Issue #2098, opened 2017-07-10
Add "immediateFlush" configuration for FileAppenderFactory


pavelkokush:
Would be nice to use "immediateFlush" feature from ch.qos.logback.core.encoder.LayoutWrappingEncoder in FileAppenderFactory

nickbabcock:
This was merged in #2193

PR #2100, opened 2017-07-10
Issue #2098: Add "immediateFlush" configuration for FileAppenderFactory

pavelkokush:


nickbabcock:
Hey, thanks for the PR!

Couple of things:

1. I'm assuming that you want the ability to customize `immediateFlush` to increase logging performance, but at the cost of safety? Is this an optimization that you have found to actually be tangible?
2. A test will be needed that confirms that property is actually set through the yaml configuration.
3. Documentation will need to be added in the configuration reference (and javadocs) that clearly states the appropriate trade-offs.

pavelkokush:
1. Yep, on prod sometimes we generate a lot of logs in short period of time. I verified that with "immediateFlush=false" it spend much less time on writing to log file (~ 2-3 times). We also write logs in Redis, so it is not big problem that we may lost some logs in file in very edge case.

2. Added `in FileAppenderFactoryTest` . Have to refactor impl and test a bit, hope it is ok:) . How to test it with real yaml configuration?

3. Added documentation and javadoc.

arteam:
The general issue with the PR is that it has a lot of unrelated changes. Does the ability disabling immediate flushing warrant them? I would expect that setting the correspondending attribute on the file appender will be enough.

pavelkokush:
Yep, setting attribute on the file appender should be enough. As I remember, I just found that it is not easy to write clean test for it. Also I found some code duplications in existing impl. So I improved code a bit. But If you do not think that is better or if you feel that it not safe improvement, then I can remove "refactoring" part and leave only "feature" part

pavelkokush:
@arteam simple implementation posted in https://github.com/dropwizard/dropwizard/pull/2193

Issue #2097, opened 2017-07-07
Malformed Accept-Language header causes server error (dropwizard-views)


BenScholl:
To reproduce:

    curl -H 'Accept-Language:;' localhost:8080/some_view

This returns an HTTP 500 and dumps a stack trace to the app's server log.

It happens because `HttpHeaders.getAcceptableLanguages()`, as called by `ViewMessageBodyWriter.detectLocale()` can throw a `HeaderValueException`.

I think a "permissive" approach probably makes sense here: catch the exception and just fall back to the default `Locale` if the header is malformed.

nickbabcock:
Your viewpoint makes sense. A server error should definitely not be the current behavior. Feel free to make a PR for it!

FredDeschenes:
I agree that this shouldn't return a server error, but I don't agree that the error should just be ignored and the default language used.

IMO, a `400 - Bad Request` should be returned and the client should either not send the header, or send a properly formatted one.

BenScholl:
I don't think it is a big deal either way.

However, just looking at the Alexa top 5:

    $ curl -s -o /dev/null -w "%{http_code}\n" -H 'Accept-Language:;' http://www.google.co.uk
    200

    $ curl -s -o /dev/null -w "%{http_code}\n" -H 'Accept-Language:;' https://www.youtube.com
    200

    $ curl -s -o /dev/null -w "%{http_code}\n" -H 'Accept-Language:;' https://www.facebook.com
    200

    $ curl -s -o /dev/null -w "%{http_code}\n" -H 'Accept-Language:;' https://www.baidu.com
    200

    $ curl -s -o /dev/null -w "%{http_code}\n" -H 'Accept-Language:;' https://www.wikipedia.org
    200

FredDeschenes:
Weird, I would've thought that most sites wouldn't let this go through.

I've also thought about having a configuration, but the views configuration is just a Map. We could have an overrideable function in "ViewBundle", but that's a bit weird. Also I think the feature is only available in Freemarker and not in Mustache, which makes this even trickier to configure.

arteam:
Resolved by #2103

PR #2103, opened 2017-07-12
Handle badly formed 'Accept-Language' headers (#2097)

FredDeschenes:
ViewMessageBodyWriter::detectLocale will now catch
`HeaderValueException`s thrown when retrieving the languages from the
headers and return a `400 - Bad Request` to clients instead of a server
error.

The issue (#2097) talked about using the default locale in that case, but I don't agree. Not my decision in the end though, I can change this to use the default locale if you guys prefer this.


arteam:
Applied to the 1.1 branch as 48b88252.

PR #1900, opened 2017-01-24
Support Dropwizard applications without logback

arteam:
This is a prototype for splitting the tight integration between Dropwizard and Logback. This feature was requested several times (See #1367). The idea is too provide users the ability to exclude Logback from Dropwizard and replace it with an external logging solution, if they really want to. The main work of this PR is too initialize the Logback infrastructure lazily, so the user can swap it to an external logging mechanism in the config file.

With this change users can swap logback to other logging implementation like that:
```xml
 <dependencies>

        <dependency>
            <groupId>io.dropwizard</groupId>
            <artifactId>dropwizard-core</artifactId>
            <version>1.1.0-SNAPSHOT</version>
            <exclusions>
                <exclusion>
                    <groupId>ch.qos.logback</groupId>
                    <artifactId>logback-classic</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>ch.qos.logback</groupId>
                    <artifactId>logback-access</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>org.slf4j</groupId>
                    <artifactId>log4j-over-slf4j</artifactId>
                </exclusion>
            </exclusions>
        </dependency>

        <dependency>
            <groupId>log4j</groupId>
            <artifactId>log4j</artifactId>
            <version>1.2.17</version>
        </dependency>

        <dependency>
            <groupId>org.slf4j</groupId>
            <artifactId>slf4j-log4j12</artifactId>
            <version>1.7.21</version>
        </dependency>

    </dependencies>
``` 

```yml
server:
  type: simple
  applicationContextPath: /application
  adminContextPath: /admin
  connector:
    type: http
    port: 8080
  requestLog:
    type: external
logging:
  type: external
```

```java
public class ExampleApplication extends Application<ExampleConfiguration> {

    @Override
    protected void bootstrapLogging() {
    }
}
```

PR #2112, opened 2017-07-27
Remove Dropwizard's Jackson dependency on Logback

arteam:
In the move to allow users plug in custom logging implementations except Logback, we should make all Logback dependencies in Dropwizard optional. Currently, our Jackson mapper is dependent on Logback for
deserializing logging `Level` in the configuration. It's forces every Dropwizard app to have a runtime dependency on Logback. To avoid this we could remove `Level` deserializers and move converting a string text level to Logback's level to `DefaultLoggingFactory`. Since `DefaultLoggingFactory` can be replaced by `ÈxternalLoggingFactory` this allows users to run Dropwizard app without a Logback dependency.


arteam:
Ah, I forgot about `threshold` in `AbstractAppenderFactory`, but we can't change its type because it's `protected` and part of the public API.

jplock:
@arteam this good to merge?

arteam:
I guess this should be good to merge. The issue was reported by @jsauvain in #1900 and the change in the pull requests should fix it. My biggest concert is that we have to change the public API which could break third-party appenders. I had a look at them and most of them use `threshold` variable directly or don't use at all. 

I am leaning to merge this MR and cut a release candidate. In case this change breaks something important, we can revert it.

swagle:
This change is missing in the 1.2.x line. Hit this issue when trying to use this feature on 1.2.0 bits published to maven central.

joschi:
@swagle Nope, the commit (c28ca30d400bf50a9b93209a7887a20b68a4c209) is also part of Dropwizard 1.2.0.

swagle:
Hi @joschi, you are right the patch is present, this breaks at io.dropwizard.Application#bootstrapLogLevel

arteam:
For those who experience this issue, you should override `bootstrapLogging()` and then `bootstrapLogLevel(Level)` won't be invoked.

```java
public class ExampleApplication extends Application<ExampleConfiguration> {

    @Override
    protected void bootstrapLogging() {
    }

    @Override
    public void run(ExampleConfiguration exampleConfiguration, Environment environment) throws Exception {
        LoggerFactory.getLogger(ExampleApplication.class).info("Test: {}", exampleConfiguration.getTest());
    }

    public static void main(String[] args) throws Exception {
        new ExampleApplication().run(args);
    }
}
```

jdimeo:
I must be missing something because I just pulled down 1.2.0 from Maven Central today in eager anticipation of this PR, and the source code for Application.java has `import ch.qos.logback.classic.Level;` and APIs that reference `Level` thus triggering indirect class not found compilation errors when I try to use Log4j. 
The SHA1 of the sources.jar I pulled down for core is `2c35bff8aeb55631cfd21cbe9c021c4d60b9e4df`

arteam:
Yes, the import is still there, but the JVM lazily initializes classes and methods lazily, so if you override the `bootstrapLogging` method in the `Application` class (which invokes the `bootstrapLogLevel` method=, your application should successfully load. 

I created a simple demo for a Dropwizard app with log4j 2: https://github.com/arteam/dropwizard-nologback

jdimeo:
So, to be more precise, you are correct in that I can get a clean compile with `javac` (specifically `mvn clean compile`). However, my IDE (specifically Eclipse) is fooled (or too smart for its own good, depending on how you look at it). See:
image
That prevents me from running my service through the IDE.

Thank you for the demo app! A skeleton app is worth a thousand words.

rvsasseen:
I found that there are further magic incantations needed which are described at https://www.dropwizard.io/1.3.5/docs/manual/core.html#logging.

Issue #2128, opened 2017-08-09
Log resource validation failures


isaki-x:
So, it seems straightforward to log JSON processing errors (set DEBUG logging for `io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper`). However, I don't see a way to easily log validation errors on the server side.

I suspect the class that should log it is `io.dropwizard.jersey.validation.JerseyViolationExceptionMapper`.

If I'm wrong and this is logged somewhere, please point me in the right direction. If not, please tell me if I have the right class and I will submit a pull request to add logging at the DEBUG level.

Thanks!

nickbabcock:
Ah yes, that does look like to be an oversight, they should be more aligned! Yeah, I'll accept a PR for this.

isaki-x:
Is JerseyViolationExceptionMapper the right spot?

nickbabcock:
Yes it is!

isaki-x:
Closing this issue; further discussion will be in the PR.

PR #2129, opened 2017-08-09
Exception Mapper Logging Enhancements

isaki-x:
Please see: #2128

Commit notes:

```
Date:   Wed Aug 9 16:10:01 2017 -0400

    Exception mapper logging enhancements
    
    1) Implemented debug logging for JerseyViolationExceptionMapper.
    2) Fixed compiler warning regarding serialization for
    JerseyViolationException.
    3) Wrapped debug statement in JsonPropcessingExeptionMapper with
    "isDebugEnabled"; this will ensure the JVM does not bother loading and
    caching the log string unless debug is actually enabled. This will also
    allow more aggressive code optimization when running the server JVM.
```


isaki-x:
Well that was unexpected. I will work on getting the code coverage up. While I'm at it, I noticed that the validation mapper and its unit test class do not share a name; is this intentional? If not, I can rename it as part of my changes.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
```diff
@@ -38,12 +38,14 @@ public Response toResponse(JsonProcessingException exception) {
             return super.toResponse(exception); // LoggingExceptionMapper will log exception
         }
 
-        final String message = exception.getOriginalMessage();
-
         /*
          * Otherwise, it's those pesky users.
          */
-        LOGGER.debug("Unable to process JSON", exception);
+        if (LOGGER.isDebugEnabled()) {
```
nickbabcock: It looks to me, since nothing fancy is going on with the log format, `isDebugEnabled` is not necessary, considering that `debug` calls `isDebugEnabled` as well before actually logging.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/JerseyViolationExceptionMapper.java
```diff
@@ -12,8 +14,14 @@
 
 @Provider
 public class JerseyViolationExceptionMapper implements ExceptionMapper<JerseyViolationException> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(JerseyViolationExceptionMapper.class);
+
     @Override
     public Response toResponse(final JerseyViolationException exception) {
+        if (LOGGER.isDebugEnabled()) {
```
nickbabcock: Same thing here.

isaki-x: If you look at the byte code you realize it is necessary. It's splitting hairs for predefined static strings; you are basically saving a one time allocation for the error log string. It gets interesting in the server JVM when branch prediction can optimize away the "if (LOG.isDebugEnabled)" check. I'm not sure LOG.debug itself would be removed in this case; likely you end up still invoking the call and having the method body optimized away (unless your LOG.debug ended up inlined by the JIT).

From a best practice perspective, however, it is ideal to have consistency.

Take this example:

```Java
LOG.debug("Something to say " + result.toString());
```

In this case, what actually happens under the hood is that the toString() method is invoked on result (may or may not be cheap) following by String concatenation (not cheap) before LOG.debug is actually invoked. In the case where `LOG.isDebugEnabled()` would return false, you do a bunch of work for no logging output. This is unlikely to be optimized away as "toString()" varies wildly across implementations and may even have side effects (if it is poorly designed).

If you wrap the call in LOG.isDebugEnabled(), on a cold JVM you pay the cost of a boolean to save some potentially expensive operations. On a hot server JVM the entire block is eliminated. On a hot server JVM in the case debug is true, the if check is simply eliminated.

By always doing this, whether it is needed or not, you ensure that if someone changes the log line to become costly you don't have to worry the developer remembered to add the if {} decoration.

I personally do this for all log levels as a stylistic choice. Typically if you are certain you will always be logging at INFO, it is sufficient to do it only for DEBUG. Since I do turn logging way down from time to time, it makes sense.

Given this is for error handling code and generally shouldn't be hit outside of development or improper handled edge cases (unless you are under attack), if you prefer not having the if {} paradigm for this I will acquiesce as this isn't my codebase.

Thanks!

dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
isaki-x: See my response to your other mention of this. TL;DR: Done.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/JerseyViolationExceptionMapper.java
nickbabcock: The difference between

```java
LOG.debug("Something to say", throwable);

// and

if (LOG.isDebugEnabled()) {
    LOG.debug("Something to say", throwable);
}
```
should be so minimal in this case (we're talking about in the middle of a jersey exception handler!) that I would prefer more the concise approach. 

Thank you for making the changes.

isaki-x: No problem! Hopefully removing those lines buys back the line coverage.

isaki-x:
I'm not sure how to clear out the "changes requested" failure. I thought committing the requested changes would fix it.

nickbabcock:
Don't worry, it's something the reviewer has to do on their end 😄 

I've approved it.

isaki-x:
Good to know, thanks!

isaki-x:
Would it be possible to this merged into 1.1.X as well as 1.2.X if 1.2.X is a ways out? Thanks!

arteam:
@isaki-x LGTM, Thank you!

arteam:
I will make a RC for 1.2 with this enhancement.

Issue #2122, opened 2017-08-03
Commands and customizable exception handling


nickbabcock:
With 1.2.0-rc1 out I tried rolling out to all my apps in testing, but I hit a snag:

#2003 introduced printing the stack trace on an exception which is good for the `server` command but the wrong call for `check` and other commands.

Let's look at the differences between v1.1.3 and v1.2.0-rc1 with dropwizard-example by `check banner.txt`

v1.1.3

```
classes\banner.txt has an error:
  * Failed to parse configuration; Can not construct instance of com.example.helloworld.HelloWorldConfiguration: no String-argument constructor/factory method to deserialize from String value ('web-scale hello world dP for the web 88 .d8888b. dP.  .dP .d8888b. 88d8b.d8b. 88d888b. 88 .d8888b. 88ooood8  `8bd8'  88'  `88 88'`88'`88 88'  `88 88 88ooood8 88.  ...  .d88b.  88.  .88 88  88  88 88.  .88 88 88.  ... `88888P' dP'  `dP `88888P8 dP  dP  dP 88Y888P' dP `88888P' 88 dP')
 at [Source: N/A; line: -1, column: -1]
```

v1.2.0

```
io.dropwizard.configuration.ConfigurationParsingException: classes\banner.txt has an error:
  * Failed to parse configuration; Cannot construct instance of `com.example.helloworld.HelloWorldConfiguration` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('web-scale hello world dP for the web 88 .d8888b. dP.  .dP .d8888b. 88d8b.d8b. 88d888b. 88 .d8888b. 88ooood8  `8bd8'  88'  `88 88'`88'`88 88'  `88 88 88ooood8 88.  ...  .d88b.  88.  .88 88  88  88 88.  .88 88 88.  ... `88888P' dP'  `dP `88888P8 dP  dP  dP 88Y888P' dP `88888P' 88 dP')
 at [Source: UNKNOWN; line: -1, column: -1]

        at io.dropwizard.configuration.ConfigurationParsingException$Builder.build(ConfigurationParsingException.java:276)
        at io.dropwizard.configuration.BaseConfigurationFactory.build(BaseConfigurationFactory.java:157)
        at io.dropwizard.configuration.BaseConfigurationFactory.build(BaseConfigurationFactory.java:92)
        at io.dropwizard.cli.ConfiguredCommand.parseConfiguration(ConfiguredCommand.java:124)
[...]

PR #2133, opened 2017-08-18
Allow commands to customize exception handling

nickbabcock:
Fix for #2122

Allows for commands to catch configuration, environment, and command exceptions in one place (eg. forcing all commands to print the stack trace is less than desirable)

Below is an example of a custom command that will print the stack trace to stderr if someone passed in `--debug`, else print the error message to stdout.

```java
public static final class CustomCommand extends Command {
    @Override
    public void configure(Subparser subparser) {
        subparser.addArgument("--debug").action(Arguments.storeTrue());
    }

    @Override
    public void run(Bootstrap<?> bootstrap, Namespace namespace) throws Exception {
        throw new RuntimeException("I did not expect this!");
    }

    @Override
    public void onError(Cli cli, Namespace namespace, Throwable e) {
        if (namespace.getBoolean("debug")) {
            super.onError(cli, namespace, e);
        } else {
            cli.getStdOut().println(e.getMessage());
        }
    }
}
```

Reviews, suggestions welcomed. Some things to think about:

- I am unsure whether to keep `Cli` argument in `onError` or replace it with `stdOut` and `stdErr` directly from `Cli`
- Should configured commands get another argument for onError, except it has configuration (if parsed correctly?). Same thing with environment commands.
- I decided to not allow customization of HelpScreenException logging

EDIT: I should clarify, this `onError` gets called before `onFatalError` so `onError` doesn't need to do `System.exit(1)`

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/cli/CheckCommand.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/cli/CheckCommand.java
@@ -37,4 +37,10 @@ public class CheckCommand<T extends Configuration> extends ConfiguredCommand<T> 
                        T configuration) throws Exception {
         LOGGER.info("Configuration is OK");
     }
+
+    /* The stacktrace is redundant as the message contains the yaml error location */
+    @Override
+    public void onError(Cli cli, Namespace namespace, Throwable e) {
+        cli.getStdErr().println(e.getMessage());
+    }
 }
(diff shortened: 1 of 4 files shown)


arteam:
I like it! I think it provides great control for those user who want to do more fine-agreed exception handling.
I also agree that the check command doesn't need the whole stacktrace to be dumped, because most of the lines are useless and hide the real error.

PR #2137, opened 2017-08-24
docs: Fix broken link in user manual to Hibernate grouped constraints docs

evnm:
I noticed today that a link in the user manual to a section of Hibernate Validator's user manual is broken. This change updates the link to point to the appropriate section of Hibernate's stable docs.

Not sure if we care about pinning the user manual to a specific version of the docs. The only past version of their docs that I can find online is [5.1](https://docs.jboss.org/hibernate/validator/5.1/reference/en-US/html/chapter-groups.html#chapter-groups), which is a few minor versions behind where Dropwizard is, at 5.4.


nickbabcock:
What about this link? http://docs.jboss.org/hibernate/validator/5.4/reference/en-US/html_single/#chapter-groups

Issue #898, opened 2015-02-25
Request Logs do not provide valid user field


abuchanan920:
It seems that the authentication schemes of Jetty and Dropwizard don't communicate enough information for the request logs to include the user field.

The user field would definitely be useful for identifying how much usage was being generated by different api users or identifying which users are still using an older version of the api so they can be targeted for upgrade, etc.

It looks like io.dropwizard.jetty.Slf4jRequestLog extends org.eclipse.jetty.server.AbstractNCSARequestLog which does the actual creation of the message.

In that class we see:

```
            Authentication authentication = request.getAuthentication();
            if (authentication instanceof Authentication.User)
                buf.append(((Authentication.User)authentication).getUserIdentity().getUserPrincipal().getName());
            else
                buf.append(" - ");
```

In all of my logs, I only ever see "-" for the user, although dropwizard basic authentication seems to be working correctly. I can't find anywhere in the code where it attempts to push the user info back into the Jetty world for logging to work correctly though.

Am I missing something?

Thanks!
Andrew

glennmcallister:
You aren't missing anything. The Dropwizard and Jetty authentication
mechanisms are currently more or less completely disconnected. When I was
satisfying auditing requirements I had to add my own logging handler class
to write out auth results in JSON format to a separate file, based on a
separate logger I had instantiated in my authentication class.

And yes, I agree this is something that could be improved upon. However,
the problem is that we are effectively doing authentication as part of
parameter processing rather than as a filter, as it's typically done. We
are planning on looking into this in more detail to address a known defect
(see issue #768) post the 0.8 release.

Glenn

yvandb:
Is there any update on this issue? @glennmcallister? It seems dw 0.9 moved to filter based authentication, however jetty still doesn't pick up the user identity for the requestlog. 
Thank you very much for your attention.

asiletto:
@yvandb I think if you are using a filter based authentication, you can try to set the user name as said here: https://groups.google.com/forum/#!topic/dropwizard-user/lVP0nDI-NT4

glennmcallister:
Sorry, I've long since been doing next to nothing with Dropwizard; no time
and a bit handcuffed by my current employer for various reasons. That being
said, from what I know, given the current authentication model employed by
DW there is likely no reason why this can't be supported, it just hasn't
been looked at.

yvandb:
Thanks @asiletto @glennmcallister 
[...]

PR #2138, opened 2017-08-28
added @Priority(Priorities.AUTHENTICATION) to WebApplicationExceptionCatchingFilter

tmszdmsk:
When `@Auth Optional<UserOrWhatever>` is used. Registered `WebApplicationExceptionCatchingFilter` has no Priority set up and in the result, filter is executed at the end of the filter chain.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/WebApplicationExceptionCatchingFilter.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/WebApplicationExceptionCatchingFilter.java
@@ -3,6 +3,9 @@ package io.dropwizard.auth;
 import com.google.common.base.Preconditions;
 
 import java.io.IOException;
+
+import javax.annotation.Priority;
+import javax.ws.rs.Priorities;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
@@ -12,6 +15,7 @@ import javax.ws.rs.container.ContainerRequestFilter;
  * WebApplicationException WebApplicationExceptions} thrown by an
  * underlying {@code ContextRequestFilter}.
  */
+@Priority(Priorities.AUTHENTICATION)
 class WebApplicationExceptionCatchingFilter implements ContainerRequestFilter {
     private final ContainerRequestFilter underlying;
 
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/OptionalAuthFilterOrderingTest.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/OptionalAuthFilterOrderingTest.java
@@ -1,0 +1,120 @@
+package io.dropwizard.auth;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.dropwizard.auth.principal.NullPrincipal;
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
+import org.glassfish.jersey.servlet.ServletProperties;
+import org.glassfish.jersey.test.DeploymentContext;
+import org.glassfish.jersey.test.JerseyTest;
+import org.glassfish.jersey.test.ServletDeploymentContext;
+import org.glassfish.jersey.test.TestProperties;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.security.Principal;
+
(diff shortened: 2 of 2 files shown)


tmszdmsk:
I consider writing a test for that. What are your suggestions in the context of test level & packaging?

lutzhorn:
Two question to clarify why this change would be needed:

* Do any other JAX-RS components in Dropwizard use `javax.ws.rs.Priorities`?
* What happens if the change is *not* made`?

joschi:
> What are your suggestions in the context of test level & packaging?

You could write a resource test for this which triggers an exception. See http://www.dropwizard.io/1.1.4/docs/manual/testing.html#testing-resources for details.

> Do any other JAX-RS components in Dropwizard use `javax.ws.rs.Priorities`?

;)
```
$ grep -r @Priority */src/main/java
dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java:@Priority(Priorities.AUTHENTICATION)
dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentialAuthFilter.java:@Priority(Priorities.AUTHENTICATION)
dropwizard-auth/src/main/java/io/dropwizard/auth/chained/ChainedAuthFilter.java:@Priority(Priorities.AUTHENTICATION)
dropwizard-auth/src/main/java/io/dropwizard/auth/oauth/OAuthCredentialAuthFilter.java:@Priority(Priorities.AUTHENTICATION)
dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/CharsetUtf8Filter.java:@Priority(Priorities.HEADER_DECORATOR)
dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java:@Priority(Priorities.USER)
dropwizard-jersey/src/main/java/io/dropwizard/jersey/gzip/ConfiguredGZipEncoder.java:@Priority(Priorities.ENTITY_CODER)
dropwizard-jersey/src/main/java/io/dropwizard/jersey/gzip/GZipDecoder.java:@Priority(Priorities.ENTITY_CODER)
```

> What happens if the change is not made?

The default priority is `Priorities.USER` (5000).

joschi:
@tmszdmsk 
> Registered `WebApplicationExceptionCatchingFilter` has no Priority set up and in the result, filter is executed at the end of the filter chain.

And what's the consequence of this?

tmszdmsk:
@joschi, @lutzhorn 

My use case:
I have `dropwizard-auth` set up to do authentication in `Priorities.AUTHETNICATION` = `1000`  phase (by extending [AuthFilter.java#L16](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java#L16)). 

I have another filter that requires authentication data to be present in `SecurityContext`. It is marked with `Priorities.AUTHORIZATION` = `2000`.

It works completely fine unless I wrap `Principal`-like object in `Optional` wrapper. Because of that my filter [is wrapped in WebApplicationExceptionCatchingFilter](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthDynamicFeature.java#L59). It has no Priority annotation, hence executes in default `Priorities.USER` = `5000` phase. 

@joschi 
> You could write a resource test for this which triggers an exception. See http://www.dropwizard.io/1.1.4/docs/manual/testing.html#testing-resources for details.

👍

tmszdmsk:
@joschi @lutzhorn 
I've committed a test that shows how my use case looks like.
On `master` only one of them passes.

ianferguson:
I've run into this issue recently as well, my use case is that when using a filter to add the user id/principal name to the jetty container request so it will be included in access logs (using logback-access similar to the code snippet/proposal here: https://github.com/dropwizard/dropwizard/issues/898#issuecomment-163342838), 

if that filter is set to a `Priorities.AUTHETNICATION + 1` (1001), on resource methods where an `@Auth Optional<Principal>` is injected, the auth filter will be run at priority 5000 because it is wrapped in the WebApplicationExceptionCatchingFilter and will not run until after the user id filter runs, causing the user id filter to not set the user id, and the access logs to just show "-" for the user id, instead of the authenticated user (if there is one)

arteam:
Looks good to me. A very nice unit test, thanks for the contribution! I will release a new 1.2.0-RC.

Issue #2142, opened 2017-09-01
Is it safe to enable prudent mode on FileAppender ?


genz10:
Hello,

Is it safe to enable (by overriding) prudent mode in FileAppender ? 

because dropwizard disabled prudent mode by default shows in this [commit](https://github.com/dropwizard/dropwizard/blob/7322af7c3a32531ed3c05b188c128c73f611f3a9/dropwizard-core/src/main/java/com/yammer/dropwizard/logging/FileLoggingOutput.java)

Thanks.

isaki-x:
That is some rather old code; I think the class you want is ```dropwizard-logging/src/main/java/io/dropwizard/logging/FileAppenderFactory.java```.

You should read the documentation on prudent: https://logback.qos.ch/manual/appenders.html

There are serious ramifications to setting this to true. If you are willing to pay the costs, it is probably safe. I will defer to the owners of this library to correct me if I am wrong.

Provided it is safe, it would be an easy PR to add an option for configuring prudent, with ```false``` being the correct default and documentation telling people not to change this unless they know what they are doing.

genz10:
Yes,

I think I can accept those trade-off. Soon i'll submit PR regarding this configuration.

Thank for your answer @isaki-x

isaki-x:
No problem! I took a look at your PR and while I'm not an owner of this repo it looks reasonable.

You will however need unit tests; your PR failed the coverage requirements.

arteam:
Tracked in #2143

PR #2143, opened 2017-09-02
Configurable prudent mode

genz10:
Issue refference : https://github.com/dropwizard/dropwizard/issues/2142

Users will able to enable prudent mode in JSON config file.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -694,7 +694,7 @@ neverBlock                   false        Prevent the wrapping asynchronous appe
                                           Set to true to disable blocking.
 bufferSize                   8KB          The buffer size of the underlying FileAppender (setting added in logback 1.1.10). Increasing this
                                           from the default of 8KB to 256KB is reported to significantly reduce thread contention.
-prudent                      false         In prudent mode, FileAppender will safely write to the specified file,
+prudent                      false        In prudent mode, FileAppender will safely write to the specified file,
                                           even in the presence of other FileAppender instances running in different JVMs,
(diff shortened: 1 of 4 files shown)


genz10:
I've missed some important part behind prudent mode, such as 

`In prudent mode, file compression is not supported nor allowed. (We can't have one JVM writing to a file while another JVM is compressing it.)`

In my opinion, it will trigger another problem such as archiving and massive log size (CMIIW). Reason behind I want to enable prudent mode was I have multiple JVM writing on the same log file.

Do anyone here have more simple and elegant solution ? 

Thanks!

mattnelson:
> Do anyone here have more simple and elegant solution ?

I would try to avoid having multiple process write to the same file. It would likely be easier to join those files together after the fact. But if you really must have this feature, then I would recommend using logrotate for compression/rotation instead of logback's built in implementation.

isaki-x:
I agree that using logrotate for the compression is way better (and if this PR goes through, it should probably throw an exception or something if prudent is enabled with logback's compression enabled). Another option would be use a syslog appender and no need to deal with prudent at all.

genz10:
Yeah I think these trade off are not balance, once someone enable prudent mode (if this PR goes through) there's possibility they will suffer another problem, i'll also consider to find other solution instead of enabling prudent mode.

Should we close this PR or continue goes through ?

arteam:
Okay, let's put this PR on hold. We can continue, if there is more interest from the community and clear how to tackle the rotation issue.

joschi:
The PR seems to be stale and hasn't seen any activity in almost 2 years.
I'm closing it for now until someone picks up the work and addresses the remaining issues.

Issue #2140, opened 2017-08-31
Add alternative to ValidationMethod


manuel-hegner:
Hey,
I would like to propose an alternative to the somewhat clunky ValidationMethod included in dropwizard. If this discussion leads to something I would also be willing to implement this new feature.

My proposal is to enable more detailed annotations by allowing users to use validationMethods that look like this:
```java
    @NewValidationMethod
    public void validateSomething(ConstraintValidatorContext context) {}
```
This would allow users a better control over the validation (e.g. better messages and multiple violations in one method) and would no longer interfere with Jackson (requiring no @JsonIgnore).

To make this work every Class that has this kind of validation methods would need to be annotated with a new @HasAnnotationMethods annotation, which would be a real hibernate constraint and would find every new-type validation method and execute them.

A caveat that I see is performance (on the other hand hibernate validation performance is not great as it is). To not use slow reflection access, code generation would be the better alternative. Does dropwizard already use a library for that? I would not like adding new dependencies just for this feature.

What do you think?

nickbabcock:
Based on your description, I'm not in favor of removing `ValidationMethod` as the proposed method doesn't seem to make it irrelevant.

However, I approve of adding the proposed method to complement the current validation method. If this can fit within the current structure of the dropwizard then we can accept it as is. Else if third party dependencies need to be pulled in or if there is a significant performance penalty for those who don't opt-in to the new method, then I'd recommend making another module for encapsulation.

manuel-hegner:
Okay, I started looking at how to implement this. I would need two dependencies to implement this cleanly in dropwizard-validation. The first one is slf4j, which I think would not be a problem. The other one is Javassist to implement the code generation. This is already a transitive dependency of dropwizard-jersey (through jersey and hk2-locator). I am not sure if it would be okay to add this dependency explicitely to dropwizard-validation. What are your thoughts?

isaki-x:
There are alternative solutions to the code generation problem. For example [ReflectASM](https://github.com/EsotericSoftware/reflectasm) has proven to be pretty performant (it is used by Kryo, which makes sense since the same authors maintain Kryo and ReflectASM).

manuel-hegner:
Yeah, Reflectasm is definitely the better choice for this kind of problem. I just thought that with javassist all of this could be done within java-validation. I don't really see anyone using this if it is in a new project.

evnm:
[...]

PR #2150, opened 2017-09-13
SelfValidation annotation as an alternative to ValidationMethod

manuel-hegner:
This pull request solves my feature request issue #2140 .
It adds the two Annoations @SelfValidation and @SelfValidating. An example of how to use it:
```java
@SelfValidating
public class Example {
    private int a = 3;
    private int b = 6;

    @SelfValidation
    public void validOnlyIfEven(ViolationCollector col) {
        if(a+b % 2 != 0)
            col.addViolation("a+b must be an even value. At the moment it is "+a+"+"+b+"="+(a+b));
    }
}
```

What do you think of it? If this solution is okay with you (especially the names of annotations) I would write a paragraph for the documentation.

### DIFF
--- a/dropwizard-benchmarks/pom.xml
+++ b/dropwizard-benchmarks/pom.xml
@@ -10,7 +10,7 @@
     <properties>
         <!-- see https://issues.apache.org/jira/browse/MCOMPILER-235 -->
         <maven.compiler.useIncrementalCompilation>false</maven.compiler.useIncrementalCompilation>
-        <jmh.version>1.15</jmh.version>
+        <jmh.version>1.19</jmh.version>
         <!-- Skip deployment of this sub-module -->
         <maven.source.skip>true</maven.source.skip>
         <maven.javadoc.skip>true</maven.javadoc.skip>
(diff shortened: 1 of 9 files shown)


manuel-hegner:
Ich think performance should not be a problem with how the code turned out. Since I use code generation instead of reflection the runtime overhead should be only once per class. But it is never a bad idea to have a benchmark and I look into writing one.

manuel-hegner:
Finally got around to writing a benchmark. As predicted the code generation overhead is negligible. @SelfValidating is still slower than ValidationMethod because Hibernate needs to reevaluate the given error string (templating engine) every time instead of only once per annotation, but it is still much faster than some other hibernate validation annotations.

Here the results:
```
SelfValidatingBenchmark.selfValidating    avgt    8  3243,249 ± 41,973  ns/op
SelfValidatingBenchmark.validationMethod  avgt    8   568,652 ± 10,330  ns/op
```

nickbabcock:
It just occurred to me. Would you use a custom `ExceptionMapper<JerseyViolationException>` for your `@SelfValidation` example, due to the caching nature of the default exception mapper? (Don't change your PR, just curious, as this got me thinking)

arteam:
Wow, that looks pretty great! @manuel-hegner Thanks for the contribution!

Very interesting to see how `@SelfValidation` performs in comparasion to `@ValidationMethod`. I think it maybe very useful for users who like put validation in domain objects and concerned about performance.

Issue #2153, opened 2017-09-15
equality on different units


daniv:
`Assertion.assertThat( reason, CoreMatchers.equalTo( Size.terabytes( 0 ) ), Size.parse( "0M" ) )
java.lang.AssertionError: parse( '0M' ) matches :<0 terabytes>
Expected: <0 terabytes>
     but: was <0 megabytes>`

`Assertion.assertThat( reason, CoreMatchers.equalTo( Size.bytes( 0 ) ), Size.terabytes( 0 ) );
java.lang.AssertionError: Size.bytes( 0 ) matches :<0 terabytes>
Expected: <0 bytes>
     but: was <0 terabytes>
`

looks like 0 ( zero ) no equals to 0 zero.........

lutzhorn:
Yes, because `Size.equals` is

````java

    @Override
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }
        if ((obj == null) || (getClass() != obj.getClass())) {
            return false;
        }
        final Size size = (Size) obj;
        return (count == size.count) && (unit == size.unit);
    }
````

Both the count and the unit must be equal.

This does not look like a bug to me.

**Comment by nickbabcock** [comment]
500 bytes can't be represented as a whole megabyte, so one would have to round to 0 or 1.

**Comment by isaki-x** [comment]
There may be a nice way to fix this; I do something similar for working with some of my own date and time constructs (for example, 60 seconds and 1 minute are equal in spite of using different units of measurement). I'll take a look at this and see if the logic I utilized is applicable here.

**Comment by isaki-x** [comment]
Actually, this class is actually bugged. It violates the Comparable contract. Two objects may show compareTo == 0 while equals returns false.

**Comment by isaki-x** [comment]
I have added a unit test to my local code base that fails (which should succeed).

```Java
    @Test
    public void verifyComparableContract() {
        final Size kb = Size.kilobytes(1024L);
        final Size bytes = Size.bytes(kb.toBytes());
        
        assertThat(bytes.compareTo(kb)).isEqualTo(0);
        assertThat(kb.compareTo(bytes)).isEqualTo(0);
        
        // If comparator == 0, then the following must be true
        assertThat(bytes.equals(kb)).isTrue();
        assertThat(kb.equals(bytes)).isTrue();
    }
```

I will submit a PR with a fix.

joschi:
Closed via #2155.

PR #2155, opened 2017-09-16
Fixed violation of Comparable contract in Size

isaki-x:
Added unit test to verify comparable contract.

This addresses the bug alluded to in #2153. It turns out that the reporter of the issue uncovered a violation of a rule set forth when implementing the Comparable interface.

### DIFF
--- a/dropwizard-util/src/main/java/io/dropwizard/util/Size.java
+++ b/dropwizard-util/src/main/java/io/dropwizard/util/Size.java
@@ -120,7 +120,7 @@ public class Size implements Comparable<Size> {
             return false;
         }
         final Size size = (Size) obj;
-        return (count == size.count) && (unit == size.unit);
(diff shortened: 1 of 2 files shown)


isaki-x:
This is the rule (which apparently is just a very strong suggestion) from the Java 8 documentation for Comparable:

```It is strongly recommended (though not required) that natural orderings be consistent with equals. This is so because sorted sets (and sorted maps) without explicit comparators behave "strangely" when they are used with elements (or keys) whose natural ordering is inconsistent with equals. In particular, such a sorted set (or sorted map) violates the general contract for set (or map), which is defined in terms of the equals method.```

nickbabcock:
From that angle, it does look like the current behavior is flawed. Good find. To me, either the `compareTo`  or `equals` needs to change, and I'm leaning towards @isaki-x's solution.

PR #2160, opened 2017-09-21
Update documentation for checkConnectionOnConnect

bryanlb:
This should bring the documentation in line with the actual default value, as on line [389](https://github.com/dropwizard/dropwizard/pull/2160/files#diff-e5126a75b90a2bf1ce64a60b923c8eadL389).


rferreira:
Great find!

PR #2161, opened 2017-09-22
Support fromString in FuzzyEnumParamConverter

reftel:
This is a continuation of Stefan Zhelyazkov 's work in PR #1987

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
@@ -5,6 +5,8 @@ import com.google.common.base.Strings;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import io.dropwizard.util.Enums;
 
+import java.lang.reflect.Method;
+import java.lang.reflect.InvocationTargetException;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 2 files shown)


reftel:
Anything I can do to move this one forward? @arteam?

reftel:
@stefan-zh: does this patch still solve your problem?
@BorisNaguet: does this version of https://github.com/dropwizard/dropwizard/pull/1987 address all your comments?

arteam:
I will review the merge request today. Thanks for the reminder!

BorisNaguet:
Hello,
looks good to me. 
My unit tests are also passing with this version.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
```diff
@@ -47,7 +56,35 @@ public T fromString(String value) {
...
+                            String.format("%s is not a valid %s", parameterName, rawType.getSimpleName());
+                        throw new WebApplicationException(getErrorResponse(errMsg));
+                    } catch (IllegalAccessException e) {
+                        final String errMsg =
+                            String.format("Not permitted to call fromString on %s", rawType.getSimpleName());
+                        LOGGER.debug(errMsg, e);
+                        throw new WebApplicationException(getErrorResponse(errMsg));
+                    } catch (InvocationTargetException e) {
+                        if (e.getCause() instanceof WebApplicationException) {
+                            throw (WebApplicationException) e.getCause();
```
stefan-zh: I am only confused as to in what cases does this occur?

arteam:
@reftel Thanks for the contribution! I will make a new 1.1 release shortly.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/FuzzyEnumParamConverterProvider.java
reftel: In the APIs for our applications, we have a standardized format for error messages, including things like extended error codes, which are part of the API documentation. In the `fromString` implementation for an enum that is part of the API (it's only ever referenced by the resource class), we then throw a `WebApplicationException` with the correct extended error code.
The lines here ensures that that exception itself is propagated, and not wrapped, so that the error code etc. that the `fromString` implementation indicated are used in the response.

reftel:
Thank you! =)

PR #2164, opened 2017-09-30
Applying EP transformation using Rascal

jmarcelonunes:
Using Rascal Software retrieved from https://github.com/refactoring-towards-language-evolution/rascal-Java8. EP transformation was applied and the result was the following log:

- Number of files:  786
- Processed Files: 4
- Exported Files:   4
- Total of files changed: 4
- Total of transformations: 4
- Errors: 15


joschi:
@jmarcelonunes Please describe why these changes would improve the code.

Additionally, your changes don't even compile.

isaki-x:
Additionally, while lambda based looping is nice and clean to read, often times it is not as fast as a direct loop implementation (depending on several factors). Just because the language supports new ways of expressing old concepts that doesn't mean that every old construct should be replaced with the newer syntax.

PR #2167, opened 2017-10-03
Add exclusive mode to MinDuration and MaxDuration annotations

### LINKED ISSUE #2130: Change DatasourceFactory.maxWaitForConnection min duration
Currently the value is set to

```java
    @NotNull
    @MinDuration(value = 1, unit = TimeUnit.SECONDS)
    private Duration maxWaitForConnection = Duration.seconds(30);
```
It should be a minimum of 1 MILLISECONDS

if it is set to 0 then it seems to hang indefinitely.

below is a test snippet demonstrating behavior. 

```java
    @Test
    public void databaseIsFailFastIfPoolIsExausted() throws Exception {
        DataSourceFactory dsFactory = objectMapper.readValue(objectMapper.writeValueAsString(config.getDataSourceFactory()), DataSourceFactory.class);
        dsFactory.setMaxSize(2);
        dsFactory.setMinSize(1);

        //We cannot set this in the config because dropwizard does
        //@MinDuration(value = 1, unit = TimeUnit.SECONDS)
        dsFactory.setMaxWaitForConnection(Duration.milliseconds(1));
        ManagedDataSource ds = dsFactory.build(registry, "TESTING-failfast");
        try {
            Connection c1 = ds.getConnection();
            Connection c2 = ds.getConnection();
            Stopwatch sw = Stopwatch.createStarted();
            try {
                ds.getConnection();
                fail("should fail");
            } catch (PoolExhaustedException e) {
                assertThat(sw.elapsed(TimeUnit.SECONDS)).isLessThan(1);
            }
        } finally {
            ds.stop();
        }
    }
```

We would like to fail fast as possible if the connection pool is full as hanging threads is not a good default.

FredDeschenes:
A lot of the configurations with a `@MinDuration` seem to actually need to validate that a duration has been set and is greater than zero, but were stuck to use some kind of arbitrary minimum value (some of them used 1 millisecond, which should be good enough for most cases, but some used 1 second which is pretty high). I don't think Dropwizard should be forcing minimum durations on its configurations, just as no `@MaxDuration` is used anywhere.

I've changed the ones in `DataSourceFactory` as mentioned in the issue, but 6 more remain throughout the project. I wasn't comfortable enough with these configurations to change them myself, but I can update the PR if you feel like they can also be changed to a `@PositiveDuration`.

I'm also not totally sold on the name 'PositiveDuration', but I feel like it's clearer than the 'NonZeroDuration' I proposed in #2130 (which would imply that "-1 nanoseconds" is valid).


evnm:
Thanks for responding to #2130, @FredDeschenes.

Instead of introducing a new annotation, what if we added an boolean `exclusive` element to `MinDuration` (defaulting to `false`) which would control whether the value was treated inclusively or exclusively? This element could feasibly be added to the rest of the min/max validation annotations in Dropwizard as well.

In this case, the usage in `DataSourceFactory` would become

```
@MinDuration(value = 1, unit = TimeUnit.SECONDS, exclusive=true)
private Duration maxWaitForConnection = Duration.seconds(30);
```

FredDeschenes:
Good idea, but that seems like a different issue.

#2130 is about some `@MinDuration` values being set to pretty high values when all we want to validate is that a duration was actually set. We could of course not add the `@PositiveDuration` and just make sure the values in the `@MinDuration`s are set to low enough values.

evnm:
As I understand it, the issue raised in #2130 is that the defined minimum duration is too large. The solution is to lower the default such that any positive real number is valid.

I believe our two solutions be to equivalent—the set of positive real numbers is equal to the set of real numbers with an exclusive minimum of zero.

Does that make sense?

FredDeschenes:
Yeah I guess it does make sense, just might be a little harder to get what's going on when reading the annotation.

Anyway I've added the `exclusive` mode to `MinDuration` and `MaxDuration`, should I modify `MinDuration` usages to be something like `@MinDuration(value = 0, unit = TimeUnit.NANOSECONDS, exclusive = true)`?

evnm:
I think 1 ms is a reasonable default. Do you agree?

I don't want to unilaterally drive the bus here. @nickbabcock, what do you think?

nickbabcock:
Oh man, if I turn my nitpickiness up to 11, I prefer an `inclusive` argument over `exclusive` to avoid the explicit `exclusive=false` double negative. I believe "inclusive" also has fewer definitions than "exclusive", which might ease the impact on readability (as @FredDeschenes pointed out)

The error message becomes shorter

> must be less than or equal (or less than if in 'exclusive' mode) to {value} {unit}

becomes

> must be less than (or equal to in 'inclusive' mode) to {value} {unit}

Ideally the correct operator (`>=` / `>`) would be present in the error message, so the user would know if upper bound is valid instead of the user checking the source to see if inclusive is enabled 🤔 

I'd prefer if this PR went through without touching all the duration annotations to keep potential bikeshedding down. I'm not sure how many annotations are affected, but if it's anymore than a handful then another PR is best.

FredDeschenes:
I switched 'exclusive' to 'inclusive'.

As for displaying the operator, I can either add them to the current text or replace words by them (ex: "greater" -> ">"), but since the message is in an annotation it needs to be constant so we can't change the message depending on the mode.

As for the original issue of the `@MinDuration` annotations, there's around 30 of them but #2130 was specifically for the ones in `DataSourceFactory` so 7. I'll create another PR once this is merged and you decide on what the minimum value should be.

dropwizard-validation/src/main/java/io/dropwizard/validation/MaxDuration.java
```diff
@@ -41,4 +41,9 @@
      * @return unit of the value the element must be higher or equal to
      */
     TimeUnit unit() default TimeUnit.SECONDS;
+
+    /**
+     * @return if the validation should include the set value (default is true).
```
evnm: Perhaps something worded like the following, for this and the other equivalent javadocs.

"True if the validation is to allow values equal to {@code value}. False if the validation is to be exclusive. Defaults to {@code true}."

dropwizard-validation/src/main/java/io/dropwizard/validation/MaxDuration.java
```diff
@@ -26,7 +26,7 @@
 @Documented
 @Constraint(validatedBy = MaxDurationValidator.class)
 public @interface MaxDuration {
-    String message() default "must be less than or equal to {value} {unit}";
+    String message() default "must be less than (or equal to, if in 'inclusive' mode) to {value} {unit}";
```
evnm: Rm that second "to"

dropwizard-validation/src/main/java/io/dropwizard/validation/MinDuration.java
```diff
@@ -26,7 +26,7 @@
 @Documented
 @Constraint(validatedBy = MinDurationValidator.class)
 public @interface MinDuration {
-    String message() default "must be greater than or equal to {value} {unit}";
+    String message() default "must be greater than (or equal to, if in 'inclusive' mode) to {value} {unit}";
```
evnm: ditto

dropwizard-validation/src/test/java/io/dropwizard/validation/DurationValidatorTest.java
```diff
@@ -67,10 +79,12 @@ public void returnsASetOfErrorsForAnObject() throws Exception {
             assertThat(errors)
                     .containsOnly(
                             "outOfRange must be between 10 MINUTES and 30 MINUTES",
-                            "tooBig must be less than or equal to 30 SECONDS",
-                            "tooSmall must be greater than or equal to 30 SECONDS",
-                            "maxDurs[0].<collection element> must be less than or equal to 30 SECONDS",
-                            "minDurs[0].<collection element> must be greater than or equal to 30 SECONDS",
+                            "tooBig must be less than (or equal to, if in 'inclusive' mode) to 30 SECONDS",
```
evnm: ditto, for all of these

FredDeschenes:
I fixed those conflicts this morning in 3b8413d, everything says there's no conflict anymore on my side ("This branch has no conflicts with the base branch"), is it showing you conflicts?

evnm:
Ah, my bad. The "rebase and merge" option was selected and is a lot less forgiving when it comes to merge conflicts. Squash+merge is green, so we're good to go.

FredDeschenes:
Thanks, I'll go ahead and make another PR to actually fix #2130 !

PR #2170, opened 2017-10-09
Fix link to Jackson Annotation in Getting Started

herrphon:
The old link seems to not work any longer...


joschi:
@herrphon Thanks!

PR #2180, opened 2017-10-20
JSON log support for Dropwizard - https://github.com/dropwizard/dropwizard/issues/1451

sabarivasan:
This PR adds JSON log support for Dropwizard. You can see the discussion at https://github.com/dropwizard/dropwizard/issues/1451

It starts off from the following branch and completes support for JSON logs for  ILoggingEvent and IAccessEvent.
https://github.com/dropwizard/dropwizard/compare/master...jamisonhyatt:AddJsonFileAppender?expand=1

### DIFF
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -504,6 +504,16 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>ch.qos.logback.contrib</groupId>
+                <artifactId>logback-json-classic</artifactId>
+                <version>0.1.5</version>
+            </dependency>
+            <dependency>
+                <groupId>ch.qos.logback.contrib</groupId>
+                <artifactId>logback-jackson</artifactId>
+                <version>0.1.5</version>
+            </dependency>
 
             <!-- Test dependencies -->
             <dependency>
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -605,33 +605,33 @@ Console
           discardingThreshold: 0
           timeZone: UTC
           target: stdout
-          logFormat: # TODO
+          logFormat: "%-5p [%d{ISO8601,UTC}] %c: %m%n%rEx"
           filterFactories:
             - type: URI
 
 
-====================== ===========  ===========
-Name                   Default      Description
-====================== ===========  ===========
-type                   REQUIRED     The appender type. Must be ``console``.
-threshold              ALL          The lowest level of events to print to the console.
-queueSize              256          The maximum capacity of the blocking queue.
-discardingThreshold    51           When the blocking queue has only the capacity mentioned in
-                                    discardingThreshold remaining, it will drop events of level TRACE,
-                                    DEBUG and INFO, keeping only events of level WARN and ERROR.
-                                    If no discarding threshold is specified, then a default of queueSize / 5 is used.
-                                    To keep all events, set discardingThreshold to 0.
-timeZone               UTC          The time zone to which event timestamps will be converted.
-                                    To use the system/default time zone, set it to ``system``.
-target                 stdout       The name of the standard stream to which events will be written.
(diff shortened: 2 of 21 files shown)


arteam:
@sabarivasan  Thanks for working on it! I hacked a little bit on your patch, trying to extract everything into own module, so the users don't need to pull additional dependencies if they don't use JSON logging. 
Please see: https://github.com/dropwizard/dropwizard/commit/6d9705bc45a50749cdc7cfac10598c04ce86bea7

Maybe there is something interesting can arise from it, which will be contributed to the upstream.

sabarivasan:
Hi @arteam , will these changes get included into the dropwizard 1.3.0 milestone ? I am not sure what the next steps toward that direction are. We at Cvent would like this feature to be included in a release version of dropwizard. As you can see from the original ticket, there are others who are interested in JSON logging as well.

mattnelson:
Is this proposal blending appenders and format? The appender shouldn't care that the format is json, it should only be concerned of which stream to write the message(file,console,syslog). Which also highlights that the other appender types would require additional implementations. With containerization happening across the industry the recommendation is often to log to console, at a minimum json console support should be introduced as well

But I have a different proposal. What if we enhanced the existing appender `logFormat` property to flex and accept a more complex log format configuration? That would enable new log formats to be implemented without tight coupling to the appender.
```yaml
logging:
  appenders:
    - type: console
       logFormat: "%-5level %logger: %msg%n"
    - type: console
       logFormat:
         type: json
         prettyprint: true
```

The loggers already support this type of overloaded configuration data type. The class logger accepts a level or a more complex object.
http://www.dropwizard.io/1.2.0/docs/manual/configuration.html#logging
```yaml
loggers:
    "io.dropwizard": INFO
    "org.hibernate.SQL":
      level: DEBUG
      additive: false
```

sabarivasan:
@mattnelson, the proposal is to decouple the LayoutFactory from a pattern-based layout. JSON is not a pattern-based layout. The discussion in https://github.com/dropwizard/dropwizard/issues/1451 has the context.

Regarding your other point, given that console logging is usually human-read while JSON is processed by log processing utilities (which read them from files), I am not sure I see the value in adding support for JSON console logging.

mattnelson:
> the proposal is to decouple the LayoutFactory from a pattern-based layout. JSON is not a pattern-based layout. The discussion in #1451 has the context.

It may not be a "layout" in the traditional logback sense, but there are numerous flags to indicate which items should be included with the json layout which is not that different from the traditional pattern based layout. The difference is that logback has codified the ability to flex the format.
https://github.com/qos-ch/logback-contrib/blob/master/json/classic/src/main/java/ch/qos/logback/contrib/json/classic/JsonLayout.java

> Regarding your other point, given that console logging is usually human-read while JSON is processed by log processing utilities (which read them from files), I am not sure I see the value in adding support for JSON console logging.

This is a broad assumption to make. The reason I brought it up is because that is exactly what I am doing, console logging in a docker container which is then forwarded to a log aggregator. A human never looks at the direct console logs in this scenario.

brentryan:
@arteam Is this going to get pulled into 1.3 release?  We're waiting on this internally before we upgrade to latest dropwizard version.  We require JSON support before proceeding forward.  We'd rather not do this off of a fork so any timeline that can be provided here would be useful.

dropwizard-bom/pom.xml
```diff
@@ -504,6 +504,16 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>ch.qos.logback.contrib</groupId>
+                <artifactId>logback-json-classic</artifactId>
+                <version>0.1.5</version>
```
jplock: Can this reference a property version? Same comment below

arteam:
Yes, I'm working on it. I like @mattnelson's proposal about a generic `logFormat`. Let me come up with a solution.

arteam:
Closed via #2232

PR #2186, opened 2017-10-26
Use InstrumentedQueuedThreadPool for admin endpoint

### LINKED ISSUE #2179: Add possibility to instrument admin threadpool just like application threadpool
Application threadpool is `com.codahale.metrics.jetty9.InstrumentedQueuedThreadPool` and it gives some valuable metrics. But sometimes in microservice-oriented infrastructure you have quite a number of different things calling your admin endpoints. In that case developers may want to configure `adminMinThreads` and `adminMaxThreads`, but to do that properly, they need information about number of non-idle threads. So would be nice to use `com.codahale.metrics.jetty9.InstrumentedQueuedThreadPool` for admin threadpool too.

patrox:
This will allow to monitor the usage of the thread pool which is used to handle the requests landing at admin endpoint.

Closes #2179.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/server/DefaultServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/server/DefaultServerFactory.java
@@ -1,6 +1,7 @@
 package io.dropwizard.server;
 
 import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.jetty9.InstrumentedQueuedThreadPool;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.google.common.base.MoreObjects;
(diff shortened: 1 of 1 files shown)


patrox:
@joschi Can you please have a look ?

isaki-x:
Is there a way to make this optional/configurable?

patrox:
To be honest I didn't consider it, as I haven't seen any downsides of using it, but I guess it could be made configurable by some additional boolean flag, i.e. `isAdminThreadPoolMonitored`.

arteam:
I don't think it's needed. The overhead of instrumenting should be fairly small to warrant adding an additional configuration option.

arteam:
The biggest concern for me is the increase of reported metrics by default for a Dropwizard app.

arteam:
thread-pools

arteam:
But, my opinion, 4 additional metrics should not be a big problem. They are quite useful metrics, especially for applications with a lot of admin tasks.

arteam:
@patrox Thank you for the contribution!

PR #2189, opened 2017-10-26
Close PackageNamesScanner in ScanningHibernateBundle

### LINKED ISSUE #2188: Potential resource leak in ScanningHibernateBundle
While playing a bit with [SonarQube](https://www.sonarqube.org/), I noticed (thanks to it), that there is [one place](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/ScanningHibernateBundle.java#L43) where an instance of [`PackageNamesScanner`](https://github.com/jersey/jersey/blob/master/core-server/src/main/java/org/glassfish/jersey/server/internal/scanning/PackageNamesScanner.java) class (which implements `AutoClosable`) is not closed anywhere, which will result in not closing the resource which is referenced by this class [`compositeResourceFinder`](https://github.com/jersey/jersey/blob/master/core-server/src/main/java/org/glassfish/jersey/server/internal/scanning/PackageNamesScanner.java#L181).

A simple fix for that would be to create this instance of `PackageNamesScanner` within a try-with-resources block, so it will get automagically closed.

If this will be confirmed, I'm happy to provide a PR addressing this issue.

patrox:
... in order to close the referenced resource.

Resolves #2188.


arteam:
Thanks! Nice to catch such errors with static analyzer tools.

PR #2192, opened 2017-10-31
Create PULL_REQUEST_TEMPLATE.md

clickthisnick:
###### Explanation About What Code Achieves:

  - Adds a pull request template so code review are easier for maintainers

###### Steps To Test:

  - When you create a pull request this template will show up automatically

###### TODOs:

  - None

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+###### Explanation About What Code Achieves:
+<!-- Please explain why this code is necessary / what it does -->
+  - Explain
+
+###### Steps To Test:
+<!-- What would someone do to be able to see the effects of your code? -->
+  - Steps
+
+###### TODOs:
+<!-- Is there any tests or logic that isn't in the pr that you want the reviewer to know about? -->
+  - None
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -2,9 +2,14 @@
 <!-- Please explain why this code is necessary / what it does -->
   - Explain
 
-###### Steps To Test:
-<!-- What would someone do to be able to see the effects of your code? -->
-  - Steps
+###### Problem:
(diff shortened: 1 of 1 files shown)


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+###### Explanation About What Code Achieves:
```
evnm: What's the reasoning behind using H6? Seems a little too deep to me. [H4s](https://gist.github.com/evnm/0fc670b2f7c55962a0c21254191a1585) have a similar weight and would allow folks to still use sub-headers.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
...
+<!-- Please explain why this code is necessary / what it does -->
```
evnm: I tend to like [the format used by Twitter's Finagle project](https://github.com/twitter/finagle/blob/develop/.github/PULL_REQUEST_TEMPLATE.md), which first frames the problem that the change sets about to solve and then describes the chosen solution. This has the benefit of encouraging contributors to think concretely about what problem they're solving, rather than making changes for change's sake.

In this model, the solution section could include a description of the testing done.

PULL_REQUEST_TEMPLATE.md
clickthisnick: H6 are shrunk and gray in github
H4 are normal sized and bold

I generally use H6 so that people can tell the difference between the template and the text that the user has entered

clickthisnick:
@evnm I believe I addressed your comments, please take a look :)

PULL_REQUEST_TEMPLATE.md
clickthisnick: Update to use that model~ Thanks!

arteam:
I am going to merge it and then amend the format to keep only problem, solution, and result sections as proposed by @evnm. Thank you!

PR #2197, opened 2017-11-10
Build and run Dropwizard on JDK9

arteam:
This change adds a Java profile which allows to compile and run the tests of Dropwizard on JDK9. TravisCI support it on their machines, so we can run the tests automatically for changes and pull requests. 

Most of the Dropwizard dependencies are compatible with JDK9 with exceptions of:
* [jadira](https://github.com/JadiraOrg/jadira). I created a pull request there: https://github.com/JadiraOrg/jadira/pull/66, but there's no response. To workaround it, I [forked](https://github.com/arteam/jadira) Jadira, applied the patch and published it via [jitpack](https://jitpack.io/#arteam/jadira). 
* findbugs (I think we are better of with error-prone anyway)
* The JDBI module had a bug with converting Instants (in JDK8 Instant had a millesecond resultion, in JDK9 - nanosecond)

So, this change makes sure that we are testing that Dropwizard apps can be successfuly run on JDK9 with the `java.xml.bind` module enabled.


isaki-x:
Does this mean that future DropWizard releases will be compiled with the JDK9 compiler or is this purely for compatibility testing?

arteam:
No, Dropwizard will stay compatible with JDK8. It will be compiled with the JDK9 compiler only if it's set as the default in the system.

isaki-x:
Awesome, thank you for the information!

pom.xml
```diff
@@ -579,23 +609,6 @@
                 </configuration>
             </plugin>
             <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>findbugs-maven-plugin</artifactId>
-                <configuration>
-                    <effort>Max</effort>
-                    <threshold>Default</threshold>
-                    <xmlOutput>true</xmlOutput>
-                    <excludeFilterFile>${basedir}/../findbugs-exclude.xml</excludeFilterFile>
```
jplock: should we also remove the `findbugs-exclude.xml` file too?

arteam: Yes, that makes sence!

PR #2203, opened 2017-11-11
Use version 3.2.5 of dropwizard metrics

georgespalding:
Don't see any reason not to stay up to date on metrics version in 1.1.x


joschi:
@georgespalding We don't plan to do another Dropwizard 1.1.x release as Dropwizard 1.2.0 has already been released.

georgespalding:
I see. Planning to move to dw 1.2 anyway, so I'll close this, then.

PR #2204, opened 2017-11-11
Enable NullAway errorprone checks

jplock:
https://github.com/uber/NullAway

PR #2205, opened 2017-11-12
Add `@Nullable` annotations or initialize nonnull fields

arteam:
We now run a static analyzer during our build (#2204) which helps us to catch NPE during compilation. To make it work, we should annotate actual nullable fields with the `@Nullable` annotation, otherwise the analyzer will always consider such fields as non-null by default and will report violations when a null value is assigned to them.

### DIFF
--- a/dropwizard-assets/src/test/java/io/dropwizard/assets/AssetsBundleTest.java
+++ b/dropwizard-assets/src/test/java/io/dropwizard/assets/AssetsBundleTest.java
@@ -24,8 +24,8 @@ public class AssetsBundleTest {
     private final ServletEnvironment servletEnvironment = mock(ServletEnvironment.class);
     private final Environment environment = mock(Environment.class);
 
-    private AssetServlet servlet;
-    private String servletPath;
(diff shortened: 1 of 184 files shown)


jplock:
If we uncomment the line I commented in the pom, we can have Travis enforce this too. Do you think we should do that?

arteam:
Yes, I went through all null check violation reports (even in the tests) and fixed them. I've changed the reporting level to ERROR. Now the build will stop if the compiler finds a violation.

dropwizard-benchmarks/pom.xml
```diff
@@ -23,6 +23,21 @@
     <artifactId>dropwizard-benchmarks</artifactId>
     <name>Dropwizard Benchmarks</name>
 
+    <build>
+        <plugins>
```
jplock: The nullaway configuration also allows you to exclude packages there too if we want to excludes these packages in the main pom

arteam: Yes, it will be probably a better approach.

PR #2210, opened 2017-11-14
Set database config '@MinDuration' to 1 millisecond

### LINKED ISSUE #2130: Change DatasourceFactory.maxWaitForConnection min duration

FredDeschenes:
Fixes #2130, more info in #2167.

This actually makes it to be >= 1 of any TimeUnit (the unit specified in the annotation is irrelevant when the value is 0), I just set them to milliseconds to be clearer that it can be a low value.

### DIFF
--- a/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
+++ b/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
@@ -373,22 +373,22 @@ public class DataSourceFactory implements PooledDataSourceFactory {
 
     private boolean logValidationErrors = false;
 
-    @MinDuration(value = 1, unit = TimeUnit.SECONDS)
+    @MinDuration(value = 0, unit = TimeUnit.MILLISECONDS, inclusive = false)
     @Nullable
     private Duration maxConnectionAge;
 
     @NotNull
-    @MinDuration(value = 1, unit = TimeUnit.SECONDS)
+    @MinDuration(value = 0, unit = TimeUnit.MILLISECONDS, inclusive = false)
     private Duration maxWaitForConnection = Duration.seconds(30);
 
     @NotNull
-    @MinDuration(value = 1, unit = TimeUnit.SECONDS)
+    @MinDuration(value = 0, unit = TimeUnit.MILLISECONDS, inclusive = false)
     private Duration minIdleTime = Duration.minutes(1);
 
     @NotNull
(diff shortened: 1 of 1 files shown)


dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
```diff
@@ -403,19 +403,19 @@ public int get() {
     private boolean autoCommentsEnabled = true;
 
     @NotNull
-    @MinDuration(1)
+    @MinDuration(value = 0, unit = TimeUnit.MILLISECONDS, inclusive = false)
     private Duration evictionInterval = Duration.seconds(5);
 
     @NotNull
-    @MinDuration(value = 50, unit = TimeUnit.MILLISECONDS)
+    @MinDuration(value = 0, unit = TimeUnit.MILLISECONDS, inclusive = false)
```
evnm: I don't think we want to let this one be that low. #2051 reduced the minimum from one second to 50ms and I don't see a reason to deviate from this.

PR #2219, opened 2017-11-24
Fixed typo. Lister -> Listener

GuiSim:
###### Problem:
Slight typo in private method name

###### Solution:
Fixed the typo

###### Result:
Not much. I just stumbled upon that and thought I'd fix it ;)


arteam:
Thanks for your attentiveness!

PR #2220, opened 2017-11-24
Add possibility to possibility to extend `DropwizardApacheConnector`

arteam:
###### Problem:
Some users want to have more fine-grained Jersey-Apache HTTP client integration. For example, more elaborate behavior during retrying of with the chunked encoding enabled. 

###### Solution:
Make the `getEntity` method of `DropwizardApacheConnector` protected and provide a new method for creating an own implementation of `DropwizardApacheConnector` in `JerseyClientBuilder`.

###### Result:
Powerful users are more happy with Dropwizard and don't need to copy-past classes to customize them.


jplock:
@arteam should we merge this into `master` as well?

arteam:
Applied to master as e40ed6c

PR #2221, opened 2017-11-24
Don't shut down asynchronous executor in Jersey

### LINKED ISSUE #2218: DropwizardExecutorProvider not shareable but it is shared
Hello,

I have the following behavior in dw 1.1.4 and 1.2.0. I create a JerseyClient and bind it to hk2 via:

```
final Client client = new JerseyClientBuilder(environment)
    .using(configuration.getJerseyClientConfiguration())
    .build(smartMobileApplication.getName());
bind(client).to(Client.class);
```

Eventually I perform some async calls:

```
    client.target("http://example")
                .path("/example")
                .register(HttpAuthenticationFeature.basic(
                        configurationCredentials.getLogin(),
                        configurationCredentials.getPass()
                ))
                .request()
                .async()
                .post(Entity.entity(entity.getPayload(), MediaType.APPLICATION_JSON), callback);
    }
```

After some calls the Executor service is shutdown by hk2 leading to a `RejectedExecutionException`.

If found that removing the line
```
                .register(HttpAuthenticationFeature.basic(
                        configurationCredentials.getLogin(),
                        configurationCredentials.getPass()
                ))
```

from the code above fixes this issue.

I assume that the "register" call will alter the underlying `ClientConfig`, which leads to a copy from the jersey runtime (i believe they have a copy on write). The jersey runtime will cleanup these copys and the `DropwizardExecutorProvider` will shutdown. In the javadoc for the `DropwizardExecutorProvider` it is stated that the `DropwizardExecutorProvider` should not be shared. However, if think it is pretty intransparent that the `DropwizardExecutorProvider` is shared just by adding a `register` call to the `WebTarget`.

I am pretty new to dropwizard and java in general. Is this a bug or expected behaviour? Or do I misuse the client api?

Kind regards
Timo

arteam:
###### Problem:
If a user customized a Jersey client, Jersey creates a new copy of the instance. Jersey has a mechanism for detecting leaks, and Jetty will invoke the `dispose` method on the old client. Unfortunately,
during a copy the executor is shared. Therefore, the user will not be able to perform async requests anymore.

###### Solution:
Remove the code for shutting down the executor from `DropwizardExecutorProvider`.

###### Result:
Users can use the Jersey client to perform async requests with customizing.  Fixes #2218.


jplock:
@arteam should we merge this into `master` as well?

arteam:
Applied to master as 9cf696f

PR #2222, opened 2017-11-29
Return a 400 when a request body can't be decompressed

ehrmann:
###### Problem:
Currently, making a request with `content-encoding` set to `gzip` (or `deflate`) with content that's either not gziped, corrupt, or truncated results in a 500 because the exception from `GZipInputStream` is never handled.

###### Solution:
Gzip request decompression is handled by `BiDiGzipHandler`. This change decorates the GZipInputStream it creates so that `ZipException` and `EORException` are handled, and the exception is propagated as a Jersey BadMessageException. Similar changes were made to handle
deflate.

###### Result:
Making a request like

```
echo plaintext | curl -X POST -d @- http://localhost:8080/endpoint
```
will return a 400 rather than a 500. Additionally, `ZipExceptions` won't be logged as unhandled exceptions.

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/BiDiGzipHandler.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/BiDiGzipHandler.java
@@ -12,6 +12,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
 import javax.servlet.http.HttpServletResponse;
 import java.io.BufferedReader;
+import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -22,6 +23,7 @@ import java.util.Enumeration;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.Inflater;
 import java.util.zip.InflaterInputStream;
+import java.util.zip.ZipException;
 
 /**
  * An extension of {@link GzipHandler} which decompresses gzip- and deflate-encoded request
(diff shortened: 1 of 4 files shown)


ehrmann:
Could someone look at codeclimate? It's getting to the point where it's making the test less readable.

arteam:
> Could someone look at codeclimate? It's getting to the point where it's making the test less readable.

It looks like you are all set! I agree that sometimes codeclimate is too aggresive in its analysys.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/BiDiGzipHandler.java
```diff
@@ -87,18 +89,33 @@ private Inflater buildInflater() {
...
-        return new WrappedServletRequest(request, input);
+        try {
+            final InflaterInputStream input = new InflaterInputStream(request.getInputStream(), inflater, inputBufferSize) {
+                @Override
+                public void close() throws IOException {
+                    super.close();
+                    localInflater.set(inflater);
+                }
+            };
+            InputStream exceptionHandlingInput = new ZipExceptionHandlingInputStream(input, DEFLATE);
```
arteam: I think we could just inline the stream to the constructor without declaring a variable.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/BiDiGzipHandler.java
```diff
@@ -87,18 +89,33 @@ private Inflater buildInflater() {
...
+            throw ZipExceptionHandlingInputStream.buildBadDataException(DEFLATE, e);
+        } catch (EOFException e) {
+            throw ZipExceptionHandlingInputStream.buildPrematureEofException(DEFLATE, e);
+        }
     }
 
     private WrappedServletRequest wrapGzippedRequest(HttpServletRequest request) throws IOException {
-        return new WrappedServletRequest(request, new GZIPInputStream(request.getInputStream(), inputBufferSize));
+        try {
+            GZIPInputStream input = new GZIPInputStream(request.getInputStream(), inputBufferSize);
```
arteam: Nit: please declare `input` as `final` for consistency with the rest of the codebase.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/BiDiGzipHandler.java
```diff
@@ -87,18 +89,33 @@ private Inflater buildInflater() {
...
+            InputStream exceptionHandlingInput = new ZipExceptionHandlingInputStream(input, GZIP);
```
arteam: It seems this variable could be inlined, too.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/ZipExceptionHandlingInputStream.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.InputStream;
+import java.util.zip.ZipException;
+
+import org.eclipse.jetty.http.BadMessageException;
+
+/**
+ * This InputStream is used to decorate a GZIPInputStream or InflaterInputStream, intercept decompression
+ * exceptions, and rethrow them as BadMessageExceptions
+ */
+class ZipExceptionHandlingInputStream extends FilterInputStream {
```
arteam: I would prefer to extend this class from `InputStream` and declare the `InputStrem` as a field instead of extending from `FilterInputStream`. Using `FilterInputStream` doesn't seem to provide any advantage considering we override all the methods anyway to handle exceptions.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/ZipExceptionHandlingInputStream.java
```diff
@@ -0,0 +1,97 @@
...
+    ZipExceptionHandlingInputStream(InputStream in, String format) {
+        super(in);
+        this.format = format;
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+        try {
+            return super.read(b, off, len);
+        } catch (ZipException e) {
```
arteam: I think it makes sense  to add a generic method for handling an `IOException` instead of repeat the logic in every method. Something like:
```java
static BadMessageException handleException(String format, IOException e) throws IOException {
        if (e instanceof ZipException) {
            return buildBadDataException(format, e);
        } else if (e instanceof EOFException) {
            return buildPrematureEofException(format, e);
        } else {
            throw e;
        }
}
```

dropwizard-jetty/src/test/java/io/dropwizard/jetty/ZipExceptionHandlingInputStreamTest.java
```diff
@@ -0,0 +1,138 @@
...
+        public void testReset() throws Exception {
+            doNothing().when(delegate).reset();
+            in.reset();
+            verify(delegate).reset();
+            verifyNoMoreInteractions(delegate);
+        }
+    }
+
+    @RunWith(Parameterized.class)
+    public static class ExceptionTest {
```
arteam: Could you also add a test case that an `IOException` from the delegate is propagated?

dropwizard-jetty/src/main/java/io/dropwizard/jetty/ZipExceptionHandlingInputStream.java
ehrmann: It did cover `mark()` and `markSupported()`, but reimplementing them was pretty easy.

arteam:

arteam:
Cherry-picked to the 1.2.x branch as b70492524190201d394a2c16941098d1054305eb

PR #2228, opened 2017-12-13
Update dependencies

arteam:
###### Problem:
Some dependencies have outdated versions.

###### Solution:
Upgrade them to the latest ones.

###### Result:
Dropwizard users are less affect by bugs in the dependencies.


nickbabcock:
Upgrading Hibernate to 5.4.2.Final fixes #1990. I'm not sure if another 1.2.x release will occur, but this would be a good candidate to include if present.

PR #2230, opened 2017-12-13
Add an example of using Dropwizard with Conscrypt

arteam:
###### Problem:
Jetty 9.4.8 supports native SSL via [Conscrypt](https://github.com/google/conscrypt), but it's not clear how to use it in Dropwizard.

###### Solution:
Add an integration test and upgrade the SSL documentation.

###### Result:
Dropwizard users can see how they can take advantage of native SSL.

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -322,6 +322,49 @@ connection similar to pre-Dropwizard 1.0.
 
 .. _man-core-bootstrapping:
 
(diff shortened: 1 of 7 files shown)


GautamGupta:
It's probably also a good idea to link to that part of the documentation in the [configuration section](https://github.com/dropwizard/dropwizard/blob/60505741ab054a18ace8cac7e874db19b2a8e9b4/docs/source/manual/configuration.rst#http2-over-tls) (or to update it).

joschi:
@GautamGupta Good idea! Would you mind creating a pull request for it?

GautamGupta:
Sure thing, here you go: #2305

Issue #2195, opened 2017-11-07
Early end of file results in 500 instead of 400


scho:
From time to time, we see some 500s in our dropwizard applications when looking at the metric `io.dropwizard.jetty.MutableServletContextHandler.5xx`. Looking at the logs, doesn't show any errors, also the `io.dropwizard.jersey.errors.LoggingExceptionMapper` is not hit.

Then we found out, that at the same time, these mysterious 500s happen, also this warning shows up in the logs:
```
WARN  [2017-11-06 16:21:57,507] [] org.eclipse.jetty.servlet.ServletHandler:
! org.eclipse.jetty.io.EofException: Early EOF
! at org.eclipse.jetty.server.HttpInput$3.noContent(HttpInput.java:877)
! at org.eclipse.jetty.server.HttpInput.read(HttpInput.java:189)
! at org.eclipse.jetty.server.HttpInput.read(HttpInput.java:141)
! at org.glassfish.jersey.message.internal.EntityInputStream.isEmpty(EntityInputStream.java:200)
! ... 46 common frames omitted
! Causing: javax.ws.rs.ProcessingException: org.eclipse.jetty.io.EofException: Early EOF
! at org.glassfish.jersey.message.internal.EntityInputStream.isEmpty(EntityInputStream.java:217)
! at org.glassfish.jersey.message.internal.InboundMessageContext.hasEntity(InboundMessageContext.java:772)
! at org.glassfish.jersey.servlet.WebComponent.filterFormParameters(WebComponent.java:670)
! at org.glassfish.jersey.servlet.WebComponent.initContainerRequest(WebComponent.java:520)
! at org.glassfish.jersey.servlet.WebComponent.serviceImpl(WebComponent.java:471)
! ... 42 common frames omitted
! Causing: javax.servlet.ServletException: javax.ws.rs.ProcessingException: org.eclipse.jetty.io.EofException: Early EOF
! at org.glassfish.jersey.servlet.WebComponent.serviceImpl(WebComponent.java:489)
! at org.glassfish.jersey.servlet.WebComponent.service(WebComponent.java:427)
! at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:388)
! at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:341)
! at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:228)
[...]

PR #2235, opened 2017-12-20
Add e2e test for eof exception mapper

nickbabcock:
###### Problem:
There is no end to end (e2e) test that ensures that a exception mapper registered for Jetty's EofException (ex. `ExceptionMapper<EofException>` / `EarlyEofExceptionMapper`) is triggered when an `EofException` occurs. This is in effort to track down the cause of #2195

###### Solution:
Add e2e test that simulates a client disconnecting before sending all the data

###### Result:
The exception mapper is hit, so there may be another case that is triggering #2195, but this test is still good.


dropwizard-e2e/src/main/java/com/example/app1/App1.java
```diff
@@ -8,13 +8,16 @@
...
 import io.dropwizard.views.ViewBundle;
+import org.eclipse.jetty.io.EofException;
 import org.glassfish.jersey.spi.ExtendedExceptionMapper;
 
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 
 public class App1 extends Application<Configuration> {
+    public boolean wasEofExceptionHit = false;
```
arteam: Shouldn't the variable be volatile? It's written in a Jetty's app thread and read in a test executor thread.

nickbabcock: Ah yes, I have to keep reminding myself [about the correct usage](https://www.ibm.com/developerworks/java/library/j-jtp06197/). Since this is an example of a status flag, volatile should be used here. I have updated the PR

PR #2236, opened 2017-12-27
add console command

javasoze:
###### Problem:

Currently to start a Dropwizard application, you do: ``` java -jar server app.jar```, this in turn reads the config.yml and setups all the environment niceties.

There are cases I would like to not run the application in server mode, while still enjoy the config and environment bootstrapping.

###### Solution:

Add a new command type: ```console```, so you can do: ```java -jar console myapp.jar```

###### Result:

No changes in current behavior. Just a new command type.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/cli/ConsoleCommand.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/cli/ConsoleCommand.java
@@ -1,0 +1,67 @@
+package io.dropwizard.cli;
+
+import org.eclipse.jetty.util.component.LifeCycle;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.dropwizard.Application;
+import io.dropwizard.Configuration;
+import io.dropwizard.cli.EnvironmentCommand;
+import io.dropwizard.setup.Environment;
+import net.sourceforge.argparse4j.inf.Namespace;
+
+public class ConsoleCommand<C extends Configuration> extends EnvironmentCommand<C> {
+
+    private static final Logger logger = LoggerFactory.getLogger(ConsoleCommand.class);
+
+    private final Application<C> application;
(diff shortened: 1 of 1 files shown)


evnm:
Hello John. I'm not sure that I understand the use case. Use of the word "console" makes me think that the command will drop you into some sort of interactive console, which this change does not provide.

Can you extrapolate on the cases in which you want to parse configuration and start the server's lifecycle without starting the server itself? I wonder if there's an alternative way your problem(s) could be solved.

javasoze:
Hey Evan, perhaps ```console``` is not an ideal name...

My use case is that I have a command line application (not a server), e.g. standard ```public static void main()```. But I find it very useful to leverage some of the functionality that Dropwizard provides, e.g. configuration management, encapsulation of logging etc. So by creating this command, I can just run my program inside ```run()``` with access to a parsed configuration and other goodies that come with Dropwizard.

I understand it is not a standard usecase.

nickbabcock:
I believe we've had this discussion before in a past issue / PR -- I tried searching for a couple mins, but couldn't find it. I can't remember what we concluded 🤔

joschi:
> My use case is that I have a command line application (not a server)

You can already have that with [`Command`](http://www.dropwizard.io/1.2.2/dropwizard-core/apidocs/io/dropwizard/cli/Command.html) or [`ConfiguredCommand<T>`](http://www.dropwizard.io/1.2.2/dropwizard-core/apidocs/io/dropwizard/cli/ConfiguredCommand.html) respectively:
http://www.dropwizard.io/1.2.2/docs/manual/core.html#commands

evnm:
Thanks for putting the change together, but it sounds to me that this is best left within your application rather than the framework. As @joschi points out, the existing `ConfiguredCommand` class should already do most of the heavy lifting involved in running a non-daemonized program that reads a config file.

Issue #1632, opened 2016-07-12
PolymorphicAuthDynamicFeature Usage/Improvement


psamsotha:
[`PolymorphicAuthDynamicFeature`][2] was added in an earlier 1.0 RC, and it seems to cover use cases in [#1579][1], but I'm looking at the code, and there seems to be a flaw in the feature. In the Javadoc it states

>Note that this feature only pertains to _authentication_. To enable authorization, use `AuthDynamicFeature`.

This sounds like the author would like us to use the `PolymorphicAuthDynamicFeature` in concert _with_ `AuthDynamicFeature`. This doesn't really make much sense. The `AuthDynamicFeature` registers a filter that handles both authentication _and_ authorization. It is not used _only_ for authorization. Using both features would end up registering the filters twice, and having them called twice. 

And also how are we expected to register `AuthDynamicFeature`. Say we have two filters for polymorphic feature

```           
final BasicCredentialAuthFilter filterOne = new BasicCredentialAuthFilter.Builder<PrincipalOne>()
        .setAuthenticator(authenticatorOne)
        .buildAuthFilter();
            
final BasicCredentialAuthFilter filterTwo = new BasicCredentialAuthFilter.Builder<PrincipalTwo>()
        .setAuthenticator(authenticatorTwo)
        .buildAuthFilter();
            
register(new PolymorphicAuthDynamicFeature<>(ImmutableMap.of(
         PrincipalOne.class, filterOne,
         PrincipalTwo.class, filterTwo
)));
```

What do we do now with the `AuthDynamicFeature`? It only takes one filter. Are we supposed to register _two_ `AuthDynamicFeature`s? There's isn't any documentation on intended usage. 

@evnm It looks like you were the author of this. Did you have an intended usage for this being used along side authorization, or did your javadoc comment mean that if you need authorization don't use this, instead use the other auth feature?

With the current implementation of the `PolymorphicAuthDynamicFeature`, was there something stopping you from simply using the same registration algorithm used in the `AuthDymamicFeature`, where the filter registration handle both authentication and authorization.

Without testing anything out, it seems to me that we _can_ use the poly feature for both authentication and authorization. We just need to use the same registration algo used in the `AuthDynamicFeature`. I'll check out the possibility later, but I'd like to hear what @evnm has to say. Maybe I am missing something.

[1]: https://github.com/dropwizard/dropwizard/issues/1579

[...]

PR #2238, opened 2018-01-05
Add documentation for `@Priority` for polymorphic auth

FredDeschenes:
###### Problem:
Any custom `AuthFilter` needs to have the `@Priority` annotation set
otherwise authorization will run before the auth filters and will
use the default security context set by Jersey.

I was having issues with authorization always allowing my `@RolesAllowed` annotated methods to be called and my auth filter not running (was basing my code on #1632, which has a test case that worked perfectly fine).

Not sure if this is specific to polymorphic auth as I've never used `@RolesAllowed` with custom `AuthFilter`s before. Should we also add documentation about adding the annotation to all custom `AuthFilter`s (there doesn't seem to be anything about this at the moment)?

###### Solution:
Added documentation.

###### Result:
Clearer documentation.

### DIFF
--- a/docs/source/manual/auth.rst
+++ b/docs/source/manual/auth.rst
@@ -377,6 +377,9 @@ Note that with the above example, only *authentication* is configured. If you al
 
 * Make sure you add ``Authorizers`` when you build your ``AuthFilters``.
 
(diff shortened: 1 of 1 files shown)


jplock:
Thanks!

PR #2239, opened 2018-01-05
Remove metrics-ganglia

arteam:
###### Problem:
We can't update Dropwizard to Metrics 4.0.* with the Ganglia module.

###### Solution:
Metrics 4.0 doesn't support Ganglia, and AFAIK the project itself not very actively developed. I had a look at the stats from Maven Central, and the module is barely downloaded from Central (around 100 downloads per month). For example, the metrics-graphite module has around 10,000
downloads per month. I think this warrants removing maintenance of this module. If anyone it still using Ganglia with Dropwizard, it should be relatively easy to copy `GangliaReporterFactory` to own
project.

###### Result:
We can update Dropwizard to Metrics 4.0.*


mattnelson:
What is the target release for metrics-4.0? We attempted to pull it in but ran into non-passive package renames with the jmx/jvm modules. As well as a type change on Timer.Context.

arteam:
The type change has been fixed in 4.0.1 - https://github.com/dropwizard/metrics/issues/1243.
The JMX reporter was extracted from the core and now is distributed in the `metrics-jmx` package. Practically it should be only just an import change.

jplock:
Should we put out another v1.2.x release where we deprecate these classes before v1.3.0?

arteam:
I think we can just remove the module. I don't think people who use Ganglia are checking out Dropwizard release candidate builds.

PR #2243, opened 2018-01-16
Add 'dropwizard-jdbi3' module

FredDeschenes:
###### Problem:
jDBI3 as been released for a couple of months and with the recent update to Metrics 4.X Dropwizard can now use it without losing any of the features from jDBI2. This should of course not be a forced upgrade and allow users to decide which jDBI version they wish to use in their project. jDBI now also supports most of the custom mappers that were in `dropwizard-jdbi` out of the box or in a separate module, removing some code from Dropwizard.

###### Solution:
Create a `dropwizard-jdbi3` module alongside the existing `dropwizard-jdbi` module.

###### Result:
Added a basic implementation of `dropwizard-jdbi3`, without any of the previous mappers or documentation.

TODO:

- [x] Add documentation (how should this be approached? a new page for the module? have documentation for both modules in the current page?)
- [x] Check if any existing mappers need to be implemented for jDBI3 (jDBI has a module for Guava which seems to cover every mapper Dropwizard had, Joda time is effectively deprecated, we might need the OptionalInt/Long/Double/etc)
- [x] Add unit tests for `JdbiFactory`


isaki-x:
For `JdbiFactory`, can you please follow the configure paradigm in `io.dropwizard.jdbi.DBIFactory`? I had implemented that to inject custom configuration in my own subclass of DBIFactory. Thanks.

FredDeschenes:
@isaki-x : Will do.

FredDeschenes:
I validated that all Dropwizard-specific mappers are now not required as they are either supported out of the box by jDBI or one of its plugins (Joda or Guava).

arteam:
I think it's better to leave them out. For users who uses Guava and JodaTime, the modules easy to include to their projects.

joschi:
@FredDeschenes I'd vote for including these plugins by default because Dropwizard usually comes with "batteries included" for the libraries it's using itself (i. e. JodaTime and Guava).

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/JdbiFactory.java
```diff
@@ -0,0 +1,118 @@
...
+     * @param name
+     * @return A fully configured {@link Jdbi} object
+     */
+    public Jdbi build(Environment environment,
+                      PooledDataSourceFactory configuration,
+                      ManagedDataSource dataSource,
+                      String name) {
+
+        // Create the instance
+        final Jdbi jdbi = this.newInstance(dataSource);
```
arteam: `this` seems to be redundant, because there's no superclass for `JdbiFactory` which has a method with the same name.

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/NamePrependingTemplateEngine.java
```diff
@@ -0,0 +1,46 @@
...
+    public NamePrependingTemplateEngine(TemplateEngine originalEngine) {
+        this.originalEngine = Objects.requireNonNull(originalEngine, "'originalEngine' cannot be null'");
+    }
+
+    @Override
+    public String render(String template, StatementContext ctx) {
+        final ExtensionMethod extensionMethod = ctx.getExtensionMethod();
+        final String originalResult = originalEngine.render(template, ctx);
+
+        if ((extensionMethod != null)
```
arteam: I think it would be more clear to make a check that the extension method _doesn't_ exist and exit early.
```java
if (extensionMethod == null || extensionMethod.getType() == null || extensionMethod.getMethod() == null) {
            return originalResult;
}
```

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/NamePrependingTemplateEngine.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.core.statement.TemplateEngine;
+
+import java.util.Objects;
+
+/**
+ * Jdbi {@link TemplateEngine} that prepends the SQLObject's type and method name in front of every
+ * statement sent to the database.
+ */
+public class NamePrependingTemplateEngine implements TemplateEngine {
```
arteam: It would be nice to have a simple unit test for `NamePrependingTemplateEngine` to verify that it actually inserts comments to the query.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/bundles/JDBIExceptionsBundleTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Test
+    public void test() {
+        Environment environment = mock(Environment.class);
+        JerseyEnvironment jerseyEnvironment = mock(JerseyEnvironment.class);
+        when(environment.jersey()).thenReturn(jerseyEnvironment);
+
+        new JdbiExceptionsBundle().run(environment);
+
+        verify(jerseyEnvironment, times(1)).register(isA(LoggingSQLExceptionMapper.class));
```
arteam: `times(1)` seems to be redundant, because `verify` by default checks that the call has happened only once.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+public class JdbiHealthCheckTest {
+
+    @Test
+    public void testItTimesOutProperly() throws Exception {
+        String validationQuery = "select 1";
+        Jdbi jdbi = mock(Jdbi.class);
+        Handle handle = mock(Handle.class);
+        when(jdbi.open()).thenReturn(handle);
+        when(jdbi.withHandle(Mockito.any())).thenCallRealMethod();
+        Mockito.doAnswer(invocation -> {
```
arteam: The `handle.execute` method returns `int` instead of `void`, so we can use the `when().doAnswer` form.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+            try {
+                TimeUnit.SECONDS.sleep(10);
```
arteam: The `answer` method can throw exceptions, so we can avoid catching the exception.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+        }).when(handle).execute(validationQuery);
+
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        JdbiHealthCheck jdbiHealthCheck = new JdbiHealthCheck(executorService,
+                Duration.milliseconds(5),
+                jdbi,
+                validationQuery);
+        HealthCheck.Result result = jdbiHealthCheck.check();
+        executorService.shutdown();
+        assertThat("is unhealthy", false, is(result.isHealthy()));
```
arteam: Could you please use an AssertJ assertions for consistency with rest of the codebase?

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/jersey/LoggingJDBIExceptionMapper.java
```diff
@@ -0,0 +1,35 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.ext.Provider;
+import java.sql.SQLException;
+
+/**
+ * Iterates through a {@link JdbiException}'s cause if it's a {@link SQLException} otherwise log as normal.
+ */
+@Provider
+public class LoggingJdbiExceptionMapper extends LoggingExceptionMapper<JdbiException> {
```
arteam: I guess the file also should be named `LoggingJdbiExceptionMapper` instead `LoggingJDBIExceptionMapper`

arteam:
> I'd vote for including these plugins by default because Dropwizard usually comes with "batteries included" for the libraries it's using itself (i. e. JodaTime and Guava).

On second thought, I think it makes sense. The modules are small and it will make transition from `dropwizard-jdbi2` a whole lot simpler.

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/JdbiFactory.java
FredDeschenes: Removed 'this'

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/NamePrependingTemplateEngine.java
FredDeschenes: Turns out `ExtensionMethod` has `requireNonNull` for type and method so I'll remove those unnecessary checks as well.

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/jersey/LoggingJDBIExceptionMapper.java
FredDeschenes: Yeah the joys of file paths on Windows with Git... I'll rename those properly.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
FredDeschenes: It's doing this for the `sleep` in the invocation, I don't think I can change this (there's no `when(handle...).doAnswer`, I'd have to do `.thenReturn` which wouldn't sleep). No sure why it's sleeping anyway so I can probably change this. What do you think?

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
FredDeschenes: Yeah this was copied from the DBIHealthcheck in the jdbi2 module, I'll convert them.

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/NamePrependingTemplateEngine.java
FredDeschenes: Test added

FredDeschenes:
I added the JodaTime and Guava plugins and also added the jDBI3-bom to Dropwizard's bom, but I'm not super used to Maven, just make sure I haven't messed up the dependencies to everything else ;)

FredDeschenes:
Concerning the documentation, I thought about having a new 'JDBI3' page with updated documentation content (although it's going to be pretty similar) and flagging the 'jdbi2' page to tell new projects that they should use the new 'jdbi3' module but existing projects can update at their discretion. Sounds good?

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Test
+    public void testNoTimeoutReturnsHealthy() throws Exception {
+        String validationQuery = "select 1";
+        Jdbi jdbi = mock(Jdbi.class);
+        Handle handle = mock(Handle.class);
+        when(jdbi.open()).thenReturn(handle);
+        when(jdbi.withHandle(Mockito.any())).thenCallRealMethod();
+
+        when(handle.execute(validationQuery)).thenAnswer((Answer<Integer>) invocation -> null);
```
arteam: Because the method returns an integer, I think we could just use `thenReturn(0)`

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class JdbiHealthCheckTest {
+
+    @Test
+    public void testNoTimeoutReturnsHealthy() throws Exception {
+        String validationQuery = "select 1";
```
arteam: It seems that we could extract preparing of `executorService`, `validationQuery`, `jdbi` and `handle` to a `setup` method and share it between the tests.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+        when(handle.execute(validationQuery)).thenAnswer((Answer<Integer>) invocation -> null);
+
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        JdbiHealthCheck jdbiHealthCheck = new JdbiHealthCheck(executorService,
+            Duration.milliseconds(5),
+            jdbi,
+            validationQuery);
+        HealthCheck.Result result = jdbiHealthCheck.check();
+        executorService.shutdown();
```
arteam: I think it would be nice to shutdown the executor in a `tearDown` method, so even if the test fails, we correctly close resources.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
```diff
@@ -0,0 +1,65 @@
...
+        Jdbi jdbi = mock(Jdbi.class);
+        Handle handle = mock(Handle.class);
+        when(jdbi.open()).thenReturn(handle);
+        when(jdbi.withHandle(Mockito.any())).thenCallRealMethod();
+
+        when(handle.execute(validationQuery)).thenAnswer((Answer<Integer>) invocation -> null);
+
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        JdbiHealthCheck jdbiHealthCheck = new JdbiHealthCheck(executorService,
+            Duration.milliseconds(5),
```
arteam: I would increase timeout from 5ms to 100ms. 5ms is too little to get a result from a new thread. The test may fail with a timeout while the new thread is being scheduled to run.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/JdbiHealthCheckTest.java
```diff
@@ -0,0 +1,65 @@
...
+        when(jdbi.withHandle(Mockito.any())).thenCallRealMethod();
+
+        when(handle.execute(validationQuery)).thenAnswer((Answer<Integer>) invocation -> {
+            TimeUnit.SECONDS.sleep(10);
+            return null;
+        });
+
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        JdbiHealthCheck jdbiHealthCheck = new JdbiHealthCheck(executorService,
+                Duration.milliseconds(5),
```
arteam: I would increase the timeout, because a test with 5ms timeout can fail sporadically, not because we actually wait for a result from the database.

dropwizard-jdbi3/src/test/java/io/dropwizard/jdbi3/NamePrependingTemplateEngineTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+        assertThat(result).isEqualTo(originalRendered);
+    }
+
+    @Test
+    public void testPrependsCorrectName() throws NoSuchMethodException {
+        final TemplateEngine original = mock(TemplateEngine.class);
+        final StatementContext ctx = mock(StatementContext.class);
+        final String template = UUID.randomUUID().toString();
+        final String originalRendered = UUID.randomUUID().toString();
```
arteam: I think we could extract preparing of the variables to a `setup` method and share it between the tests.

dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/JdbiHealthCheck.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public JdbiHealthCheck(Jdbi jdbi, String validationQuery) {
+        this(MoreExecutors.newDirectExecutorService(), Duration.seconds(0), jdbi, validationQuery);
+    }
+
+    @Override
+    protected Result check() throws Exception {
+        return timeBoundHealthCheck.check(() ->
+            jdbi.withHandle((handle) -> {
```
arteam: On second thought, I think it would be better to use ` try (Handle handle = jdbi.open())` instead of `withHandle`. The biggest advantages of lambda-expressions is that we could use a short-form for one-line calls or even reference a method. Here we use two commands, so basically the both expressions require the same amount of code. At the same time the former version is easier to test.

arteam:
> Concerning the documentation, I thought about having a new 'JDBI3' page with updated documentation content (although it's going to be pretty similar) and flagging the 'jdbi2' page to tell new projects that they should use the new 'jdbi3' module but existing projects can update at their discretion. Sounds good?

Sounds good to me!

arteam:
I think we also need to add an integration test test to verify that Dropwizard and Jdbi3 together and you can write database access code with `JdbiFactory`. For example, for Jdbi2 we've got `JDBITest`. In my personal project, I wrote a couple tests which showcase Jdbi features, you can get an idea:
* https://github.com/arteam/dropwizard-jdbi3/blob/master/src/test/java/com/github/arteam/jdbi3/JdbiTest.java
* https://github.com/arteam/dropwizard-jdbi3/blob/master/src/test/java/com/github/arteam/jdbi3/GameDao.java
* https://github.com/arteam/dropwizard-jdbi3/blob/master/src/test/resources/com/github/arteam/jdbi3/GameDao.sql.stg

isaki-x:
One note on the configure implementation, you can drop the `PooledDataSourceFactory` argument; the was something I hadn't wanted there in the first place but had no other simple way of getting some of the things required to configure the legacy mappers/translators. There is no harm in leaving it, it just was something added as a shortcut and due to design requirements. Thanks!

arteam:
@FredDeschenes Thank you very much for the contribution, great work! I'm going to make a RC shortly.

isaki-x:
This is super exciting, I look forward to migrating when this is ready. Thanks for the work @FredDeschenes !

Issue #2245, opened 2018-01-18
Custom messages in constraint validators are incorrectly cached


peterklijn:
Hi,

It seems DropWizard caches the message of validators but does not correctly invalidate cache. Take the following example:

```java
public class SomeService extends Application<SomeServiceConfiguration> {
  
  public static void main(final String[] args) throws Exception {
    new SomeService().run(args);
  }

  @Override
  public void run(final SomeServiceConfiguration configuration, final Environment environment) throws Exception {
    environment.jersey().register(new SomeResource());
  }
}
```

```java
@Path("/something")
@Produces(MediaType.APPLICATION_JSON)
@Consumes(MediaType.APPLICATION_JSON)
public class SomeResource {

  public SomeResource() {}

  @PUT
  @Path("/")
  public Something something(@NotNull @Valid Something something) {
    return something;
  }
}
```

```java
import javax.validation.constraints.Min;

public class Something {

  public int getNumber() {
    return number;
  }

  public void setNumber(final int number) {
    this.number = number;
  }

  @Min(value = 42, message = "The value ${validatedValue} is less then {value}, you are wrong.")
  public int number;
}
```

Now sending something like `{ "number": 1 }` will great me with a 422 (unprocessable entity) with the error: 
```json
{
    "errors": [
        "number The value 1 is less then 42, you are wrong."
    ]
}
```

Awesome so far, but, when sending another body to this endpoint, for example `{ "number": 2 }` it will keep greeting me with the old message (containing the number 1).

It seems to be comparable to [this issue](https://github.com/dropwizard/dropwizard/issues/2090), but I fail to see how this is not a bug?

I'm using Dropwizard 1.2.1, but have detected this problem over a year ago in older versions of DropWizard as well.

peterklijn:
Btw, this is also the case with custom validators, like so:

```java
import javax.validation.ConstraintValidator;
import javax.validation.ConstraintValidatorContext;

import org.joda.time.DateTime;

public class CustomComponentValidator implements ConstraintValidator<ValidCustomComponent, CustomComponent> {

  @Override
  public void initialize(final ValidCustomComponent constraintAnnotation) {}

  @Override
  public boolean isValid(final CustomComponent value, final ConstraintValidatorContext context) {
    final DateTime now = DateTime.now();
    context.disableDefaultConstraintViolation();
    context
        .buildConstraintViolationWithTemplate("The time is ==> " + now.toLocalTime())
        .addConstraintViolation();

    return false;
[...]

Issue #2090, opened 2017-06-27
Due to caching of constraint violation messages wrong message is being displayed.


imochurad:
I have a bean, that has a field annotated with several validation annotations.
```@NotEmpty```
```@NotBlankElements```
One of them is a standard annotation, one is custom built. 
The problem is that if in the first call to API I place request that causes first validation to fail, it's constraint violation message gets cached for 1 hour time. Then I place request with slightly modified payload that causes second annotation to fail. But I get exactly the same constraint validation message. 
I would like these messages to exactly reflect what part of payload was wrong.

`ConstraintMessage.getMessage(violation, invocable);`
This is the place where I obtain bad error message. `violation` object itself is holding proper message.

**Comment by nickbabcock** [comment]
What version of dropwizard are you running? If it is not 1.0 or 1.1, can you try either of these versions?

I created a test and everything worked as expected:

```java
public class MultiRepresentation {
    @NotNull
    @MaxSize(value = 10)
    @JsonProperty
    private Size size;

    public Size getSize() {
        return size;
    }
}
```

The `MaxSize` annotation is a custom annotation like yours.

- When `size` is null, the `@NotNull` triggers
- When `size` is more than 10 bytes, `MaxSize` triggers

imochurad:
My version of DW is 1.0

nickbabcock:
Ok, I'll need a either a test case that demonstrates this issue or a more fleshed out example.

rtti:
I have similar issue. 

```
@StringEnumeration(enumClass = ComparisonOperator.class)
    private String operator;
```

```
@Override
    public boolean isValid(String value, ConstraintValidatorContext context) {
        if (!AVAILABLE_ENUM_NAMES.contains(value)) {
            String message = String.format("%s is invalid. Valid values are: %s", value, AVAILABLE_ENUM_NAMES);
            context.disableDefaultConstraintViolation();
            context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
            return false;
        }
        return true;
    }
```

I always get the same error message that was generated at the first time, regardless of new constraint message value.

Please take a look at ConstraintMessage.MESSAGES_CACHE. Probably it invalidates cache in wrong way

nickbabcock:
Ah, I see, it's a custom annotation. Adding the invalid value to the violation message would defeat all caching attempts. I'm torn 😕 

Does this mean you in favor of removing the cache?

imochurad:
[...]

PR #2246, opened 2018-01-18
Constraint validation caches error prefixes

nickbabcock:
Previously, once a validation constraint was evaluated the entire string
was cached. The impetus for a cache is that reflection is used to
determine if we need to prefix an error message with additional context
(eg. "query parameter x ..."). Repeatedly calculating this prefix is
wasteful -- thus the cache was born.

The biggest issue with this cache is that while it worked with all
Hibernate Validation default message templates, it failed for message
templates that contained the invalid value (eg. "query parameter x
({validatedValue}) is greater than {value}"), as the same error would be
returned regardless of validatedValue. It would also fail for custom
constraint validators that didn't return the same message everytime.

The fix is to switch the cache from caching the entire error message to
just the error prefix (eg. "query parameter x", "return value", no
prefix).

This has no effect on performance and will allow for more
custom validation messages in Dropwizard.

Closes https://github.com/dropwizard/dropwizard/issues/2245
Ref https://github.com/dropwizard/dropwizard/issues/2090
CC'ing users who expressed interest: @imochurad @rtti @peterklijn

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
@@ -24,7 +24,7 @@ import java.util.concurrent.TimeUnit;
 
 public class ConstraintMessage {
 
-    private static final Cache<Pair<Path, ? extends ConstraintDescriptor<?>>, String> MESSAGES_CACHE =
+    private static final Cache<Pair<Path, ? extends ConstraintDescriptor<?>>, String> PREFIX_CACHE =
             CacheBuilder.newBuilder()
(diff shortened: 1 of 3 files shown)


arteam:
LGTM!

arteam:
Applied to 1.2.x as 76ebbb0e04d44ccdda12f6076eac0932f85c0a80

nickbabcock:
Hey @arteam, I realize now that the PR title "Constraint validation caches error prefixes" may not give dropwizard users an apt description of what changes for them. Maybe something along the lines of "Allow dynamic constraint validation messages" should be in the release notes / new title of this PR.

arteam:
Sounds good. I've changed the release notes for 1.3.0-rc4 and the name of the issue.

PR #2247, opened 2018-01-19
Add `dropwizard-jdbi3` to bom

FredDeschenes:
Oops >_>

###### Problem:
I dun goofed and forgot to add `dropwizard-jdbi3` to the project's bom

###### Solution:
Added it to the bom

###### Result:
It's now in the bom


FredDeschenes:
Thanks for the quick merge @joschi :P

PR #2248, opened 2018-01-19
Support dynamic current session to allow separate read and write connections

sam-berry:
###### Problem:

I am working on a project which utilizes Dropwizard Hibernate. My current mission is to distribute database operations across two endpoints, one for reads and another for writes. DW Hibernate is great because it provides the UnitOfWork framework which allows for multiple bundles to be configured and for those bundles to be specified on an individual basis by using `UnitOfWork#value`. This accomplishes exactly what I need - I will setup a bundle for reading, a bundle for writing, then on my read-only resources throw a `@UnitOfWork(“read-only”)` and then the resources that require writing `@UnitOfWork(“write”)`.

Currently, UnitOfWorkAspect will determine the appropriate session factory based on the current UnitOfWork (see [UnitOfWork:62](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java#L62)), which is great. The problem is, the session factory and session are then set privately inside of the aspect and made unavailable to other components. This is particularly a problem inside of [AbstractDAO](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java), where the session factory is required as an instance parameter. This presents limitations in client code outside of Dropwizard, such as in event listeners that perform database operations.

In summary, I propose two things inside of this PR:
1. The context determined inside of UnitOfWorkAspect be made publicly available to other Dropwizard and non-Dropwizard client components
2. AbstractDAO be modified (a breaking change unfortunately) to not require a SessionFactory for construction, and instead use the UnitOfWork context

###### Solution:

The primary modification includes adding a UnitOfWorkContext class with ThreadLocals to store the context that is determined inside of `UnitOfWorkAspect#beforeStart`. This context is then used to determine the current session AbstractDAO, rather than the injected property that it previously depended on.

###### Result:

The change to UnitOfWorkAspect alone is non-breaking. It simply makes internal context publicly available. The proposed change for AbstractDAO is, unfortunately, breaking. It will be a potentially tedious change but hopefully not complicated. Clients that are only using one bundle currently (I imagine this is the majority case otherwise this issue would have been hit) will have to update their app configuration, or wherever they instantiate their DAOs, to no longer provide a session factory. In my project we have over 100 DAOs and this was a smooth transition that took about 20 minutes.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java
@@ -35,7 +35,7 @@ public class AbstractDAO<E> {
      * @return the current session
      */
     protected Session currentSession() {
-        return UnitOfWorkAspect.getCurrentSession();
+        return UnitOfWorkContext.getCurrentSession();
     }
 
     /**
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java
@@ -5,14 +5,12 @@ import org.hibernate.Criteria;
 import org.hibernate.Hibernate;
 import org.hibernate.HibernateException;
 import org.hibernate.Session;
-import org.hibernate.SessionFactory;
 import org.hibernate.query.Query;
 import org.hibernate.query.internal.AbstractProducedQuery;
 
+import javax.persistence.criteria.CriteriaQuery;
 import java.io.Serializable;
 import java.util.List;
-
-import javax.persistence.criteria.CriteriaQuery;
 
 import static java.util.Objects.requireNonNull;
 
@@ -22,26 +20,22 @@ import static java.util.Objects.requireNonNull;
  * @param <E> the class which this DAO manages
  */
 public class AbstractDAO<E> {
-    private final SessionFactory sessionFactory;
     private final Class<?> entityClass;
 
     /**
-     * Creates a new DAO with a given session provider.
-     *
-     * @param sessionFactory    a session provider
+     * Creates a new DAO for {@code <E>}.
      */
-    public AbstractDAO(SessionFactory sessionFactory) {
-        this.sessionFactory = requireNonNull(sessionFactory);
+    public AbstractDAO() {
         this.entityClass = Generics.getTypeParameter(getClass());
     }
 
     /**
-     * Returns the current {@link Session}.
+     * Returns the current {@link Session} according to {@link UnitOfWorkAspect}.
      *
      * @return the current session
      */
     protected Session currentSession() {
-        return sessionFactory.getCurrentSession();
+        return UnitOfWorkAspect.getCurrentSession();
     }
 
     /**
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
@@ -7,8 +7,8 @@ import org.hibernate.context.internal.ManagedSessionContext;
 
 import javax.annotation.Nullable;
 import java.util.Map;
+import java.util.Optional;
 
-import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -36,6 +36,7 @@ import static java.util.Objects.requireNonNull;
  * </pre>
  */
 public class UnitOfWorkAspect {
+    private static final ThreadLocal<UnitOfWorkContext> CONTEXT = new ThreadLocal<>();
 
     private final Map<String, SessionFactory> sessionFactories;
 
@@ -43,23 +44,12 @@ public class UnitOfWorkAspect {
         this.sessionFactories = sessionFactories;
     }
 
-    // Context variables
-    @Nullable
-    private UnitOfWork unitOfWork;
-
-    @Nullable
-    private Session session;
-
-    @Nullable
-    private SessionFactory sessionFactory;
-
     public void beforeStart(@Nullable UnitOfWork unitOfWork) {
         if (unitOfWork == null) {
             return;
         }
-        this.unitOfWork = unitOfWork;
 
-        sessionFactory = sessionFactories.get(unitOfWork.value());
+        SessionFactory sessionFactory = sessionFactories.get(unitOfWork.value());
         if (sessionFactory == null) {
             // If the user didn't specify the name of a session factory,
             // and we have only one registered, we can assume that it's the right one.
@@ -69,20 +59,22 @@ public class UnitOfWorkAspect {
                 throw new IllegalArgumentException("Unregistered Hibernate bundle: '" + unitOfWork.value() + "'");
             }
         }
-        session = sessionFactory.openSession();
+
+        Session session = sessionFactory.openSession();
         try {
+            setContext(unitOfWork, session);
             configureSession();
-            ManagedSessionContext.bind(session);
             beginTransaction(unitOfWork, session);
         } catch (Throwable th) {
             session.close();
-            session = null;
-            ManagedSessionContext.unbind(sessionFactory);
+            clearContext();
             throw th;
         }
     }
 
     public void afterEnd() {
+        UnitOfWork unitOfWork = getUnitOfWork();
+        Session session = getCurrentSession();
         if (unitOfWork == null || session == null) {
             return;
         }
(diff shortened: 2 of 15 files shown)


dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/AbstractDAO.java
```diff
@@ -22,26 +20,22 @@
...
 
     /**
-     * Returns the current {@link Session}.
+     * Returns the current {@link Session} according to {@link UnitOfWorkAspect}.
      *
      * @return the current session
      */
     protected Session currentSession() {
-        return sessionFactory.getCurrentSession();
+        return UnitOfWorkAspect.getCurrentSession();
```
sam-berry: I believe that this change requires some architectural insight that I do not have; I am not sure if the intent of the dropwizard-hibernate module is to be centered around UnitOfWork. It appears to me that it currently is geared that way and I think that makes a lot of sense - but this change creates a literal coupling that did not previously exist.

There are certainly a number of options that could be explored to offer both a DAO implementation that requires a session factory to be provided and one that is driven off of the context.

But, if the goal is to direct clients toward using UnitOfWork for configurations than perhaps it is best to stick to a single DAO.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
```diff
@@ -36,30 +36,20 @@
  * </pre>
  */
 public class UnitOfWorkAspect {
+    private static final ThreadLocal<UnitOfWorkContext> CONTEXT = new ThreadLocal<>();
```
sam-berry: Another option here would be to create multiple ThreadLocals, one for each piece of state.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
```diff
@@ -155,12 +149,38 @@ private void commitTransaction(UnitOfWork unitOfWork, Session session) {
...
 
-    protected SessionFactory getSessionFactory() {
-        return requireNonNull(sessionFactory);
+    private static void clearContext() {
+        ManagedSessionContext.unbind(getSessionFactory());
+        CONTEXT.remove();
     }
 
+    @Nullable
+    public static SessionFactory getSessionFactory() {
```
sam-berry: It is kind of a drag to force clients to handle the nullable - curious if anyone thinks it would be better to throw an exception here if any of these properties are fetched before they are initialized.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkContext.java
```diff
@@ -0,0 +1,24 @@
...
+        this.sessionFactory = requireNonNull(sessionFactory);
+    }
+
+    public UnitOfWork getUnitOfWork() {
+        return unitOfWork;
+    }
+
+    public SessionFactory getSessionFactory() {
+        return sessionFactory;
+    }
```
sam-berry: I don't think that equals/hashcode/toString make sense here but correct me if I'm wrong.

sam-berry:
In the first iteration here, the session context was set internal to UnitOfWorkAspect and made available through getters. This puts a bit of a burden on testing since it requires instantiating a UnitOfWorkAspect and running it's `beforeStart` function in order to set the current session.

I have updated the ThreadLocals to exist in their own class, so that tests are able to set the session factory directly without dealing with the aspect.

See ed2f3b4 for the change. I've left this commit independent for reviewing purposes but if this ends up being accpeted it probably makes most sense to squash it up into f5a92fb9f8d2d64ff311a28481a63101ae1ba71f.

arteam:
Thank you for the pull request! 

Do I understand correctly that you want to have access to the current session in an event listener? Would it be possible to start a new session/transaction? You could do it even declaratively with `UnitOfWorkAwareProxyFactory`.

Regarding the implementation, I feel that removing `SessionFactory` from `AbstractDAO` would be a big issue preventing integrating this pull request to the upstream. We're trying to maintain backward-compatibility between minor releases, and such a breaking change will need to have a good justification.

sam-berry:
@arteam Thanks for the response. I understand this is a big change so let me try and explain the situation a bit more.

To clarify, I want to allow UnitOfWork to dynamically assign a session factory per request and use that factory for all operations inside of the entire request. The problem with hardcoding session factories is that it may not be safe to use two separate bundles inside of a request. Initially, I thought I could avoid the suggested change by abandoning AbstractDAO for something like the following:

```
public class ReportDAO {
  private final SessionFactory readSessionFactory;
  private final SessionFactory writeSessionFactory;

  public ReportDAO(
    SessionFactory readSessionFactory,
    SessionFactory writeSessionFactory
  ) {
    this.readSessionFactory = readSessionFactory;
    this.writeSessionFactory = writeSessionFactory;
  }

  public List findAllReports() {
    return readSessionFactory.getCurrentSession()
      .createQuery(“…”)
      .list();
  }

  public void createReport(DailyOccupancyReportModel report) {
    writeSessionFactory.getCurrentSession().persist(report);
  }
}
```

Some downsides of delegating session factory like this is that it ignores what UnitOfWork determines, rendering UnitOfWork a mechanism for starting and ending transactions. It also can produce data inconsistencies inside of a single request depending on the nature of the read and write database instances. For clustered environments, read instances may not always be up-to-date. Sometimes there is a brief lag in replication. This puts a bit of a burden on the client code to remain conscious that writing something and immediately reading it is not entirely reliable. Because of this, it seems important to support assigning a session factory to be used globally inside of a single request. So that reads can be directed at master for requests that make modifications.

The issue with AbstractDAO being dependent on SessionFactory at the time of construction is that it prevents the session factory from being dynamically assigned without creating new instances of your DAOs.

I think it is a pretty nice pattern to set an application up with singleton DAOs, and right now that requires defining a session factory at build time. It seems that in order to use AbstractDAO with multiple bundles you need to either instantiate DAOs at runtime inside of requests, or assign and access the session factory statically by thread. 

[HelloWorldApplication.java:83](https://github.com/dropwizard/dropwizard/blob/3165d668dd7fecbd46d6a8540d064202772adf70/dropwizard-example/src/main/java/com/example/helloworld/HelloWorldApplication.java#L83) highlights the issue that I am trying to resolve here. A Hibernate bundle is specific to a single datasource, so in order to support a dynamic datasource based on request (i.e. UnitOfWork) it cannot be hardcoded at build time.

Please let me know if this is more clear or if there are better ways to communicate the problem I'm trying to solve.

sam-berry:
Let me provide an example to try and explain this more thoroughly.

This resource utilizes the UnitOfWork annotation to determine session factory at the request level. `staleResponse` is an example of the kind of error code that can be written when multiple session factories are used in a single request. Specifically, when a read instance is used inside of an unsafe request, e.x. a record is created then immediately retrieved to be returned.

```java
@Path("/reports")
public class ReportResource {
  private final ReportDAO reportDAO;

  public ReportResource(ReportDAO reportDAO) {
    this.reportDAO = reportDAO;
  }

  @GET
  @Valid
  @Timed
  @UnitOfWork("read-only") // session factory with connections to read replica databases
  public ImmutableSet<Report> fetchReports() {
    return reportDAO.fetchReports();
  }

  @POST
  @Valid
  @Timed
  @UnitOfWork("write") // session factory connected to master
  public Report createReport(@NotNull @Valid Report request) {
    return reportDAO.createReport(request);
  }

  @POST
  @Path("/stale-response")
  @Valid
  @Timed
  @UnitOfWork("write")
  public ImmutableSet<Report> staleResponse(@NotNull @Valid Report request) {
    reportDAO.createReport(request);
    // If connected to a read replica there is a chance that this fetch
    // will not include the newly created record. If there is replication
    // lag in the system then only the master instance can be trusted
    // as up-to-date.
    return reportDAO.fetchReports();
  }
}
```

sam-berry:
Another thing that may be useful; this pattern for handling session factory is similar to what Hibernate does internally. From what I can tell `org.hibernate.context.internal.ManagedSessionContext` is what Hibernate uses to track current session internally and it is driven by a ThreadLocal.

A difference to note is that Hibernate supports multiple sessions per thread. Notice [`ManagedSessionContext.java:43`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/context/internal/ManagedSessionContext.java#L43) uses a map and tracks by factory. Seems like a convenient purpose of UnitOfWork is that it handles the work of figuring out which specific datasource to use. It doesn't seem to me that supporting multiple sessions per thread is necessary.

scottaj:
@arteam Did you have any more thoughts on this change? We are trying to use dropwizard and hibernate to connect to a DB cluster that has separate connections for the read master and write replicas. The fact that a DAO instance is statically linked to a single connection despite the fact that you can change the connection in the `@UnitOfWork` annotation makes that challenging.

sam-berry:
Closing this in favor of #2428

PR #2213, opened 2017-11-16
Update Freemarker and support multiple extensions for views (#2196)

FredDeschenes:
Fixes #2196 but breaks compatibility for external `ViewRenderer`s (although it's a simple fix) and configuration files (also simple to change).

- Updated Freemarker to 2.3.27
- Renamed `ViewRenderer::getSuffix` to `getConfigurationKey`
- Changed configuration keys in yaml/json file:
  * `.ftl` -> `freemarker`
  * `.mustache` -> `mustache`
- Use regexes to match renderable templates in `ViewRenderer`s (Freemarker's `.ftlh`/`.ftlx` extensions can now be used, as well as `.ftl.html` and similar paths)

PR #2251, opened 2018-01-22
Enable auto escaping of strings in Freemarker templates

willp-bl:
## Problem
Output was not being auto escaped by Freemarker

## Solution
Freemarker output format was not set [as HTML] which meant that auto escaping was not enabled. With this change it is enabled by default.  

This might affect users who are using formats other than HTML, but 
there should be enough benefit of making this a default.  The current method for getting the output format setting through to the underlying freemarker library is not very pleasant.

Users relying on the existing behaviour may need to set individual
freemarker values as `?unsafe`

## Question
I couldn't find security contact information - is that something you are planning on adding?

### DIFF
--- a/dropwizard-views-freemarker/src/main/java/io/dropwizard/views/freemarker/FreemarkerViewRenderer.java
+++ b/dropwizard-views-freemarker/src/main/java/io/dropwizard/views/freemarker/FreemarkerViewRenderer.java
@@ -4,6 +4,7 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableMap;
+import freemarker.core.HTMLOutputFormat;
 import freemarker.template.Configuration;
 import freemarker.template.DefaultObjectWrapperBuilder;
 import freemarker.template.Template;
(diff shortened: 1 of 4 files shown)


arteam:
Thank you for the pull request! We've made some progress in this area in 1.3.* : with #2213 you can use `ftlh` templates with automatic escaping. Anyway, I think it makes sense to use the HTML escaping for `ftl` templates by default. I will make a maintenance release for the 1.2 branch.

arteam:
> I couldn't find security contact information - is that something you are planning on adding?

Unfortunately, we don't have a security policy like Rails, Django, or Spring and don't provide a email address for secure communications. I guess Dropwizard isn't big enough for that, but having a security policy is definitely is a good practice for a mature project. For the time being, I think the best way to disclose security issues is to post a request to disclose to the [dropwizard-dev](https://groups.google.com/forum/#!forum/dropwizard-dev) mailing list and/or contact on of the [maintainers](http://www.dropwizard.io/1.2.2/team-list.html) privately.

mattnelson:
> I think the best way to disclose security issues is to post a request to disclose to the dropwizard-dev mailing list and/or contact on of the maintainers privately.

Are there email services for dropwizard.io?

joschi:
@mattnelson See http://www.dropwizard.io/1.2.2/docs/about/index.html

* https://groups.google.com/forum/#!forum/dropwizard-user
* https://groups.google.com/forum/#!forum/dropwizard-dev

If there's the need for it, we could probably create a "write-only" Google Group which only the maintainers can read.

arteam:
Unfortunately, Sonatype is down and I can't deploy the release. Will wait when they're up again.

arteam:
1.2.3 has been [released](https://repo1.maven.org/maven2/io/dropwizard/dropwizard-forms/1.2.3/)

PR #2253, opened 2018-01-25
Add a new Maven profile to support Jetty ALPN on jdk8u161

arteam:
###### Problem:
Currently the HTTP2 tests fail on Travis CI, because they use jdk-8u161.

###### Solution:
Add a new profile which links to the corresponding Jetty ALPN library.

###### Result:
The HTTP2 test correctly run on Travis CI.


jplock:
@arteam you added jdk-1.8.0_162, do we need jdk-1.8.0_161 as well?

arteam:
Yes! I've updated the POM.

PR #2254, opened 2018-01-26
Add context path to logged endpoints

afrin216:
###### Problem:
Endpoints logged out take into consideration jersey's URL pattern, but not the context path.
```
GET /endpointA
POST /endpointA
```
Copy pasting this URL gives 404s if you have context-path

###### Solution:
Append context path if present to the end points along with jersey URL pattern

###### Result:
```
GET /context-path/endpointA
POST /context-path/endpointA
```

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -11,8 +11,6 @@ import io.dropwizard.jersey.caching.CacheControlledResponseFeature;
 import io.dropwizard.jersey.params.AbstractParamConverterProvider;
 import io.dropwizard.jersey.sessions.SessionFactoryProvider;
 import io.dropwizard.jersey.validation.FuzzyEnumParamConverterProvider;
-
-import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.server.ServerProperties;
 import org.glassfish.jersey.server.model.Resource;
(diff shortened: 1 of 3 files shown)


nickbabcock:
Just to double check if I understand you right, now everyone that uses

```yaml
server:
  type: simple
  applicationContextPath: /application
```

Will see

```
GET /application/endpointA
```

instead of

```
GET /endpointA
```

afrin216:
@nickbabcock Yes, that's it. So that consumers can copy pasting from the console directly get the context path without looking anywhere else (We have quite a few including myself)

arteam:
I think this change is justified. I can imagine situations when operations staff tries to diagnose a service by the logs and expect that the logged endpoints correspondent to actual endpoints.

arteam:
@afrin216 Thank you for the contribution!

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -177,8 +185,10 @@ String cleanUpPath(String path) {
         private final String rootPath;
         private final Class<?> klass;
 
-        EndpointLogger(String urlPattern, Class<?> klass) {
-            this.rootPath = urlPattern.endsWith("/*") ? urlPattern.substring(0, urlPattern.length() - 1) : urlPattern;
+        EndpointLogger(String contextPath, String urlPattern, Class<?> klass) {
+            final String rootPattern = urlPattern.endsWith("/*") ? urlPattern.substring(0, urlPattern.length() - 1) : urlPattern;
```
Alexey1Gavrilov: Looks like by default the printed path starts with double slashes "//".
If urlPattern == '/*', then rootPattern = '/', and if contextPath == '/', rootPath = '//'

arteam: Thanks for reporting this! I've created #2279 to fix the output.

PR #2256, opened 2018-02-02
Doc fix: h2 conscrypt pom update

nickbabcock:
###### Problem:
Docs needs a little fixin' up when mentioning HTTP2 support and conscrypt.

###### Solution:
Updated docs. Don't need to specify jetty version as it in brought in with our bom. Also one doesn't want it as only a test dependency 😉


arteam:
Nice!

PR #2258, opened 2018-02-02
Fix Graphite reporter documentation

FredDeschenes:
###### Problem:
Incorrect default value for port was used, "transport" was missing when comparing with the actual [configuration class](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporterFactory.java).

###### Solution:
Changed default value for port, added "transport"

###### Result:
Doc now displays the correct configuration options for the Graphite reporter.


arteam:
Thank you.

PR #2260, opened 2018-02-08
Add support for providing a custom layout during logging bootstrap

### LINKED ISSUE #2250: Support bootstrap logging json format?
Pulled in `1.3.0-rc4` to experiment with json logging before the release. Bootstrap logging[1] defaults to a traditional format. Should there be an option to flex bootstrap logging to json? In the use cases that all logs are sent to log processors (splunk, elk, etc..) they may not be configured to recognize the non-json log event.

> 22:10:37.601 [main] INFO org.eclipse.jetty.util.log - Logging initialized @414ms to org.eclipse.jetty.util.log.Slf4jLog
{"level":"INFO","logger":"io.dropwizard.server.DefaultServerFactory","thread":"main","message":"Registering jersey handler with root path prefix: /","timestamp":"2018-01-21T22:10:40.131Z"}

[1] https://github.com/dropwizard/dropwizard/blob/v1.3.0-rc4/dropwizard-logging/src/main/java/io/dropwizard/logging/BootstrapLogging.java

arteam:
###### Problem:
Dropwizard users have the ability to configure their application to use JSON as the logging format. Unfortunately, the default logging bootstrap mechanism which kicks off before Dropwizard logging
configuration, uses the default logging layout. As a result, some internal logging messages (Jetty, Hibernate initialization) are produced in a not recognized format.

###### Solution:
This change add possibility to specify a layout factory during the bootstrap process. Users can override the bootstrap method of `Application` and provide an own layout format. For example,

```java
BootstrapLogging.bootstrap(Level.INFO, new EventJsonLayoutBaseFactory());
```

###### Result:
As a result, the all applications logs will produced in the JSON format.
Resolves #2250

### DIFF
--- a/dropwizard-json-logging/src/test/java/io/dropwizard/logging/json/LayoutIntegrationTests.java
+++ b/dropwizard-json-logging/src/test/java/io/dropwizard/logging/json/LayoutIntegrationTests.java
@@ -40,7 +40,7 @@ import static org.mockito.Mockito.when;
 public class LayoutIntegrationTests {
 
     static {
-        BootstrapLogging.bootstrap(Level.INFO);
+        BootstrapLogging.bootstrap(Level.INFO, new EventJsonLayoutBaseFactory());
     }
 
     private final ObjectMapper objectMapper = Jackson.newObjectMapper();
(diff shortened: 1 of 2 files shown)


li-a:
This also gives us a way to change exceptions back to the standard Logback format (prefixed w/ tab, no more exclamation points) by implementing a custom version of DropwizardLayout. Thanks!

PR #2265, opened 2018-02-09
Refactor tests for JSON layout log factories

arteam:
Make them deserialized directly without using the console appender. This way the coverage tool should be able to recognize that the class is covered during a test run. Let's check whether it's true.


arteam:
Doesn't seem to make any difference. Okay, it's probably some weird case of handling reflection in JaCoCo.

PR #2278, opened 2018-02-19
Run code coverage only if needed

arteam:
###### Problem:
The jacoco agent runs by default alongs with the tests which creates additional overhead. We need only on Travis CI, where we measure code coverage, we don't in another CI or during local development.

###### Solution:
Move the jacoco agent to a separate profile which is invoked by the build script at Travis.

###### Result:
More fast tests during development.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: java
 before_install:
   - ulimit -c unlimited -S
 script:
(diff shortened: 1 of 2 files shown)


arteam:
As an example, the time to run the test suite on AppVeyour went from 8 to 7 minutes.

PR #2290, opened 2018-03-03
Fixed template broken after enabling html escaping of strings in Free…

Silvmike:
…marker templates by default, added integration tests.

See also: https://github.com/dropwizard/dropwizard/pull/2251

Before you could see this in logs trying to render 'view_freemarker' template:

```
Causing: io.dropwizard.views.ViewRenderException: freemarker.core.ParseException: Syntax error in template "com/example/helloworld/views/freemarker/person.ftl" in line 8, column 38:
! Using ?html (legacy escaping) is not allowed when auto-escaping is on with a markup output format (HTML), to avoid double-escaping mistakes.
```

###### Problem:

Seems that auto-escaping for strings in Freemarker templates is enabled by default.
But an example template still contains **?html** , so we've got **ViewRenderException** mentioned in comment to commit, and a http status **500**.

###### Solution:

Fixed template assuming auto-escape enabled.
Added small tests verifying that **/people/{personId}/{view_name}** returns **200 OK**.

### DIFF
--- a/dropwizard-example/src/main/resources/com/example/helloworld/views/freemarker/person.ftl
+++ b/dropwizard-example/src/main/resources/com/example/helloworld/views/freemarker/person.ftl
@@ -5,7 +5,7 @@
     </head>
     <body>
         <!-- calls getPerson().getFullName() and sanitizes it -->
-        <h1>Hello, ${person.fullName?html}!</h1>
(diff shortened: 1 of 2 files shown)


jplock:
Thanks!

PR #2291, opened 2018-03-03
Add a TCP and UDP log appenders to Dropwizard

arteam:
Add support for sending log events to a log management server via TCP or UDP. This is useful for applications that want ship logs to a remote machine for processing/alerting. We recently have added support for formatting messages as JSON and in combination with TCP/UDP logging it can be an interesting solution for a direct log shipping from Dropwizard without 3rd-party modules. 

From the implementation side `TcpSocketAppenderFactory` and `UdpSocketAppenderFactory` are both discoverable and work with generic logging events. That means they can be configured in a YAML config file for general and request logging.
```yaml
logging:
  level: INFO
  appenders:
    - type: tcp
      host: localhost
      port: 24562
```

```yaml
server:
  requestLog:
    appenders:
      - type: tcp
        host: localhost
        port: 24562
```

From the reliability side, the appenders extend `ResilientOutputStreamBase`. That means in case if the connections drops because of network glitch or a DNS failure, it will back off and try to reopen a new connection automatically.

### DIFF
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -33,6 +33,7 @@ v1.3.0: Unreleased
 * Upgrade handling an unknown value in FuzzyEnumModule `#2266 <https://github.com/dropwizard/dropwizard/pull/2266>`_
 * Add support for providing a custom layout during logging bootstrap `#2260 <https://github.com/dropwizard/dropwizard/pull/2260>`_
 * Add context path to logged endpoints `#2254 <https://github.com/dropwizard/dropwizard/pull/2254>`_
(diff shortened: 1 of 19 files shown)


jplock:
This looks really good, one question though. What ends up closing the socket connections? Do we need to worry about that?

arteam:
> This looks really good, one question though. What ends up closing the socket connections? Do we need to worry about that?

Both appenders extend `OutputStreamAppender` which closes the underneath output stream when it's stopped. So, a TCP/UDP socket should be closed during a shut down of Dropwizard/Logback.

jplock:
Can you please update the relnotes file with this too?

arteam:
Sure, will do.

PR #2296, opened 2018-03-06
add OWASP dependency check

1605200517:
This is the simplest way to introduce dependency check; it will generate reports on all module levels where it can extract a dependency tree.


pom.xml
```diff
@@ -665,6 +665,22 @@
                     </execution>
                 </executions>
             </plugin>
+            
+            <plugin>
+	            <groupId>org.owasp</groupId>
+	                <artifactId>dependency-check-maven</artifactId>
```
joschi: Please fix the indentation of this `<plugin>` block and add a `<version>` tag for the dependency-check-maven plugin.

1605200517:
travis job fails as previously with "could not connect to Central" I think it is some bug in the dependency check plugin https://github.com/jeremylong/DependencyCheck/issues/978 

Additionally, I can't understand the problem with org.glassfish.jersey.ext.rx version in dropwizard.client, this was not me  (c)

joschi:
@1605200517 Please revert the unrelated changes.

joschi:
I'm closing this PR now.

If you're still interested in adding the OWASP plugin to Dropwizard, please provide a pull request which:

* builds without error messages (check Travis CI and CircleCI for that)
* doesn't add new compile or build warnings
* contains a minimum of necessary changes (i. e. no unnecessary changes in whitespace or indentation, no unnecessary version changes)

Issue #2127, opened 2017-08-07
Upgrade to hibernate validator 6.0


mattnelson:
The upgrade to hibernate validator 6.0 will not be a drop in replacement as some of the classes in dropwizard-validation need to migrate to javax.transaction interfaces.

https://developer.jboss.org/wiki/HibernateValidatorMigrationGuide#jive_content_id_60x

jplock:
Maybe we do this in a Dropwizard 2.0 that also includes JDBI3?

debraj-manna:
`hibernate-validator` 5.4.x has security vulnerability [CVE-2017-7536](https://nvd.nist.gov/vuln/detail/CVE-2017-7536) . Is there any plan to upgrade it in 1.2.x and 1.3.x also?

joschi:
@debraj-manna Dropwizard 1.2.6 and 1.3.2 are both using Hibernate Validator 5.4.2.Final, which includes a fix for the referenced issue.

https://github.com/hibernate/hibernate-validator/blob/5.4.2.Final/changelog.txt#L8
https://github.com/dropwizard/dropwizard/blob/be9b592857e85dc50a3181b6b5efc23bdbc384d8/dropwizard-bom/pom.xml#L68-L72
https://github.com/dropwizard/dropwizard/blob/f22ae5293ec400e9faafd1cf18944f5ed60b3c0e/dropwizard-bom/pom.xml#L71-L75

mszabo-wikia:
Leaving a note here that this is blocked by https://github.com/eclipse-ee4j/jersey/issues/3871 on Jersey side (as per https://github.com/dropwizard/dropwizard/pull/2298 conversation), so that nobody starts working on it prematurely :)

pkwarren:
We encountered a pretty significant perf issue with Hibernate 5.4.2 (https://hibernate.atlassian.net/browse/HV-1642) that is confirmed resolved in 6.0. Looking forward to HV 6.x in DropWizard.

mattnelson:
The jersey blocked has been merged and is targeted for 2.28.

https://github.com/eclipse-ee4j/jersey/pull/3908

twbecker:
Is this still planned for 2.0? We've been using 6.x for some time in non-dropwizard apps and it's a substantial improvement.

usulkies:
HI, I'm working on the upgrade to hibernate validator 6.0.14.Final along with the new jersey 2.28 in my fork.
Hope to open pull request soon.

[...]

PR #2298, opened 2018-03-07
Upgrade to hibernate validator 6.0 (#2127)

1605200517:
This is a contribution towards the 2.0.0 milestone.

Some tests are now green, but in one case I am not sure what to do.

Namely, a comment in the sources (`OptionalValidatedValueUnwrapperTest`) says,
```

// Dropwizard used to supply its own Java 8 optional validator but since
// Hibernate Validator 5.2, it's built in, so the class was removed but
// the test class stays to ensure behavior remains
```
So I'd consider, if something is built in, not so much needs to be done: but obviously if it still there, something needs to be done to activate it.

My first guess was to use constraint mapping but I must admit I have no clue how to approach this.
https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#section-programmatic-constraint-definition


joschi:
@1605200517 Please revert all unrelated whitespace changes and remove unused code instead of commenting it out.

joschi:
@1605200517 Any news or interest in continuing with this PR?

alex-shpak:
I'm not sure that upgrade to validation api `2.0.1` is possible at the moment, since jersey validation package still depends on `1.1.0.Final` version.
See https://github.com/eclipse-ee4j/jersey/blob/master/pom.xml#L1949

PR #2299, opened 2018-03-08
Allow to use some custom security provider in HTTP client

mgtriffid:
###### Problem:
Dropwizard allows to configure JCE provider for server side, passing `jceProvider` property to Jetty. This allows to use some non-standard provider, like Conscrypt. But there is no such setting on client side.

###### Solution:
Add new field `provider` to `TlsConfiguration` and pass it to Apache HTTP client builder.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -440,7 +440,7 @@ enableCRLDP                      false               Whether or not CRL Distribu
 enableOCSP                       false               Whether or not On-Line Certificate Status Protocol (OCSP) support is enabled.
 maxCertPathLength                (unlimited)         The maximum certification path length.
 ocspResponderUrl                 (none)              The location of the OCSP responder.
-jceProvider                      (none)              The name of the JCE provider to use for cryptographic support.
+jceProvider                      (none)              The name of the JCE provider to use for cryptographic support. See `Oracle documentation <https://docs.oracle.com/javase/8/docs/technotes/guides/security/SunProviders.html>`_ for more information.
 validateCerts                    false               Whether or not to validate TLS certificates before starting. If enabled, Dropwizard
                                                      will refuse to start with expired or otherwise invalid certificates. This option will
                                                      cause unconditional failure in Dropwizard 1.x until a new validation mechanism can be
@@ -1315,6 +1315,7 @@ protocol                     TLSv1.2            The default protocol the client 
                                                 See
                                                 `here <http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#SSLContext>`_ for more information.
 provider                     (none)             The name of the JCE provider to use on client side for cryptographic support (for example, SunJCE, Conscrypt, BC, etc).
+                                                See `Oracle documentation <https://docs.oracle.com/javase/8/docs/technotes/guides/security/SunProviders.html>`_ for more information.
 verifyHostname               true               Whether to verify the hostname of the server against the hostname presented in the server certificate.
 keyStorePath                 (none)             The path to the Java key store which contains the client certificate and private key.
 keyStorePassword             (none)             The password used to access the key store.
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -1294,7 +1294,7 @@ TLS
     httpClient:
       tls:
         protocol: TLSv1.2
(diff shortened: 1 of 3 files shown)


dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
```diff
@@ -71,7 +71,8 @@ private SSLContext buildSslContext() throws SSLInitializationException {
         final SSLContext sslContext;
         try {
             final SSLContextBuilder sslContextBuilder = new SSLContextBuilder();
-            sslContextBuilder.useProtocol(configuration.getProtocol());
+            sslContextBuilder.setProtocol(configuration.getProtocol());
```
mgtriffid: Method `useProtocol` is marked as deprecated, so changing to `setProtocol`.

docs/source/manual/configuration.rst
```diff
@@ -1313,6 +1314,7 @@ Name                         Default            Description
 protocol                     TLSv1.2            The default protocol the client will attempt to use during the SSL Handshake.
                                                 See
                                                 `here <http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#SSLContext>`_ for more information.
+provider                     (none)             The name of the JCE provider to use on client side for cryptographic support.
```
joschi: Could you please add some examples for valid values here, e. g. "BC" for [BouncyCastle](https://www.bouncycastle.org/) or [Conscrypt](https://github.com/google/conscrypt)?

docs/source/manual/configuration.rst
```diff
@@ -1313,6 +1314,7 @@ Name                         Default            Description
 protocol                     TLSv1.2            The default protocol the client will attempt to use during the SSL Handshake.
                                                 See
                                                 `here <http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#SSLContext>`_ for more information.
+provider                     (none)             The name of the JCE provider to use on client side for cryptographic support (for example, SunJCE, Conscrypt, BC, etc).
```
joschi: Are these really valid values for the `SSLContextBuilder#setProvider()` method?

And if so, why is "SUN" being used in the example above?

mgtriffid: `SSLContextBuilder#setProvider()` calls `Security.getProvider()` under the hood.
```java
    public SSLContextBuilder setProvider(final String name) {
        this.provider = Security.getProvider(name);
        return this;
    }
```
Oracle Standard JDK comes with 10 providers, SUN and SunJCE among them. SUN is what `Security.getProviders()[].getName()` returns on my JDK. Index 0 means it has the highest priority. That's why "SUN" in the example above. But I think now that it's better to specify default value as `SunJSSE`, as the provider which will be used in fact with TLSv1.2 (in JDK 8) is SunJSSE.
`BC` and `Conscrypt` will work if they are previously configured using `Security.addProvider()` or `Security.insertProviderAt`.

dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
```diff
@@ -71,7 +71,8 @@ private SSLContext buildSslContext() throws SSLInitializationException {
...
+            sslContextBuilder.setProvider(configuration.getProvider());
```
joschi: Would it make sense to only set the provider if `configuration.getProvider()` wasn't `null`?

docs/source/manual/configuration.rst
joschi: Thanks for the explanation! 👍

Please add a link to https://docs.oracle.com/javase/8/docs/technotes/guides/security/SunProviders.html for a more detailed overview.

dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
mgtriffid: There is no desperate need in it, because `Security.getProvider(null)` doesn't throw NPE, but you're right, better to check. Updated.

mgtriffid:
Thanks for such a quick merge, @joschi! :+1:

PR #2304, opened 2018-03-13
Allow lazy initialization of resources in ResourceTestRule

### LINKED ISSUE #1800: Add method ResourceTestRule#addResource(Supplier<?> lambda)
I have injection f/w which imitates the CDI container using my `InjectionRunner`.
The f/w does not provide the instances of mocked beans at the creation time of `@Rule` `ResourceTestRule` because it is technically not possible in JUnit.
Simply update the `apply()` method in the way that the resources are taken from suppliers at the beginning of the method. This means we can lazily fetch the resources to the rule.
```
        public Builder addResource(Supplier<?> resource) {
            singletons.add(resource); -> change to Set of Suppliers
            return this;
        }
```

joschi:
Instead of eagerly initializing resource instances added to `ResourceTestRule` or `Resource`, we now allow using a `Supplier<T>` which can be used to lazily initialize the resources or providers used in `ResourceTestRule`.

Additionally, this change set adds an adapter for `MockitoJUnit` which allows to use a Mockito JUnit Rule in `RuleChain` for defining an execution order for rules.

Closes #1800


joschi:
> I don't like bring Mockito as a compile dependency again, but Maven/Gradle should be able to pick up a correct Mockito version for end users.

It's an optional dependency, so it will only be pulled in if people are using Mockito anyway.

arteam:
Right, I was thinking about version conflicts. But Maven/Gradle/Enforcer plugin should be able to resolve them.

PR #2310, opened 2018-03-16
Remove metrics-ganglia completely

t-tsutsumi:
###### Problem:
metrics-ganglia was removed in #2239, but it is still referenced from BOM and documentation.

###### Solution:
Remove metrics-ganglia completely.


jplock:
Thanks!

PR #2311, opened 2018-03-19
Helpful error logging on JSON mapping

llyys:
When this error is logged as debug level, by default developer will not see root cause of the error, and I struggled several hours because the only response was "Unable to process JSON" and logs for io.dropwizard ERROR level did not show any problems.

Invalid jackson mappings are quite common and should be more visible at least Error level.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -41,9 +41,9 @@ public class JsonProcessingExceptionMapper extends LoggingExceptionMapper<JsonPr
         /*
          * Otherwise, it's those pesky users.
          */
-        LOGGER.error("Unable to process JSON", exception);
-
         final String message = exception.getOriginalMessage();
+        LOGGER.error("Unable to process JSON", message);
+
         final ErrorMessage errorMessage = new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(),
                 "Unable to process JSON", showDetails ? message : null);
         return Response.status(Response.Status.BAD_REQUEST)
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapper.java
@@ -41,7 +41,7 @@ public class JsonProcessingExceptionMapper extends LoggingExceptionMapper<JsonPr
         /*
          * Otherwise, it's those pesky users.
          */
-        LOGGER.debug("Unable to process JSON", exception);
+        LOGGER.error("Unable to process JSON", exception);
 
         final String message = exception.getOriginalMessage();
         final ErrorMessage errorMessage = new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(),


nickbabcock:
Hey, thank you for your interest in supporting dropwizard.

The reason why it's under debug is for the same reason you mention:

> Invalid jackson mappings are quite common

I'd be annoyed with all the stack traces in my logs (and being woken up at 2am) due to client issues.

If you'd like your clients to see the details, you can register (pseudo-code) `new JsonProcessingExceptionMapper(showDetails: true)`

arteam:
If I remember correctly, this issue popped up a couple of years ago. Please see the discussion in #1052.

llyys:
`showDetails: true`
Is not sufficient, I don't want to expose to the world about error details in json response. This part is fine.

But I need to have information about Error that caused this.
So if there was some error at 2am, I'm unable to see anything from error log, 
even if I configure logger `io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper: DEBUG`
This will be still logged as debug not an error.

If you annoyed about these errors, you can turn off logger for JsonProcessingExceptionMapper, and sleep well ;)

But rest of developers should not suffer and struggle by swallowed exceptions in log.

nickbabcock:
I'm not convinced.

If you want to see the details, enable debug logging.

> So if there was some error at 2am, I'm unable to see anything from error log,

Because it's logged under debug. Ensures there's an appender which receives those events.

> But rest of developers should not suffer and struggle by swallowed exceptions in log.

No one other than the client is suffering from client inflicted problems.

If the solutions provided thus far are still insufficient, you can always override the exception mapper, but we will not be changing the default level.

llyys:
The appender will still output log message starting with 
`... DEBUG : ....` 
And you should avoid logging DEBUG level messages in prod.

> If you want to see the details, enable debug logging.

I would argue that its much easier to pinpoint from log output what class logs you don't want to see and add those classes to certain level.
as  `io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper: NEVER`
rather than specifically googling and figuring out that there should be some DEBUG level logging switched on.

Right now initially it's hard to figure out what's wrong with api, `Unable to process JSON` clearly indicates that there is something wrong and error log should be place to take a look, but if developer has not specifically configured
`io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper: DEBUG` it's only guessing and stumbling without any help from logs or anywhere.

Also, by seeing something like this in prod log config
`io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper: DEBUG` 
is wierd and it feels wrong.

What I'm trying to say is that helpful error messages are really important, and useful for rapid issue resolving. 

This issue caused me approx 4 hours of stumbling without any clue what's wrong with the api. 
I had no clue that there is needed such hack, to set specific class to DEBUG level, to see an internal binding error real cause, and I'm trying to help other developers in future, avoiding being stuck with no clue what's wrong with JSON binding.

nickbabcock:
Yes, I agree, 4 hours debugging client issues would not be fun. If your clients can't figure out the error, and you don't want to return details to the client so they can fix the issue -- we've presented several methods to help.

Logging exceptions under a debug level is not uncommon -- here's an [example from Jetty](https://github.com/eclipse/jetty.project/blob/afd067771937bd46eb72bbd2c4d27956e0874386/jetty-io/src/main/java/org/eclipse/jetty/io/WriteFlusher.java#L347) among many.  I'm fine if you want to pivot the discussion to allowing a configurable log level for client `JsonProcessingException` issues, but we won't be changing the default.

joschi:
I'm closing this PR due to lack of activity.

@llyys Please leave a comment if you want us to reopen this pull request.

PR #2317, opened 2018-03-26
Add TLS socket logging appender

arteam:
###### Problem:
Many organizations mandate using of SSL/TLS for communications between applications to make sure that no sensitive data is sent over the wire in plain text. Currently, Dropwizard supports sending logs only via plain TCP.

###### Solution:
Add support for sending logs to a server that requires a secure means of communication via SSL/TLS. This change adds a new logging factory `TlsSocketAppenderFactory` which produces a new socket factory for creating secure client SSL/TLS sockets.

* The default configuration allows to communicate with a server with a certificate signed by a CA. In that case the administrator only need to change the port and the type of the appender.

```yaml
appenders:
  - type: tls
    host: ${tls.public_server_host}
    port: ${tls.public_server_port}
```

* If the server is deployed in a internal network with a self-signed certificate, the factory allows to specify a path to a trust store with the server certificate.

```yaml
appenders:
  - type: tls
    host: ${tls.server_host}
    port: ${tls.server_port}
    trustStorePath: ${tls.trust_store.path}
    trustStorePassword: ${tls.trust_store.pass}
```

* If the server requires the client to authenticate itself, the factory allows to specify a path to a key store with a client certificate that will be sent to the server during negotiation.

```yaml
appenders:
  - type: tls
    host: ${tls.server_host}
    port: ${tls.server_port}
    keyStorePath: ${tls.client_key_store.path}
    keyStorePassword: ${tls.client_key_store.pass}
```

* The factory allows to configure allowed cipher suites and SSL/TLS protocols.

###### Result:
Dropwizard users should be able to send logs to log servers with enabled SSL/TLS.


dropwizard-logging/src/main/java/io/dropwizard/logging/TlsSocketAppenderFactory.java
```diff
@@ -0,0 +1,429 @@
...
+ * </td>
+ * </tr>
+ * </table>
+ * <p/>
+ * For more configuration parameters, see {@link TcpSocketAppenderFactory}.
+ *
+ * @see TcpSocketAppenderFactory
+ */
+@JsonTypeName("tls")
+public class TlsSocketAppenderFactory<E extends DeferredProcessingAware> extends TcpSocketAppenderFactory<E> {
```
nickbabcock: I feel like the TLS configuration code is duplicated in three places: server, client, and now logging. It's fine for this PR, but I'm curious if we could ever consolidate these options so that none of the implementations are ever out of date (like we saw recently with lack of a jce provider for clients)

arteam: I've been thinking about this too. I guess the biggest issue is that we use different mechanisms for building `SSLContext`'s: Jetty's and Apache HTTP Client's and most of the code is Jackson deserialization code. There's also a mismatch between using it on the server and client sides. I think for the time being, it should be good enough as it is, but we can extract it, if we implement another SSL/TLS client or server.

PR #2320, opened 2018-03-30
Remove prerequisites from archetype pom

nickbabcock:
Since Maven 3.5, maven will warn when using `<prerequisites>..</prerequisites>` for a non maven-plugin project. Since Dropwizard projects are not maven plugins, we remove it from the
archetype pom.

https://maven.apache.org/docs/3.5.0/release-notes.html


arteam:
LGTM

PR #2323, opened 2018-04-05
Add owasp dependency check.

suryatej16:
###### Problem:
Currently, the dropwizard project does not run a check for security vulnerabilities during its build process. This results in dependencies being included that contain security risks (e.g. Jackson versions below 2.9.5)

###### Solution:
I have added the dependency-check plugin to the maven build process (https://jeremylong.github.io/DependencyCheck/) so CVEs can be identified.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,8 @@
 
     <properties>
         <dropwizard.url>http://www.dropwizard.io/${project.version}</dropwizard.url>
+        <dropwizard.check.fail-owasp>true</dropwizard.check.fail-owasp>
+        <dropwizard.check.skip-owasp>false</dropwizard.check.skip-owasp>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <maven.site.deploy.skip>true</maven.site.deploy.skip>
@@ -662,6 +664,27 @@
                         <configuration>
                             <failOnViolation>true</failOnViolation>
                         </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.owasp</groupId>
+                <artifactId>dependency-check-maven</artifactId>
+                <version>3.1.2</version>
+                <configuration>
+                    <failBuildOnAnyVulnerability>${dropwizard.check.fail-owasp}</failBuildOnAnyVulnerability>
+                    <skip>${dropwizard.check.skip-owasp}</skip>
+                    <showSummary>true</showSummary>
+                    <format>ALL</format>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>security-check</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
                     </execution>
                 </executions>
             </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -53,8 +53,6 @@
 
     <properties>
         <dropwizard.url>http://www.dropwizard.io/${project.version}</dropwizard.url>
-        <dropwizard.check.fail-owasp>true</dropwizard.check.fail-owasp>
-        <dropwizard.check.skip-owasp>false</dropwizard.check.skip-owasp>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <maven.site.deploy.skip>true</maven.site.deploy.skip>
@@ -673,10 +671,9 @@
                 <artifactId>dependency-check-maven</artifactId>
                 <version>3.1.2</version>
                 <configuration>
-                    <failBuildOnAnyVulnerability>${dropwizard.check.fail-owasp}</failBuildOnAnyVulnerability>
-                    <skip>${dropwizard.check.skip-owasp}</skip>
+                    <failBuildOnAnyVulnerability>true</failBuildOnAnyVulnerability>
                     <showSummary>true</showSummary>
-                    <format>ALL</format>
+                    <suppressionFiles>suppressed-cves.xml</suppressionFiles>
                 </configuration>
                 <executions>
                     <execution>
--- a/suppressed-cves.xml
+++ b/suppressed-cves.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+    <suppress>
+        <notes><![CDATA[file name: dropwizard-views-1.4.0-SNAPSHOT.jar]]></notes>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -53,6 +53,8 @@
 
     <properties>
         <dropwizard.url>http://www.dropwizard.io/${project.version}</dropwizard.url>
+        <dropwizard.check.fail-owasp>true</dropwizard.check.fail-owasp>
+        <dropwizard.check.skip-owasp>false</dropwizard.check.skip-owasp>
```
joschi: No need to introduce another property here. The OWASP dependency check plugin provides the `dependency-check.skip` property out of the box with the same default.

pom.xml
```diff
@@ -53,6 +53,8 @@
 
     <properties>
         <dropwizard.url>http://www.dropwizard.io/${project.version}</dropwizard.url>
+        <dropwizard.check.fail-owasp>true</dropwizard.check.fail-owasp>
```
joschi: No need to introduce another property here. The OWASP dependency check plugin provides the `failOnError` and `failBuildOnAnyVulnerability` properties out of the box.

pom.xml
```diff
@@ -665,6 +667,27 @@
...
+
+            <plugin>
+                <groupId>org.owasp</groupId>
+                <artifactId>dependency-check-maven</artifactId>
+                <version>3.1.2</version>
+                <configuration>
+                    <failBuildOnAnyVulnerability>${dropwizard.check.fail-owasp}</failBuildOnAnyVulnerability>
+                    <skip>${dropwizard.check.skip-owasp}</skip>
+                    <showSummary>true</showSummary>
+                    <format>ALL</format>
```
joschi: Do we really need all report formats here? I think HTML (default) should be sufficient.

https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html

nickbabcock:
I find it bizarre that the jdk10 build doesn't report the same false positive, gets a bit farther in the build, and then reports a different false positive: [CVE-2015-8862](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8862) 😅

suryatej16:
@joschi Added a suppression file to ignore false positives. Just a general question, how did you distinguish those as false positives?

joschi:
> how did you distinguish those as false positives?

@suryatej16 From the fact that Dropwizard is not using the PHP CMS Drupal. 😉

suppressed-cves.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+    <suppress>
+        <notes><![CDATA[file name: dropwizard-views-1.4.0-SNAPSHOT.jar]]></notes>
+        <gav regex="true">^io\.dropwizard:dropwizard-views(-.*)?:.*$</gav>
+        <cve>CVE-2015-3378</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[file name: dropwizard-views-1.4.0-SNAPSHOT.jar]]></notes>
```
joschi: That's a bit specific for the current (SNAPSHOT) version. Maybe add a comment that it is a false positive instead.

suppressed-cves.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+    <suppress>
+        <notes><![CDATA[file name: dropwizard-views-1.4.0-SNAPSHOT.jar]]></notes>
```
joschi: That's a bit specific for the current (SNAPSHOT) version. Maybe add a comment that it is a false positive instead.

suppressed-cves.xml
```diff
@@ -0,0 +1,18 @@
...
+        <gav regex="true">^io\.dropwizard:dropwizard-views(-.*)?:.*$</gav>
```
joschi: I think it's safe to ignore CVE-2015-3378 completely and not just for this artifact.

suppressed-cves.xml
```diff
@@ -0,0 +1,18 @@
...
+        <cve>CVE-2015-3378</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[file name: dropwizard-views-1.4.0-SNAPSHOT.jar]]></notes>
+        <gav regex="true">^io\.dropwizard:dropwizard-views(-.*)?:.*$</gav>
```
joschi: I think it's safe to ignore CVE-2015-3379 completely and not just for this artifact.

suppressed-cves.xml
```diff
@@ -0,0 +1,18 @@
...
+        <cve>CVE-2015-3379</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[file name: compiler-0.9.5.jar]]></notes>
+        <gav regex="true">^com\.github\.spullara\.mustache\.java:compiler:.*$</gav>
```
joschi: I think it's safe to ignore CVE-2015-8862 completely and not just for this artifact.

suppressed-cves.xml
```diff
@@ -0,0 +1,18 @@
...
+        <gav regex="true">^io\.dropwizard:dropwizard-views(-.*)?:.*$</gav>
+        <cve>CVE-2015-3378</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[file name: dropwizard-views-1.4.0-SNAPSHOT.jar]]></notes>
+        <gav regex="true">^io\.dropwizard:dropwizard-views(-.*)?:.*$</gav>
+        <cve>CVE-2015-3379</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[file name: compiler-0.9.5.jar]]></notes>
```
joschi: That's a bit specific for the version of the Mustache template engine currently used by Dropwizard. Maybe add a comment that it is a false positive instead

suryatej16:
Updated with review comments.

suppressed-cves.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+    <suppress>
+        <notes><![CDATA[false positive, dropwizard does not use Drupal]]></notes>
+        <cve>CVE-2015-3378</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[false positive, dropwiard does not use Drupal]]></notes>
```
joschi: Typo: dropwiard -> Dropwizard

suppressed-cves.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+    <suppress>
+        <notes><![CDATA[false positive, dropwizard does not use Drupal]]></notes>
```
joschi: Typo: dropwiard -> Dropwizard

PR #2331, opened 2018-04-18
Fix javadoc for ViewBundle

PeterWippermann:
###### Problem:
The formatting of [ViewBundle's JavaDoc](http://www.dropwizard.io/1.3.1/dropwizard-views/apidocs/index.html) is broken.

The formatting breaks after the @see Link to the FreeMarker reference. The rest of the javadoc relating to Mustache is totally broken.

###### Solution:
@see may not be used for inline links, but will always start a new section at the end of the javadoc ([reference](https://docs.oracle.com/javase/6/docs/technotes/tools/windows/javadoc.html#@see)). @link in turn is used to link to other classes and thus is also no good substitute ([reference](https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#link)). 

Inline links should thus be created with plain HTML ([stackoverflow](https://stackoverflow.com/questions/1082050/linking-to-an-external-url-in-javadoc)).

###### Result:
Javadoc is properly formatted.

### DIFF
--- a/dropwizard-views/src/main/java/io/dropwizard/views/ViewBundle.java
+++ b/dropwizard-views/src/main/java/io/dropwizard/views/ViewBundle.java
@@ -51,8 +51,8 @@ import static com.google.common.base.MoreObjects.firstNonNull;
  * <p>A resource method with a view would looks something like this:</p>
  *
  * <pre><code>
- * \@GET
- * public PersonView getPerson(\@PathParam("id") String id) {
+ * @GET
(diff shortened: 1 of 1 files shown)


nickbabcock:
Sounds good. Thank you for the fix!

PeterWippermann:
You're welcome! This was a nice chance to learn a bit more about javadoc ;-)
Thank you for maintaining this project!

Issue #2334, opened 2018-04-28
@SelfValidating annotation results in 500 when used with @BeanParam 


matt-snider:
When using the new [@SelfValidating](https://github.com/dropwizard/dropwizard/blob/v1.3.1/dropwizard-validation/src/main/java/io/dropwizard/validation/selfvalidating/SelfValidating.java) annotation on a class that is used as a `@BeanParam`, a 500 error is raised when a [@SelfValidation](https://github.com/dropwizard/dropwizard/blob/v1.3.1/dropwizard-validation/src/main/java/io/dropwizard/validation/selfvalidating/SelfValidation.java) fails instead of 400/422.

This is caused by execution falling through to [ConstraintMessage#getMemberName()](https://github.com/dropwizard/dropwizard/blob/release/1.3.x/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java#L120) where an `IllegalArgumentException` is raised in `FieldUtils#getField()` since `member.getName()` returns an empty string. 

I believe execution should not get to this point and [@SelfValidating](https://github.com/dropwizard/dropwizard/blob/v1.3.1/dropwizard-validation/src/main/java/io/dropwizard/validation/selfvalidating/SelfValidating.java) should be handled like [@ValidationMethod](https://github.com/dropwizard/dropwizard/blob/v1.3.1/dropwizard-validation/src/main/java/io/dropwizard/validation/ValidationMethod.java).

Here is an example showing this behaviour:
```java
@SelfValidating
public class FailingSelfValidation {

    @SelfValidation
    public void validate(ViolationCollector collector) {
         collector.addViolation("failing self validation");
    }

}
```

And the following resource method:
```java
@GET
@Path("failingSelfValidation")
public Response failingSelfValidation(@Valid @BeanParam FailingSelfValidation) {
        return Response.ok().build();
}
```

Since I have a already written a test exemplifying this issue, and I have a fix in mind, I will submit a pull request shortly

PR #2335, opened 2018-04-28
Fix bug when using @SelfValidating with @BeanParam (#2334)

matt-snider:
###### Problem:
When using the new [@SelfValidating](https://github.com/dropwizard/dropwizard/blob/v1.3.1/dropwizard-validation/src/main/java/io/dropwizard/validation/selfvalidating/SelfValidating.java) annotation on a class that is used as a `@BeanParam`, a 500 error is raised instead of 400/422.

This is described in more detail in #2334. 

###### Solution:
Handle [@SelfValidating](https://github.com/dropwizard/dropwizard/blob/v1.3.1/dropwizard-validation/src/main/java/io/dropwizard/validation/selfvalidating/SelfValidating.java) violations like [@ValidationMethod](https://github.com/dropwizard/dropwizard/blob/v1.3.1/dropwizard-validation/src/main/java/io/dropwizard/validation/ValidationMethod.java).

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
@@ -6,6 +6,7 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.Iterables;
 import io.dropwizard.validation.ValidationMethod;
+import io.dropwizard.validation.selfvalidating.SelfValidating;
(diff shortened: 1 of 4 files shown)


matt-snider:
@manuel-hegner since you contributed this feature it would be great to get some feedback. Maybe constraint violations were intended to be handled differently and this is the wrong spot for it? Let me know.

nickbabcock:
I like it, seems sensible. I'll pause for a couple days to merge in case there are comments. Thank you very much for the excellent pull request.

manuel-hegner:
I would agree that a 4XX is the expected behaviour. This is an excellent addition.

PR #2337, opened 2018-05-01
Skip an OWASP dependency check in the dev profile

arteam:
###### Problem:
In the dev profile we want to build and install artifacts as fast as possible, and an OWASP check is quite slow.

###### Solution:
Disable the dependency-check plugin in the dev profile.

###### Result:
Faster builds during development

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -301,6 +301,13 @@
                             <perCoreThreadCount>true</perCoreThreadCount>
                         </configuration>
                     </plugin>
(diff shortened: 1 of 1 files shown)


nickbabcock:
Side note, I'm assuming the dev profile works for you? I ask as I have tried multiple times and it keeps failing. Looks like some of the tests that rely on logging output get mixed up when multi-threaded.

PR #2340, opened 2018-05-02
Http2 ALPN and bootclasspath update

nickbabcock:
###### Problem:
- JDK 171 & 172 have missing alpn values, so add them to the pom
- I have a machine that contains special characters in `${user.home}`, so quotes are necessary to reference the ALPN jar correctly. @arteam I see that you reverted this behavior in 3b8c255d929e and I just want to make sure I'm not missing something.

###### Result:
I can run http2 tests on this machine again 😄


arteam:
It didn't work for me with the quotes on my Linux machine at some moment of time, but I'm willing to believe it was a configuration issue in my OS.

Issue #2336, opened 2018-04-30
@SelfValidation methods in base class are not recognised


dave1e6:
I have a base class with some general validations that apply to all subclasses and so I would like do the following:
```
@SelfValidating
public class BaseClass {
    @SelfValidation
    public void appliesToAllSubclasses(ViolationCollector collector) {
    }
}

public class SubClass extends BaseClass {
}
```

I expected that validating an instance of SubClass would call BaseClass.appliesToAllSubClasses, it doesn't and I get the following message in the logs:
```
08:16:45.188 [main] ERROR io.dropwizard.validation.selfvalidating.SelfValidatingValidator - The class class SubClass is annotated with @SelfValidating but contains no valid methods that are annotated with @SelfValidation
```

```
    @Test
    public void selfValidatingSubclass() {
        BaseClass baseClass = new SubClass();
        Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
        validator.validate(baseClass);
    }
```

If I annotate SubClass with @SelfValidating as well then I get two log messages rather than one.

manuel-hegner:
I fixed the inheritance problem but I am not sure in which case the problem with two log messages occurs. @dave1e6  Could you give me a test case (maybe based on my PR #2344) that shows that problem?

nickbabcock:
Closing this as #2344 was merged. If the problem wasn't resolved, please reply with more info. Thanks @manuel-hegner for the implementation.

isaki:
The issue still remains for a specific corner case. I will look into putting together some tests to reproduce it and possibly submit a PR to fix it (or at least share the tests so someone else can fix it if I can't find the time to patch it in the near term).

manuel-hegner:
If you can write a test for this corner case or describe it well enough I will write the fix, no problem.

isaki:
I can get you a pretty simple object model to use to reproduce it fairly quickly. I should have some time today to dedicate to this. I'll reply back in a bit.

isaki:
I've broken down the problematic classes into three simple cases (you really only need two, but my original issue involved 3, so I've left it that way).

1) A base abstract class that contains the primary validation method.
2) One or more child classes that implement an extended validation method.

I'm sure this isn't the best design. Ideally I'd like to just hit each class with its own `@SelfValidating` method and have them all be run down the entire hierarchy. My workaround has each class overriding the base class `validate` and invoking `super.validate`.

Anyway, here is a super simplified version of what triggered the issue:

[...]

PR #2344, opened 2018-05-03
improved self validation #2336

manuel-hegner:
###### Problem:
As stated in #2336 the SelfValidation system did not handle class hierarchies as one would expect.

###### Solution:
This pull request:
- discoveres @SelfValidation annotated methods on superclasses
- handles overriding these methods as expected
- adds tests for these cases
- extends the testing to test the logged errors and warnings in case of incorrectly used annotations

### DIFF
--- a/dropwizard-validation/pom.xml
+++ b/dropwizard-validation/pom.xml
@@ -46,14 +46,14 @@
             <artifactId>javassist</artifactId>
         </dependency>
         <dependency>
-        	<groupId>org.slf4j</groupId>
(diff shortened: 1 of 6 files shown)


manuel-hegner:
I am not sure yet if all the problems mentioned in #2336 are solved by this PR and it should not be merged until this was clarified.

manuel-hegner:
Since there was no more comment concerning #2336 I think this could be merged now. If there is another problem I can handle it in another PR.

PR #2346, opened 2018-05-06
Upgrade Jetty to 9.4.10.v20180503

arteam:
###### Problem:
We use not the latest version of Jetty in the upstream

###### Solution:
Bump the version

###### Result:
We take advantage of many small bugfixes in Jetty, especially in the HTTP/2 module. See

https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.9.v20180320
https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.10.v20180503

### DIFF
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -22,7 +22,7 @@
         <guava.version>24.0-jre</guava.version>
         <jersey.version>2.25.1</jersey.version>
         <jackson.version>2.9.5</jackson.version>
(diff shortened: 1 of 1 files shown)


debraj-manna:
@arteam - Will this be in 1.3.2 and 1.2.6 also ?

arteam:
Yes, shouldn't be a problem.

debraj-manna:
Thanks @arteam

debraj-manna:
Thanks @arteam the releases are there I missed it.

PR #2360, opened 2018-05-18
Add support for multiple session factories at the same time

damagode:
###### Problem:
Sometimes we need to use multiple Hibernate bundles on a resource endpoint at the same time. In some projects, this is a requirement and we can't refactor the code to use only one. We found that in the current code we are able to register multiple Hibernate bundles but we can only use one at a time (With the `@UnitOfWork` annotation). This PR is to add the ability to use more than one bundle in the same `@UnitOfWork` context.

###### Solution:
We changed the `@UnitOfWork` annotation to accept multiple values for the bundle name. Also, we changed the `UnitOfWorkAspect` to process all the selected bundles (instead of processing only one) in the same way as before. So when we perform any operation in the aspect we execute it in all the selected bundles. 

This allows us to see all the selected bundles as a group in the way that if an error occurs and the transaction needs to be reverted they are going to be rollbacked in all the selected bundles, the same happens with the commit and clean operations.

###### Result:
The result is that we can select multiple bundles in the same `@UnitOfWork` by passing the values as follow:

`@UnitOfWork(value={'bundle1', 'bundle2'})`
or just
`@UnitOfWork({'bundle1', 'bundle2'})`

The change is totally retro-compatible and we can specify only one bundle by:

`@UnitOfWork('bundle')`
or use the default one (if we only register one bundle) by not passing any value.
`@UnitOfWork`

The rest of the system will remain exactly the same and also the use of Hibernate will not change at all.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
@@ -1,14 +1,22 @@
 package io.dropwizard.hibernate;
 
+import org.apache.commons.lang3.tuple.Pair;
+import org.hibernate.HibernateException;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
 import org.hibernate.context.internal.ManagedSessionContext;
 
 import javax.annotation.Nullable;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -47,11 +55,8 @@ public class UnitOfWorkAspect {
     @Nullable
     private UnitOfWork unitOfWork;
 
-    @Nullable
-    private Session session;
-
-    @Nullable
-    private SessionFactory sessionFactory;
+    private Map<String, Session> sessions = new HashMap<>();
+    private Map<String, SessionFactory> selectedSessionFactories = new HashMap<>();
 
     public void beforeStart(@Nullable UnitOfWork unitOfWork) {
         if (unitOfWork == null) {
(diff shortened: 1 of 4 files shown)


dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
```diff
@@ -47,120 +55,189 @@ public UnitOfWorkAspect(Map<String, SessionFactory> sessionFactories) {
...
+
+    private void rollbackTransactions(UnitOfWork unitOfWork, Map<String, Session> sessions) {
         if (!unitOfWork.transactional()) {
             return;
         }
-        final Transaction txn = session.getTransaction();
-        if (txn != null && txn.getStatus().canRollback()) {
-            txn.rollback();
+
+        //We need to ensure that the changes in all the sessions have been rollbacked
```
kewpiedoll99: i suggest you change "rollbacked" to "rolled back"

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
```diff
@@ -47,120 +55,189 @@ public UnitOfWorkAspect(Map<String, SessionFactory> sessionFactories) {
...
+        //and also we need to inform about the exceptions
+        List<Optional<RuntimeException>> errors = sessions.values().stream()
+            .map(UnitOfWorkAspect::rollbackTransaction)
+            .filter(Optional::isPresent)
+            .collect(Collectors.toList());
+
+        if (!errors.isEmpty()) {
+            throw errors.iterator().next().get();
```
kewpiedoll99: does this just throw the first error in the list?

damagode: This is a tricky situation. We tried to mimic the previous behavior so If an exception is thrown during the process we should let the exception bubble up. The thing with multiple sessions is that for certain operations such as `rollback`or `session.close` we need to ensure that the operation is executed for all the sessions before letting the exception bubbles up. If we want to change this behavior in the future we are prepared to notify multiple exceptions.

joschi:
@dropwizard/committers Does anyone want to invest some time into this PR to make it ready for Dropwizard 2.0.0?

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
mzamani1: Would it make sense to throw an exception that contains references to all of the errors, if there is more than one, so no information is lost when propagating?

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
```diff
@@ -47,120 +55,189 @@ public UnitOfWorkAspect(Map<String, SessionFactory> sessionFactories) {
...
 
     public void onFinish() {
         try {
-            if (session != null) {
-                session.close();
+            //We need to ensure that all the sessions have been closed
+            //and also we need to inform about the exceptions
+            List<Optional<RuntimeException>> errors = sessions.values().stream()
+                .map(UnitOfWorkAspect::closeSession)
+                .filter(Optional::isPresent)
```
mzamani1: Bit of a nitpick, but it might be a little cleaner if you call `.map(Optional::get)`, since you’re already filtering by `isPresent` here.

PR #2367, opened 2018-05-24
Add JUnit 5 Example to Testing Clients

RohanNagar:
###### Problem:

Testing client implementations in JUnit 5 looks a lot different, and I had to dig through the source code to get it to work. 

###### Solution:

This PR adds documentation for testing client implementations with JUnit 5 so that it is easier to achieve.

###### Result:

New documentation example.


arteam:
LGTM, Thanks for improving the docs.

PR #2375, opened 2018-06-01
Support expirable log level configurations

mattnelson:
###### Problem:
With the log configuration task. My typical use case is to temporarily change the log level, get the relevant information, then revert the change. I would like the option to set a duration at which the log change reverts to the previous state. The example below takes in an ISO8601 duration as optional parameter.

https://www.dropwizard.io/1.3.2/docs/manual/core.html#logging-configuration-via-http
https://github.com/dropwizard/dropwizard/blob/v1.3.2/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/LogConfigurationTask.java

###### Solution:
Support an optional ISO 8601 duration on the log level task.

```
curl -X POST -d "logger=com.example.helloworld&level=INFO&duration=PT10M" http://localhost:8081/tasks/log-level
```

###### Result:
Log level configuration will support automatic reverting to the default level.


dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/LogConfigurationTask.java
```diff
@@ -57,10 +62,28 @@ public LogConfigurationTask(ILoggerFactory loggerContext) {
...
-            output.println(String.format("Configured logging level for %s to %s", loggerName, loggerLevel));
+            Logger logger = ((LoggerContext) loggerContext).getLogger(loggerName);
+
+            String message = String.format("Configured logging level for %s to %s", loggerName, loggerLevel);
+
+            if (loggerLevel != null && duration != null) {
+                // Get the effective level before setting the new level
+                final Level effectiveLevel = logger.getEffectiveLevel();
+                final long millis = duration.toMillis();
+                timer.schedule(new TimerTask() {
```
joschi: What would happen if a user wants to override a previously set timeout for a logger or remove the timeout?

Currently, the logger would be reset to the initial log level by the first scheduled task anyway and the second scheduled task would restore the previously set log level.

Example:

* Initial log level is "WARN".
* User sends logger level "DEBUG" and duration P10m.
* -- 8 minutes pass --
* User sends logger level "DEBUG" and duration P30m.
* -- 2 minutes pass --
* Scheduled task resets log level to "WARN"
* -- 28 minutes pass --
* Scheduled task resets log level to "DEBUG"
* ???
* PROFIT!

mattnelson: > Currently, the logger would be reset to the initial log level by the first scheduled task anyway and the second scheduled task would restore the previously set log level.

I went back and forth on this. Should it be restored to the previous effective level or `null`? When `null` is supplied logback will set to the [effective parent level](https://github.com/qos-ch/logback/blob/4dd08e68321548cd85251ec389f5a832505e1268/logback-classic/src/main/java/ch/qos/logback/classic/Logger.java#L157-L158). This would eliminate some of the ordering issues, as 

>What would happen if a user wants to override a previously set timeout for a logger or remove the timeout?

I had a hunch this would come up. I had originally gone down that path and it was getting a little more complex than I wanted to introduce. My second pass implementation was to keep it really simple, with some of those caveats on ordering of operations.

We could push the lifecycle from the environment down to the admin environment.
https://github.com/dropwizard/dropwizard/blob/2cfdd6dc5bd390e3f5b1ef47d8f7e7fd9efd7fd9/dropwizard-core/src/main/java/io/dropwizard/setup/Environment.java#L76-L78

Use that to create a scheduled executor service for the log configuration task
https://github.com/dropwizard/dropwizard/blob/2cfdd6dc5bd390e3f5b1ef47d8f7e7fd9efd7fd9/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java#L70

Then use weak concurrent hash map to store package/future mappings, so that we can support `doDelete` on the task servlet to support removal. Could also enforce that there is only one "active" log level change per package at a time, or when a new one comes it replaces the existing override.
https://github.com/dropwizard/dropwizard/blob/2cfdd6dc5bd390e3f5b1ef47d8f7e7fd9efd7fd9/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java#L96

nickbabcock: I haven't looked too close at the impl but I'm in agreement that when believe the situation where the initial level is `WARN` and there are two tasks to set the level to `DEBUG` and `INFO` then no matter when the two tasks are executed or finished the final log level should be set back to `WARN`

mattnelson:
Addressed comment on f7d4302. Updated to set log level to null to revert to parent effective level.

jplock:
Looks like one of the throttling tests from https://github.com/dropwizard/dropwizard/pull/2384 is causing the appveyor builds to fail.

nickbabcock:
@jplock I had some thoughts on making those tests less flaky. I'll submit a PR

jplock:
Is there a way to rerun the appveyor tests without committing anything?

jplock:
@mattnelson after merging in the guava removal PR, we have some conflicts on this PR now.  Would you mind resolving those?  Thanks!

dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/LogConfigurationTask.java
```diff
@@ -34,6 +38,7 @@
 public class LogConfigurationTask extends Task {
 
     private final ILoggerFactory loggerContext;
+    private final Timer timer = new Timer(LogConfigurationTask.class.getSimpleName(), true);
```
mattnelson: Thinking about wrapping this in a lazy init block to delay creating the timer until the first expireable log level is submitted.

jplock:
@joschi @nickbabcock any additional feedback?

dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/LogConfigurationTask.java
```diff
@@ -1,15 +1,22 @@
 package io.dropwizard.servlets.tasks;
 
 import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.LoggerContext;
+import com.google.errorprone.annotations.concurrent.LazyInit;
+import java.sql.Time;
```
joschi: Unused import.

mattnelson:
> With the caveat that overwriting the trigger to reset the log level isn't possible.

documented on b67f44d

Issue #2376, opened 2018-06-04
Optional throttling for log appenders


ochedru:
Situation: logs are sent to an external managed service through a specific log appender. Something goes wrong in the application and it generates a lot of logs very quickly. You do not notice the problem immediately. You end up with a hefty bill.

I was thinking of implementing a simple throttling mechanism in `AbstractAppenderFactory` and `AsyncAppenderBase`: maximum M messages over a given duration. Extra messages are discarded and not put into the queue.
Of course, this is disabled by default.

Do you think this is something of interest to Dropwizard users? If yes, I will submit a PR.

nickbabcock:
Yeah I've definitely been bitten by overly chatty logs caused by unfortunate situations. I'm curious if this can't already be solved through `queueSize` and `discardingThreshold` logging configurations. Let me know your thoughts 😄

ochedru:
If the remote log service has very good response times, I am afraid limiting the queue capacity will not help.

nickbabcock:
Very true. Yeah go ahead and file a PR. Having the ability for an opt-in throttle for any appender sounds good.

nickbabcock:
Closed by #2384

PR #2384, opened 2018-06-11
Throttling logging appender

ochedru:
See https://github.com/dropwizard/dropwizard/issues/2376.

This PR implements a throttling mechanism for logging appenders.

The existing parameters limiting the queue size (`queueSize` and `discardingThreshold`) do not prevent the application from flooding a remote logging service. Such services usually become expensive when their usage quota is exceeded.

The proposed feature aims at providing a safety net when the application logging goes out of control.

Two new logging configuration parameters are introduced and can apply to any logging appender:
- `throttlingTimeWindow` is a `Duration` defining a sliding window for throttling. By default, it is not set and throttling is disabled.
- `maxMessagesPerThrottlingTimeWindow` is the maximum number of messages sent during the throttling time window. Once this number is reached, messages are silently discarded until there is room for new messages in the sliding time window.

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
@@ -183,8 +183,8 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
 
     @JsonProperty
     public void setTimeZone(String zoneId) {
-        this.timeZone = nullToEmpty(zoneId).equalsIgnoreCase("system") ? TimeZone.getDefault()
-                : TimeZone.getTimeZone(zoneId);
+        this.timeZone = nullToEmpty(zoneId).equalsIgnoreCase("system") ? TimeZone.getDefault() :
+            TimeZone.getTimeZone(zoneId);
     }
 
     @JsonProperty
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
@@ -16,14 +16,19 @@ import io.dropwizard.logging.async.AsyncAppenderFactory;
 import io.dropwizard.logging.filter.FilterFactory;
 import io.dropwizard.logging.layout.DiscoverableLayoutFactory;
 import io.dropwizard.logging.layout.LayoutFactory;
+import io.dropwizard.util.Duration;
+import io.dropwizard.validation.MaxDuration;
+import io.dropwizard.validation.MinDuration;
 
 import javax.annotation.Nullable;
+import javax.validation.Valid;
 import javax.validation.constraints.DecimalMin;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 import java.util.List;
 import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 
 import static com.google.common.base.Strings.nullToEmpty;
 
@@ -78,10 +83,10 @@ import static com.google.common.base.Strings.nullToEmpty;
  *         </td>
  *     </tr>
  *     <tr>
- *         <td>{@code maxMessagesPerSecond}</td>
+ *         <td>{@code messageThrottle}</td>
  *         <td>
- *             Throttling: maximum number of messages per second.
- *             Throttled messages are discarded.
+ *             Average duration between messages. Throttled messages are discarded.
+ *             Maximum acceptable duration is 1 minute.
  *             By default, this is not set and throttling is disabled.
  *         </td>
  *     </tr>
@@ -116,8 +121,9 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
     private int discardingThreshold = -1;
 
     @Nullable
-    @DecimalMin(value="0", inclusive = false)
-    private Double maxMessagesPerSecond;
+    @MinDuration(value = 0, unit = TimeUnit.SECONDS, inclusive = false)
+    @MaxDuration(value = 1, unit = TimeUnit.MINUTES)
+    private Duration messageThrottle;
 
     private boolean includeCallerData = false;
 
@@ -147,13 +153,13 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
 
     @JsonProperty
     @Nullable
-    public Double getMaxMessagesPerSecond() {
-        return maxMessagesPerSecond;
+    public Duration getMessageThrottle() {
+        return messageThrottle;
     }
 
     @JsonProperty
-    public void setMaxMessagesPerSecond(Double maxMessagesPerSecond) {
-        this.maxMessagesPerSecond = maxMessagesPerSecond;
+    public void setMessageThrottle(Duration messageThrottle) {
+        this.messageThrottle = messageThrottle;
     }
 
     @JsonProperty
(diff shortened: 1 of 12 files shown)


nickbabcock:
I haven't looked too close at the impl, but have you tested to see if [Gauva's RateLimiter](https://google.github.io/guava/releases/25.1-jre/api/docs/) is sufficient?

ochedru:
The problem with Guava's [RateLimiter](https://google.github.io/guava/releases/snapshot-jre/api/docs/com/google/common/util/concurrent/RateLimiter.html) is that it will not allow small bursts of messages without blocking, so I went for a simple implementation based on a ring buffer of timestamps. It does not block until the time window is "full".
I still have to fix the tests in the PR, though.

nickbabcock:
Oh man, now I'm kinda torn between the two philosophies. I have an example in my head:

Let's say you set a throttle 100 messages over 10 seconds (ie: average 10 messages a second), but the actual log rate is 100 per second.

With this impl, you see all 100 messages for the first second, but the logs are then silent for 9 seconds.

With guava, there is no silent period, but if the log rate of 100 per second lasts only 1 second and subsequently drops to 0 then guava will drop 90% of the logs in that 10 second period.

So each impl appears to have pros and cons. But in my opinion, the guava ratelimiter seems in better control of bursts -- especially if the infrastructure hosting the logs is measured / charged in ops, the ratelimiter would ensure those ops weren't exceeded. There is also no period of time with the ratelimiter where all logs are dropped until the next window.

Can you comment as to why you prefer one side over the other, so we get the whole story 😄

ochedru:
Your approach using ops makes sense as well. However, I have two problems with `RateLimiter`:

- Guava will block temporarily even if we are under the limit on average. For example, if the limit is 10 logs per second, but the application is logging every second 5 messages in a row, that will take 0.5 second with the rate limiter.
Ideally, we would implement throttling in the worker thread and this would not be a problem. But this is not possible because the worker code is deep inside `logback`.
That is why I have to throttle before enqueing the message. At this point, I think it is not acceptable to wait: this makes the whole async logging useless. My throttle implementation decides without waiting if the incoming message is kept and sent to the queue, or discarded.
(Yes, `RateLimiter` has the `tryAcquire` method to avoid blocking, but then we will likely discard messages whereas we are under the limit on average.)

- `RateLimiter` is configured with an integer number of permits per second. This is less flexible that a time window and a number of messages. If I want to configure throttling for 50000 messages per day, I cannot do that directly with `RateLimiter`.

nickbabcock:
> (Yes, RateLimiter has the tryAcquire method to avoid blocking, but then we will likely discard messages whereas we are under the limit on average.)

I think if you're implying that the RateLimiter prefers under-utilization -- I believe the opposite is true:

You can see in this sample that even though I only allow one qps, I can immediately acquire 1000 seconds worth:

```java
final RateLimiter rateLimiter = RateLimiter.create(1);
final boolean acquired = rateLimiter.tryAcquire(1000);
assertThat(acquired).isTrue();
```

Subsequent requests will not be successfully acquired for the next 999 seconds. So for those 999 seconds, the rate limiter is over-utilized. The example is a bit contrived (we're not trying to acquire 1000 at a time), but is demonstrative.

On the topic of under-utilization and bursts, the `RateLimiter` rolls over some of that under utilization to handle a burst:

```java
final RateLimiter rateLimiter = RateLimiter.create(1);
Thread.sleep(1000);
final boolean acquired = rateLimiter.tryAcquire();
final boolean acquired2 = rateLimiter.tryAcquire();
assertThat(acquired).isTrue();
assertThat(acquired2).isTrue();
```

Without the `sleep` simulating under utilization, the previous sample will fail. There is a quite a [lengthy design javadoc](https://github.com/google/guava/blob/c2f8967beb5a6feb9b104739d453dc116982a76a/guava/src/com/google/common/util/concurrent/SmoothRateLimiter.java#L26) on utilization.

> `RateLimiter` is configured with an integer number of permits per second

I don't think this is accurate. The [javadoc](https://google.github.io/guava/releases/19.0/api/docs/index.html?com/google/common/util/concurrent/RateLimiter.html) has `create` taking in a double. You can set the `permitsPerSecond` to be less than one. For your example: `RateLimiter.create(50000 / 86400)` should do it.

---

Not trying to push the use of `RateLimiter` here, I just thought it fit nicely into this situation. It has O(1) space requirements whereas the PR, as it stands, has a O(n) space requirement where n is the number of message per unit of time. A pitfall for those who may want to constrain the number of messages per day.

ochedru:
My bad, I did not read thoroughly the documentation: `RateLimiter` is indeed suitable for our use case. Thank you for pointing me to this page!
I updated the PR but I still have to fix the tests regarding timing differences between the CI platforms... I will try to get rid of the ugly `Thread.sleep()` calls using a `CountDownLatch` or something.

dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
```diff
@@ -137,6 +144,17 @@ public void setDiscardingThreshold(int discardingThreshold) {
         this.discardingThreshold = discardingThreshold;
     }
 
+    @JsonProperty
+    @Nullable
+    public BigDecimal getMaxMessagesPerSecond() {
```
nickbabcock: Are there benefits to using a `BigDecimal` when it is converted into a `double` straightaway?

dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
```diff
@@ -111,6 +114,8 @@
 
     private int discardingThreshold = -1;
 
+    private double maxMessagesPerSecond = -1;
```
nickbabcock: I can't help but think if someone set

```yaml
maxMessagesPerSecond: -2
```

that this should fail fast instead of ignoring it (eg. the dash was a typo). So I was thinking of using a validation constraint for the range (0, ∞). Unfortunately the `Min` annotation cannot represent an exclusive range, and `DecimalMin` isn't technically supported for `double` according to the bean validation spec -- but Hibernate Validator might support it

so you might want to try something like:

```java
    @DecimalMin(value = "0", inclusive = false)
    private Double maxMessagesPerSecond;
```

And write a test that ensures that validation fails if someone enters 0 or below. What do you think?

ochedru: Sounds good. It must be `@Nullable` as well, I guess.

dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
```diff
@@ -111,6 +115,10 @@
 
     private int discardingThreshold = -1;
 
+    @Nullable
+    @DecimalMin(value="0", inclusive = false)
+    private Double maxMessagesPerSecond;
```
jplock: does it make sense to support a fractional number of messages per second instead of making this an `Integer` or even an `OptionalInt`?

nickbabcock: Yeah if you want 30 messages per minute, you'd set to this `0.5`

jplock: It’s too bad a Duration wouldn’t work here somehow

nickbabcock: I think you could be on to something, you could have it like (may not be the best name):

```yaml
messageThrottle: 1ms
```

Here you'd be throttling logging to 1 message per millisecond.

jplock: Exactly what I thought. Duration has a count and time unit which we need in this case even though it’s not the best name.

ochedru:
Following your comments, I amended the PR to use a throttle defined by a `Duration`.

jplock:
`messageRate` or `messageFrequency` might make sense.  We should probably also add a comment (with an example) of how to understand the usage of `Duration`

> A `Duration` of `1m` means only allow one message per minute, while a `Duration` of `30s` means allow 30 messages per second

nickbabcock:
> Duration of 30s means allow 30 messages per second

Slight slip-up, `30s` would mean 1 message every 30 seconds 😄

jplock:
@nickbabcock this looks good to me 👍 

Great job @ochedru!

jplock:
We also need to update the release notes with this change.

nickbabcock:
Updated in https://github.com/dropwizard/dropwizard/commit/bfce31f6bbfeb41b12ec0ef629e4300a8b6fc245

PR #2390, opened 2018-06-13
Add Custom JCE support for Client side TLS connections (keystore and truststore)

nheitz:
###### Problem:
Client side TLS support in Dropwizard currently only supports custom providers equivalent to SunJSSE (secure socket implementations) but not for JCE pieces (eg keystore, key implementations).  The default SunJCE implementation implement weak password protection for PKCS12 keystores, and it might be preferable to upgrade to BouncyCastle in such situations.

###### Solution:
TlsConfiguration now supports specifying providers for keystore and truststore by named provider string. 

###### Result:
This should have no impact on business as usual code execution, but if you use stronger keystores, you now have the option of using alternate providers.


dropwizard-client/src/main/java/io/dropwizard/client/DropwizardSSLConnectionSocketFactory.java
```diff
@@ -121,11 +125,28 @@ private void loadTrustMaterial(SSLContextBuilder sslContextBuilder) throws Excep
...
+        KeyStore keyStore;
+
+        if (provider == null) {
+            keyStore = KeyStore.getInstance(type);
+
+        } else {
+            try {
+                keyStore = KeyStore.getInstance(type, provider);
+
+            } catch (KeyStoreException ignore) {
```
nickbabcock: If the configured provider doesn't work, should this exception be propagated instead of falling back -- in order to alert the user their provider is incorrect?

nheitz: I was emulating an initial implementation proposed by @arteam ....I suspect the motivation was to provide as soft an integration as possible, falling back to default providers.  
If you feel the WARN log is insufficient, however, I'm happy to propagate and enforce a more strict configuration correctness.

arteam: I agree, that WARN should be good enough. If a keystore is still supported by the default JVM providers, I don't see a reason not to load it.

arteam:
@nheitz Great work, thank you!

PR #2391, opened 2018-06-13
Update release notes with latest versions

nickbabcock:
Release notes were a bit out of date. I noticed a couple of our releases (v1.3.1 / v1.2.5) and (v1.3.2 / v1.2.6) had the exact same changelog so I condensed them into a single entry. Is this ok, or should I expand them out.

### DIFF
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -4,13 +4,30 @@
 Release Notes
 #############
 
+.. _rel-1.3.3:
+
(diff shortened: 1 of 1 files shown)


jplock:
I usually prefer them all separate but this looks fine to me. Thanks for taking care of this.

joschi:
I've updated the release notes in `master`, `release/1.3.x`, and `release/1.2.x`.

Each version is individually linkable (and it simplified cherry-picking in `release/1.2.x`).

PR #2395, opened 2018-06-15
Update to Jersey 2.27

nickbabcock:
###### Problem:
Took another crack at updating jersey. I've gotten most of the way there, the only issue remaining is that the Jersey update has a couple of outstanding issues:

https://github.com/jersey/jersey/issues/3670
https://github.com/jersey/jersey/issues/3675

Also, weirdly I had to play around with switching out `Feature` and `AbstractBinders`

EDIT: All tests are now passing


jplock:
Great progress @nickbabcock!

alex-shpak:
This is great!

Regarding https://github.com/jersey/jersey/issues/3675
I guess Jersey's team intention was for everybody to use `Binder` from jersey package in future. So maybe it will be good to get rid of direct hk2 usage in code.

alex-shpak:
Flickering `customJsonProviderClient` and `customJsonProviderRoundtrip` in `e2e` tests look like an ordering issue, similar to mentioned https://github.com/jersey/jersey/issues/3670

Adding `Rank(1)` from hk2 to `CustomJsonProvider` to force ordering solves that issue. But it doesn't look like a decent solution.

nickbabcock:
(I was going to comment on your PR, but figured I'd leave a comment here for full transparency)

@alex-shpak Excellent, thank you for looking into this with me 😄!

I came across the `rank` solution as well and the following diff also fixed the flaky tests

```diff
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JacksonFeature.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jackson/JacksonFeature.java
@@ -1,7 +1,9 @@
 package io.dropwizard.jersey.jackson;

 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.glassfish.jersey.internal.inject.AbstractBinder;

+import javax.inject.Singleton;
 import javax.ws.rs.core.Feature;
 import javax.ws.rs.core.FeatureContext;
 import javax.ws.rs.ext.MessageBodyReader;
@@ -19,7 +21,16 @@ public class JacksonFeature implements Feature {

     @Override
     public boolean configure(FeatureContext context) {
-        context.register(new JacksonMessageBodyProvider(mapper), MessageBodyReader.class, MessageBodyWriter.class);
+        context.register(new AbstractBinder() {
+            @Override
+            protected void configure() {
+                bind(new JacksonMessageBodyProvider(mapper))
+                    .to(MessageBodyReader.class)
+                    .to(MessageBodyWriter.class)
+                    .in(Singleton.class)
+                    .ranked(10);
+            }
+        });
         return true;
     }
 }
```

But I'm not knowledgeable about manipulating the order of components. Can you speak to why you prefer `@Produces` and `@Consumes` over `rank`? The ramifications of this means that whenever this PR lands, dropwizard users will need to modify their own code and I was hoping this could be a backwards compatible PR (though, I completely understand if it ends up not being backwards compatible).

alex-shpak:
I think there are 2 points for media type annotations:
 - Rank is not obvious enough, we then have to document this hardcoded number `10` so enduser have to put over 10 to add own provider.
 - By [javadoc](https://github.com/eclipse-ee4j/jersey/blob/12a057315292a624ceb31b58bd04eec3cb19a819/core-common/src/main/java/org/glassfish/jersey/message/MessageProperties.java#L105) that I mentioned before looks like `declaration distance` is default for ordering since `JAX-RS 2.x`, I think it's wise to order providers by how it meant by `JAX-RS`

On another hand we also potentially need to document how to make wildcard provider work.
Personally, I think in project that uses dropwizard I would use rank to register my own provider, since I control that code.

Anyway It is a breaking change.

nickbabcock:
> Anyway It is a breaking change.

When I used `rank` as shown in the snippet earlier, I didn't need to modify any e2e code, which is why I arbitrarily chose the rank of `10` because it didn't matter what rank was chosen (as long as it was `> 0`) -- a user supplied provider would take precedence.

alex-shpak:
This is interesting... while I was testing rank ordering turned that using `AbstractBinder` and skipping `raking` makes tests pass without need for mediatype annotations.
```java
context.register(new AbstractBinder() {
    @Override
    protected void configure() {
        bind(new JacksonMessageBodyProvider(mapper))
            .to(MessageBodyReader.class)
            .to(MessageBodyWriter.class)
            .in(Singleton.class);
    }
});
```

So maybe that is a solution 👍

alex-shpak:
>  a user supplied provider would take precedence.
I debugged it a bit.
User's custom provider will take precedence if registered via `ResourceConfig`. If registered via `AbstractBinder` it should have at least same rank as `JacksonMessageBodyProvider`

nickbabcock:
> If registered via AbstractBinder it should have at least same rank as JacksonMessageBodyProvider

IIRC this causes the tests to be flaky as then it is nondeterministic which provider will take precedence.

 What are you current thoughts on this? An internal `ranked` vs forcing users to annotate providers with `@Produces` and `@Consumes`. While it can be argued that users should have been annotating since the beginning, it is a break in behavior so will need to be considered.

I suppose jersey 2.26+ is already a breaking change for those using reactive jersey clients, but imo it's a relatively minor break. Users using hk2 in dropwizard may experience a bigger break, but I'd hate for this to be delayed for the foreseeable future if Dropwizard 2.0 is a significantly far from release.

alex-shpak:
After sleeping with it over weekend,

I think I still worry about hardcoded value for rank, even tho it brings less breaks into update. I can presee how one can spend hours debugging and then hit into hardcoded value.

Optionally we can keep rank now for compatibility, and change to media types annotations in `2.0.0` release, If merging Jersey `2.27` into `1.x.x` is an options.
I really would like to see Jersey 2.27 soon!

Back on ranking: [here](https://github.com/javaee/hk2/blob/14b92ddd06a69d71f80f354ed2c5bfba184cdd8d/hk2-locator/src/main/java/org/jvnet/hk2/internal/DescriptorComparator.java#L63) is how descriptors are ordered.
So two descriptors with same rank will come by order they were added, which is risky.

So by the end we have such options:
1. Move `JacksonMessageBodyProvider` to `AbstractBinder` instead of jersey resource config, which will put it from `customMbrs` to `mbrs` and eventually will move it down in list. [here](https://github.com/eclipse-ee4j/jersey/blob/12a057315292a624ceb31b58bd04eec3cb19a819/core-common/src/main/java/org/glassfish/jersey/message/internal/MessageBodyFactory.java#L214)
2. Set `@Consumes` and `@Produces` in `CustomJsonProvider` to lower declaration distance.
3. Use `@Rank(1)` annotation on `CustomJsonProvider` to move this provider upfront
4. Use `@Priority(Priorities.ENTITY_CODER)` on `CustomJsonProvider`, same as 3 but not dependent on `hk2`

I'm not sure that option 1 is less breaking than 2-4, even if there is no need to change test. It highly depends on how user registered his custom provider in first place.
Option 3, meanwhile, is relying on `hk2` which I think is bad.

nickbabcock:
I went with option 2 (annotate `CustomJsonProvider` with `Consumes` and `Produces`). All tests are now passing. Hopefully a migration guide will ensure the handful of dropwizard users who use custom providers to annotate them with `Consumes` and `Produces`. I'm not sure when this PR will land, but here is all the breaking changes I see:

- Dropwizard users registering hk2 components may need to update to the new API
- Dropwizard users using reactive jersey clients will need to rewrite to conform to the new reactive client api
- Dropwizard users writing custom message body writers / readers will need to ensure those classes are annotated with `Produces` and `Consumes`

Let me know if I missed anything. Thank you for the investigation @alex-shpak 😄

jplock:
@nickbabcock do you think this change should go into 2.0.0 or 1.4.0?  We should also start a wiki page on how to upgrade.

joschi:
@jplock @nickbabcock I'd rather move this to Dropwizard 2.0.0 since moving to Jersey 2.27 is quite a breaking change in public classes/methods and it will affect users building applications with Dropwizard and other DI-frameworks such as Spring DI or Guice.

In fact, I would skip Dropwizard 1.4.0 completely if there is nothing we eagerly want to release.

nickbabcock:
> In fact, I would skip Dropwizard 1.4.0 completely if there is nothing we eagerly want to release.

Yeah looks like @jplock and I also agree 2.0 is the next logical release

jplock:
I guess let's merge this in and start prepping 2.0.0-rc1

Tibor17:
@jplock 
When you want to cut new Version 2.0.0 or 2.0.0-RC1?

jplock:
@Tibor17 Nick opened https://github.com/dropwizard/dropwizard/issues/2418 with some Jersey issues we have to keep an eye on. I guess we could create an RC whenever. I’d like to submit a dropwizard-guava and dropwizard-jodatime modules but haven’t had time to work on them yet.

joschi:
> It seems there is a fix commit since 2017

The pull request (eclipse-ee4j/jersey#3817) hasn't been merged and we won't use a custom fork of Jersey in Dropwizard.

Tibor17:
@joschi 
I know you do not use it in dropwizard. I only said to use it local PC and announce this PR with test result.

PR #2400, opened 2018-06-19
Remove Guava

joschi:
There have been regular questions about the removal of Guava from Dropwizard on the mailing list and on GitHub.

Guava has been (or was; see [New Guava compatibility policy](https://groups.google.com/forum/#!msg/guava-discuss/rX-QXo-67ZU/gLEvfV4CAwAJ)) breaking backward compatibility quite often.

That imposed problems to people using Dropwizard to implement an application using another framework pulling in a different version of Guava, for example Apache Spark.

With the removal of Guava, these use cases would be easier to implement, and with Java 8 and later, there isn't too much critical functionality left which Guava contributes to Dropwizard.

This PR removes Guava from Dropwizard and replaces the [Guava Cache](https://github.com/google/guava/wiki/CachesExplained) with [Caffeine](https://github.com/ben-manes/caffeine).

If we wanted to be independent of a specific caching implementation, we could probably use JCache (JSR-107) and [Caffeine's JCache adapter](https://github.com/ben-manes/caffeine/wiki/JCache) internally.


joschi:
I haven't removed support for Guava from Jackson, JDBI, and Jersey (JAX-RS) yet.
We could extract that into their own modules, though.

jplock:
I was thinking we could introduce a dropwizard-guava and a dropwizard-jodatime modules, assuming it doesn’t make the rest of the code unmaintainable.  But let users opt in to using those.

dropwizard-assets/src/main/java/io/dropwizard/assets/AssetsBundle.java
```diff
@@ -91,8 +89,14 @@ public AssetsBundle(String resourcePath, String uriPath, String indexFile) {
      * @param assetsName          the name of servlet mapping used for this assets bundle
      */
     public AssetsBundle(String resourcePath, String uriPath, String indexFile, String assetsName) {
-        checkArgument(resourcePath.startsWith("/"), "%s is not an absolute path", resourcePath);
-        checkArgument(!"/".equals(resourcePath), "%s is the classpath root", resourcePath);
+        if (!resourcePath.startsWith("/")) {
```
jplock: Should we check if resourcePath is null here too?

joschi: It wasn't checked before. 🤷‍♂️

isaki-x: The first removed line is not null safe but the second is. So maybe the original intent was to be null safe? Also, it seems that this is no longer ensuring that the path is not the root. Is that the intention?

isaki-x:
@jplock I think that separating into their own modules is ideal or provide easy instructions for users how to inject Guava support into the Jackson/JDBI support provided by DropWizard. Additionally, it would be great if Jodatime was also split out so that those who don't use it (and instead use Java 8's time functionality) don't have to include Jodatime JARs and dependencies in their builds (or uber jars for those that shade).

dropwizard-assets/src/main/java/io/dropwizard/assets/AssetsBundle.java
joschi: @isaki-x I'd suggest checking the full diff of `AssetsBundle`. 😉

isaki-x: @joschi /facepalm

My bad.

ben-manes:
Thanks for this! 😄 

I don't know if its helpful, but I did notice an instrumented cache. A minor enhancement over Guava was to allow for a custom `StatsCounter`, with the conical [example](https://github.com/ben-manes/caffeine/tree/master/examples/stats-metrics) using Dropwizard. This allows for push-based stats collection if that helps for code reuse.

I would implore you to not use JCache if that is seriously considered in a later iteration. I think it is mostly harmful, as it was designed for selling vendorware and not for helping developers. I can go into technical details later if necessary.

mattnelson:
> I would implore you to not use JCache if that is seriously considered in a later iteration. I think it is mostly harmful, as it was designed for selling vendorware and not for helping developers. I can go into technical details later if necessary.

I would love to hear your thoughts on this topic, but this issue doesn't seem like the best forum for that type of discussion.

jplock:
@joschi did you want to handle introducing a `dropwizard-guava` module in this PR or do another PR into this branch?

nickbabcock:
Some thoughts as a I scrolled through:

- Though guava has been removed in all situations, I'm fine with guava in tests, dropwizard-example, and the e2e module. End users don't see these modules, so if Guava brings something to table in these situations, I can't think of reason to remove it.
- Replacing `ByteStreams::toByteArray` and `CharStream::toString` with a manual while loop copying data distracts from the meat of the function and should be extracted out to another function.
- Having two different `buildThreadFactory` that are close in implementation -- hopefully there is a way to reconcile them, so that one doesn't have to remember to update all instances when bugs / features come along.
- MediaType, oh man, is there no alternative than copying the class and the CharMatcher dependency? Looks like AssetServlet and TaskSevlet want it. IMO, it's not worth maintaining 3k lines of code for 20 lines of usage in AssetServlet
- In AssetServlet, `Pattern.compile("^/*(.*?)/*$").matcher(s)` is less readable (and may perform worse than) `CharMatcher.is('/')`, but I suppose there needs to be sacrifice somewhere 😅 
- I see some introduced classes appear unused like `Throwables::getRootCause`, or maybe I just didn't see the usage. 

Personally, I'm willing to give guava the benefit of the doubt on backwards compatibility, but I was pleasantly surprised that were you able to replace it so cleanly! Nice job! I will miss immutable classes, but the momentum appears to be removing guava, so I have no qualms

dropwizard-util/src/main/java/io/dropwizard/util/Sets.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public static <T> Set<T> of(T e1, T e2, T e3, T e4, T e5) {
+        final Set<T> set = new HashSet<>(5);
+        set.add(e1);
+        set.add(e2);
+        set.add(e3);
+        set.add(e4);
+        set.add(e5);
+        return set;
```
RohanNagar: Just a style suggestion for these methods:

```java
Set<T> set = Stream.of(e1, e2, e3, e4, e5).collect(Collectors.toSet());
```

isaki-x: The downside of that suggestion is that invoking `Stream.of` actually causes an extra object allocation which isn't really needed (and thus hurts performance). I personally feel that the current implementation is perfectly readable and thus changing it to use stream doesn't add any value and only brings detriments.

RohanNagar: @isaki-x sounds good, didn't think of the performance hit but that makes sense.

arteam:
@joschi I took the liberty pushing this change to the branch and running the tests on top of it.

jplock:
Should we change guava to be a test scoped dependency before merging?

arteam:
I think we still have some Guava-related code it in the Jersey and Jackson modules. We can get rid of the dependency after we move it to a Guava module.

alexey-wg2:
There is still io.dropwizard.util.Optionals class in the dropwizard-util module. Is it required?

Issue #2287, opened 2018-03-01
Jersey test client does not support PATCH requests


aharin:
Jersey test client provided by `DropwizardAppRule` does not support `PATCH` requests (Exception stack trace is attached). I suggest setting `JerseyClientBuilder.property(HttpUrlConnectorProvider.SET_METHOD_WORKAROUND, true)`  in `DropwizardAppRule.clientBulder` to fix this issue. Happy to submit a pull request.

```
avax.ws.rs.ProcessingException: java.net.ProtocolException: Invalid HTTP method: PATCH

	at org.glassfish.jersey.client.internal.HttpUrlConnector.apply(HttpUrlConnector.java:287)
	at org.glassfish.jersey.client.ClientRuntime.invoke(ClientRuntime.java:252)
	at org.glassfish.jersey.client.JerseyInvocation$3.call(JerseyInvocation.java:724)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:315)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:297)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:228)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:444)
	at org.glassfish.jersey.client.JerseyInvocation.invoke(JerseyInvocation.java:720)
	at org.glassfish.jersey.client.JerseyInvocation$Builder.method(JerseyInvocation.java:459)
	at co.wrisk.scoring.it.QuoteTestIT.shouldSelectQuoteItem(QuoteTestIT.kt:743)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #2410, opened 2018-06-25
Add support for PATCH request to ResourceTestRule client when using Grizzly

danpeter:
###### Problem:
When using the client from ResourceTestRule with the Grizzly container it is not possible to use PATCH via the custom annotation io.dropwizard.jersey.PATCH. (It seems to work with  the in-memory container)
Similar to: #2287 
###### Solution:
Since the builder takes a `Consumer<ClientConfig>`  instead of a property, like in #2287  I choose to set it in the Resource instead of manipulating the builder. This solution is up for discussion. 

```
   protected Resource buildResource() {
   ...
   Consumer<ClientConfig> extendedConfigurator = config -> {
                clientConfigurator.accept(config);
                config.property(HttpUrlConnectorProvider.SET_METHOD_WORKAROUND, true);
            };
```

###### Result:
Now possible to use the client and make a PATCH reuqest
client.target(...).request(...).method("PATCH", Entity.text("Patch is working"), String.class)

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/common/Resource.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/common/Resource.java
@@ -7,6 +7,7 @@ import io.dropwizard.jersey.validation.Validators;
 import io.dropwizard.logging.BootstrapLogging;
 import io.dropwizard.testing.junit5.ResourceExtension;
 import org.glassfish.jersey.client.ClientConfig;
+import org.glassfish.jersey.client.HttpUrlConnectorProvider;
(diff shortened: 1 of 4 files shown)


nickbabcock:
Looks our CI failed due to an unrelated test. Should be fixed with a commit I just pushed: https://github.com/dropwizard/dropwizard/commit/ff69c2dc23c2779c13f5bd87a96fb47ff3b458aa. Can you rebase off master? 😄

danpeter:
@nickbabcock I rebased but it is still failing

PR #2411, opened 2018-06-25
Fix for Issue 2408 (InvalidKeyException: Illegal key size)

nheitz:
###### Problem:
Previous development that introduced BouncyCastle as an "alternate" provider for PKCS12 handling also introduced the need for an unlimited strength key provider.  

###### Solution:
I've replaced the stronger keystore/truststore vairant with a weaker set, which should comply with Oracle key strength policy restrictions.

###### Result:
Unit tests should be able to run successfully in key strength-constrained environments.


nheitz:
wow...from java.security (this area has gotten weird and confusing):
# 3.  If the jar files are not present in the legacy location
#     and the "crypto.policy" Security property is not defined,
#     then the JDK will use the unlimited settings (equivalent to
#     crypto.policy=unlimited)

This is true from jdk8u152, apparently, and as of java 9, unlimited will be the default.

Glad it worked.

PR #2413, opened 2018-06-26
Simplify SelfValidatingValidator

joschi:
Remove use of Javassist from SelfValidatingValidator in favor of plain old Java.


jplock:
Can https://github.com/dropwizard/dropwizard/blob/master/dropwizard-validation/pom.xml#L46 be removed or do we use that in other places within the validation module?

joschi:
@jplock We're still using Javassist in [`UnitOfWorkAwareProxyFactory`](https://github.com/dropwizard/dropwizard/blob/v1.3.5/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java).

joschi:
@dropwizard/committers Maybe someone else could review the changes. I'm not sure that I didn't miss some side effects.

jplock:
I meant just remove the dependency from the validation pom for now

PR #2414, opened 2018-06-27
Document that the Timed and Metered can't be used at the same time.

tsundberg:
###### Problem:

The documentation didn't tell me that I can't annotate a method with `@Timed` and `@Meterd` at the same time.

###### Solution:

I added an Important box that says that `@Timed` includes `@Metered` and that an `IllegalArgumentException` is thrown when both are used at the same time. 

###### Result:

It is easier for a user to understand that `@Timed` and `@Meterd` can't be used at the same time.

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -1300,6 +1300,11 @@ Dropwizard augments Jersey to automatically record runtime information about you
 * ``@Metered`` measures the rate at which the resource is accessed
 * ``@ExceptionMetered`` measures how often exceptions occur processing the resource
 
+.. important::
+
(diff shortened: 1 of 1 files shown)


jplock:
Thanks!

joschi:
@tsundberg That's not completely correct. `@Timed` and `@Metered` can be used on the same resource method, as long as you override the name of at least one of them.

See also 99e307da9ced3d31dc5cc71847bfa5ca6c3ebcba.

jplock:
Thanks @joschi 👍

tsundberg:
@joschi, I had no idea :)

But I also see that you clarified the docs a bit more which is what I really was looking for.
Every small step in a good direction counts.

joschi:
@tsundberg Thanks for bringing attention to this issue. 😃

Issue #1700, opened 2016-08-19
Only TLSv1.2 Seems Supported


nickbabcock:
Running [sslyze](https://github.com/nabla-c0d3/sslyze) on our production Dropwizard 1.0.0 machines returns the following:

```
  * SSLV2 Cipher Suites:
      Server rejected all cipher suites.
  * SSLV3 Cipher Suites:
      Server rejected all cipher suites.

  * TLSV1 Cipher Suites:
      Server rejected all cipher suites.
  * TLSV1_1 Cipher Suites:
      Server rejected all cipher suites.
  * TLSV1_2 Cipher Suites:
      Preferred:                       
        TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256             ECDH-570 bits  128 bits                                                                  
      Accepted:                        
        TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256             ECDH-570 bits  128 bits                                                                  
        TLS_DHE_RSA_WITH_AES_128_CBC_SHA256               DH-1024 bits   128 bits                                                                  
        TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256             ECDH-570 bits  128 bits                                                                  
        TLS_DHE_RSA_WITH_AES_128_GCM_SHA256               DH-1024 bits   128 bits                                                                  
        TLS_RSA_WITH_AES_128_GCM_SHA256                   -              128 bits                                                                  
        TLS_RSA_WITH_AES_128_CBC_SHA256                   -              128 bits                                                                  

```

Issues detected so far:

- Default `curl` on CentOS 5.11 will not work
- `openssl s_client -prexit -tls1 -connect <host>:<port>` can't communicate with the server
- Some Java 6/7 clients are unable to communicate with the services.

The error in our logs (under DEBUG) are

```
DEBUG [2016-08-19 13:38:40,761] org.eclipse.jetty.server.HttpConnection:
! javax.net.ssl.SSLHandshakeException: no cipher suites in common
! at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)
! at sun.security.ssl.SSLEngineImpl.fatal(SSLEngineImpl.java:1666)
! at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:304)
! at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:292)
```

[...]

PR #2417, opened 2018-06-27
Disable protocols less secure than tls v1.2 by default

nickbabcock:
I'm feeling this out -- feel free to discuss 😄 

###### Problem:
Thanks to https://github.com/dropwizard/dropwizard/pull/2405 we'll see the following on startup for an HTTPS connector:

```
Enabled protocols: [TLSv1, TLSv1.1, TLSv1.2]
Disabled protocols: [SSLv2Hello, SSLv3]
```

You may recall that nearly two years ago I tripped up upgrading to Dropwizard 1.0 as [only TLSv1.2 seemed supported](https://github.com/dropwizard/dropwizard/issues/1700), so why is Dropwizard logging that TLSv1 and TLSv1.1 is enabled? Jetty would serve TLSv1.1 requests when the server runtime and client could agree on a cipher suite (nothing surprising here).  But by default, openssl could not communicate via TLSv1.1 with a Dropwizard server running on the latest oracle / openjdk versions. This can be verified with [sslyze](https://github.com/nabla-c0d3/sslyze), where it only lists TLS v1.2 cipher suites as possible. While one can use an alternate provider on the server (bouncycastle) -- if by default Dropwizard doesn't expose a non-TLSv1.2 cipher suite, what's the point of enabling TLSv1 and TLSv1.1? 

###### Solution:
Double down on the statement that [that only TLSv1.2 is enabled by default](https://github.com/dropwizard/dropwizard/issues/1700), by disabling everything that is less secure.

###### Result:
Secure by default! If someone has allowed TLSv1.1 cipher suites they'll now need to reset the `excludeProtocols` to the defaults

```yaml
server:
    applicationConnectors:
        - type: https
          port: 0
          keyStorePath: keystore.jks
          excludedProtocols:
```

On startup we'll see:

```
Enabled protocols: [TLSv1.2]
Disabled protocols: [SSLv2Hello, SSLv3, TLSv1, TLSv1.1]
```

When TLS v1.3 becomes supported, it will automatically be picked up and enabled.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -417,52 +417,52 @@ Extends the attributes that are available to the :ref:`HTTP connector <man-confi
           allowRenegotiation: true
           endpointIdentificationAlgorithm: (none)
 
-================================ ==================  ======================================================================================
-Name                             Default             Description
-================================ ==================  ======================================================================================
-keyStorePath                     REQUIRED            The path to the Java key store which contains the host certificate and private key.
-keyStorePassword                 REQUIRED            The password used to access the key store.
-keyStoreType                     JKS                 The type of key store (usually ``JKS``, ``PKCS12``, ``JCEKS``,
-                                                     ``Windows-MY``}, or ``Windows-ROOT``).
-keyStoreProvider                 (none)              The JCE provider to use to access the key store.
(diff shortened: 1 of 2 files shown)


jplock:
I think as long as we document this in an upgrading document, 2.0.0 would be a good time to do this.

PR #2426, opened 2018-07-09
Fix ResourceURL.isDirectory to handle Jar paths that are URL-escaped.

ewanmellor:
The heuristic inside the "jar" codepath in ResourceURL.isDirectory is
splitting the filename from the URL by String indexing.  This means that
if the file name would be URL-escaped (e.g. if it has a space in it) then
relativeFilePath is still escaped.  This manifests as zipEntry being null,
which then causes NullPointerException -> IOException ->
ResourceNotFoundException.

Fix this by using entry.getName() instead of taking a URL substring.
This works because jarConnection has been formed by correctly parsing the URL.


joschi:
@ewanmellor Thanks for your contribution!

Could you please add some unit tests validating the change and making sure that you didn't break the previous behavior?

joschi:
@ewanmellor Thanks for bringing this issue up! This will be fixed in #2674.

PR #2427, opened 2018-07-10
Allow constraint validators access to context types

### LINKED ISSUE #2234: Jersey @Context injection is not possible on ConstraintValidator instances
Jersey runtime allows for `@Context` injection on `ConstraintValidator` instances ([see manual](https://jersey.github.io/documentation/latest/bean-validation.html#bv.injecting)) via its `InjectingConstraintValidatorFactory` that is configured as the default `ConstraintValidatorFactory` implementation.

Somewhere between Dropwizard 0.9.2 and 1.1.3, the implementation of `dropwizard-validation` has changed such that these injections are no longer possible. To reproduce, use `@Context` injection in the constructor or field of a custom `ConstraintValidator`, and set it to validate a resource method parameter.

Ideally, it should be possible to continue to make use of this useful Jersey feature.

nickbabcock:
###### Problem:
Currently one can't use `@Context` injection in constraint validators (and other hooks in hibernate validator). (#2234)

###### Solution:
Everytime Jersey requests a new `ConfiguredValidator` we create a new one with the current context:

```java
final ValidatorContext context = factory.usingContext();
context.constraintValidatorFactory(resourceContext.getResource(InjectingConstraintValidatorFactory.class));
return new DropwizardConfiguredValidator(context.getValidator());
```

And just like that, our constraint validator receive context injections!

Not so fast.

For every request, a validator will need to be newly instantiated with the request context. The issue is that Dropwizard already sends in a fully configured validator that can't be used across multiple requests simultaneously for context injection to work. This required changing the `Environment` and `JerseyClientBuilder` public APIs which I am loathe to do (but had to).

###### Result:
Can use `@Context` inside constraint validators

public APIs that have had their signature changed (breaking changes):

```
JerseyClientBuilder::using(Validator) -> JerseyClientBuilder::using(ValidatorFactory)
Environment::new(..., Validator, ...) -> Environment::new(..., ValidatorFactory, ...)
Environment::setValidator(Validator) -> Environment::setValidatorFactory(ValidatorFactory)
Resource::setValidator(Validator) -> Resource::setValidatorFactory(ValidatorFactory)
```

To help those who may not care about context injection and just want a validator, I have decided to keep `Environment::getValidator`. It returns a single instance of a validator that created by the validator factory in `Environment` constructor or when a new factory is passed to `Environment::setValidatorFactory`

Closes #2234

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java
@@ -22,7 +22,7 @@ import org.glassfish.jersey.client.spi.ConnectorProvider;
 
 import javax.annotation.Nullable;
 import javax.net.ssl.HostnameVerifier;
-import javax.validation.Validator;
+import javax.validation.ValidatorFactory;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
 import javax.ws.rs.client.RxInvokerProvider;
@@ -61,7 +61,7 @@ public class JerseyClientBuilder {
     private JerseyClientConfiguration configuration = new JerseyClientConfiguration();
 
     private HttpClientBuilder apacheHttpClientBuilder;
-    private Validator validator = Validators.newValidator();
+    private ValidatorFactory validatorFactory = Validators.newValidatorFactory();
 
     @Nullable
     private Environment environment;
@@ -151,13 +151,13 @@ public class JerseyClientBuilder {
     }
 
     /**
-     * Use the given {@link Validator} instance.
+     * Use the given {@link ValidatorFactory} instance.
      *
-     * @param validator a {@link Validator} instance
+     * @param validatorFactory a {@link ValidatorFactory} instance
      * @return {@code this}
      */
-    public JerseyClientBuilder using(Validator validator) {
-        this.validator = validator;
+    public JerseyClientBuilder using(ValidatorFactory validatorFactory) {
+        this.validatorFactory = validatorFactory;
         return this;
     }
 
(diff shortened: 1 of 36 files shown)


alex-shpak:
Maybe you would be interested, I came up with some alternative (hacky?) solution
https://github.com/alex-shpak/dropwizard-hk2bundle/blob/master/src/main/java/net/winterly/dropwizard/hk2bundle/validation/InjectValidatorBundle.java

Idea is to set custom `ConstraintValidatorFactory` which reuses default `ConstraintValidatorFactoryImpl` before jersey initialisation and `InjectingConstraintValidatorFactory` after.

nickbabcock:
Wow, and that works with constraint validators with context? This may be grounds for a less API breaking implementation.

I had assumed that a `ResourceContext` shouldn't be resolved once (if I'm reading your code correctly), and that each request had their own `ResourceContext`. But a smoke test did show that only one resource context was allocated across multiple requests, so it looks like you're right.

As a side note, I based this PR off a small part of Jersey's own validation binder, but that could be overkill for our situation.

alex-shpak:
Yes, in my case ResourceContext resolved once. I think single `ConfiguredValidator` created per request, but not `ResourceContext`

I don't have confidence that it's fully functional, more testing definitely required.
But according to [this](https://github.com/jersey/jersey/blob/2.25.x/core-server/src/main/java/org/glassfish/jersey/server/internal/JerseyResourceContext.java#L278) it looks like `ResourceContext` is singleton in jersey `2.25`

nickbabcock:
I feel like there will need to be a breaking change somewhere, as if someone uses `Environmnet::setValidator` there is no chance to inject the factory (eg. `JerseyClientBuilder` uses `Environment::getValidator`).

alex-shpak:
But maybe distribute it as bundle? So that users who need injection just add it

alex-shpak:
@nickbabcock 
Here I made an example https://github.com/dropwizard/dropwizard/pull/2432
Looks like it works, one can use `@Context` as well `@Inject` in custom `ConstraintValidator`

nickbabcock:
Closing in favor of #2432

PR #2428, opened 2018-07-10
Seperate readonly session

scottaj:
###### Problem:
Our company uses dropwizard with hibernate, but our production database is actually a cluster with a write master and read replicas. We want our dropwizard app to be able to connect to either the master or the replicas on a per-request basis depending on the endpoint.

Someone from our company previously opened #2248 which was basically a version of the internal fork of dropwizard hibernate we maintain, and works by decoupling the DAO from a specific session factory.

After some changes we've made internally, we realized the problem with that approach is that it breaks an existing workflow that allows you to have multiple datasources in your app for totally different databases, and have different DAOs connect to those. 

###### Solution:
This is a second attempt to solve the problem stated above with a different approach. Instead of decoupling the DAO from the session factory, we replace the session factory provided to DAOs with a Dropwizard class that can hold two different connections, one to a write master and one to a read replica. That class also has a `ThreadLocal` that is set based on the `readOnly` property on the `UnitOfWork`, and if it is true, returns the reader connection. 

This is all configured at the datasource configuration level, where `PooledDataSourceFactory` now has an optional parameter for a read only URL. If that property is specified, two managed data sources will be created instead of one, and the second will be connected to the read URL.

###### Result:
For projects using dropwizard-hibernate, you can now specify a `readOnlyUrl` property on your database configuration, and add `readOnly=true` to the `UnitOfWork` annotations on your endpoints to have those endpoints hit a read replica connection you specify. 

If you do not specify a `readOnlyUrl`, it will only make one connection and use that regardless of the `readOnly` setting (as it does currently). So there should be no performance or resource overhead added if you do not use this feature.

Unfortunately, this is a breaking change, because the signature of `AbstractDAO` has changed to take in a different class. The migration should be pretty easy for most projects, unless they are manually doing something with the hibernate session factory that used to get passed in there. I could not think of a way to accomplish this change without changing the AbstractDAO.

Other changes:

Since I changed the base database configuration that is shared by hibernate, liquibase, and JDBI, I also made some minor changes to the latter two integrations.

*Liquibase:*
No real change, but always uses the write connection and never a read one you specify.

*JDBI and JDBI3:*
Added a `buildReadOnly` method to both factory classes that will build a instance using the `readOnlyUrl` if it exists.


nickbabcock:
Sorry I've let this PR languish for so long, hibernate-orm is not my strong suit, but I don't want to leave you out in the cold! 😅 

Some questions:

1. You mention a fork of this code. Have you used this readonly feature successfully (ie: how confident are you)
2. What's the reasoning behind the `hibernate-testing` package?
3. Are there alternatives to this approach? I know this isn't hibernate-orm, but could one have multiple JDBI configurations -- one just points to the readonly instance?

scottaj:
Sorry, I missed the email update of this PR!

1. We have a fork of some similar but more hacked together code running in production. We have a chore to update that app to use this version, but haven’t had the resources to do that. I do have a test app very bare-bones drop wizard hibernate app (I could look into publishing it) that demonstrates the feature working. I am pretty confident in the code because the approach is pretty simple. It is just some extra logic around handling multiple database connections (which drop wizard already supports) and some changes to DAOs to not tightly couple them to a single database connection.
I’m happy to take any steps you think are necessary to demonstrate it is working properly

2. Circular dependency. Dropwizard-hibernate depends on dropwizard-testing, and was able to get away with that by having drop wizard-testing only depend on hibernate-core and not the drop wizard hibernate package. I replaced some publicly exposed hibernate-core classes with drop wizard classes that wrap them, so to keep the hibernate tests helper classes around would have required making dropwizard-testing depend on dropwizard-hibernate. So I moved those test helper classes to their own package to resolve that. Hopefully that explanation was not too confusing

3. We’ve looked a lot into alternatives, none are available that really work for us. There are some drivers for certain DBs that claim to do it, but we didn’t really have any luck with that, and it is not a general purpose solution. From the dropwizard side, I don’t think there is a way to do this with the current dropwizard-hibernate implementation. You can have multiple connections, but the way DAO classes are designed in hibernate right now means they are fundamentally tightly coupled to one particular database connection, regardless of what connection you set up in your `@UnitOfWork` annotation. This means that to have read and write connections right now, you would have to have separate DAOs and duplicate code across them for any read actions. From a practical perspective (at least in our case) this would lead to an unacceptable amount of duplicated code in our codebase, because many of our write operations also require reads.

joschi:
@scottaj Thanks for opening this PR and sorry for the late response!

Would it also solve your issue, if Dropwizard allowed multiple sessions in `@UnitOfWork`, as discussed in #2360?

scottaj:
@joschi That feature is cool, but unfortunately I don't think it addresses our use case. The biggest technical limitation we are encountering is that a DAO class has to be tightly coupled to a particular `SessionFactory`. 

#2360 is nice if you want to transactionally access multiple distinct datasources (with their own distinct DAO classes) in the same request cleanly. But we want to access what is essentially the same datasource, using the same DAO's, just at two different endpoints depending on context.

PR #2430, opened 2018-07-12
Fix incorrect reading of somaxconn for tcp backlog on linux

### LINKED ISSUE #2429: acceptQueueSize defaults to an incomplete value from somaxconn on Linux
A call to io.dropwizard.jetty.NetUtil.getTcpBacklog() reads just the first digit of /proc/sys/net/core/somaxconn on Linux, effectively making the default acceptQueueSize just 1 or 2 (depending on the actual value of somaxconn; verified with 128 and 2048). This in turn results in increased number of failed connections to DW (connection timeouts) when the server is under stress. This seems to be a regression between 0.9.x. and 1.0.x (because in 0.9.x, somaxconn was not explicitly used as a source of a default value).

The issue seems to be specifically in Files.readAllBytes(paths.get('/proc/sys/net/core/somaxconn')), in that the SeekableByteChannel in readAllBytes is treating the /proc fs as a regular file, which it is not (seeks aren't possible, nor filesize can be read ahead). Thus, in effect it only reads the first byte of somaxconn (instead of the entire string value), passing it without any further warning into acceptQueueSize as a default.

nickbabcock:
###### Problem:
Dropwizard drops connections during a burst of short lived connections, as the accept queue is set to (most commonly) 1 or 2 due to `Files.readAllBytes` unable to work with pseudo files like somaxconn

###### Solution:
[Adopt Netty's implementation](https://github.com/netty/netty/blob/77ec8397927e3ceb9b9a447a74e718f625ed9976/common/src/main/java/io/netty/util/NetUtil.java#L261-L269)

###### Result:
Test passed and dropwizard should be more burst resistant. Closes #2429

Dropwizard users currently afflicted by this bug, can override this default behavior by specifying an `acceptQueueSize` in their configuration.

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/NetUtil.java
@@ -15,13 +15,13 @@
  */
 package io.dropwizard.jetty;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


OndraZizka:
Just a note. You say that affected users can override it in settings. However, the system limit still applies.
Users trying beyond that will not get over that effectively, AFAICT. To override:

```
sysctl -w net.core.somaxconn=65535
```
Typically added to `sudo nano /etc/rc.local`.

OndraZizka:
I suspect there is still another bug in Jetty in how it handles the sockets with refused connections. Setting higher accept queue will make it close to impossible to hit in real world scenarios though.

When we hit this issue, we see connection timeouts on the client side after 60 seconds. If it was just the accept queue full, the client would get the rejection and fail the connection right away. But it does not - instead, it keeps waiting. I think what is happening is described here: https://blog.cloudflare.com/this-is-strictly-a-violation-of-the-tcp-specification/
In short, Jetty doesn't `close()` the rejected connections properly.

OndraZizka:
Or maybe not. Petr Kacer said he checked with `ss` and there were no `CLOSE_WAIT` pending.

PR #2432, opened 2018-07-13
Inject validator feature

### LINKED ISSUE #2234: Jersey @Context injection is not possible on ConstraintValidator instances

alex-shpak:
Following discussion at https://github.com/dropwizard/dropwizard/pull/2427

###### Problem:
Somewhere during update to `1.0.0` we lost ability to inject into custom `ConstraintValidator`s because validator factory was moved into Dropwizard.

###### Solution:
PR Introduces `InjectValidatorBundle` to enable validator injections inside jersey resources. Note that validators for dropwizard configuration are not affected because `ConstraintValidatorFactory` is updated after jersey initialisation.

Under the hood: There is new `MutableValidatorFactory` registered by default in validator configuration that allows to swap factory at runtime. So after jersey initialisation is possible to change `ConstraintValidatorFactoryImpl` with `InjectingConstraintValidatorFactory`

###### Result:
Closes #2234, making injections available in ConstraintValidators

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/validation/GetLocatorFeature.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/validation/GetLocatorFeature.java
@@ -1,0 +1,26 @@
+package io.dropwizard.validation;
+
+import org.glassfish.hk2.api.ServiceLocator;
+
+import javax.inject.Inject;
+import javax.ws.rs.core.Feature;
+import javax.ws.rs.core.FeatureContext;
+import java.util.function.Consumer;
+
+public class GetLocatorFeature implements Feature {
+
+    private final Consumer<ServiceLocator> action;
+
+    @Inject
+    private ServiceLocator serviceLocator;
+
+    public GetLocatorFeature(Consumer<ServiceLocator> action) {
+        this.action = action;
+    }
+
+    @Override
+    public boolean configure(FeatureContext context) {
+        action.accept(serviceLocator);
+        return true;
+    }
+}
--- a/dropwizard-core/src/main/java/io/dropwizard/validation/GetResourceContextFeature.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/validation/GetResourceContextFeature.java
@@ -1,26 +1,25 @@
 package io.dropwizard.validation;
 
-import org.glassfish.hk2.api.ServiceLocator;
-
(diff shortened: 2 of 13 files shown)


nickbabcock:
Excellent, this does appear to be a less intrusive change! Dare I say this backwards compatible!?

> Somewhere during update to 1.0.0 we lost ability to inject into custom ConstraintValidators because validator factory was moved into Dropwizard.

You're (probably) looking at them (#1251) 😄 

As you've already stated tests will be needed for a more thorough review. Ideally there will be two types of tests: one to show that context injections works with this bundle, and one to show that context injection doesn't work without.

Still mulling over if the bundle should be registered by default (eg. any downsides) 🤔

alex-shpak:
I added tests:
 - In core: `InjectValidatorBundleTest` unit tests for bundle
 - in e2e
   - `InjectValidatorBundleTest` test that injection work as well as usual validators.
   - `DefaultValidatorTest` regression test, that validators work with/without bundle.

I'm not sure where to place them correctly, so in case if something not on it's place please guide me.

From tests it looks like change is backward compatible, even with bundle registered. The only option that comes to my mind that if somebody had `@Inject` in constraint validator from before `1.0.0` and kept it until now.

About registering bundle by default:
While examining code I haven't found other bundles registered by default. I'm not sure is there reason for that or convention.

Also to mention, the only reason of having this functionality in bundle is because of shared instance of `MutableValidatorFactory` between `initialize` and `run` methods.

jplock:
@nickbabcock any more thoughts on this one?  Enabling by default sounds ok to me in 2.0.0

nickbabcock:
Yeah this PR is good, I'll open another issue discussing making this the default.

PR #2453, opened 2018-08-03
Rewrite of throttling logging appender testing

isaki-x:
- Renamed test class to match the class it is testing
(ThrottlingAppenderWrapper).
- Rewrote class to play nicer with slower processing and to improve long
term supportability.
- Relied on system property for line seperation instead of hardcoded
regex.
- Removed unused imports (and in the class being tested as well).

###### Problem:
When CI takes awhile, the unit test fails. Also, I like cleaning up compiler warnings.

###### Solution:
I've rewritten key pieces of the original unit test to be less dependent on the test finishing in a certain time window.  These changes aim to dynamically address the assertion targets based on how long the test took to run.

###### Result:
The unit test passes even if the CI machine is running slow.


isaki-x:
You may notice two commits; I accidentally checked in a POM change to make my Eclipse happy (as I didn't feel like moving the plugin to a profile; which I will do in a different PR at some point).

isaki-x:
Too many commits, will submit clean PR.

PR #2460, opened 2018-08-18
Extend from AbstractHandlerContainer instead of AbstractHandler

### LINKED ISSUE #2455: Dropwizard sessions leak memory
From what I have gathered, the recommended way to support sessions in Dropwizard is to add a `SessionHandler()` to the environment:

```java
environment.servlets().setSessionHandler(new SessionHandler());
```

This however, leads to a memory leak with `org.eclipse.jetty.server.session.SessionData` objects not being cleaned up. I have set the max inactive interval and session cache eviction policy to `SessionCache.EVICT_ON_INACTIVITY`, which means that session should expire after 1s of inactivity. Profiling the live objects on the heap in jmap show the following, after repeated `curl` requests to the server:

```console
num   #instances      #bytes  class name
-----------------------------------------------------------------------------------------------
31:         1315      126240  org.eclipse.jetty.server.session.SessionData
43:         1315       63120  org.eclipse.jetty.server.session.Session
44:         1315       63120  org.eclipse.jetty.server.session.Session$SessionInactivityTimeout
```

These session objects are never cleaned up and eventually lead to an OOM.

After a little bit of investigation, the issue seems to be coming from https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jetty/src/main/java/io/dropwizard/jetty/ContextRoutingHandler.java#L18
The class extends `AbstractHandler` instead of `AbstractHandlerContainer` (and doesn't override https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractHandlerContainer.java#L74). This is necessary for the session cleanup, because the children `SessionHandler` for all contexts need to be discovered by the jetty server.

The cleanup is triggered by jetty hosuekeeper's call to `scavenge()` in the `SessionHandler()`
- https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java#L1276
- https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionIdManager.java#L472
- https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionIdManager.java#L525

The call to `_server.getChildHandlersByClass(SessionHandler.class);` returns an empty list because when examining the chain of handlers, `ContextRoutingHandler` hides the fact that it contains `MutableServletContextHandler` because it doesn't expose its children handlers correctly.

I think the solution would be to have `ContextRoutingHandler` extend `AbstractHandlerContainer` and override `expandChildren(...)`, but I will defer to the maintainers, who will probably understand this better!

---
Update:
It seems that more than just the session data contains a memory leak - there is a also a leak with the `SSLSessionImpl`, which shows in jmap:

```console
num   #instances      #bytes  class name
-----------------------------------------------------------------------------------------------
 6:       188528    19661252  sun.security.ssl.SSLSessionImpl
28:       188528     3095728  sun.security.ssl.SSLSessionId
```

jplock:
Fixes #2455

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/ContextRoutingHandler.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/ContextRoutingHandler.java
@@ -4,7 +4,6 @@ import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.handler.AbstractHandlerContainer;
 import org.eclipse.jetty.util.ArrayTernaryTrie;
-import org.eclipse.jetty.util.Trie;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
@@ -16,7 +15,7 @@ import java.util.Map;
  * A Jetty router which routes requests based on context path.
  */
 public class ContextRoutingHandler extends AbstractHandlerContainer {
-    private final Trie<Handler> handlers;
+    private final ArrayTernaryTrie<Handler> handlers;
 
     public ContextRoutingHandler(Map<String, ? extends Handler> handlers) {
         this.handlers = new ArrayTernaryTrie<>(false);
@@ -57,6 +56,6 @@ public class ContextRoutingHandler extends AbstractHandlerContainer {
 
     @Override
     public Handler[] getHandlers() {
-        return handlers.keySet().stream().map(key -> handlers.get(key)).toArray(Handler[]::new);
+        return handlers.entrySet().stream().map(e -> e.getValue()).toArray(Handler[]::new);
     }
 }
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/ContextRoutingHandler.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/ContextRoutingHandler.java
@@ -2,7 +2,7 @@ package io.dropwizard.jetty;
 
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.Request;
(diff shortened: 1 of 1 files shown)


dropwizard-jetty/src/main/java/io/dropwizard/jetty/ContextRoutingHandler.java
```diff
@@ -54,4 +54,9 @@ protected void doStop() throws Exception {
             handlers.get(key).stop();
         }
     }
+
+    @Override
+    public Handler[] getHandlers() {
+        return handlers.keySet().stream().map(key -> handlers.get(key)).toArray(Handler[]::new);
```
mattnelson: Changing `Trie` to `ArrayTernaryTrie` introduces `entrySet` which will eliminate the lookup in a loop.

```java
private final ArrayTernaryTrie<Handler> handlers;
handlers.entrySet().stream().map(e -> e.getValue()).toArray(Handler[]::new);
```

https://www.eclipse.org/jetty/javadoc/9.4.11.v20180605/org/eclipse/jetty/util/ArrayTernaryTrie.html#entrySet--

jplock:
Thanks @mattnelson, fixed!

jplock:
AppVeyor failed with:

```
[ERROR] Errors: 
5910[ERROR]   JerseyClientIntegrationTest.testBufferedGzipPost:111->postRequest:209 � Processing
```

jraylee:
Would it be possible to have this included in the 1.3.6 release?

PR #2462, opened 2018-08-20
Fixed deprecated class usage (Apache StrSubstitutor)

isaki-x:
###### Problem:
The `StrSubstitutor` class is deprecated as of Apache commons-text 1.3. This will be removed in 2.0. See: https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StrSubstitutor.html

**NOTE:** This is a breaking change.

###### Solution:
I have replaced the deprecated classes with their replacements as per the Apache documentation (with the exception of `StrLookup`; its documentation states to use `StringLookupFactory` but in our case, we needed to instead implement `StringLookup` which is the interface that replaced the abstract `StrLookup` class).

###### Result:
This eliminates compiler warnings related to using a deprecated class. It also will allow a smooth upgrade path for commons-text when the time comes.


jplock:
Looks to me, thanks for taking care of this!

jplock:
@isaki-x would you mind adding a release notes entry about this in https://github.com/dropwizard/dropwizard/blob/master/docs/source/about/release-notes.rst ?

PR #2463, opened 2018-08-22
Fix new jersey 2.27 warnings

jbarnettwomply:
Register instances using an AbstractBinder instead of
registering them directly

###### Problem:
This should fix all the warnings in https://github.com/dropwizard/dropwizard/issues/2418

###### Solution:
Instead of registering Resource instances directly, register the .class and let jersey instantiate it.
In the case where there are constructor args, use an AbstractBinder that binds the instance to the class.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -1,5 +1,7 @@
 package io.dropwizard.jersey;
 
+import static java.util.Objects.requireNonNull;
+
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.jersey2.InstrumentedResourceMethodApplicationListener;
 import com.fasterxml.classmate.ResolvedType;
@@ -9,6 +11,11 @@ import io.dropwizard.jersey.params.AbstractParamConverterProvider;
 import io.dropwizard.jersey.sessions.SessionFactoryProvider;
 import io.dropwizard.jersey.validation.FuzzyEnumParamConverterProvider;
 import io.dropwizard.util.Strings;
+import javassist.ClassPool;
+import javassist.CtClass;
+import org.glassfish.hk2.utilities.Binder;
+import org.glassfish.jersey.internal.inject.AbstractBinder;
+import org.glassfish.jersey.internal.inject.Providers;
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.server.ServerProperties;
 import org.glassfish.jersey.server.model.Resource;
@@ -20,7 +27,6 @@ import org.glassfish.jersey.server.monitoring.RequestEventListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -30,8 +36,10 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.UUID;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 public class DropwizardResourceConfig extends ResourceConfig {
     private static final Logger LOGGER = LoggerFactory.getLogger(DropwizardResourceConfig.class);
@@ -124,8 +132,50 @@ public class DropwizardResourceConfig extends ResourceConfig {
         return allClasses;
     }
 
+    @Override
+    public ResourceConfig register(final Object component) {
+        final Object object = requireNonNull(component);
+        Class<?> clazz = object.getClass();
+        // If a class gets passed through as an object, cast to Class and register directly
+        if (component instanceof Class<?>) {
+            return super.register((Class<?>) component);
+        } else if (Providers.isProvider(clazz) || Binder.class.isAssignableFrom(clazz)) {
+            // If jersey supports this component's class (including hk2 Binders), register directly
+            return super.register(object);
+        } else {
+            // Else register a binder that binds the instance to its class type
+            try {
+                // Need to create a new subclass dynamically here because hk2/jersey
+                // doesn't add new bindings for the same class
+                ClassPool pool = ClassPool.getDefault();
+                CtClass cc = pool.makeClass(SpecificBinder.class.getName() + UUID.randomUUID());
+                cc.setSuperclass(pool.get(SpecificBinder.class.getName()));
+                Object binderProxy = cc.toClass().getConstructor(Object.class, Class.class).newInstance(object, clazz);
+                super.register(binderProxy);
+                return super.register(clazz);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
     static String cleanUpPath(String path) {
         return PATH_DIRTY_SLASHES.matcher(path).replaceAll("/").trim();
+    }
+
+    public static class SpecificBinder extends AbstractBinder {
+        private Object object;
+        private Class clazz;
+
+        public SpecificBinder(Object object, Class clazz) {
+            this.object = object;
+            this.clazz = clazz;
+        }
+
+        @Override
+        public void configure() {
+            bind(object).to(clazz);
+        }
     }
 
     private static class EndpointLogLine {
--- a/dropwizard-jersey/src/test/java/io/dropwizard/jersey/DropwizardResourceConfigTest.java
+++ b/dropwizard-jersey/src/test/java/io/dropwizard/jersey/DropwizardResourceConfigTest.java
@@ -1,5 +1,7 @@
 package io.dropwizard.jersey;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import io.dropwizard.jersey.dummy.DummyResource;
 import io.dropwizard.logging.BootstrapLogging;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
@@ -16,8 +18,6 @@ import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Application;
 import javax.ws.rs.core.MediaType;
-
-import static org.assertj.core.api.Assertions.assertThat;
 
 public class DropwizardResourceConfigTest {
     static {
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -536,7 +536,7 @@
                 <plugin>
                     <groupId>org.owasp</groupId>
                     <artifactId>dependency-check-maven</artifactId>
-                    <version>3.3.0</version>
+                    <version>3.3.1</version>
```
jbarnettwomply: bumped this version to avoid https://github.com/jeremylong/DependencyCheck/issues/1394

dropwizard-testing/src/main/java/io/dropwizard/testing/common/DropwizardTestResourceConfig.java
```diff
@@ -46,7 +46,7 @@
         register(new JacksonFeature(configuration.mapper));
         register(new HibernateValidationBinder(configuration.validator));
         for (Supplier<?> singleton : configuration.singletons) {
-            register(singleton.get());
+            register(singleton.get().getClass());
```
serhiypal: Not sure this will work with `PeopleResourceTest`/`PersonResourceTest`

jbarnettwomply: yep, meant to have an abstractbinder here too, but missed from the diff.  still failing with that, so working through it.  pushing updated wip diff now

jbarnettwomply:
still a couple of PersonResource constructors to clean up -- doing so from the DropwizardTestResourceConfig isn't quite working for me yet.

dropwizard-core/src/test/java/io/dropwizard/server/DefaultServerFactoryTest.java
```diff
@@ -150,7 +151,13 @@ public void testGracefulShutdown() throws Exception {
         CountDownLatch requestReceived = new CountDownLatch(1);
         CountDownLatch shutdownInvoked = new CountDownLatch(1);
 
-        environment.jersey().register(new TestResource(requestReceived, shutdownInvoked));
+        environment.jersey().register(new AbstractBinder() {
```
jplock: this is pretty crazy this is the only way to do this now.  :(

dropwizard-testing/src/main/java/io/dropwizard/testing/common/DropwizardTestResourceConfig.java
```diff
@@ -46,7 +46,21 @@
         register(new JacksonFeature(configuration.mapper));
         register(new HibernateValidationBinder(configuration.validator));
         for (Supplier<?> singleton : configuration.singletons) {
-            register(singleton.get());
+            Object object = singleton.get();
+//            Class<?> clazz = object.getClass();
```
serhiypal: Do we want to clean this thing?

jbarnettwomply: yep, that's where I was struggling with getting the PersonResource tests to pass & not warn, updated PR now.

dropwizard-testing/src/test/java/io/dropwizard/testing/app/PersonResource.java
```diff
@@ -32,6 +33,10 @@ public PersonResource(PeopleStore store) {
         this.store = store;
     }
 
+    public PersonResource() {
+        this.store = Mockito.mock(PeopleStore.class);
+    }
```
jbarnettwomply: Without this, the `PersonResourceExceptionMapperTest` is unable to instantiate a PersonResource, despite the bytebuddy enhanced binder that should be able to bind PersonResource to the constructed instance passed to the `ResourceTestRule`.  No idea why :(

dropwizard-testing/src/main/java/io/dropwizard/testing/common/DropwizardTestResourceConfig.java
```diff
@@ -48,6 +64,27 @@
         for (Supplier<?> singleton : configuration.singletons) {
             register(singleton.get());
         }
+        for (Supplier<?> resource : configuration.resources) {
```
jbarnettwomply: without separating out Resources from Provider singletons, dropwizard-example's `Protected*ResourceTest`s fail trying to bind/register a `Class`

jbarnettwomply:
Sorry for the lots-of-intermediate-versions of this PR, but the current one now both passes and does not throw any of the `org.glassfish.jersey.internal.inject.Providers` warnings

jbarnettwomply:
It's possible that we could add something like the `DropwizardTestResourceConfig` block that's using bytebuddy to `.jersey().register()` -- I'm not sure how safe or effective that would really be given the trouble I've had getting it to work (the 2 comments I added today to explain some of the changes makes me not super comfortable with that area of the work so far).   If i understood more about why the `.jersey().register()`s work better than the `ResourceTestRule` invoked registrations, I might go down that path.  But it's worth a shot, just not sure I have time to do so at the moment.

jbarnettwomply:
I think you're right though -- I know we do `jersey().register()` on resource instances, and assume other downstreams do as well.

nickbabcock:
Sorry I won't have the time to look at this for the next several days, but last I knew the jersey team recognized this shortfall and would be willing to allow a fix for the next version. So I don't want to spend additional time mapping out a migration guide that this PR would entail for users who will inevitably wind up here due to the warnings they will now see in their apps -- only to have the migration become obsolete as soon as the jersey version is released.

I haven't checked everything out, so if this PR has become necessary, let me know.

jbarnettwomply:
@nickbabcock based on the comments in https://github.com/eclipse-ee4j/jersey/issues/3700 I'm not sure they're keen to fix this shortfall.  However, based on @jplock's suggestion, updated the change to be much smaller and not require downstream migrations.

Couple semi-strange things to note, but functional and I think ok; would love feedback from someone who knows jersey/hk2 better:
1)  Need to explicitly check for hk2 Binders (as opposed to jersey.internal Binders), as Providers.isProvider() only recognizes the jersey.internal ones.
2)  Need to dynamically subclass the binder, as jersey appears to only actually register the first instance of a given class (ComponentBag.registerModel keys duplicates off the componentClass, and then the instance doesn't get added to the bag)

dropwizard-jersey/pom.xml
```diff
@@ -113,5 +113,10 @@
             <groupId>javax.activation</groupId>
             <artifactId>javax.activation-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy</artifactId>
+            <version>1.8.13</version>
```
jplock: this should probably be defined up in `dropwizard-bom`

jbarnettwomply: updated

dropwizard-bom/pom.xml
```diff
@@ -53,6 +53,11 @@
                 <artifactId>guava</artifactId>
                 <version>${guava.version}</version>
             </dependency>
+            <dependency>
+                <groupId>net.bytebuddy</groupId>
+                <artifactId>byte-buddy</artifactId>
```
mattnelson: Is the usage of bytebuddy introducing possible fragility in relation to the transitive dependency on asm which generally requires changes based on each new version of java. Which is now occurring at a more rapid pace.

jbarnettwomply: That's a valid question -- if there's another (safer) way to dynamically subclass the SpecificBinder, or to register it in such a way that multiple instances of it actually get bound by hk2, I'd love suggestions on that.

jbarnettwomply: Also note that mockito-core and hibernate-core already both pull in byte-buddy as a dependency (mockito only in test scope though)

nickbabcock: hibernate-core is only brought in with dropwizard-hibernate, so this would still be a new dependency to non-hibernate orm users.

You may be able to accomplish the [same thing using Javassist](https://stackoverflow.com/a/17282014/433785), which is already a dependency of dropwizard-core.

nickbabcock:
@jbarnettwomply thank you. While considering instance based resource classes incorrect is working as designed according to the jersey team, I still find their lack of reasons for this decision unfortunate. No one wants to dance with a DI framework and walk deeper into annotation hell just to complete Dropwizard's getting started guide. So, yes, I believe the solution here presents the best way forward.

jbarnettwomply:
there's a `register(Class)` method that handles that case mostly -- I suppose I could add to the exception on line 139 to check for `instanceof Class` in case there's something like a `Collection<Object>` being iterated through and registered via `register(Object)` , and that collection containing a Class instance

isaki:
If there is already a `register(Class)` implementation (which I admittedly missed), then you are golden; the JVM should choose the Class one for Classes and the Object one for everything else automatically.

isaki:
That might be a good idea. I wrote a quick test to see if it is required and it is. Java only chooses the specific when it is being called on the type that implements the more specific method and when it is not lost due to type erasure.

```Java
import java.util.ArrayList;
import java.util.Collection;

import org.junit.Test;

public class PolyTest {

    @Test
    public void testA() {
        final TestIFace test = new TestHelper("A");

        final Collection<?> x = this.getCollection();
        for (final Object o : x) {
            test.handle(o);
        }
    }

    @Test
    public void testB() {
        final TestHelper test = new TestHelper("B");

        final Collection<?> x = this.getCollection();
        for (final Object o : x) {
            test.handle(o);
        }
    }

    @Test
    public void testC() {
        final TestIFace test = new TestHelper("C");
        test.handle(this.getClass());
        ((TestHelper) test).handle(this.getClass());
    }

    private Collection<?> getCollection() {
        final Collection<Object> ret = new ArrayList<>(3);
        ret.add(Integer.valueOf(100));
        ret.add("This string is a string");
        ret.add(this.getClass());

        return ret;
    }

    private interface TestIFace {

        void handle(Object o);
    }

    private static final class TestHelper implements TestIFace {

        private final String a;

        TestHelper(final String a) {
            super();
            this.a = a;
        }

        @Override
        public void handle(final Object o) {
            System.out.println(this.a + " Object: " + o.toString());
        }

        public void handle(final Class<?> clazz) {
            System.out.println(this.a + " Class: " + clazz.getName());
        }
    }
```

Gives the output:

```
A Object: 100
A Object: This string is a string
A Object: class io.isaki.jsandbox.PolyTest
B Object: 100
B Object: This string is a string
B Object: class io.isaki.jsandbox.PolyTest
C Object: class io.isaki.jsandbox.PolyTest
C Class: io.isaki.jsandbox.PolyTest
```

jbarnettwomply:
updated to check for Class instances, can look into javassist later if that's desired

nickbabcock:
> updated to check for Class instances

Is it possible to write a test for this case, so we don't possibly regress on it in the future?

jbarnettwomply:
Added test (which fails without the `instanceof Class<?>` check) and migrated bytecode instrumentation to javassist.

jplock:
@nickbabcock any additional feedback on this one?

jplock:
@nickbabcock with this merged, what do you think about starting to create some release candidates, or should we wait until Jersey 2.28 is released?  /cc @dropwizard/committers

nickbabcock:
As far as 2.0 proper is concerned, we'll have to wait for jersey 2.28 to update to hibernate validator 6.0, but as far aiding people testing features so far, it's always good to get code in front of more people (imo release candidate may be too strong of wording as we still have breaking changes on the horizon).

serhiypal:
does this work align with the latest comment at https://github.com/eclipse-ee4j/jersey/issues/3700#issuecomment-424596364?

PR #2466, opened 2018-08-24
Compiler Warning Cleanup

isaki-x:
###### Problem:
Eclipse logs over 300 compiler warnings.

###### Solution:
This branch was done in two commits. The first was addressing POM based warnings. The second, here is the commit message:

```
    This commit addresses the following compiler warnings:
    
    - Unncessary @SuppressWarnings
    - Unused imports
    - Unused local variables
    - Missing serialization long
    - Parameterized type warnings
    - Unchecked casts
    - Raw type usage
    
    This commit did not touch:
    
    - @SuppressWarnings not understood by Eclipse
    - Deprecated code
    - Restricted code access
```

Note that all of these changes were made by hand. My personal auto-cleanup setup is too aggressive for DropWizard coding conventions and touched too much code (like marking nearly every parameter in a method declaration final). I experimented briefly with trying to set something up that was DropWizard specific and was unhappy with the results (it removed all of the unsupported `@SuppressWarnings` declarations even with the least aggressive setup). As such, by hand was the only way I could ensure the bare minimum was changed to meet my goal, which was to eliminate the easiest warnings to fix.

###### Result:
DropWizard now has a little over 150 warnings remaining. Some of these are from annotations that my barebones Eclipse did not understand (so may or may not be valid warnings). However, the ones referring to deprecated method and class usage and the restricted access to JRE internals will need to be addressed at some point.


jplock:
Looks good to me!  Thanks!

PR #2469, opened 2018-08-27
ConnectorProvider Not Set Silently

ipropper:
###### Problem:

I wanted to build two different clients with the same builder, so I had some code sort of like this:

```
final private JerseyClientConfiguration clientConfigA = new JerseyClientConfiguration();
final private JerseyClientConfiguration clientConfigB = new JerseyClientConfiguration();
...
JerseyClientBuilder jerseyClientBuilder = new JerseyClientBuilder(environment);
Client a = jerseyClientBuilder.using(clientConfigA).build("client-A"));
Client b = jerseyClientBuilder.using(clientConfigB).build("client-B"));
```

I ran into these problems:
1. Only metrics for "client-A" were being reported
2. The connection timeout I set for client-A was the same for client-B. This especially confused me because client B's config had a different timeout value.  

I realized when `connectorProvider` is not explicitly set in the builder, then it gets silently set on the first build. From then on that builder will use the same `connectorProvider`. 

Current solutions to this are:
- instantiating a new builder for every client
- creating a unique `connectorProvider` for each client. 

However, I thought people might want to reuse the same builder without explicitly setting the `connectorProvider`. I'm not sure if this is a worthy change, as the current process may be the desired state. I think you all will know better.  

###### Solution:

When `connectorProvider` is null in the `JerseyClientBuilder`, build will not auto set it.
Instead, a new `connectorProvider` will be instantiated on each build. If users want to use the same `connectorProvider` across builds, they only need to explicitly set it once.

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java
@@ -412,12 +412,14 @@ public class JerseyClientBuilder {
         }
 
         config.register(new DropwizardExecutorProvider(threadPool));
+
         if (connectorProvider == null) {
             final ConfiguredCloseableHttpClient apacheHttpClient =
                     apacheHttpClientBuilder.buildWithDefaultRequestConfiguration(name);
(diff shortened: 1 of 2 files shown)


arteam:
LGTM, thank you for the conitribution!

PR #2474, opened 2018-08-28
Updated documentation and comments that mentioned StrSubstitutor.

Sounie:
###### Problem:
StrSubstitutor was recently replaced by StringSubstitutor in the code, but some mentions in the comments and documentation were not updated.

###### Solution:
Replaced mentions, including naming of library that the replacement came from.

###### Result:
Docs a little more up to date with the code.


isaki-x:
Good catch! I missed the doc references when making the change. Thank you!

PR #2487, opened 2018-09-11
Documenting how to escape periods

zmarois:
Documenting how to escape periods in system property config overrides

###### Problem:

I've always wanted an easy way to tweak logging levels of individual loggers with a low touch approach (so I can debug issues without pushing an entire deployable change),  but had difficulty because the property for it contains periods. Finally, I spent some time today going through the code that makes these system property overrides happen, and found that there is a way to escape the period, it just wasn't explicitly documented.

###### Solution:
Document how periods are escaped in parsing system property overrides


jplock:
Thanks for the contribution!

PR #2490, opened 2018-09-14
Update to jetty 9.4.12

### LINKED ISSUE #2480: Upgrade to Jetty 9.4.12
Jetty 9.4.12 is now out.

It includes many bug fixes and improvements to HTTP/2 and general security.

If you choose to upgrade to Jetty 9.4.12, please be aware of the changed default Cipher exclusion list.

See eclipse/jetty.project#2807 for details.

nickbabcock:
What this PR entails:

- Update Jetty to latest (9.4.12) from (9.4.11)
- Remove deprecated configuration knobs (blockingTimeout and soLingerTime)
- Add minResponseDataRate

As for reasoning:

- Dropwizard should start with a clean slate with deprecations
- soLingerTime was literally a noop and a warning, so it has been removed (and not replaced with an alternate method)
- blockingTimeout has been rendered unnecessary (https://github.com/eclipse/jetty.project/issues/2525), and those interested in some sort rate limit  should see minResponseDataRate and minRequestDataRate (which was already part of 1.3.0 (https://github.com/dropwizard/dropwizard/pull/2184))
- Since minResponseDataRate was not present in our configuration, I added it, in case someone was using blockingTimeout in that fashion.

I'm torn on the what types minResponseDataRate / minRequestDataRate should be. They are currently numeric, but seeing just an `100` in a configuration file is not conducive to its meaning (number of bytes per second, as in if someone is sending less than `100` bytes per second, respond with a 408). I like how in https://github.com/dropwizard/dropwizard/pull/2384 we were able to use a `Duration` to convey how to throttle logging messages. I was looking into migrating to `Size` but I couldn't think of a descriptive name (and it would also be a breaking change), so I left it. But if we're going to make the switch from long to Size we should probably do it soon as the minRequestDataRate is a relatively new know (1.3.0), so it's usage may not be terribly widespread. (cc @patrox in case he has any thoughts on this)

###### Result:

Closes #2480 

Afterwards:

- [ ] update release notes
- [ ] write migration guide for removed configuration settings

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -303,11 +303,11 @@ HTTP
           bufferPoolIncrement: 1KiB
           maxBufferPoolSize: 64KiB
           minRequestDataRate: 0
+          minResponseDataRate: 0
           acceptorThreads: 1
           selectorThreads: 2
           acceptQueueSize: 1024
           reuseAddress: true
-          soLingerTime: 345s
           useServerHeader: false
           useDateHeader: true
           useForwardedHeaders: true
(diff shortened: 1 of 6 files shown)


arteam:
Agree with moving it to `Size`

arteam:
LGTM, merged!

debraj-manna:
Can the same change be done for 1.2.x branch? Below is the issue for upgrading jetty-server to 9.4.12 in 1.2.x release

https://github.com/dropwizard/dropwizard/issues/2495

PR #2493, opened 2018-09-15
Migrate tests to JUnit 5.4.0

arteam:
JUnit5 has been slowly getting among the community over the past years and it looks like stable enough for us to migrate.

### DIFF
--- a/dropwizard-core/src/test/java/io/dropwizard/setup/BootstrapTest.java
+++ b/dropwizard-core/src/test/java/io/dropwizard/setup/BootstrapTest.java
@@ -11,8 +11,8 @@ import io.dropwizard.configuration.DefaultConfigurationFactoryFactory;
 import io.dropwizard.configuration.FileConfigurationSourceProvider;
 import io.dropwizard.jackson.Jackson;
 import org.hibernate.validator.HibernateValidator;
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 
 import javax.validation.Validation;
 import javax.validation.ValidatorFactory;
@@ -27,7 +27,7 @@ public class BootstrapTest {
     };
     private Bootstrap<Configuration> bootstrap;
 
-    @Before
+    @BeforeEach
     public void setUp() {
         bootstrap = new Bootstrap<>(application);
     }
--- a/dropwizard-e2e/src/test/java/com/example/app1/App1Test.java
+++ b/dropwizard-e2e/src/test/java/com/example/app1/App1Test.java
@@ -4,10 +4,11 @@ import io.dropwizard.Configuration;
 import io.dropwizard.client.JerseyClientBuilder;
 import io.dropwizard.jackson.Jackson;
 import io.dropwizard.testing.ResourceHelpers;
-import io.dropwizard.testing.junit.DropwizardAppRule;
-import org.junit.BeforeClass;
-import org.junit.ClassRule;
-import org.junit.Test;
+import io.dropwizard.testing.junit5.DropwizardAppExtension;
+import io.dropwizard.testing.junit5.DropwizardExtensionsSupport;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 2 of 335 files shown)


dropwizard-core/src/test/java/io/dropwizard/setup/BootstrapTest.java
```diff
@@ -106,7 +106,7 @@ public void defaultsToDefaultValidatorFactory() throws Exception {
 
         ValidatorFactoryImpl validatorFactory = (ValidatorFactoryImpl) bootstrap.getValidatorFactory();
 
-        // It's imperative that the NonEmptyString validator come before the general param validator
+        // It's imperative that the NonEmptyString validator come setUp the general param validator
```
nickbabcock: Looks like this was changed accidentally

dropwizard-e2e/src/test/java/com/example/app1/App1Test.java
```diff
@@ -56,7 +57,7 @@ public void custom404OnViewRenderMissingMustacheTemplate() {
 
     @Test
     public void earlyEofTest() throws IOException, InterruptedException {
-        // Only eof test so we ensure it's false before test
+        // Only eof test so we ensure it's false setUp test
```
nickbabcock: Looks like there are multiple occurrences of the before/setUp swap 😄 

I'll assume you'll fix all of them, so I'll stop commenting when I see one.

dropwizard-e2e/src/test/java/com/example/badlog/BadLogTest.java
```diff
@@ -48,20 +45,20 @@ public void teardown() {
...
     public void thatLoggingIsntBrokenOnCleanup() throws Exception {
         BadLogApp.runMe(new String[]{"server"});
-        LOGGER.info("I'm after the test");
+        LOGGER.info("I'm tearDown the test");
         Thread.sleep(100);
 
         assertThat(new String(out.toByteArray(), UTF_8))
             .contains("Mayday we're going down")
-            .contains("I'm after the test");
+            .contains("I'm tearDown the test");
```
nickbabcock: Looks like after/tearDown suffered the same fate

dropwizard-e2e/src/test/java/com/example/badlog/BadLogTest.java
```diff
@@ -28,18 +28,15 @@
     private static PrintStream oldOut = System.out;
     private static PrintStream oldErr = System.err;
 
-    @Rule
-    public TemporaryFolder folder = new TemporaryFolder();
```
nickbabcock: Does this mean the temporary file allocated in the test will need to be deleted?

dropwizard-e2e/src/test/java/com/example/app1/App1Test.java
arteam: Rolled back

dropwizard-e2e/src/test/java/com/example/badlog/BadLogTest.java
arteam: Rolled back

dropwizard-e2e/src/test/java/com/example/badlog/BadLogTest.java
arteam: Those files will be created in the `/tmp` directory on UNIX-systems and the OS will automatically take care of cleaning them. I agree that explicit removing is better, but we use temporary files sparingly in the test suite, so that shouldn't be an issue.

Tibor17:
You should use higher version `maven-surefire-plugin:2.22.1` because this is able to work with any JUnit5 version. The JUnit5 provider should not be used. It's enough to use the JUnit5 engine in POM dependencies.

dropwizard-e2e/src/test/java/com/example/badlog/BadLogTest.java
joschi: JUnit 5.4.0 re-introduced the [`TempDirectory` extension](https://junit.org/junit5/docs/5.4.0/user-guide/index.html#writing-tests-built-in-extensions-TempDirectory).

jplock:
Is there a way to re-run the tests on Appveyor?

Issue #2489, opened 2018-09-13
Support json log exception type configuration


mattnelson:
The exception log format defaults[1,2] to root cause first for all types(console,file,json,etc...). With the non-json types the layout can be overridden to standard stack traces. JSON log format it not currently configurable. Can the option to configure exception implementations be added?

* `exception`
  * https://logback.qos.ch/manual/layouts.html#ex
  * https://github.com/qos-ch/logback/blob/v_1.2.3/logback-classic/src/main/java/ch/qos/logback/classic/pattern/ThrowableProxyConverter.java
* `xException`
  * https://logback.qos.ch/manual/layouts.html#xThrowable
  * https://github.com/qos-ch/logback/blob/v_1.2.3/logback-classic/src/main/java/ch/qos/logback/classic/pattern/ExtendedThrowableProxyConverter.java
* `rootException`
  * https://logback.qos.ch/manual/layouts.html#rootException
  * https://github.com/qos-ch/logback/blob/v_1.2.3/logback-classic/src/main/java/ch/qos/logback/classic/pattern/RootCauseFirstThrowableProxyConverter.java

`exception` and `xException` appear be identical due to the extended packaging data[3] being disabled by default.

[1] https://www.dropwizard.io/1.3.5/docs/manual/configuration.html#logging
[2] https://github.com/dropwizard/dropwizard/blob/v1.3.5/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/EventJsonLayoutBaseFactory.java#L92
[3] https://logback.qos.ch/manual/configuration.html#packagingData

nickbabcock:
It looks like json exception logging is somewhat configurable if someone where to override `EventJsonLayoutBaseFactory::createThrowableProxyConverter` (ie define your own logging type like `json2`), but if you have something else in mind, I'm open to suggestions!

mattnelson:
@nickbabcock Any objections if I start working on this? Do you have any other suggestions for resolving this within dropwizard? I could impleent it internally with a custom type like you mentioned, but I would like to solve upstream if possible.

nickbabcock:
Certainly, go ahead! The design makes sense (keeping the same terminology as logback to make it easier to look up what each field's effect is). Hopefully you can leverage a lot of logback's code to implement your idea. If it needs 2k of code to implement, I might have reservations if the feature is really worth the maintenance cost 😄

arteam:
Closed via #2501

Issue #2488, opened 2018-09-13
Support json log exception stack filtering


mattnelson:
Logback supports the ability to filter[1] out stack frames from the exception[2] with the format template of `%exception{full, io.dropwizard,org.apache}` syntax. With json log format we lost the ability to support this type of filtering. I think it can be reintroduced by adding configuration and setting[3] those on the `ThrowableHandlingConverter`[4].

[1] http://jira.qos.ch/browse/LOGBACK-540
[2] https://logback.qos.ch/manual/layouts.html#ex
[3] https://github.com/qos-ch/logback/blob/v_1.2.3/logback-core/src/main/java/ch/qos/logback/core/pattern/DynamicConverter.java#L54
[4] https://github.com/dropwizard/dropwizard/blob/v1.3.5/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/EventJsonLayoutBaseFactory.java#L91

nickbabcock:
I realize now that you're already familiar with my suggestion in #2489.

Can the issue of configuring json logging of exception be solved in a single swoop, as these two issues seem very much related.

arteam:
Closed via #2501

PR #2501, opened 2018-09-22
Support configuration of exception details with JSON logging

mattnelson:
###### Problem:
JSON logging is missing some features that standard logging supplies due to the lack of a format override.

###### Solution:
Support configuration of exception cause placement, depth and package filtering with JSON logging.

See also #2488 and #2489


arteam:
Looks great! Thank you very much for the contribution!

PR #2506, opened 2018-09-25
Convert to lazy evaluation for json event creation

mattnelson:
###### Problem:
json event creation calls into the I*Event objects to get the value before it is known if the value should be included in the log message.

###### Solution:
* Convert to lazy evaluation for json event creation
* Initialize maps to the size of the includes list.

###### Result:
json log message creation should be more performant by delaying method invocations until necessary.


jplock:
Looks good to me, thanks!

Issue #2497, opened 2018-09-17
JSON access log URI vs URL defaults


mattnelson:
The standard request access logs default[1] to request URL. The JSON access logs introduced in #2232 default[3] to request URI. While it could be argued that a different output format doesn't have to remain passive to the previous layout. It did take a bit to trace down the issue as I assumed switching the format would produce logs with the same composite parts. What this change between uri/url intentional? 

[1] https://github.com/dropwizard/dropwizard/blob/v1.3.5/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/layout/LogbackAccessRequestLayout.java#L27
[2] https://github.com/qos-ch/logback/blob/v_1.2.3/logback-access/src/main/java/ch/qos/logback/access/PatternLayout.java#L113-L114
[3] https://github.com/dropwizard/dropwizard/blob/v1.3.5/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/AccessJsonLayoutBaseFactory.java#L43

nickbabcock:
Can you give an example of the difference?

If one format is not inherently better than the other, than I would agree that it makes sense for them to be consistent.

mattnelson:
Discovered another discrepancy that may throw a wrench into this. The [AccessEvent](https://github.com/qos-ch/logback/blob/v_1.2.3/logback-access/src/main/java/ch/qos/logback/access/spi/AccessEvent.java#L18) interprets URL as method+uri+querystring+protocol. Since three of those items are already available as discrete parts, it doesn't make sense to use URL via logback's interpretation of it. As that would require users to parse the string to break it back down to the parts. I think dropwizard should interpret the `requestUrl` configuration value to be uri+querystring.

```json
{
  "method": "PUT",
  "uri": "/v2/apps/EUREKA/a7eb6f43-8510-4196-9aab-82d947401312",
  "params": {
    "lastDirtyTimestamp": [
      "1537398169789"
    ],
    "status": [
      "UP"
    ]
  },
  "protocol": "HTTP/1.1",
  "url": "PUT /v2/apps/EUREKA/a7eb6f43-8510-4196-9aab-82d947401312?status=UP&lastDirtyTimestamp=1537398169789 HTTP/1.1"
}
```

```yaml
# example dw config
# provides uri and query param via url
includes:
- protocol
- method
- requestUrl 
# provides uri and query param as distinct items
includes:
- protocol
- method
- requestUri
- requestParameters
```

[...]

PR #2232, opened 2017-12-14
Add support for JSON logs in Dropwizard

arteam:
The new `dropwizard-json-log` module allows users to configure a specific layout to convert logging message to JSON. The module supports event logs and access logs like that:

For general logging:
```yaml
    logging:
      appenders:
        - type: console
          layout:
            type: json
```

For request logging:

```yaml
    server:
      requestLog:
        appenders:
          - type: console
            layout:
              type: access-json
```

The default format produces logs like:
```
 {"level":"INFO","logger":"org.eclipse.jetty.server.Server","thread":"main","message":"Started @6505ms","timestamp":"2018-01-03T19:01:37.674Z"}
```

and access logs like
 ```
    {"method":"GET","userAgent":"Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:57.0) Gecko/20100101 Firefox/57.0","uri":"/hello-world","requestTime":5,"protocol":"HTTP/1.1","contentLength":37,"remoteAddress":"127.0.0.1","timestamp":"2018-01-03T19:04:48.448Z","status":200}
```

Solves #1388, #1451 as an alternative to #2180

PR #2509, opened 2018-09-27
Add support for pathQuery json access log attribute

mattnelson:
###### Problem:
Implementation from #2497 

> The standard request access logs default[1] to request URL. The JSON access logs introduced in #2232 default[3] to request URI. While it could be argued that a different output format doesn't have to remain passive to the previous layout. It did take a bit to trace down the issue as I assumed switching the format would produce logs with the same composite parts. What this change between uri/url intentional?

###### Design:

Been thinking about this quite a bit more trying to reason about the differences between uri/url. It isn't very clear from the [spec](https://tools.ietf.org/html/rfc3986#section-3) if there is a canonical name for the path+query part. The current dropwizard namings seem misleading according to spec, not sure if it was intended to align or if it carried over the logback naming conventions. [Logback](https://github.com/qos-ch/logback/blob/v_1.2.3/logback-access/src/main/java/ch/qos/logback/access/spi/AccessEvent.java#L147) appears to mostly align with [javax.servlet](https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getRequestURI--).

```
         foo://example.com:8042/over/there?name=ferret#nose
         \_/   \______________/\_________/ \_________/ \__/
          |           |            |            |        |
       scheme     authority       path        query   fragment
          |   _____________________|__
         / \ /                        \
         urn:example:animal:ferret:nose
```

p.s. sorry for the bikeshedding...

###### Solution:
Introduce `pathQuery` attribute that includes the path/query segments as defined in RFC3986. Left all the existing attributes unchanged for passivity reasons and the possibility of backpatching depending on projected 2.0 release date.

Will defer to the maintainers if any more drastic resolutions should be considered for 2.x up to and including:
* Deprecate uri/url
* Change default of uri to pathQuery
* Rename uri to path
* Rename params to query

###### Result:
JSON access logs can include the combined path/query in a single named parameter.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -953,7 +953,8 @@ includes                 (timestamp, remoteAddress,
                                                       - ``userAgent``         *true*     Whether to include the user agent of the request as the ``userAgent`` field.
                                                       - ``requestParameters`` *false*    Whether to include the request parameters as the ``params`` field.
                                                       - ``requestContent``    *false*    Whether to include the body of the request as the ``requestContent`` field.
-                                                      - ``requestUrl``        *false*    Whether to include the request URL (method, URI, query parameters, protocol) as the ``contentLength`` field.
+                                                      - ``requestUrl``        *false*    Whether to include the request URL (method, URI, query parameters, protocol) as the ``url`` field.
+                                                      - ``pathQuery``         *false*    Whether to include the URI and query parameters of the request as the ``pathQuery`` field.
                                                       - ``remoteHost``        *false*    Whether to include the fully qualified name of the client or the last proxy that sent the request as the ``remoteHost`` field.
                                                       - ``responseContent``   *false*    Whether to include the response body as the ``responseContent`` field.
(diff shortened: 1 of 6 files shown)


mattnelson:
Trying to validate these changes and the snapshot doesn't appear to have been updated. Possibly related to #2505?

https://oss.sonatype.org/content/repositories/snapshots/io/dropwizard/dropwizard-json-logging/1.4.0-SNAPSHOT/
`dropwizard-json-logging-1.4.0-20180927.140524-72.pom`

```
commit b584bf3ddb4e22abd9b147e349f47e8a29534d0a
Merge: 0c59c0c50 14732c5bc
Author: Justin Plock <jplock@users.noreply.github.com>
Date:   Fri Sep 28 17:41:38 2018 -0400

    Merge pull request #2509 from mattnelson/json_uri_query

    Add support for pathQuery json access log attribute
```

arteam:
Let me cut a 2.0. release candidate

PR #2512, opened 2018-09-30
Merge Bundle and ConfiguredBundle

alex-shpak:
###### Solution:
In short terms,
1. `Bundle` interface is gone
2. `ConfiguredBundle` interface is renamed to `Bundle`

###### Result:
There will be only one bundle to extend from which will make bundles simpler


jplock:
@dropwizard/committers do we want to do this?  I maintain some bundles that don't require any configuration where I'm using `Bundle`.  I think this would cause a lot of churn for people waiting for their bundles to be updated.

evnm:
Sounds like a job for an `@Deprecated` annotation.

jplock:
@evnm @alex-shpak I'd be in favor of releasing 1.3.7 with Bundle deprecated and then remove it in 2.0.0, but I don't think we should remove in 2.0.0 without deprecating first.  Or we deprecate in 2.0.0 and remove it in 2.1.0 or something.

evnm:
Removing it would warrant a major version bump, so it seems to me like the right thing to do is your first option, @jplock—deprecate in 1.3.7 and remove in 2.0.0.

alex-shpak:
Also we can start with way proposed in #1360 and mark `Bundle` as `@Deprecated` and then remove it later.

joschi:
I've given the thing a try in #2516 with a backward-compatible change.

PR #2516, opened 2018-10-03
Deprecate Bundle in favor of ConfiguredBundle<T>

joschi:
Instead of supporting two types of bundles, `Bundle` and `ConfiguredBundle<T>`, this change deprecates `Bundle` in favor of `ConfiguredBundle<T>` in a backward-compatible way.

Closes #1360
Closes #2512

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
@@ -39,7 +39,6 @@ import static java.util.Objects.requireNonNull;
  */
 public class Bootstrap<T extends Configuration> {
     private final Application<T> application;
-    private final List<Bundle> bundles;
     private final List<ConfiguredBundle<? super T>> configuredBundles;
     private final List<Command> commands;
 
@@ -61,7 +60,6 @@ public class Bootstrap<T extends Configuration> {
     public Bootstrap(Application<T> application) {
         this.application = application;
         this.objectMapper = Jackson.newObjectMapper();
-        this.bundles = new ArrayList<>();
         this.configuredBundles = new ArrayList<>();
         this.commands = new ArrayList<>();
         this.validatorFactory = Validators.newValidatorFactory();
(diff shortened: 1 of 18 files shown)


dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
```diff
@@ -128,16 +126,6 @@ public void setClassLoader(ClassLoader classLoader) {
         this.classLoader = classLoader;
     }
 
-    /**
-     * Adds the given bundle to the bootstrap.
-     *
-     * @param bundle a {@link Bundle}
-     */
-    public void addBundle(Bundle bundle) {
```
evnm: I think removing the `Bundle` hooks in `Bootstrap` is apt to surprise people who expect their existing bundles to work after a minor version bump. Would you be opposed to holding off on removing this until 2.0.0?

dropwizard-core/src/main/java/io/dropwizard/sslreload/SslReloadBundle.java
```diff
@@ -14,19 +14,17 @@
 import java.util.HashSet;
 import java.util.Set;
 
-/** Bundle that gathers all the ssl connectors and registers an admin task that will
- *  refresh ssl configuration on request */
-public class SslReloadBundle implements Bundle {
+/**
+ * Bundle that gathers all the ssl connectors and registers an admin task that will
+ *  refresh ssl configuration on request
```
evnm: -1 space indentation in the comment.

dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
joschi: They will still work because `Bundle` implements `ConfiguredBundle<Configuration>`.

The change should be completely backwards-compatible (also see `io.dropwizard.BundleTest`).

jplock:
Should we add an entry to the release notes for this as well?

PR #2525, opened 2018-10-11
fix typos

nitram509:
I just stumbled upon some typos in the documentation.

In Detail:
* A few times, 'Dropwizard' did not start with upper case.
* 'subcommand' is not known to leo.org

Any Feedback welcome.


joschi:
@nitram509 Thanks!

Issue #2500, opened 2018-09-19
Include default requestLog format string in documentation


novaterata:
I'm trying to just slightly tweak the requestLog, but I can't for the life of my find the default format string. Please include it in the entry for requestLog

nickbabcock:
Yes, it is missing and should be added to the documentation. For the time being, the request log format is:

```
%h %l %u [%t{dd/MMM/yyyy:HH:mm:ss Z, <tz-id>] "%r" %s %b "%i{Referer}" "%i{User-Agent}" %D
```

PR #2526, opened 2018-10-11
Fix issue #2500 Include default requestLog format string in documentation

nitram509:
Picked up issue #2500 and added the default log pattern into documentation.
I did copy it directly from class LogbackAccessRequestLayoutTest, in order to avoid copy/paste errors.

Any feedback welcome.


jplock:
Thanks!

PR #2528, opened 2018-10-17
WIP: Update build dependencies

jplock:
NullAway just merged in Errorprone 2.3.2 support the other day, so a new release may be imminent.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,9 +66,9 @@
         <argLine>-Duser.language=en -Duser.region=US</argLine>
 
         <junit.version>4.12</junit.version>
-        <junit5.version>5.2.0</junit5.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -464,7 +464,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-deploy-plugin</artifactId>
-                    <version>2.8.2</version>
+                    <version>3.0.0-M1</version>
```
joschi: Should we really be using milestone (i. e. pre-release) versions of Maven plugins?

jplock: Probably not, I was just testing what worked and what didn't.  I need to fix the new Nullaway errors in this PR still.

Issue #2535, opened 2018-10-26
Guava still presents in project. Please use scope=test for tests but not compile.


Tibor17:
This is the list of `pom.xml` which directly specify guava dependency.
dropwizard-util
dropwizard-jackson
dropwizard-lifecycle
dropwizard-jdbi3
dropwizard-bom

This is the list of modules where Java sources use guava transitively from modules in the previous list.
dropwizard-jdbi
dropwizard-jersey
dropwizard-validation

What happens if we use `dropwizard-testing` as a dependency in our project POM and we exclude the `guava` dependency from it? The dropwizard still needs to have guava on classpath?

Tibor17:
closing as following the discussion on google discussion.

PR #2538, opened 2018-10-27
Make the usage of Guava optional in all modules

joschi:
Refs #2535


dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -67,17 +67,20 @@ public DropwizardResourceConfig(@Nullable MetricRegistry metricRegistry) {
...
-        register(io.dropwizard.jersey.guava.OptionalMessageBodyWriter.class);
-        register(new io.dropwizard.jersey.guava.OptionalParamBinder());
-        register(io.dropwizard.jersey.optional.OptionalMessageBodyWriter.class);
-        register(io.dropwizard.jersey.optional.OptionalDoubleMessageBodyWriter.class);
-        register(io.dropwizard.jersey.optional.OptionalIntMessageBodyWriter.class);
-        register(io.dropwizard.jersey.optional.OptionalLongMessageBodyWriter.class);
+        register(new CacheControlledResponseFeature());
+        register(new io.dropwizard.jersey.guava.OptionalMessageBodyWriter());
+        register(new io.dropwizard.jersey.optional.OptionalMessageBodyWriter());
+        register(new io.dropwizard.jersey.optional.OptionalDoubleMessageBodyWriter());
```
jplock: Should we register these as instances or classes going forward? I thought part of the issue with the latest Jersey was that it no longer supported registering instances?

mattnelson:
> Copy RateLimiter (and dependent classes) from Guava 27.0

I think a shade with [include](https://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html) and [relocate](https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html) would be a more preferable option to strait copies in order to continue picking up upstream changes to guava.

Tibor17:
Yes, I mentioned `maven-shade-plugin` too. We use it in Apache Maven but you have to check the license of guava if you want to inline it in dropwizard.

minisu:
Did you consider moving the Guava stuff to a `dropwizard-guava` module instead (similar to how Java 8 specifics used to reside in [dropwizard-java8](https://github.com/dropwizard/dropwizard-java8))?

Tibor17:
There are still 8 open issues in version `2.0`.
Can the Dropwizard committers make a decision about this issue and make a bigger progress on it?

joschi:
@dropwizard/committers Given that many people are still using Guava when building Dropwizard applications and Guava's improved focus on backward-compatibility, I'm not sure if pursuing the removal in Dropwizard 2.x is still worth it.

Any opinions?

jplock:
@joschi makes sense to me. I think we can leave Guava for now.

nickbabcock:
Right now we're kinda in a halfway state, where we've incorporated a subset of Guava classes (like `Lists`). I'm fine with continuing with guava or removing it -- but a halfway state doesn't seem like the best long term solution

jplock:
I think of there are jdk8 compatible replacements, we should opt for using those whenever possible. For anything else using guava is ok.

Tibor17:
Guava is breaking backwards compatibility. It is found as a problematic library nevertheless historically many people have used it before. One way would be to remove it completely or write utilities in Dropwizard which substitutes Guava and the last alternative would be to use `maven-shade-plugin` and shift Java packages and inline into dropwizard libs.

joschi:
@Tibor17 Every library is eventually breaking backward compatibility. Pretty much every single dependency of Dropwizard did backward-incompatible changes in the past, so that's not a really good argument in my opinion.

Sure, we could shade it, but following that precedent, we'd also have to shade other libraries. 🤷‍♂️

Tibor17:
@joschi 
Removing Guava happens elsewhere but here important is to make stable Dropwizards towards users since it is a big project. As you can see Google breaks typical rules of what major.minor.incremental version means and you can really expect broken compatibility with increased minor version. Next bad with Guava is the modularity because this is libbrary having multiple responsibilities instead of a good design with modular Guava which was in past.

PR #2546, opened 2018-11-10
Remove entries covered by the jdbi3-bom

jplock:
I think these three entries can be removed since we're importing the `jdbi3-bom` above them.


jplock:
Not sure what is up with CircleCI lately

PR #2548, opened 2018-11-10
Set maven.compiler.release to 8 when building on JDK9+

jplock:
Also removed skipping jacoco on JDK11 since https://github.com/jacoco/jacoco/issues/663 is now resolved.


jplock:
We can probably skip this.

PR #2549, opened 2018-11-10
Add @JsonProperty to AbstractServerFactory setters

### LINKED ISSUE #2545: registerDefaultExceptionMappers is not a JSON Property and there is not way to reconfig it
This should be expose as JSON property so that default/simple could configure it.

https://github.com/dropwizard/dropwizard/blob/82e62e9c91b5acb3a47c04783f1204fc2aed34be/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java#L445

minisu:
A few setters were missing @JsonProperty, which makes deserialization fail
if the ObjectMapper is configured to not look at setter names.

Closes #2545

###### Problem:
A few setters were missing `@JsonProperty` in AbstractServerFactory, which makes deserialization fail if the ObjectMapper is configured to not look at setter names (e.g. configured like `mapper.setVisibility(PropertyAccessor.ALL, Visibility.NONE)`).

###### Solution:
Add `@JsonProperty` to the 2 setters that were missing it. This also makes all setters consistent in this class.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java
@@ -446,6 +446,7 @@ public abstract class AbstractServerFactory implements ServerFactory {
         return registerDefaultExceptionMappers;
     }
 
+    @JsonProperty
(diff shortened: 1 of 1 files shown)


minisu:
Thanks @nickbabcock! How about now? I also added a corresponding test in `SimpleServerFactoryTest`.

PR #2550, opened 2018-11-12
Fix flaky DefaultLoggingFactoryPrintErrorMessagesTest.testLogbackStatusPrinterPrintStreamIsRestoredToSystemOut

OrDTesters:
###### Problem:

The test DefaultLoggingFactoryPrintErrorMessagesTest.testLogbackStatusPrinterPrintStreamIsRestoredToSystemOut fails when run after ThrottlingAppenderWrapperTest.overThrottlingLimit.

Steps to reproduce:
```
git clone https://github.com/dropwizard/dropwizard
cd dropwizard
mvn install -DskipTests
cd dropwizard-logging
mvn test -Dtest="ThrottlingAppenderWrapperTest#overThrottlingLimit,DefaultLoggingFactoryPrintErrorMessagesTest#testLogbackStatusPrinterPrintStreamIsRestoredToSystemOut" -Dsurefire.runOrder=reversealphabetical
```

###### Solution:
Reconfigure in the teardown of ThrottlingAppenderWrapperTest so that the default logging factory resets the StatusPrinter's print stream to the default System.out.

###### Result:
The tests run by the command listed will pass in both orders.

### DIFF
--- a/dropwizard-logging/src/test/java/io/dropwizard/logging/ThrottlingAppenderWrapperTest.java
+++ b/dropwizard-logging/src/test/java/io/dropwizard/logging/ThrottlingAppenderWrapperTest.java
@@ -95,6 +95,8 @@ public class ThrottlingAppenderWrapperTest {
         }
 
         this.bos.reset();
+
+        new DefaultLoggingFactory().configure(new MetricRegistry(), "test-logger");
     }
 
     @Test


nickbabcock:
I can replicate the issue, but I did notice that CI did fail in one of the effected tests, so your solution may need tweaking

OrDTesters:
@nickbabcock I don't believe this failure is related to my changes, as this issue seems to occasionally occur in other builds that don't include my changes, so it's probably flaky for some other reason (https://travis-ci.org/dropwizard/dropwizard/jobs/455793243, https://travis-ci.org/dropwizard/dropwizard/jobs/455616746, https://travis-ci.org/dropwizard/dropwizard/jobs/455605635, https://travis-ci.org/dropwizard/dropwizard/jobs/455555882). Furthermore, in my own Travis builds I don't see the same issue (https://travis-ci.org/OrDTesters/dropwizard/builds/455185698).

nickbabcock:
Yes, you're right. I'm sorry I didn't circle back to this PR to let you know things should be better with the recent CI improvements. Can you please rebase your changes on master?

Also other notes:

- It is odd looking to see `new DefaultLoggingFactory().configure` as the last line of a teardown for the throttling test. Were there other approaches that you considered? The line stands out, and out of context it is not clear what it is doing. Is there a more clear way to reset the status printer?
- Have you found any other test interactions that fail this way? I feel like the throttling test shouldn't be an isolated incident as I think there are similar logging tests that redirect stdout

nickbabcock:
Thank you for all the testing. Your logic seems sound. Can you add a comment in the code basically summarizing this PR, so the importance is apparent?

isaki:
I'm not 100% convinced this is the right approach. I wrote the current throttling logger test (the original was horribly flaky and if run on a machine with any amount of load would straight up fail). Let me play with it a bit and get back to you. As @nickbabcock mentioned, it feels bad referring to a completely unrelated class in this unit test.

isaki:
I have an alternate fix. It passes @OrDTesters reproducing steps and I'm currently running the full build. I will submit a PR and you guys can choose which approach you like better.

arteam:
Fixed via #2753

Issue #2496, opened 2018-09-17
DropwizardTestSupport sets ConfigurationFactoryFactory to early


manuel-hegner:
in [DropwizardTestSupport](https://github.com/dropwizard/dropwizard/blob/fb58944deddfa2925adb2fd4a3b4c3dba5998cd7/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java#L221) the test support class sets the bootstrap ConfigurationFactoryFactory before initializing the application. Should this not happen after the initialization since the application itself could change the FactoryFactory?

nickbabcock:
Is there a use case where an application sets their factory in initialization and then a test overrides that one? I have a thought brewing, but if you have a specific example that you're encountering, I'd love to hear it. All tests pass either way.

manuel-hegner:
We set the factory to a JSONConfigurationFactoryFactory in our app so that the config is parsed as json instead of yaml.

nickbabcock:
Yes that would lead to an incorrect order. Makes sense, feel free to create a PR with the change + test so that we don't regress on that behavior

PR #2551, opened 2018-11-12
DropwizardTestSupport sets ConfigurationFactoryFactory to early

manuel-hegner:
see #2496 
###### Problem:
in DropwizardTestSupport the test support class sets the bootstrap ConfigurationFactoryFactory before initializing the application

###### Solution:
moved application initialization before overwriting the configurationfactoryfactory in the TestSupport

###### Result:
ConfigurationFactoryFactory can be changed in app and the app can still be tested


nickbabcock:
Oh yeah, thanks for bringing this up again. Is it possible to write a test that demonstrates how the previous behavior was incorrect (app parses config as JSON and the test wants to override it)?

PR #2558, opened 2018-11-14
Allow reporting Metrics on stop

### LINKED ISSUE #2557: Allow option to report metrics on stopping dropwizard
We use dropwizard extensively and are fans of how relatively quick it starts and stops.

When testing locally, we start our application from its all-in-one jar file and shut it down again after the tests, configuring metrics as local .csv files. However, we never see them :disappointed: 
This is because dropwizard starts, we run tests and dropwizard stops again without ever getting the `report()` called from `ScheduledReporterManager` .

For us, it would be great if we could configure a `reportOnStop` setting to `true` that would cause one last `report()` when stopping. We have no problems defaulting this to `false` to remain backward-compatible.

I will submit a pull request for this myself.
Thanks!

sjoerdtalsma:
###### Problem:
We use dropwizard extensively and are fans of how relatively quick it starts and stops.

When testing locally, we start our application from its all-in-one jar file and shut it down again after the tests, configuring metrics as local .csv files. However, we never see them :disappointed:
This is because dropwizard starts, we run tests and dropwizard stops again without ever getting the `report()` called from `ScheduledReporterManager `.

###### Solution:
Add a `reportOnStop` boolean setting in `MetricsFactory` that would cause one last report() by the `ScheduledReporterManager` when stopping.

###### Result:
This should have no other impact, as the default for this boolean is `false` to remain backwards-compatible.  
This fixes #2557

### DIFF
--- a/dropwizard-metrics/src/main/java/io/dropwizard/metrics/ScheduledReporterManager.java
+++ b/dropwizard-metrics/src/main/java/io/dropwizard/metrics/ScheduledReporterManager.java
@@ -10,16 +10,30 @@ import io.dropwizard.util.Duration;
 public class ScheduledReporterManager implements Managed {
     private final ScheduledReporter reporter;
     private final Duration period;
+    private final boolean reportOnStop;
 
     /**
      * Manages the given {@code reporter} by reporting with the given {@code period}.
      *
      * @param reporter the reporter to manage.
-     * @param period the frequency to report metrics at.
+     * @param period   the frequency to report metrics at.
+     * @see #ScheduledReporterManager(ScheduledReporter, Duration, boolean)
      */
     public ScheduledReporterManager(ScheduledReporter reporter, Duration period) {
+        this(reporter, period, false);
+    }
+
+    /**
+     * Manages the given {@code reporter} by reporting with the given {@code period}.
+     *
+     * @param reporter     the reporter to manage.
+     * @param period       the frequency to report metrics at.
+     * @param reportOnStop whether the reporter should send one last report upon stopping
+     */
+    public ScheduledReporterManager(ScheduledReporter reporter, Duration period, boolean reportOnStop) {
         this.reporter = reporter;
         this.period = period;
+        this.reportOnStop = reportOnStop;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


dropwizard-metrics/src/main/java/io/dropwizard/metrics/ScheduledReporterManager.java
```diff
@@ -10,16 +10,32 @@
...
     /**
      * Manages the given {@code reporter} by reporting with the given {@code period}.
      *
      * @param reporter the reporter to manage.
-     * @param period the frequency to report metrics at.
+     * @param period   the frequency to report metrics at.
+     * @see #ScheduledReporterManager(ScheduledReporter, Duration, boolean)
+     * @deprecated Please specify the {@code reportOnStop} parameter as well.
      */
+    @Deprecated
```
nickbabcock: IMO, no need to deprecate just because it's overloaded.

nickbabcock:
I agree that the premise of flushing metrics (like logs) before shutdown is sound, though I think that this PR should be targeted towards the master branch. If it needs to be backported to a 1.x branch, it can be. Also can you include a test for deserialization, so we don't miss something obvious. Thanks!

dropwizard-metrics/src/main/java/io/dropwizard/metrics/ScheduledReporterManager.java
sjoerdtalsma: Agree. Removed deprecation but kept the `@see` reference

sjoerdtalsma:
Updated the Pull request to target `master` and made requested change + added unit test for deserialisation (missing property -> default false + minimum config with true)

sjoerdtalsma:
Can you make sure that after merging to master, this also gets backported to 1.3.x?
We're still on the latest 1.3 version 😉

nickbabcock:
Can you update the javadocs and the configuration docs with this new value?

> Can you make sure that after merging to master, this also gets backported to 1.3.x

No, I don't think that is how that works. We tend to reserve patch releases for security updates or major bugs

sjoerdtalsma:
Isn’t 1.3 the official ‘current’ release though?

sjoerdtalsma:
> Can you update the javadocs and the configuration docs with this new value?

Done 15 days ago. Is there more I should/can do before this can be merged?

nickbabcock:
This PR is good to go. Thank you for the contribution. I wouldn't think this would land in a 1.3.x release. A 1.4 (unplanned) may be better suited. 2.0 has been blocked for what seems like ages 😅

PR #2560, opened 2018-11-15
Fix for surefire crash in OpenJDK8

isaki:
###### Problem:
Some versions of 8u181 have been patched with fixes from 8u191 that have introduced problems for the surefire plugin.

###### Solution:
This disables the security fix that is not working right in the debian/ubuntu 8u181b13 package and
any other build that is using the patchset. I've also updated to the latest surefire version which is the only version I have tested this fix with.

###### Result:
Circle-ci should work without having to pin a JDK version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
     environment:
       JAVA_TOOL_OPTIONS: -Xms512m -Xmx2g
     docker:
-      - image: circleci/openjdk:8u171-jdk
+      - image: circleci/openjdk:8-jdk
     steps:
       - checkout
       - run: mvn test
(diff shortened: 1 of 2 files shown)


nickbabcock:
I couldn't get this method to work with the HTTP2 tests that require the ALPN jar 😖

isaki:
That is really odd. As much as I dislike pinning to an older JDK version, that might be the only way. Let's leave my PR open for now. I want to play with this a bit, if that is ok with you!

nickbabcock:
Thank you for your summary. It's certainly a bizarre situation!

isaki:
I don't understand why some fixes were patched into 8u181 rather than just upgrade to 8u191.

PR #2563, opened 2018-11-16
Tweak throttler test for CI

nickbabcock:
###### Problem:
Travis CI has been extra flaky with the `overThrottlingLimit` logging throttle appender. The test has been failing as it sees 12 lines instead (at max) 11 logged.

###### Solution:
Calculate the total elapsed time after the sleep to "clear" the rate limiter. By delaying the calculation, it encapsulates the time for the async log to write out the last line, which in turn allows the test to pass when 12 lines are logged.

###### Result:
If this test passes 10 times on travis CI, I'll mark this as good to go.


nickbabcock:
I'm 5/5 for tonight, so I think (Travis) CI is now stable 🎉

Issue #2567, opened 2018-11-20
Allow unknown JSON properties by default?


minisu:
Before 2.0.0 is released, what do you think of disabling `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` by default?

**Arguments for:**

 - Most, if not all, Dropwizard projects that I’ve seen disables it.
 - It follows Postel’s law (_be conservative in what you send, be liberal in what you accept_).
 - All other JSON libraries that I know allows unknown properties by default: GSON, the json package in Go and Json.NET. Spring Boot (which uses Jackson) also disables it by default.

**Arguments against:**

 - Breaks backwards compatibility (although no services should stop working).
 - Not Jackson default behavior (although e.g. the existing `FuzzyEnumModule` also makes Jackson less strict than its defaults).

Thoughts?

nickbabcock:
Huh, those are some pretty good arguments for changing the behavior. I support this change, as it seems it will bring dropwizard in line with the ecosystem at large (though becoming disjointed with default jackson behavior may be unexpected). As long as it's clearly documented how to revert the behavior, the change seems sensible. But I'd wait to see if there are additional opinions before further action.

jplock:
I’d be supportive of this.

kravchik:
Great, and now, how to disable it back?)

PR #2570, opened 2018-11-26
Allow unknown JSON properties by default

minisu:
###### Problem:
See #2567

###### Solution:
Modify the ObjectMapper provided by `Jackson.newObjectMapper()` and `Jackson.newMinimalObjectMapper()` to disable `FAIL_ON_UNKNOWN_PROPERTIES`.

###### Result:
Dropwizard no longer fails on unknown properties, except for when reading YAML configurations (in which case `FAIL_ON_UNKNOWN_PROPERTIES` is enabled by `DefaultConfigurationFactoryFactory`).


dropwizard-configuration/src/test/java/io/dropwizard/configuration/BaseConfigurationFactoryTest.java
```diff
@@ -427,21 +427,6 @@ public void throwsAnExceptionIfDefaultConfigurationCantBeInstantiated() throws E
...
-                "  * Unrecognized field at: propertis%n" +
-                "    Did you mean?:%n" +
-                "      - properties%n" +
-                "      - servers%n" +
-                "      - type%n" +
-                "      - name%n" +
-                "      - age%n" +
-                "        [2 more]%n", typoFile));
-    }
-
```
minisu: Seems like this is redundant since `ConfigurationFactoryFactoryTest` [tests the same behavior](https://github.com/dropwizard/dropwizard/blob/0d143d7bc35351a135f8dc82c5e6f78d1eb39268/dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationFactoryFactoryTest.java#L34-L44), but in a more apt place (since the behavior [is implemented by](https://github.com/dropwizard/dropwizard/blob/1e40fef4f4e8f048885b7ce6067bd9cd87a749b2/dropwizard-configuration/src/main/java/io/dropwizard/configuration/DefaultConfigurationFactoryFactory.java#L31) `DefaultConfigurationFactoryFactory`).

dropwizard-jersey/src/test/java/io/dropwizard/jersey/jackson/JsonProcessingExceptionMapperTest.java
```diff
@@ -110,11 +110,6 @@ public void returnsA400ForMalformedListInputCausingIoException() throws Exceptio
         assertEndpointReturns400("urlList", "[\"no-scheme.com\"]");
     }
 
-    @Test
-    public void returnsA400ForNonDeserializableRequestEntities() throws Exception {
-        assertEndpointReturns400("ok", new UnknownRepresentation(100));
-    }
```
nickbabcock: Any particular reason this test was removed? Should it now return a 200 or something else?

minisu: This test case seems to in practice test that Dropwizard fails on unknown JSON properties, which is no longer the expected behavior. There also seems to be other test cases covering other variants of non-deserializable request entities.

However, I see now that there's no test testing to not provide a `@NotNull` field. Maybe this test can be rewritten so that it tests that case instead?

jplock:
@minisu can you also please add a note about this to the release notes in https://github.com/dropwizard/dropwizard/blob/master/docs/source/about/release-notes.rst

minisu:
Fixed!

isaki:
I think we should add documentation on how to enable fail on unknown properties if we are removing it as the default. I haven't looked but I assume there is documentation for the inverse (which is no longer true) so we can probably just update that.

minisu:
> I think we should add documentation on how to enable fail on unknown properties if we are removing it as the default. I haven't looked but I assume there is documentation for the inverse (which is no longer true) so we can probably just update that.

I couldn't find anything in the documentation covering this, but I added it in the most recent commit.

PR #2575, opened 2018-12-13
Cleaned up flaky test in CachingAuthenticatorTest

isaki:
###### Problem:
The CachingAuthenticatorTest fails under load due to the way eviction works in Caffeine. Refer to the AppVeyor failure in PR #2573.

###### Solution:
I've changed around a few things to detect cache changes and eliminated sleeps that actually weren't required. This is not a perfect solution, but should be more robust than the current implementation.

- Removed unnecessary sleeps.
- Added retry concept to extend the window for cache eviction.

Unforuntately without invoking direct cache controls, this runs the risk of being flaky on high load systems. This implementation should be more robust than the previous, however.

###### Result:
The unit test should not fail except under the most extremely slow (i.e. high load) invocations.

### DIFF
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/CachingAuthenticatorTest.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/CachingAuthenticatorTest.java
@@ -1,14 +1,19 @@
 package io.dropwizard.auth;
 
 import com.codahale.metrics.MetricRegistry;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.CaffeineSpec;
+import com.google.common.util.concurrent.MoreExecutors;
+
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.InOrder;
+import org.mockito.stubbing.Answer;
 
 import java.security.Principal;
 import java.util.Collections;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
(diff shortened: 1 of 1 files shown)


isaki:
Travis failed on depcheck for slf4j: https://nvd.nist.gov/vuln/detail/CVE-2018-8088. This CVE has been around since March, so not sure why it is showing up now and not across all builds.

ben-manes:
fyi, you can set `Caffeine.executor(Runnable::run)` so that asynchronous work (like eviction) is run on a calling thread. This is especially convenient for unit tests, where you might inject in that configuration while keeping the default async. (The eviction work is cheap, so async here is merely to smooth out tail latencies but often not necessary).

isaki:
Using Guava's `MoreExecutors.directExecutor()` with Caffeine seems to have done the trick. Thank you for the suggestion! This is way more elegant than what I had done originally.

ben-manes:
You're welcome! 😄 

btw, when you do have to wait for asynchronous results, [Awaitility](https://github.com/awaitility/awaitility) is a very elegant alternative to `Thread.sleep`.

isaki:
This wasn't my unit test, so I was trying to go the least level of effort to make sure builds pass reliably. I shall definitely check out Awaitility for future reference!

nickbabcock:
> Oh, are you depending on an eviction order?

I think that might be the case, as the cache has `maximumSize=1`

I don't know the solution offhand, but I may revert this at the end of next week if we haven't found a solution as imo this trades one type of flakiness for another 😄

isaki:
The best thing might be to just scrap this test case.

nickbabcock:
> The best thing might be to just scrap this test case.

Are you talking about `respectsTheCacheConfiguration`? If so, removing that test case may be a good solution as there is no reason for us to test caffeine's implementation -- I'm assuming that's been tested aplenty 😆

ben-manes:
We brute force through over a million tests, taking 1hr on travis (parallelized). Yet somehow I still sneak a few bugs in there... 😞

but yeah, should be on me if it breaks 😄

PR #2576, opened 2018-12-14
Removing unnecessary import

sergioescala:
###### Problem:
There are imports in two different files that are not being used.

###### Solution:
The imports that are not being used were removed.

###### Result:
Nothing, just imports were removed.


sergioescala:
The Travis CI failed because slf4j-api-1.7.25.jar: CVE-2018-8088 were identified with vulnerabilities, but that error is not related to my change:
`[ERROR] Failed to execute goal org.owasp:dependency-check-maven:3.3.4:check (security-check) on project dropwizard-jackson: One or more dependencies were identified with vulnerabilities: slf4j-api-1.7.25.jar: CVE-2018-8088` 
any suggestions?

PR #2579, opened 2018-12-17
Optimizing code in the the class ConfiguredGZipEncoder

sergioescala:
###### Problem:
There is not a problem, just optimizing the class ConfiguredGZipEncoder

###### Solution:

1. Variable 'forceEncoding' initializer 'false' is redundant
2. Exception 'java.io.IOException' is never thrown in the method filter(ClientRequestContext context)
3. Simplifying conditional statement
4. Creating variables for GZIP and X-GZIP in order to avoid hard coded values

###### Result:
See the Solution Section

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/gzip/ConfiguredGZipEncoder.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/gzip/ConfiguredGZipEncoder.java
@@ -24,24 +24,25 @@ import java.util.zip.GZIPOutputStream;
 @Provider
 @Priority(Priorities.ENTITY_CODER)
 public class ConfiguredGZipEncoder implements WriterInterceptor, ClientRequestFilter {
-    private boolean forceEncoding = false;
+    private boolean forceEncoding;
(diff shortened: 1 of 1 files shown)


joschi:
Hi @sergioescala, thank you for your PR!

I have a few comments and questions about it.

> Variable 'forceEncoding' initializer 'false' is redundant

While that's true, the redundancy doesn't hurt and makes the default value explicit.

> Exception 'java.io.IOException' is never thrown in the method filter(ClientRequestContext context)

True, but it's the method signature of the overridden method (`ClientRequestFilter#filter(ClientRequestContext requestContext)`).

> Creating variables for GZIP and X-GZIP in order to avoid hard coded values

Why? And why not constants (`private static final´)?
The literal strings make it quite easy to follow the logic and won't change, while the variables (or constants) are simply another indirection.

sergioescala:
Hi, @joschi thanks for your comment!

I wanted to help with those changes and about the last one, you're right! I can make them private static final if you think it's better. 

I think the only that doesn't change the code drastically its the if sentence.

Please let me know which ones can be applied to keep them!

PR #2585, opened 2018-12-26
Fix date formatting pattern in test

simoneau:
"Y" is actually the year according to "week of year"-based calendars. This will be different from the calendar year on dates like 2018-12-30, which has a "y" of 2018 and a "Y" of 2019.

http://unicode.org/reports/tr35/tr35-10.html#Date_Format_Patterns


arteam:
Nice one! Thanks for fixing this.

PR #2511, opened 2018-09-30
Add safe Jackson deserializers to prevent a DoS attack

arteam:
Jackson uses `BigDecimal` for deserialization of `java.time` instants and durations. The problem is that if the users sets a very big number in the scientific notation (like `1e1000000000`), it takes forever to convert `BigDecimal` to `BigInteger`, and then to convert it to a long value. An example of the stack trace:

```java
@Test(timeout = 2000)
public void parseBigDecimal(){
    new BigDecimal("1e1000000000").longValue();
}

org.junit.runners.model.TestTimedOutException: test timed out after 2000 milliseconds
	at java.math.BigInteger.squareToomCook3(BigInteger.java:2074)
	at java.math.BigInteger.square(BigInteger.java:1899)
	at java.math.BigInteger.squareToomCook3(BigInteger.java:2053)
	at java.math.BigInteger.square(BigInteger.java:1899)
	at java.math.BigInteger.squareToomCook3(BigInteger.java:2051)
	at java.math.BigInteger.square(BigInteger.java:1899)
	at java.math.BigInteger.squareToomCook3(BigInteger.java:2049)
	at java.math.BigInteger.square(BigInteger.java:1899)
	at java.math.BigInteger.squareToomCook3(BigInteger.java:2049)
	at java.math.BigInteger.square(BigInteger.java:1899)
	at java.math.BigInteger.squareToomCook3(BigInteger.java:2055)
	at java.math.BigInteger.square(BigInteger.java:1899)
	at java.math.BigInteger.squareToomCook3(BigInteger.java:2049)
	at java.math.BigInteger.square(BigInteger.java:1899)
	at java.math.BigInteger.pow(BigInteger.java:2306)
	at java.math.BigDecimal.bigTenToThe(BigDecimal.java:3543)
	at java.math.BigDecimal.bigMultiplyPowerTen(BigDecimal.java:3676)
	at java.math.BigDecimal.setScale(BigDecimal.java:2445)
	at java.math.BigDecimal.toBigInteger(BigDecimal.java:3025)
```

A fix would be to reject big decimal values outside of the Instant and Duration ranges.

See:
[1] https://github.com/FasterXML/jackson-databind/issues/2141
[2] https://reddit.com/r/java/comments/9jyv58/lowbandwidth_dos_vulnerability_in_jacksons/

PR #2591, opened 2019-01-05
Upgrade to official Jackson fix for preventing a DoS attack

arteam:
Now Jackson automatically coerces very small and very big integers, so we can remove our custom deserializers introduced in #2511.


wurstbrot:
Hi,

thank you for including #2620 . I updated my version of dropwizard to 1.3.8, but I am still vulnerable (as the fix will come with this release). Due to the high criticality of the fixed vulnerabilities a release should be addressed soon.
When will 1.3.9 be released?

Cheers,
Timo

msymons:
@joschi, per your advice in #2595 I have upgraded dropwizard to 1.3.x

Now that I have done that I am also very interested in knowing when 1.3.9 might be released.

I do have one question about 1.3.9.  Will it use slf4j v1.7.26?  This has been updated for the 2.0.0 branch via #2652.  The reason why this matters is that v1.7.26 addresses [CVE-2018-8088](https://nvd.nist.gov/vuln/detail/CVE-2018-8088)

I see from #2578 that this is a false positive for dropwizard.  However, using 1.7.25 does give threat warnings when projects that use dropwizard are analysed.... far easier to just use 1.7.26.

(Should I log an issue/PR?)

PR #2613, opened 2019-01-27
Upgrade to Jersey 2.28

joschi:

PR #2625, opened 2019-02-03
Deprecate *Param classes (removed in 2.0.0)

jplock:
The `*Param` classes, except for `NonEmptyStringParam` will be removed in #2613.

This PR, deprecates them in the 1.3.x releases.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/IntParam.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/IntParam.java
@@ -6,7 +6,7 @@ import javax.annotation.Nullable;
  * A parameter encapsulating integer values. All non-decimal values will return a
  * {@code 400 Bad Request} response.
  * 
- * @deprecated As of release 1.3.9, will be removed in 2.0.0. Please use {@link java.util.OptionalInt} instead.
+ * @deprecated As of release 1.3.9, will be removed in 2.0.0.
  */
 @Deprecated
 public class IntParam extends AbstractParam<Integer> {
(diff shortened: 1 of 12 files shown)


joschi:
@jplock Upon reviewing the class hierarchy below `AbstractParam<T>` in Dropwizard 1.x, I think it would make sense to deprecate all of them because they can be modeled by `Optional<T>`.

What do you think?

jplock:
@joschi makes sense to me, I’ll update my PR

nickbabcock:
Since there is no direct replacement for `NonEmptyStringParam`, as @joschi noted, maybe we should not deprecate that one

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/IntParam.java
```diff
@@ -5,7 +5,10 @@
 /**
  * A parameter encapsulating integer values. All non-decimal values will return a
  * {@code 400 Bad Request} response.
+ * 
+ * @deprecated As of release 1.3.9, will be removed in 2.0.0. Please use {@link java.util.OptionalInt} instead.
```
nickbabcock: Will `OptionalInt` be a true replacement in v1.3.9? I thought it could only be replaced with the new functionality brought in Jersey 2.28 + hibernate validator 6.0

jplock: I'll remove that mention

nickbabcock:
Thinking about it, would it make more sense to deprecate in 2.0 instead of 1.3.x (and keep it around until 3.0), as I feel like it's a little iffy to deprecate a construct where there is no alternative other than bumping to the next major version, which may not be immediately attainable for a given project.

jplock:
@nickbabcock I think that makes sense.  I'll close this PR and submit another against `master`

PR #2637, opened 2019-02-09
Deprecate *Param classes, will be removed in 3.0.0

jplock:
Replaces https://github.com/dropwizard/dropwizard/pull/2625


adityamehra:
@jplock Any idea about the *Param classes in ```io.dropwizard.jersey.jsr310```?

joschi:
@adityamehra Every class extending [`AbstractParam<T>`](https://javadoc.io/static/io.dropwizard/dropwizard-project/2.0.21/io/dropwizard/jersey/params/AbstractParam.html) is deprecated and will be removed in Dropwizard 3.0.0.

You can replace them with an `Optional<T>` in almost every case.

PR #2647, opened 2019-02-14
Prefer instanceof instead of getClass in equals

nickbabcock:
###### Problem:
Another mini PR coming at you. This time fixing error prone's [EqualsGetClass](https://errorprone.info/bugpattern/EqualsGetClass) lint. From the site:

> Requiring the argument to equals to be of a specific concrete type is incorrect, because it precludes any subtype of this class from obeying the substitution principle.

###### Result:
No behavior difference inside dropwizard though users that subtype classes like `Size`, `Duration`, etc may see a difference (unlikely)

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParam.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParam.java
@@ -117,7 +117,7 @@ public abstract class AbstractParam<T> {
         if (this == obj) {
             return true;
         }
-        if ((obj == null) || (getClass() != obj.getClass())) {
+        if (!(obj instanceof AbstractParam)) {
             return false;
         }
         final AbstractParam<?> that = (AbstractParam<?>) obj;
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/PrincipalImpl.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/PrincipalImpl.java
@@ -20,7 +20,7 @@ public class PrincipalImpl implements Principal {
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (!(o instanceof PrincipalImpl)) {
             return false;
         }
 
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentials.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentials.java
@@ -51,7 +51,7 @@ public class BasicCredentials {
         if (this == obj) {
             return true;
         }
-        if (obj == null || getClass() != obj.getClass()) {
+        if (!(obj instanceof BasicCredentials)) {
             return false;
         }
         final BasicCredentials other = (BasicCredentials) obj;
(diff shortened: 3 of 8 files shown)


joschi:
Checking the equality of `getClass()` has different semantics than using `instanceof`.

For example `String.class == CharSequence.class` is false, while `String instanceof CharSequence` is true.

Are we sure that these changes are all correct?

nickbabcock:
You're right that there is / was a debate if it's too aggressive of a check: https://github.com/google/error-prone/issues/1144 

Do you expect the following test to pass (I just wrote it offhand, but I can add it to the PR):

```java
@Test
public void testEquality() {
    final HashSet<AbstractParam<Instant>> params = new HashSet<>();
    final OffsetDateTime dt = OffsetDateTime.of(2018, 1, 1, 1, 0, 0, 0, ZoneOffset.UTC);
    params.add(new InstantSecondParam(Long.toString(dt.toInstant().getEpochSecond())));

    assertThat(params)
        .contains(new InstantParam(Long.toString(dt.toInstant().toEpochMilli())));
}
```

Even though the params represent the same instant of time the test fails with the following message:

```
Expecting:
 <[2018-01-01T01:00:00Z]>
to contain:
 <[2018-01-01T01:00:00Z]>
but could not find:
 <[2018-01-01T01:00:00Z]>
```

Which can be confusing. Is the test contrived? Yes, as one should prefer `Set<Instant>` instead.

A rebuttal could be that the types in the standard library don't act that way (and this test passes):

```java
@Test
public void testEquality2() {
    final String str = "Hello World";
    final StringBuilder builder = new StringBuilder().append(str);
    final CharBuffer buffer = CharBuffer.wrap(str);
    assertThat(builder)
        .isNotEqualTo(buffer)
        .isEqualToIgnoringCase(buffer);
}
```

Reading the assertions aloud doesn't make sense (ignoring underlying assertion implementation), as how can something be equal but not equal at the same time!

In the end, while linters should not be taken as the whole truth, I want to address the noise generated on compilation and either:

- Globally suppress the `EqualsGetClass` check
- Suppress `EqualsGetClass` on a class by class basis

Looking over the effected classes, I'm not overly concerned either way as lint instances fall under classes that in tests, effectively final, aren't acted upon by users. Only `PrincipleImpl` and `AbstractParam` need close scrutiny and I don't see how they'd be abused by this change. Anyways thoughts appreciated.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParam.java
```diff
@@ -117,7 +117,7 @@ public boolean equals(Object obj) {
         if (this == obj) {
             return true;
         }
-        if ((obj == null) || (getClass() != obj.getClass())) {
+        if (!(obj instanceof AbstractParam)) {
```
rage-shadowman: FYI, these methods are now broken (they violate the equals contract of `a.equals(b) == b.equals(a)` when a and b both share the same `AbstractParam` interface and either of them overrides `equals`).

The only way to use `instanceof` in an equals check without violating the equals contract is to make your equals method `final`. If any subclass overrides the equals method (anything more than `return super.equals(object)`) you have broken symmetry (unless you have some ridiculously painful bits of code in the super implementation that guarantees that no subtype of your overridden class can ever be equal to any subtype with a different heritage -- and good luck maintaining that without constantly introducing equals symmetry bugs).

rage-shadowman:
Why not warn on `getClass()` equality checks when the `equals` method is `final` due to Liskov violation, and warn on `instanceof` checks when not `final` due to equals contract symmetry violation?

PR #2639, opened 2019-02-11
Add support for instrumented executors

dennyac:
###### Problem:
I'm attempting to improve visibility of [InstrumentedExecutorService](https://github.com/dropwizard/metrics/blob/4.1-development/metrics-core/src/main/java/com/codahale/metrics/InstrumentedExecutorService.java) / [InstrumentedScheduledExecutorService](https://github.com/dropwizard/metrics/blob/4.1-development/metrics-core/src/main/java/com/codahale/metrics/InstrumentedScheduledExecutorService.java) which is available in dropwizard metrics. I've worked with the dropwizard/metrics framework for close to four years now, and I wasn't aware that they existed. Hoping this helps others.

###### Solution:
I'm suggesting adding a method to the builders to provide the metricRegistry. If the metricRegistry is provided, it will return the instrumented executor service. 
```java
ExecutorService executorService = environment.lifecycle()
    .executorService(nameFormat)
    .maxThreads(maxThreads)
    .metricRegistry(environment.metrics()) // Returns InstrumentedExecutorService if metricRegistry is passed
    .build();

ScheduledExecutorService scheduledExecutorService = environment.lifecycle()
    .scheduledExecutorService(nameFormat)
    .metricRegistry(environment.metrics()) // Returns InstrumentedScheduledExecutorService if metricRegistry is passed
    .build();
```
I've added some documentation as well.

PR #2649, opened 2019-02-15
Use instrumented thread factory

dennyac:
###### Problem:
Currently the ```Environment``` factory methods for ```ExecutorService``` and ```ScheduledExecutorService``` don't come with instrumentation out of the box.

###### Solution:
I'm suggesting using ```InstrumentedThreadFactory``` which monitors  the number of threads created, running and terminated. I've added some documentation as well.

Please refer to https://github.com/dropwizard/dropwizard/pull/2639 as there's more context there

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -489,7 +489,7 @@ For example, given a theoretical Riak__ client which needs to be started and sto
 
     public class MyApplication extends Application<MyConfiguration>{
         @Override
-        public void run(MyApplicationConfiguration configuration, Environment environment) {
+        public void run(MyConfiguration configuration, Environment environment) {
             RiakClient client = ...;
             RiakClientManager riakClientManager = new RiakClientManager(client);
             environment.lifecycle().manage(riakClientManager);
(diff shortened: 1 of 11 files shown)


docs/source/manual/core.rst
```diff
@@ -501,8 +501,25 @@ application will not start and a full exception will be logged. If ``RiakClientM
...
-``ScheduledExecutorService`` instances which are managed. See ``LifecycleEnvironment#executorService``
-and ``LifecycleEnvironment#scheduledExecutorService`` for details.
+``ScheduledExecutorService`` instances which are managed. These managed instances use ``InstrumentedThreadFactory``
+that monitors the number of threads created, running and terminated
+
+.. code-block:: java
+
+    public class MyApplication extends Application<MyConfiguration> {
+        @Override
+        public void run(MyApplicationConfiguration configuration, Environment environment) {
```
nickbabcock: Should this be `MyApplicationConfiguration` or `MyConfiguration`?

dennyac: Good catch! I'll change that.

PR #2559, opened 2018-11-15
Pin circle ci jdk version

nickbabcock:
###### Problem:
Circle CI fails 100% of the time due to surefire exiting with an error (snipped)

```
[ERROR] Process Exit Code: 1
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:671)
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:533)
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:278)
```

###### Solution:

Turns out, a lot of people have this problem:

- [link 1](https://discuss.circleci.com/t/circleci-build-failure-on-openjdk-image/26104)
- [link 2](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911925)
- [link 3](https://stackoverflow.com/q/53010200/433785)
- [link 4](https://github.com/apache/maven-surefire/pull/197)

The temporary solution is to pin the jdk version used to one that works until maven surefire + openjdk play nicely 😅 

###### Result:
Circle CI passes 🤞

PR #2669, opened 2019-03-02
Upgrade to CircleCI 2.0

joschi:
* Use OpenJDK 11
* Use Maven wrapper
* Use build cache to speed up builds even more

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,15 +5,21 @@ jobs:
     environment:
       JAVA_TOOL_OPTIONS: -Xms512m -Xmx2g
     docker:
-      - image: circleci/openjdk:8u171-jdk
+      - image: circleci/openjdk:11-jdk
     steps:
       - checkout
-      - run: mvn test
+      - restore_cache:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -5,15 +5,21 @@ jobs:
     environment:
       JAVA_TOOL_OPTIONS: -Xms512m -Xmx2g
     docker:
-      - image: circleci/openjdk:8u171-jdk
+      - image: adoptopenjdk/openjdk8:jdk8u202-b08
```
nickbabcock: I pinned the openjdk image previously due to a bug (https://github.com/dropwizard/dropwizard/pull/2559). Do we still want to pin to a specific version? Else we can just use latest

joschi: @nickbabcock I was playing around with different images and also could reproduce the failing builds with CircleCI's OpenJDK 8u181 image.

This being said, the OpenJDK 11 image seems to work, so I'd go with this one in the future.

PR #2682, opened 2019-03-07
Update JdbiFactory to use metrics' InstrumentedSqlLogger

FredDeschenes:
###### Problem:

Metrics library was updated to use JDBI 3's `SqlLogger` instead of the deprecated `TimingCollector`, but Dropwizard is still using `TimingCollector`.

###### Solution:

Updated Dropwizard to use the new `SqlLogger`.

### DIFF
--- a/dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/JdbiFactory.java
+++ b/dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/JdbiFactory.java
@@ -16,7 +16,7 @@ import org.jdbi.v3.jodatime2.JodaTimePlugin;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 
 public class JdbiFactory {
-    protected final StatementNameStrategy nameStrategy;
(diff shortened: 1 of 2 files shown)


dropwizard-jdbi3/src/main/java/io/dropwizard/jdbi3/JdbiFactory.java
```diff
@@ -15,7 +16,7 @@
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 
 public class JdbiFactory {
-    private final StatementNameStrategy nameStrategy;
+    protected final StatementNameStrategy nameStrategy;
```
jplock: can this remain private?

FredDeschenes: Yeah I could keep it private and send it as an argument to `buildSQLLogger`.

joschi:
@FredDeschenes Thanks for your PR!

Please make sure to run the tests after your changes.

FredDeschenes:
Should be good now.

PR #2683, opened 2019-03-07
Fixed flaky test in CachingAuthorizer

FredDeschenes:
###### Problem:

Travis build failed in #2682 because sleeping for any amount of time does not guarantee that Caffeine will have performed its background cleanup before performing the next call.

###### Solution:

Made `cache` field in CachingAuthorizer package private and manually called the `cleanup` function on the Caffeine instance instead of sleeping.

###### Result:

Test should not fail randomly anymore.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
@@ -7,6 +7,7 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.CaffeineSpec;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
+import com.google.common.annotations.VisibleForTesting;
 import io.dropwizard.util.Sets;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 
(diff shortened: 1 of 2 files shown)


dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
```diff
@@ -41,7 +41,9 @@
     //
     // `null` cache values are interpreted as cache misses, and will
     // thus result in read through to the underlying `Authorizer`.
-    private final LoadingCache<ImmutablePair<P, String>, Boolean> cache;
+    //
+    // Field is package-private to be visible for unit tests
+    final LoadingCache<ImmutablePair<P, String>, Boolean> cache;
```
joschi: Please add [`@VisibleForTesting`](https://google.github.io/guava/releases/27.0.1-jre/api/docs/com/google/common/annotations/VisibleForTesting.html) to the field.

FredDeschenes: Yeah I wasn't sure about adding that since it looks like you're removing Guava. Should I use the one in `com.google.common.annotations` or `io.dropwizard.logback.shaded.guava.annotations`?

joschi: Use the one from Guava for now, please. 😃

PR #2684, opened 2019-03-08
Document asynchronous logging

### LINKED ISSUE #2681: Dropwizard loses log under highly concurrent request traffic
During a performance test I found an issue that lots of logs are missing in our log file. I have a bare minimum dropwizard service to reproduce the issue that logs are lost under highly concurrent request traffic. The only maven dependency I have is `dropwizard-core 1.3.9`, and I have a test resource: 
```java
@Path("test")
public class TestResource {

    Logger LOGGER  = LoggerFactory.getLogger(TestResource.class);

    @GET
    public Response get() throws Exception{

    	MDC.put("version", "dropwizard default");

    	String id = UUID.randomUUID().toString();

    	LOGGER.info("start " + id);

    	Thread.sleep(new Random().nextInt(50));

    	LOGGER.info("end " + id);

        return Response.ok("{\"status\":\"ok\"}").build();
    }
}
```

and my app config is 
```yaml
server:
  applicationConnectors:
    - type: http
      port: 3080
  adminConnectors:
    - type: http
      port: 3081
  registerDefaultExceptionMappers: false

logging:
  level: INFO
  loggers:
    "root":
      additive: true
      appenders:
        - type: file
          logFormat: "%d{yyyy-MM-dd HH:mm:ss.SSS} %level [%t] %logger{15} - context=[%mdc] message=%msg %n"
          currentLogFilename: '/tmp/test-app.log'
          archivedLogFilenamePattern: '/tmp/test-app-%d{yyyy-MM-dd}.log.gz'
          archivedFileCount: 1
  appenders:
    - type: console
      logFormat: "%d{yyyy-MM-dd HH:mm:ss.SSS} %level [%t] %logger{15} - context=[%mdc] message=%msg %n"
```

I used 200 concurrent users to call the endpoint, 100 times, I'm supposed to get 120000 "start" and 20000 "end", but the result: 
image

After switching to external logback by adding a custom implementation of `io.dropwizard.logging.LoggingFactory` the issue is gone.

nickbabcock:
Document that Dropwizard defaults to asynchronous and will drop lower priority logging events if the log queue buffer starts to fill up. Closes #2681

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -619,10 +619,11 @@ Name                   Default                                  Description
 type                   REQUIRED                                 The appender type. Must be ``console``.
 threshold              ALL                                      The lowest level of events to print to the console.
 queueSize              256                                      The maximum capacity of the blocking queue.
-discardingThreshold    51                                       When the blocking queue has only the capacity mentioned in
+discardingThreshold    -1                                       When the blocking queue has only the capacity mentioned in
                                                                 discardingThreshold remaining, it will drop events of level TRACE,
                                                                 DEBUG and INFO, keeping only events of level WARN and ERROR.
-                                                                If no discarding threshold is specified, then a default of queueSize / 5 is used.
+                                                                If no discarding threshold is specified (-1), then a default of 
+                                                                queueSize / 5 (logback's default ratio) is used.
                                                                 To keep all events, set discardingThreshold to 0.
 timeZone               UTC                                      The time zone to which event timestamps will be converted.
                                                                 To use the system/default time zone, set it to ``system``.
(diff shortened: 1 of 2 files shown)


docs/source/manual/configuration.rst
```diff
@@ -673,9 +673,9 @@ type                         REQUIRED                                   The appe
 currentLogFilename           REQUIRED                                   The filename where current events are logged.
 threshold                    ALL                                        The lowest level of events to write to the file.
 queueSize                    256                                        The maximum capacity of the blocking queue.
-discardingThreshold          51                                         When the blocking queue has only the capacity mentioned in discardingThreshold
+discardingThreshold          -1                                         When the blocking queue has only the capacity mentioned in discardingThreshold
                                                                         remaining, it will drop events of level TRACE, DEBUG and INFO, keeping only events
-                                                                        of level WARN and ERROR. If no discarding threshold is specified, then a default
+                                                                        of level WARN and ERROR. If no discarding threshold is specified (-1), then a default
```
mattnelson: > then a default of queueSize / 5 is used

It took me a while to track down the origin of this. It should be called out that this is based on logback's implementation and the divisor may be subject to change.

https://github.com/qos-ch/logback/blob/v_1.2.3/logback-core/src/main/java/ch/qos/logback/core/AsyncAppenderBase.java#L104-L105

nickbabcock: Sounds good, I callout that the ratio is defined by logback.

PR #2686, opened 2019-03-09
Add data size class adhering to the correct SI and IEC prefixes

### LINKED ISSUE #2152: SizeUnit needs to be extended for MibiBytes, KibiBytes, etc
Hi guys,

I noticed today that you added support for KiB, MiB, etc a couple of years ago (https://github.com/dropwizard/dropwizard/commit/9044d44857fc1c90424303732502536c84963968), but you didn't get the semantic meaning of your extensions right. For instance, see: https://en.wikipedia.org/wiki/Gigabyte#Definition.

SizeUnit.KILOBYTE should use 1000 as the multiplier, while (the currently missing) SizeUnit.KIBIBYTE should use 1024 as the multipler. Once these additional SizeUnit values are defined, they need to be used on the proper suffixes in Size.java.

See this (10 year) old post by Jeff Atwood for a more complete picture of the differences: https://blog.codinghorror.com/gigabyte-decimal-vs-binary/. Please don't listen to the negative attitude of this article - it's 10 years old. The storage industry really does use these suffixes correctly these days and enterprise customers know the difference and expect them to be accurate.

joschi:
The `Size` and `SizeUnit` classes in Dropwizard are using a common but incorrect definition of data size units, in which 1 kilobyte equals 1024 bytes.

The `DataSize` and `DataSizeUnit` classes are using the correct definitions according to SI and IEC, in which 1 kilobyte equals 1000 bytes and 1 kibibytes equals 1024 bytes.

Instead of modifying the existing `Size` and `SizeUnit` classes, this change set introduces a new set of classes, `DataSize` and `DataSizeUnit`, with the possibility to convert the old units into the new, correct units.

Example:
```
assertEquals(Size.parse("128 kb").toDataSize(), DataSize.kibibytes(128L));
```

This is a breaking change and has to be properly documented in the release notes and upgrade notes of Dropwizard.

Closes #2152
Closes #2154

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -660,7 +660,7 @@ File
           archivedFileCount: 5
           timeZone: UTC
           logFormat: "%-5p [%d{ISO8601,UTC}] %c: %m%n%rEx"
-          bufferSize: 8KB
+          bufferSize: 8KiB
           immediateFlush: true
           filterFactories:
             - type: URI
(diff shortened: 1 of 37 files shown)


nickbabcock:
It seems add a lot of code to workaround a problem that I find to be small, but I'm sure there are those out there who would be very appreciative of a seamless transition. Though I do have to wonder, if it involves a breaking change anyways, how much better is this approach?

joschi:
> It seems add a lot of code to workaround a problem that I find to be small

Good point. I'm also willing to change this PR so that it resembles #2154 more closely, if that's the consensus.

> Though I do have to wonder, if it involves a breaking change anyways, how much better is this approach?

It's not _much_ better, but I think it makes the breaking change more obvious by using a different type and it provides at least a manual migration path if people were using the `Size` and `SizeUnit` classes in their own implementations (and not only in configuration files).

nickbabcock:
Yeah sounds good, I'm ok with this.

PR #2691, opened 2019-03-11
Add unit tests for dropwizard.util.StringsTest

paulkennethkent:
###### Problem:
I've analysed your code base and noticed that `dropwizard.util.StringsTest` is not fully tested.

###### Solution:
I've written some tests for the functions in this class with the help of [Diffblue](https://www.diffblue.com/) [Cover](https://www.diffblue.com/products). Hopefully, these tests should help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other particular classes that you consider important.

### DIFF
--- a/dropwizard-util/src/test/java/io/dropwizard/util/StringsTest.java
+++ b/dropwizard-util/src/test/java/io/dropwizard/util/StringsTest.java
@@ -1,0 +1,46 @@
+package io.dropwizard.util;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
+public class StringsTest {
+
+  @Test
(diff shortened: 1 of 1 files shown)


dropwizard-util/pom.xml
```diff
@@ -45,5 +45,10 @@
             <artifactId>jackson-databind</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
```
joschi: Please use JUnit 5 for writing tests. The dependency is already included in "test" scope for all sub-modules.

jplock: Also assertj for assertions

paulkennethkent:
Thanks for the feedback,  the tests have now been modified to use JUnit 5 and AssertJ libraries.

joschi:
@paulkennethkent Thanks!

Issue #2702, opened 2019-03-21
Allow null validationQuery


slivkamiro:
tomcat jdbc allows validationQuery to be null, see https://tomcat.apache.org/tomcat-9.0-doc/jdbc-pool.html . In that case it will validate connections by isValid() method on Connection object.

In our company, we get our database logs flooded by SELECT 1 queries, causing not being able to hold more than one day of logs. This is caused by setting up checkConnectionWhileIdle and other connection checks (which we want to keep) and the fact that in dropwizard it is not possible to set validationQuery to null. This setting, causing the call to isValid(), would help for drivers which implement this method by not executing the query.

I would happily help with implementation of this.

jplock:
Happy to review a PR for this

PR #2703, opened 2019-03-21
#2702 allowing validation query to be null

slivkamiro:
###### Problem:
tomcat jdbc allows validationQuery to be null, see https://tomcat.apache.org/tomcat-9.0-doc/jdbc-pool.html . In that case it will validate connections by isValid() method on Connection object.

###### Solution:
making validation query Optional. healthchecks calling isValid when validation query is not defined.

### DIFF
--- a/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
+++ b/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
@@ -306,6 +306,9 @@ import java.util.concurrent.TimeUnit;
  * </table>
  */
 public class DataSourceFactory implements PooledDataSourceFactory {
+
+    private static final String DEFAULT_VALIDATION_QUERY = "/* Health Check */ SELECT 1";
+
     @SuppressWarnings("UnusedDeclaration")
     public enum TransactionIsolation {
         NONE(Connection.TRANSACTION_NONE),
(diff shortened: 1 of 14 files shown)


dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
```diff
@@ -306,6 +306,9 @@
  * </table>
  */
 public class DataSourceFactory implements PooledDataSourceFactory {
+
+    public static final String DEFAULT_VALIDATION_QUERY = "/* Health Check */ SELECT 1";
```
jplock: can this be `private`?

slivkamiro: Done

jplock:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project dropwizard-hibernate: Compilation failure
[ERROR] /home/circleci/dropwizard/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/HibernateBundle.java:[82,79] error: incompatible types: Optional<String> cannot be converted to String
```

slivkamiro:
`Received 'killed' signal` not sure what happened there.

jplock:
@joschi @nickbabcock thoughts on adding this to 2.0.0?

PR #2704, opened 2019-03-22
Retrieve Connector Information in ServiceLifecycleListener 1.3.x

terezivy:
###### Problem:
In the situation where the server is using dynamic port and the port numbers and protocol are not available to bundles during the initialize method. As such they are picked up during ServerLifecylesListener's ServerStarted method. There are convenience methods for getting the first and last port, but this doesn't address the situation where there are multiple connectors configured with multiple protocols.

###### Solution:
The remedy for the problem is a method that extracts and collects all the connector information to present to the listener in question. As an example, it would provide the port, interface(host/ip), protocols, and connector type(Application/Admin) such that they could be registered as metadata in a service discovery tool such as Consul or Eureka. 

###### Result:
The consolidated information(PortDescriptor) regarding the server's connectors can be presented to the bundle(or other implementors of the LifecycleListener), via the ServiceListener in a convenient manner.

### DIFF
--- a/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/PortDescriptor.java
+++ b/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/PortDescriptor.java
@@ -5,6 +5,9 @@ import com.google.common.base.MoreObjects;
 public class PortDescriptor {
 
     private static final String UNKNOWN = "UNKNOWN";
+    private static final String DEFAULT_HOST = "0.0.0.0";
+
+    private String host;
 
     private String protocol;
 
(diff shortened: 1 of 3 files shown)


jplock:
@joschi @nickbabcock thoughts on this?  Should this go into 1.3.x or `master` (2.0)?

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/PortDescriptor.java
```diff
@@ -0,0 +1,69 @@
+package io.dropwizard.lifecycle;
+
+import com.google.common.base.MoreObjects;
+
+public class PortDescriptor {
```
jplock: could the class and all of the properties be made `final` so the class is effectively immutable?

terezivy: I can certainly do that. I will make that change on the resubmitted pull request against master.

terezivy: The pull on master is up with the discussed changes. Thanks!

jplock:
Closing in favor of #2711

Issue #1176, opened 2015-07-17
add configuration option for Proxy Protocol


devmage:
Dropwizard currently provides a configuration option to read `X-Forwarded-*` headers, `useForwardedHeaders`, but provides no similar configuration option to read [HAProxy PROXY protocol](http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt) data frames, such as those able to be sent through by [AWS ELBs](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/enable-proxy-protocol.html). The NGINX blog has [a write-up](https://www.nginx.com/blog/realtime-applications-nginx/) about this as well.

In vanilla Jetty, this is accomplished using the [ProxyConnectionFactory](http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/server/ProxyConnectionFactory.html), as described in [Jetty's docs on configuring Connectors](http://www.eclipse.org/jetty/documentation/current/configuring-connectors.html#d0e4404).

Dropwizard should include a configuration option to support reading the PROXY protocol.

devmage:
After some digging, this appears to be a Jetty 9.3.x feature, so this would entail a Jetty version bump: https://github.com/eclipse/jetty.project/commit/a308c087edae79917b40dbc8e1b5f0c0b3b15b12

arteam:
#1036 is now resolved and Dropwizard HEAD supports Jetty 9.3.*

devmage:
@arteam thanks for the tip! I'll make sure my colleagues are aware.

jplock:
Do we still need to introduce a configuration option for this or does Jetty 9.3.x fix this?

nickbabcock:
Hopefully you're able to do what you want with Jetty 9.3 in Dropwizard. Let us know if not.

wedgwoodtom:
I think this is still open.  I have the same issue and see no way to configure tcp proxy protocol support in dropwizard although Jetty supports it via a startup module.

nickbabcock:
Ok I will re-open. Ideally one might get away with doing:

```java
((ServerConnector) connector).addFirstConnectionFactory(new ProxyConnectionFactory())
```

But in the `Application::run` the connectors are not yet built so there doesn't appear to be a way to interject. So it looks like there may need to be some level of dropwizard configuration to allow the proxy protocol.

tyagiakhilesh:
[...]

PR #2709, opened 2019-03-26
Add ProxyConectionFactory for enabling the proxy protocol.

tyagiakhilesh:
###### Problem:
A DW service running behind a proxy/LB need to know the real source of the request. Attempt to solve #1176 

###### Solution:
[Proxy-Protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) is one the solution for this. Jetty has inbuilt support for it. We just need to utilize that.

###### Result:
request.getRemoteAddr() and request.getRemoteHost() shall give you information about the client that is making the request to proxy not of proxy itself (which is the case as of today without this feature).

### DIFF
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -34,6 +34,7 @@ v2.0.0: Unreleased
 * Upgrade to Jersey 2.28 (`#2613 <https://github.com/dropwizard/dropwizard/pull/2613>`_)
 * Deprecate ``*Param`` classes and will be removed in 3.0.0 (`#2637 <https://github.com/dropwizard/dropwizard/pull/2637>`_)
 * Add data size class adhering to the correct SI and IEC prefixes (`#2686 <https://github.com/dropwizard/dropwizard/pull/2686>`_)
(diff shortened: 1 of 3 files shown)


tyagiakhilesh:
Tests are failing of course, as by default the proxy protocol must be disabled. Once I figure out a way to make it configurable, things shall start looking better.

jplock:
@tyagiakhilesh thanks for this, is there anyway to provide a test that it gets enabled when `useProxyProtocol` is set to true?

tyagiakhilesh:
I have not written any test yet. In my setup i have a haproxy setup with `send-proxy` backend setup and on DW server I am setting `useProxyProtocol` to true in config.yml. I am not well versed this project's test suite. I guess we can write one integration test for this if this project uses docker and all for tests ? Any pointers where to start on that ?

tyagiakhilesh:
@jplock Done. Added the test case.

jplock:
Thanks @tyagiakhilesh, sorry I should have mentioned this before as well, but would you mind adding an entry under the 2.0.0 release notes in https://github.com/dropwizard/dropwizard/blob/master/docs/source/about/release-notes.rst for this PR?  Then this should be good to merge.

tyagiakhilesh:
@jplock Updated release notes.

PR #2711, opened 2019-03-27
Retrieve Connector Information in ServiceLifecycleListener

terezivy:
###### Problem:
In the situation where the server is using dynamic port and the port numbers and protocol are not available to bundles during the initialize method. As such they are picked up during ServerLifecylesListener's ServerStarted method. There are convenience methods for getting the first and last port, but this doesn't address the situation where there are multiple connectors configured with multiple protocols.

###### Solution:
The remedy for the problem is a method that extracts and collects all the connector information to present to the listener in question. As an example, it would provide the port, interface(host/ip), protocols, and connector type(Application/Admin) such that they could be registered as metadata in a service discovery tool such as Consul or Eureka.

###### Result:
The consolidated information(PortDescriptor) regarding the server's connectors can be presented to the bundle(or other implementors of the LifecycleListener), via the ServiceListener in a convenient manner.


dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/ServerLifecycleListenerTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test
+    void getPortDescriptorList() {
+        Server server = configureMultiProtocolServer();
+        ServerLifecycleListener listener = (server1) -> {
+        };
+        List<PortDescriptor> portDescriptorList = listener.getPortDescriptorList(server);
+        PortDescriptor[] portDescriptors = buildCompletePortDescriptorsArray();
+        Assertions.assertThat(portDescriptorList).usingElementComparator((o1, o2) -> {
+            if (Objects.equals(o1.getConnectorType(), o2.getConnectorType()) &&
```
nickbabcock: Perhaps may have been cleaner to just define `PortDescriptor::equals` (and hash). Still looks good 👌

Issue #2719, opened 2019-04-04
DropwizardTestSupport change in Dropwizard 2.x RC overwrites ConfigurationSourceProvider set by application


mzamani1:
It seems that a change (https://github.com/dropwizard/dropwizard/commit/86abeb0a187fd5c80488ba21324aaa16dabde6a3) in the Dropwizard 2.x RC in the `DropwizardTestSupport.startIfRequired(...)` method breaks some of our current usages. We currently use Dropwizard 1.3.9, for context.

Specifically, if any bundles set the `ConfigurationSourceProvider`, this line in `DropwizardTestSupport` overwrites whatever the app set one line earlier: https://github.com/dropwizard/dropwizard/commit/86abeb0a187fd5c80488ba21324aaa16dabde6a3#diff-0c89d4d1401a3b757bcfd634613fc073R291, which breaks the expected behavior of our tests. 

We have a `ConfigurationSourceProvider` that is added via a bundle, which replaces tokens in the config.yml file with overrides specified elsewhere. The result is that this `ConfigurationSourceProvider` is overwritten in our tests.

mzamani1:
One approach to fixing this might be to only override the app's `ConfigurationSourceProvider` if it's provided as a constructor argument to `DropwizardTestRule`, and to otherwise defer to whatever the app sets otherwise. 

I'd be happy to submit a PR for this change if that sounds reasonable.

PR #2720, opened 2019-04-05
DropwizardTestSupport should only override the app's ConfigurationSou…

mzamani1:
…rceProvider if one is explicitly provided in a constructor (#2719)

###### Problem:

See #2719 

###### Solution:

Made `DropwizardTestSupport`s `ConfigurationSourceProvider` override nullable, and only override the app's `ConfigurationSourceProvider` if one is explicitly provided via a constructor.

###### Result:

This will fix the behavior of `DropwizardTestSupport` in tests when the app sets a different `ConfigurationSourceProvider`.

### DIFF
--- a/dropwizard-testing/src/test/java/io/dropwizard/testing/DropwizardTestSupportWithResourceConfigProviderTest.java
+++ b/dropwizard-testing/src/test/java/io/dropwizard/testing/DropwizardTestSupportWithResourceConfigProviderTest.java
@@ -1,16 +1,21 @@
 package io.dropwizard.testing;
 
 import io.dropwizard.configuration.ResourceConfigurationSourceProvider;
+import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
-import io.dropwizard.testing.app.TestApplication;
 import io.dropwizard.testing.app.TestConfiguration;
 import org.junit.jupiter.api.AfterAll;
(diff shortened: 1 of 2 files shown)


dropwizard-testing/src/test/java/io/dropwizard/testing/DropwizardTestSupportWithResourceConfigProviderTest.java
```diff
@@ -41,4 +47,36 @@ void returnsEnvironment() {
...
+    @Test
+    void doesNotOverwriteConfigSourceProviderIfNotProvidedInConstructor() throws Exception {
+
+        DropwizardTestSupport<TestConfiguration> support = new DropwizardTestSupport<>(
+                TestApplication.class, "test-config.yaml");
+        try {
+            support.before();
+            assertThat(TEST_CONFIG_SOURCE_PROVIDER.openCalled).isTrue();
+        } catch (Exception e) {
+            Assert.fail();
```
nickbabcock: This catch statement isn't needed, no?

mzamani1: Yep, definitely not needed. Pushed a fix.

mzamani1:
@nickbabcock Anything else to consider with this one? Or does this approach seem tenable?

PR #2723, opened 2019-04-06
Use ReadTheDocs theme for documentation

joschi:
Instead of hosting the Dropwizard documentation on GitHub pages and having to update the `gh-pages` branch (semi-) manually after each release, we can use https://readthedocs.org/ which will take care of building the documentation and serving it in multiple versions.

While the original Dropwizard theme could be used on ReadTheDocs, the official RTD theme providers some features which we would have to add to the original theme, such as:

* Search function
* Version selection
* Automatic sitemap generation


joschi:
Preview: https://dropwizard.readthedocs.io/en/rtd/

@dropwizard-committers When somebody wants to port the search bar and the version selection to Dropwizard's original theme, I'm happy to keep using it.

pstackle:
Overall, looks good, but one unfortunate downside I see is that the [Dropwizard Configuration Reference](https://dropwizard.readthedocs.io/en/rtd/manual/configuration.html) page becomes much harder to use due to the layout being so narrow.

joschi:
@pstackle I've made the theme a little bit wider to accommodate wide tables (although the configuration reference still doesn't fit).

@dropwizard-committers If you're okay with it, I'd merge this PR and release Dropwizard 2.0.0-rc1 in the next days.

The latest version of the docs created from this branch are at https://dropwizard.readthedocs.io/en/rtd/.

PR #2748, opened 2019-04-28
Don't use X-Forwarded-* headers by default

### LINKED ISSUE #2663: Default useForwardedHeaders setting leaves applications vulnerable to X-Forwarded-For spoofing
The `useForwardedHeaders` setting in the [HTTP connector](https://www.dropwizard.io/1.3.9/docs/manual/configuration.html#man-configuration-http) is set to true by default. The documentation specifies that having this enabled will make Dropwizard honor `X-Forwarded-*` headers added by proxies (via the usage of Jetty's [ForwardedRequestCustomizer](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-server/src/main/java/org/eclipse/jetty/server/ForwardedRequestCustomizer.java)), but it's not clear what is meant by "look at", ie, what actually is affected by this setting.

One of the undocumented implications of having this turned on is that if the `X-Forwarded-For` HTTP header is set on the incoming request, it actually changes the output of `ServletRequest::getRemoteHost`. The [servlet spec](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletRequest.html#getRemoteAddr--) for this method specifies that the output of that method should be the IP address of the client or the last proxy that sent the request (ie, the IP address of the remote peer in the TCP connection), making no reference to `X-Forwarded-For` or any other HTTP headers, so this is an unexpected change.

An implication of this is that anyone that can manipulate the `X-Forwarded-For` HTTP header can cause Dropwizard to use a spoofed value both for its internal logging and by providing this spoofed value via `ServletRequest::getRemoteHost`, potentially bypassing access control restrictions or auditing that an application might implement if it relies on this value.

With that setting on, any Dropwizard API that is exposed to the Internet would be vunerable to this type of remote IP address spoofing, unless all public facing load balancers that could potentially serve as reverse proxies to said application are actively performing sanitization of the `X-Forwarded-For` header (clearing it out fully instead of just appending its own IP address to an already existing header). With either no load balancers configured in that way or when exposed directly to the Internet, any Dropwizard application that relies on `getRemoteHost` can be tricked into using arbitrary data set by a malicious party instead of the client's remote IP address – this includes the access logs generated by Dropwizard.

Ideally, the default for this option should be `false`, to avoid undesired insecure behavior for situations where Dropwizard is not deployed behind a reverse proxy that is actively performing this sanitization. However, since that could potentially break existing applications, at the very least, documentation should be ammended to that the default setting of `true` will cause `ServletRequest::getRemoteHost` and Dropwizard's access logs to return the leftmost value in the `X-Forwarded-For` HTTP header and that sanitization of that header by a public-facing reverse proxy would be necessary if full accuracy in identifying the remote address of the client is required.

This is fairly easy to reproduce; on a sample 1.3.9 Dropwizard project, with the default configuration, with a single resource, sending a request while setting the `X-Forwarded-For` header to `dummy`, the following is logged in the access logs:
```dummy - - [26/Feb/2019:11:02:35 +0000] "GET /test HTTP/1.1" 200 19 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.119 Safari/537.36" 78```

After setting `useForwardedHeaders` to false and repeating the same request:
```127.0.0.1 - - [26/Feb/2019:11:25:56 +0000] "GET /test HTTP/1.1" 200 18 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.119 Safari/537.36" 86```

joschi:
Closes #2663

### DIFF
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -36,6 +36,7 @@ v2.0.0: Unreleased
 * Add data size class adhering to the correct SI and IEC prefixes (`#2686 <https://github.com/dropwizard/dropwizard/pull/2686>`_)
 * Added PortDescriptor class and method in ServerLifeCycleListener to provide a list of PortDescriptors, detailing all listening information for the application (`#2711 <https://github.com/dropwizard/dropwizard/pull/2711>`_)
 * Add support for proxy-protocol in http connector configuration (`#2709 <https://github.com/dropwizard/dropwizard/pull/2709>`_)
+* Disable using ``X-Forwarded-*`` headers by default (`#2748 <https://github.com/dropwizard/dropwizard/pull/2748>`_)
 
 .. _rel-1.3.9:
 
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -313,7 +313,7 @@ HTTP
           reuseAddress: true
           useServerHeader: false
           useDateHeader: true
-          useForwardedHeaders: true
+          useForwardedHeaders: false
           httpCompliance: RFC7230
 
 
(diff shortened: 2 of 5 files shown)


jplock:
Can you add this to the release notes too please?

mattnelson:
> Can you add this to the release notes too please?

I think this should be added to the 2.0 upgrade guide as well, since it is a non-passive change.

PR #2786, opened 2019-05-31
Add current class loader to javassist ClassPool

josephlbarnett:
In certain environments (eg AWS Lambda), the default/system
class loader doesn't have dropwizard classes in it.  Use the
current class's class loader to ensure that it does.

###### Problem:
In certain environments (eg AWS Lambda), the default/system
class loader doesn't have dropwizard classes in it, and we see the following:
```
javassist.NotFoundException: io.dropwizard.jersey.DropwizardResourceConfig$SpecificBinder
    at javassist.ClassPool.get(ClassPool.java:422)
    at io.dropwizard.jersey.DropwizardResourceConfig.register(DropwizardResourceConfig.java:152)
```
###### Solution:
Use the current class's class loader to ensure that it does.

###### Result:
Running dropwizard in aws lambda works (with some glue between an aws RequestHandler and the dropwizard application's resourceConfig).  Running as a standard shaded jar also still works.
Note that I don't know a good way to write a test for this, as the failure is dependent on aws lambda's underlying jvm implementation.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -13,6 +13,7 @@ import io.dropwizard.jersey.validation.FuzzyEnumParamConverterProvider;
 import io.dropwizard.util.Strings;
 import javassist.ClassPool;
 import javassist.CtClass;
+import javassist.LoaderClassPath;
 import org.glassfish.jersey.internal.inject.AbstractBinder;
 import org.glassfish.jersey.internal.inject.Binder;
 import org.glassfish.jersey.internal.inject.Providers;
@@ -148,6 +149,7 @@ public class DropwizardResourceConfig extends ResourceConfig {
                 // Need to create a new subclass dynamically here because Jersey
                 // doesn't add new bindings for the same class
                 ClassPool pool = ClassPool.getDefault();
+                pool.insertClassPath(new LoaderClassPath(this.getClass().getClassLoader()));
                 CtClass cc = pool.makeClass(SpecificBinder.class.getName() + UUID.randomUUID());
                 cc.setSuperclass(pool.get(SpecificBinder.class.getName()));
                 Object binderProxy = cc.toClass().getConstructor(Object.class, Class.class).newInstance(object, clazz);


dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -148,6 +149,7 @@ public ResourceConfig register(final Object component) {
                 // Need to create a new subclass dynamically here because Jersey
                 // doesn't add new bindings for the same class
                 ClassPool pool = ClassPool.getDefault();
+                pool.insertClassPath(new LoaderClassPath(this.getClass().getClassLoader()));
```
josephlbarnett: would it be better to make the ClassPool object a member variable instead of creating a new one for each registration?  not sure....

jplock:
Running in a lambda is an interesting idea.  I've wondered if it would make sense to provide a way not to launch jetty and instead provide a hook to do a per-request processing directly.

josephlbarnett:
Yeah, we currently bypass much of the startup infrastructure, create a `DropwizardResourceConfig`, and register all the same resources we register when we use the regular Application/Cli based startup.

Then we store the return of aws lambda-jersey's `JerseyLambdaContainerHandler.getAwsProxyHandler(dropwizardResourceConfig)` in a singleton, and have our lambda RequestHandler do `singleton.proxy(input, context)`.

Main downsides are: startup time is a bit of a beast, duplicating the startup logic is meh but could probably be cleaned up & abstracted some. And so we actually aren't running lambda in production for those and other reasons right now, but having a service that can be deployed as both a lambda function and as an executable jar and expose the same business logic / apis either way is pretty handy.

mattnelson:
> Main downsides are: startup time is a bit of a beast

I would be interested in hearing more about this. Particularly what impact all the new ClassDataSharing(CDS) features being added would have on those times.

josephlbarnett:
> I would be interested in hearing more about this. Particularly what impact all the new ClassDataSharing(CDS) features being added would have on those times.

It's been a while since I looked at it, but I believe startup times were in the ~15-20s range on lambda with 1G of memory allocated (vs 2-3s local startup time).  We use guice for DI, and switching that to dagger instead saved some time, but not much (maybe 3-5 seconds?).  But basically we get into the lambda entry point quickly, and then things are blocked while our DI resources initialize and the jersey context gets set up (possibly hk2 DI has similar reflection based approaches as guice?  but hard to tell how much time is spent loading classes vs/ actually executing code, etc).

w/r/t CDS features, I don't know what impact they'd have, but lambda is still on java 8, so not sure when it would even get those features either.

josephlbarnett:
oh, also, fwiw this is a regression from 1.3.x.  Not a blocker for us as we're not using the lambda implementation at the moment.

nickbabcock:
Does it make sense to write a test for this so that we don't regress on this (again) in the future?

josephlbarnett:
it makes sense to write a test, but I'm not seeing a good way to test for it -- it looks to me like:

a) ClassPool doesn't expose enough to assert that the DropwizardResourceConfig's classloader is in the sources list
b) testing that the dropwizard libraries are loadable through the classpool will pass on standard JVMs while still fail on lambda VMs (i think in my limited testing the system classloader is the same as the DRC's classloader).  

Unless there's a way to configure the system classloader to be effectively empty just for this test?  my classloader knowledge is limited, so open to other suggestions as well.

nickbabcock:
IMO that's fine, it's not apparent to me how to write a regression test for this either.

jplock:
Thanks for the fix @josephlbarnett !

Issue #2794, opened 2019-06-10
An opt-in Request Log Filter for configurable URLs


mzamani1:
Would others find it useful to have a request log filter factory they can use to configure specific URLs to filter out request logs for? I personally find this somewhat useful, if for instance my health check request logs don't provide *much* value, and it's helpful to filter out some of that noise to see real user traffic.

PR #2795, opened 2019-06-10
Adds a request logging uri filter 

mzamani1:
###### Problem:
#2794 
 
###### Solution:
Adds a `UriFilterFactory` class to the `dropwizard-request-logging` module, so that users of Dropwizard can add this in without pulling in any other dependencies if they find it useful.

###### Result:
If this factory is used and configured, any specified paths (only fully qualified paths, regexes don't work in this implementation) will be filtered out of request logs. 

If a user wanted to filter out their request logs for the `/health-check` path (if they were using https://github.com/dropwizard/dropwizard-health :) ), their request logs would only contain logs for requests to paths that are not equal to `/health-check`.

### DIFF
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -38,6 +38,7 @@ v2.0.0: Unreleased
 * Add support for proxy-protocol in http connector configuration (`#2709 <https://github.com/dropwizard/dropwizard/pull/2709>`_)
 * Disable using ``X-Forwarded-*`` headers by default (`#2748 <https://github.com/dropwizard/dropwizard/pull/2748>`_)
 * Fix typo by renaming ``ResilentSocketOutputStream`` to ``ResilientSocketOutputStream`` (`#2766 <https://github.com/dropwizard/dropwizard/pull/2766>`_)
+* Adds an opt-in URI request logging filter factory (`UriFilterFactory`)  (`#2794 <https://github.com/dropwizard/dropwizard/pull/2795>`_)`
 
 .. _rel-1.3.9:
 
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -1178,6 +1178,24 @@ Reference ``SecretFilterFactory`` type in our configuration.
 
 The last step is to add our class (in this case ``com.example.SecretFilterFactory``) to ``META-INF/services/io.dropwizard.logging.filter.FilterFactory`` in our resources folder.
 
+.. _man-core-request-log-url-filtering:
+
+Filtering Request Logs for a Specific URI
+-----------------------------------------
+
+Reference ``UriFilterFactory`` type in your configuration.
+
+.. code-block:: yaml
+
+    server:
+      requestLog:
+        appenders:
+          - type: console
+            filterFactories:
+              - type: uri
+                uris:
+                  - "/health-check"
+
 .. _man-core-testing-applications:
 
 Testing Applications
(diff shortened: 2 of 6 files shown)


jplock:
@mzamani1 can you also add a 2.0.0 release notes entry to https://github.com/dropwizard/dropwizard/blob/master/docs/source/about/release-notes.rst and how about an example in https://github.com/dropwizard/dropwizard/blob/master/docs/source/manual/core.rst#logging ?

mzamani1:
@jplock Sure!

dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/filter/UrlFilterFactory.java
```diff
@@ -0,0 +1,43 @@
...
+    public void setUrls(final Set<String> urls) {
+        this.urls = urls;
+    }
+
+    @Override
+    public Filter<IAccessEvent> build() {
+        return new Filter<IAccessEvent>() {
+            @Override
+            public FilterReply decide(final IAccessEvent event) {
+                if (urls.contains(event.getRequestURI())) {
```
mattnelson: Rename to uri to match up with the field from the event. Since url and uri are both available.

https://github.com/qos-ch/logback/blob/8fccc8879641be6c4ad7da0f2a6057f6bb9fbd47/logback-access/src/main/java/ch/qos/logback/access/spi/IAccessEvent.java#L83-L88

mzamani1:
@jplock Done.

dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/filter/UrlFilterFactory.java
```diff
@@ -0,0 +1,43 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import io.dropwizard.logging.filter.FilterFactory;
+
+import java.util.Collections;
+import java.util.Set;
+
+import javax.validation.constraints.NotNull;
+
+@JsonTypeName("url-filter-factory")
```
mattnelson: For consistency with the server factories this can be shortened to uri.

https://github.com/dropwizard/dropwizard/blob/c05a7aa6cff717683ccf9aa43130c5eb3a7a1c45/dropwizard-core/src/main/java/io/dropwizard/server/DefaultServerFactory.java#L69-L70

https://github.com/dropwizard/dropwizard/blob/c05a7aa6cff717683ccf9aa43130c5eb3a7a1c45/dropwizard-core/src/main/java/io/dropwizard/server/SimpleServerFactory.java#L56-L57

mzamani1: I see in the docs an example request logging filter has the suffix `-filter-factory`, so I'm a little unsure if there is a convention for log filter factories, or if I should just follow other factory naming conventions.

mzamani1:
@mattnelson 

> We only deny when the request is 200. We want to know of any abnormal response code. We also compile the list of URIs into a regex to do a predicate match.

The HTTP status code filtering seems like it should live in a separate factory. In terms of path regexes, I agree that'd be useful, but I didn't want have a need for that personally yet (I use this same request log filter in production today), so I didn't get around to it just yet.

dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/filter/UrlFilterFactory.java
mzamani1: I'll switch to `uri` as the type, unless I hear there is a different naming convention though.

jplock:
Any additional comments @mattnelson ?

mzamani1:
Unless @mattnelson has any additional feedback, I think this is safe to merge, in the interest of `2.0.0` readiness.

Issue #2553, opened 2018-11-12
Environment.lifecycle().executorService(...) might create single-threaded pools


efimovandr:
Following snippet suddenly creates a pool with at most one worker thread instead of N:
```
ExecutorService executorService = environment.lifecycle()
     .executorService("my-pool-%d")
    .maxThreads(N)
    .build();
```
Here is why: https://github.com/kimchy/kimchy.github.com/blob/master/_posts/2008-11-23-juc-executorservice-gotcha.textile

minisu:
Confirmed. This code takes 10 seconds to execute (but only 1 second if you add `.minThreads(10)`):

```java
@Override
public void run(TestingConfiguration configuration, Environment environment) throws Exception {
    ExecutorService executorService = environment
        .lifecycle()
        .executorService("testing")
        .maxThreads(10)
        .build();

    long startTime = currentTimeMillis();
    CountDownLatch countDownLatch = new CountDownLatch(10);

    for(int i = 0; i < 10; i++) {
        executorService.submit(() -> {
            try {
                Thread.sleep(1000);
            } catch (InterruptedException e) {
                e.printStackTrace();
            }
            countDownLatch.countDown();
        });
    }

    countDownLatch.await();
    System.out.println("Done in " + (currentTimeMillis() - startTime) + " ms");
}
```

xiaodong-xie:
If more fine grained controls are needed, probably need to set both `minThreads` and `workQueue`.

minisu:
When looking closer, there's actually a [log.warn](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java#L102) printed if a user does this misconfiguration.

efimovandr:
I think the API provided is misleading and people will keep making this mistake.

nickbabcock:
> I think the API provided is misleading and people will keep making this mistake.

Now would be a great time if we want to rethink this. So if you have an idea or implementation in mind, I'd love to hear it, as I would agree it is easy to fall into this trap

sergeiwp:
I think it makes sense to keep method naming closer to ThreadPoolExecutor, i.e. `coreThreads()` instead of `minThreads()`. Also worth referring [ThreadPoolExecutor doc](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html) for the fields meaning.

[...]

PR #2797, opened 2019-06-12
Use SynchronousQueue as the default queue

dennyac:
###### Problem:
I'm attempting to address the issue where Environment.lifecycle().executorService(...) might create single-threaded pools - https://github.com/dropwizard/dropwizard/issues/2553 . Since the default queue is unbounded we will end up with single-threaded pools

###### Solution:
I've changed the default queue to ```SynchronousQueue```. I gave this some thought and felt it might be appropriate to use ```SynchronousQueue``` as the default queue, as it just does a direct handoff. Libraries like [Hystrix](https://github.com/Netflix/Hystrix) use ```SynchronousQueue``` as the default queue if [maxQueueSize](https://github.com/Netflix/Hystrix/wiki/Configuration#maxQueueSize) isn't specified. ```SynchronousQueue``` is used in ```Executors.newCachedThreadPool()``` as well.

###### Result:
```ExecutorService``` will use between the specified ```minThreads``` and ```maxThreads``` by default. With this change, the warn message "Parameter 'maximumPoolSize' is conflicting with unbounded work queues" will show up only if an unbounded queue is provided explicitly

### DIFF
--- a/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
+++ b/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
@@ -105,8 +105,8 @@ public class ExecutorServiceBuilderTest {
             .maxThreads(Integer.MAX_VALUE)
             .build();
 
-            verify(log, never()).warn(anyString());
-            assertCanExecuteAtLeast2ConcurrentTasks(exe);
+        verify(log, never()).warn(anyString());
+        assertCanExecuteAtLeast2ConcurrentTasks(exe);
     }
 
     @Test
--- a/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
+++ b/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
@@ -99,22 +99,14 @@ public class ExecutorServiceBuilderTest {
         verify(log, never()).warn(anyString());
     }
 
-    /**
-     * Setting large max threads without large min threads is misleading on the default queue implementation
-     * It should warn or work
-     */
     @Test
-    public void shouldBeAbleToExecute2TasksAtOnceWithLargeMaxThreadsOrBeWarnedOtherwise() {
+    public void shouldBeAbleToExecute2TasksAtOnceWithLargeMaxThreadsAndDefaultMinThreads() {
         ExecutorService exe = executorServiceBuilder
             .maxThreads(Integer.MAX_VALUE)
             .build();
 
-        try {
-            verify(log).warn(anyString());
-        } catch (WantedButNotInvoked error) {
-            // no warning has been given so we should be able to execute at least 2 things at once
+            verify(log, never()).warn(anyString());
             assertCanExecuteAtLeast2ConcurrentTasks(exe);
-        }
     }
 
     @Test
--- a/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
+++ b/dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
@@ -42,6 +42,7 @@ public class ExecutorServiceBuilderTest {
         executorServiceBuilder
             .minThreads(4)
             .maxThreads(8)
+            .workQueue(new LinkedBlockingQueue<>())
             .build();
 
         verify(log).warn(WARNING);
--- a/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
+++ b/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
@@ -10,8 +10,8 @@ import java.util.Locale;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.atomic.AtomicLong;
@@ -39,7 +39,7 @@ public class ExecutorServiceBuilder {
         this.allowCoreThreadTimeOut = false;
         this.keepAliveTime = Duration.seconds(60);
         this.shutdownTime = Duration.seconds(5);
-        this.workQueue = new LinkedBlockingQueue<>();
+        this.workQueue = new SynchronousQueue<>();
         this.threadFactory = factory;
         this.handler = new ThreadPoolExecutor.AbortPolicy();
     }

### CODE REVIEW

**Comment by jplock** [comment]
Are there any tests that can be added for this to highlight the issue to prevent regressions?

**Comment by dennyac** [comment]
Good point. I'll try to add tests.

**Comment by dennyac** [comment]
@jplock I could actually modify one of the existing tests to prevent future regressions. Let me know if that is sufficient.

File: dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java [hunk 1]
```diff
@@ -98,22 +99,14 @@ public void shouldNotWarnWhenUsingTheDefaultConfiguration() {
...
+    public void shouldBeAbleToExecute2TasksAtOnceWithLargeMaxThreadsAndDefaultMinThreads() {
         ExecutorService exe = executorServiceBuilder
             .maxThreads(Integer.MAX_VALUE)
             .build();
 
-        try {
-            verify(log).warn(anyString());
-        } catch (WantedButNotInvoked error) {
-            // no warning has been given so we should be able to execute at least 2 things at once
+            verify(log, never()).warn(anyString());
```
jplock: can you fix the indentation on these two lines?

dennyac: Fixed indentation!

isaki:
There is a problem with this change is that current implementations which rely on quick handoffs to an executor could now block; especially when the configured executor is using a fixed thread pool (which is a use case where this change adds no value as per the original post). I think a check is necessary that if a fixed thread pool is in use (i.e. min == max) then using something like LinkedBlockingQueue should be the default.

While I appreciate the attempt to address this issue, the consequences of such a change for those who are used to DropWizard acting a certain way may not appreciate their threads suddenly blocking where they weren't before. I, for example, use DropWizards ExecutorService building capability but only ever use fixed thread pools, so I would be negatively impacted by this change unless I go find every single place all my projects are calling this and explicitly provide a backing queue.

The original blog post is here: https://github.com/kimchy/kimchy.github.com/blob/master/_posts/2008-11-23-juc-executorservice-gotcha.textile

This also covers a possible implementation to work around this issue.

It is my opinion that this PR does not address the full complexity of the issue at hand. If DropWizard wants to go this route, at the very least documentation needs to be updated and warnings put into the release notes as this will have consequences to all projects that rely on this functionality. I'd be willing to take a stab at addressing this in a minimally invasive way, but I don't want to step on OP's toes.

It is also entirely possible I missed something, so if I'm mistaken in my assessment, please let me know!

dennyac:
"especially when the configured executor is using a fixed thread pool (which is a use case where this change adds no value as per the original post). I think a check is necessary that if a fixed thread pool is in use (i.e. min == max) then using something like LinkedBlockingQueue should be the default."

I'm not sure I understand why the behavior should be any different with fixed thread pools. @isaki Could you elaborate?

Agree that there may be a small performance cost in blocking to perform the handoff, but that may be negligible (This is subjective and based on the application). In primitive tests I did on my machine the difference is in nanoseconds. In terms of throughput, SynchronousQueue seems to perform better ([Stack Overflow post](https://stackoverflow.com/questions/5102570/implementation-of-blockingqueue-what-are-the-differences-between-synchronousque)). PS - These tests will vary from system to system and based on the params passed

Correct me if I'm wrong or have missed something.

dennyac:
The current behavior isn't intuitive and there have been multiple issues raised in the past regarding this. This PR is just to address that. To be consistent, I referred to the default behavior used in other mature libraries.

dennyac:
With this PR, the default behavior will be to accept tasks between the min and max specified threads. Submitting new tasks will block for the hand-off to an available thread (or spawn a new thread). If we've reached the max threads, and all threads are busy, when a new task is submitted, behavior will depend on the RejectedExecutionHandler (default is to throw RejectedExecutionException).

nickbabcock:
I agree with @isaki.

Here's an example that demonstrates the issue:

```java
final ExecutorService service = env.lifecycle().executorService(null).maxThreads(10).build();
logger.info("Start submitting");
for (int i = 0; i < 20; i++) {
    int step = i;
    service.execute(() -> {
        try {
            Thread.sleep(1000);
            logger.info("{} done", step);
        } catch (InterruptedException ignored) {
        }
    });
}
logger.info("End submitting");
```

Notes on current dropwizard implementation:

- Elapsed time between "Start submitting" and "End submitting": 1ms
- Elapsed time for all submitted jobs to finish: 20s
- Current implementation is sub-optimal as despite specifying `maxThreads`, only one thread is used. 

Notes on PR implementation:

- Will fail with rejection exception as 20 tasks > 10 threads. This can be modified to work if a dropwizard user sets `.rejectedExecutionHandler(new ThreadPoolExecutor.CallerRunsPolicy())`, which has the following characteristics:
- Elapsed time between "Start submitting" and "End submitting": 2s
- Elapsed time for all submitted jobs to finish: 3s
- PR implementation is sub-optimal as previously the quick handoffs to `ExecutorService#execute` are gone and rejection exception can be startling.

Notes for a desired implementation:

- Elapsed time between "Start submitting" and "End submitting": 1ms
- Elapsed time for all submitted jobs to finish: 2s
- When tasks > threads, store tasks, by default, in an infinite queue, so we get the best of both worlds: backwards compatible in queuing and rejection behavior, in addition to the additional thread utilization.

How do we get to the desired implementation? As already mentioned, [kimchy's](https://github.com/kimchy/kimchy.github.com/blob/master/_posts/2008-11-23-juc-executorservice-gotcha.textile) solution seems viable. I copied, pasted, and adapted the `ForceQueuePolicy` and `ScalingQueue` and achieved the desired results. Even simpler than the `ScalingQueue` ([as pointed out by this SO answer](https://stackoverflow.com/a/24493856/433785)) is `ForceQueuePolicy` coupled with:

```java
final BlockingQueue<Runnable> queue = new LinkedTransferQueue<Runnable>() {
    @Override
    public boolean offer(Runnable e) {
        return tryTransfer(e);
    }
};
```

And as @isaki was alluding to, there can be slightly different behavior in queues / policies if it makes sense depending on min == max thread pool

Don't worry everyone, this is a confusing topic to wrap one's head around 😅

jplock:
I like the idea of including a variation of kimchy’s solution as the default and adapting it as necessary.

dennyac:
Hmm. I guess I can understand why the default behavior of rejecting tasks when all threads are busy may confuse people, especially users upgrading dropwizard.

isaki:
I actually have a solution to this problem in a code base I'm working on open sourcing. It's an executor similar to kimchi's but with a few minor fixes/changes. Worst case I can just write a new version for DropWizard as it's a process to get things open sourced where I work (but I can contribute freely to existing projects).

jplock:
@dennyac did you want to update this PR based on @nickbabcock’s feedback, or how would you like to proceed?

dennyac:
@jplock I'll attempt to update the PR based on @nickbabcock 's feedback. Will request for help if needed as well. Thanks.

dennyac:
Could also close this and raise another PR to avoid adding unnecessary commits. I'm fine either ways. And thanks for all the feedback!

isaki:
There are bugs in the proposed implementation in the blog. I have a working executor that behaves as expected, but it is is currently closed source. I hope to be open sourcing this code in the next month or two (requires some refactoring and approval from legal). If that is too long to wait, I will review any proposed solution to provide feedback to fix those issues if it contains them.

The biggest thing I remember off the top of my head is that the rejected executor proposal doesn't handle a terminating or shutdown executor.

dennyac:
Since you have the most context, I'll let you proceed with the implementation. If there are any change in plans or further delays, I will attempt to implement it with your assistance/feedback.

Thanks again @isaki !

isaki:
I've been granted permission to contribute my executor to DropWizard. I will submit a PR this week.

isaki:
There are some discussions as to if we will create an open source repo or if I'll actually put the classes directly into DropWizard. In the meantime, I'm submitting a PR to fix errors for users of the Eclipse IDE.

joschi:
@isaki Any news about the donation of your executor code?

isaki:
I actually just following up with management last week regarding our open source strategy. I'll poke them again this week.

isaki:
Update: I don't have time to open source this as an official project, however, I have the blessing to give you what you need. I'll be taking some time off at the end of December and will start the process of creating my company's open source projects, which for day one will contain only the executor infrastructure you need at this time. Down the road, I will add other stuff on the open source roadmap. Unfortunately things like this take time in the corporate world and I want to make sure I do things according to company policy (which is still being created/discussed).

PR #2811, opened 2019-06-21
Allow to setNormalizeUri on HTTP client

cyberdelia:
###### Problem:

Recent versions (>=4.5.8) of HTTPClient introduced a normalizeUri
settings that prevent URL rewrites introduce in earlier versions.

Being able to disable said settings is sometimes useful when dealing
with legacy systems.

See: https://issues.apache.org/jira/browse/HTTPCLIENT-1968

###### Solution:

This allow to enable or disable `normalizeUri` settings, the default is still true, therefore the existing behaviour is unchanged.


jplock:
Thanks for the contribution!

PR #2819, opened 2019-06-29
Allow simple logger level config to support "OFF"

### LINKED ISSUE #2818: Logger level of OFF results in "Unsupported format of logger"
When configuring a logger level of `OFF` via "logging -> loggers" using the simple form of `com.example.my.logger: OFF`, fails with "Unsupported format of logger".

This is a corner case because the yaml parsing converts `OFF` to `false` and the simple level config is only expecting a String/textual value. The other level parsing in `DefaultLoggingFactory` use `toLevel()` to handle this case. 

A workaround is to quote the level: `com.example.my.logger: "OFF"` 

PR with a fix to follow.

pstackle:
Fixes #2818

###### Problem:
`DefaultLoggingFactory` doesn't allow for `OFF` to be configured as a logger level in the simple config case. Example: 
```
logging:
  loggers:
    "com.example.my.logger": OFF
```
results in "Unsupported format of logger".

###### Solution:
Updated the if check to accept boolean values since `OFF` is translated to `false` in YAML and use existing `toLevel` method that properly handles boolean as `Level.OFF`.

###### Result:
`OFF` will be allowed as a logger level in the simple case without requiring it to be quoted.

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/DefaultLoggingFactory.java
@@ -291,9 +291,9 @@ public class DefaultLoggingFactory implements LoggingFactory {
         for (Map.Entry<String, JsonNode> entry : loggers.entrySet()) {
             final Logger logger = loggerContext.getLogger(entry.getKey());
             final JsonNode jsonNode = entry.getValue();
-            if (jsonNode.isTextual()) {
+            if (jsonNode.isTextual() || jsonNode.isBoolean()) {
                 // Just a level as a string
-                logger.setLevel(Level.valueOf(jsonNode.asText()));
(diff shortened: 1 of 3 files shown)


GuiSim:
If you do hit this issue, a temporary workaround is to write `OFF` in double quotes: 
`"my.logger": OFF #Doesn't work`
`"my.logger": "OFF" #Works!`

joschi:
@GuiSim Thanks for pointing this out.

The issue is that YAML up to version 1.2 interprets the string OFF (without quotes) as a boolean value (false). 🙄

> A Boolean represents a true/false value. Booleans are formatted as English words (“true”/“false”, “yes”/“no” or “on”/“off”) for readability and may be abbreviated as a single character “y”/“n” or “Y”/“N”. 

https://yaml.org/type/bool.html

Issue #2534, opened 2018-10-24
Authenticator : access request path


n1b0r:
Hello

Is there a way to get the requested url in my custom Authenticator ?

My auth mecanism needs the requested url to check if user is allowed to access resource.

nib

FredDeschenes:
Don't know if this is still relevant, but you could write your own `AuthFilter`. The [`authenticate`](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java#L132) method gives you access to the `ContainerRequestContext` which has the `URIInfo`.

But why not just use an [authorizer](https://www.dropwizard.io/1.3.9/docs/manual/auth.html#authorizer)?

n1b0r:
Thanks for the information.

PR #2837, opened 2019-07-10
Pass request UriInfo to Authorizer.authorize

cyberdelia:
###### Problem:

I was looking for a way to authorize a request based on both the `Principal` and path parameters, without having to rewrite an `AuthFilter` or more.

This comes up if you have a `Principal` that has different permissions or roles depending on the object it's accessing, for example: 

A user can access its `Company`, can create a `Case`, but cannot read somebody else `Case`:
```java
@RolesAllowed({"read", "write"})
@GET
public Case getPlan(@Auth User user, @PathParam("company") String company, @PathParam("case") String case) {
    return dao.findCase(user, company, case);
}

@RolesAllowed({"write"})
@POST
public Case getPlan(@Auth User user, @PathParam("company") String company) {
    return dao.createCase(user, company);
}
```

There is currently no real way to authorize this kind of logic using the existing`Authorizer`, since it only aware of `User`/`Principal`.

I've seen a few mention of this problem on the mailing list and it's also related to issue #2534.

###### Solution:

This pass the `UriInfo` from each request to `Authorizer.authorize`, so that you can access path or query parameters to decide if a request is authorized or not. The new signature default to calling `authorize(User principal, String role)`, so it should be backward compatible.

But otherwise the change would look like this:

```java
public class Authorizer implements Authorizer<User> {
    @Override
    public boolean authorize(User principal, String role) {
        return authorize(principal, role, null);
    }

    @Override
    public boolean authorize(User principal, String role, UriInfo uriInfo) {
        return uriInfo.getUriInfo().getPathParameters().getFirst("company").equals("dropwizard");
    }
}
```

It does seems strange that I would be the first one to encounter this use case, with both dropwizard and/or Jersey. So I'm sort of expecting that someone has a much better solution that does not involve rewriting the whole authentication/authorization stack, but if not, this seems like a decent compromise.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
@@ -152,7 +152,7 @@ public abstract class AuthFilter<C, P extends Principal> implements ContainerReq
 
                 @Override
                 public boolean isUserInRole(String role) {
-                    return authorizer.authorize(prince, role, requestContext.getUriInfo());
+                    return authorizer.authorize(prince, role, requestContext);
                 }
 
                 @Override
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
@@ -152,7 +152,7 @@ public abstract class AuthFilter<C, P extends Principal> implements ContainerReq
 
                 @Override
                 public boolean isUserInRole(String role) {
-                    return authorizer.authorize(prince, role);
+                    return authorizer.authorize(prince, role, requestContext.getUriInfo());
                 }
 
                 @Override
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/Authorizer.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/Authorizer.java
@@ -1,7 +1,7 @@
 package io.dropwizard.auth;
 
 import javax.annotation.Nullable;
-import javax.ws.rs.core.UriInfo;
+import javax.ws.rs.container.ContainerRequestContext;
 import java.security.Principal;
 
 /**
@@ -24,10 +24,10 @@ public interface Authorizer<P extends Principal> {
      *
      * @param principal a {@link Principal} object, representing a user
      * @param role a user role
-     * @param uriInfo a request URI information
+     * @param requestContext a request context.
      * @return {@code true}, if the access is granted, {@code false otherwise}
      */
-    default boolean authorize(P principal, String role, @Nullable UriInfo uriInfo) {
+    default boolean authorize(P principal, String role, @Nullable ContainerRequestContext requestContext) {
         return authorize(principal, role);
     }
 }
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/Authorizer.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/Authorizer.java
@@ -1,5 +1,6 @@
 package io.dropwizard.auth;
 
+import javax.ws.rs.core.UriInfo;
 import java.security.Principal;
 
 /**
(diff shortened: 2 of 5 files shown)


dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
```diff
@@ -152,7 +152,7 @@ public Principal getUserPrincipal() {
 
                 @Override
                 public boolean isUserInRole(String role) {
-                    return authorizer.authorize(prince, role);
+                    return authorizer.authorize(prince, role, requestContext.getUriInfo());
```
joschi: Why not pass the complete `ContainerRequestContext` instead of just `UriInfo`?
That way, other use cases, such as using a request header, could also be supported.

dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
```diff
@@ -45,7 +47,7 @@
     //
     // Field is package-private to be visible for unit tests
     @VisibleForTesting
-    final LoadingCache<ImmutablePair<P, String>, Boolean> cache;
+    final LoadingCache<ImmutableTriple<P, String, UriInfo>, Boolean> cache;
```
joschi: Could we create a dedicated data class for this triple, for example `AuthorizationContext`?

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
cyberdelia: The main concern I had was that `ContainerRequestContext` allow to change the request in itself, which I'm not sure should be allowed in the `Authorizer`. If that concerns is unfounded, I'm open to make the change.

dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
cyberdelia: I'll look into it.

cyberdelia: I introduced an `AuthorizationContext`.
Fair warning, my java is a bit rusty.

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
jplock: i think passing in the `ContainerRequestContext` is ok, what do you think @joschi ?

dropwizard-auth/src/main/java/io/dropwizard/auth/Authorizer.java
```diff
@@ -17,4 +18,16 @@
      * @return {@code true}, if the access is granted, {@code false otherwise}
      */
     boolean authorize(P principal, String role);
```
leoconco: This method is called by the Filter, why not delete or at least deprecate it?

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFilter.java
cyberdelia: I made the change in a separate commit.

dropwizard-auth/src/main/java/io/dropwizard/auth/Authorizer.java
cyberdelia: Done, at least in the code.
Should I mention this in the release notes and/or elsewhere?

joschi: @cyberdelia Good catch! Please add it to the release notes in `./docs/source/about/release-notes.rst`.

cyberdelia: Done!

joschi:
@cyberdelia Thanks for your contribution!

PR #2839, opened 2019-07-11
Add a JSONUnauthorizedHandler

cyberdelia:
###### Problem:

Dropwizard provides a `DefaultUnauthorizedHandler` that will returns an HTML error message. A lot of end user seems to need return JSON error message back, which they have to implement themselves, in a more or less consistent fashion with dropwizard error handling.

###### Solution:

This provides a `JSONUnauthorizedHandler` that match dropwizard JSON error message handling, and prevent every end user to reimplement one.

###### References:

- https://github.com/stardogventures/starwizard/blob/master/starwizard-core/src/main/java/io/stardog/starwizard/handlers/JsonUnauthorizedHandler.java
- https://github.com/dropwizard/dropwizard/pull/1863#issuecomment-269090433

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/JSONUnauthorizedHandler.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/JSONUnauthorizedHandler.java
@@ -1,0 +1,24 @@
+package io.dropwizard.auth;
+
+import io.dropwizard.jersey.errors.ErrorMessage;
+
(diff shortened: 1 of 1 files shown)


cyberdelia:
@jplock Should I make a PR against the 1.3x branch? Or are you planning to cherry-pick this directly?

jplock:
@cyberdelia please create another PR

PR #2844, opened 2019-07-16
Added eclipse exclusion to fix IDE errors

isaki:
###### Problem:
Eclipse displayed multiple errors due to unsupported Maven plugins.

###### Solution:
Eclipse auto-generated POM changes to fix the errors.

###### Result:
Eclipse IDE can be used for DropWizard development.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -612,6 +612,33 @@
                     <version>1.6.8</version>
                     <extensions>true</extensions>
                 </plugin>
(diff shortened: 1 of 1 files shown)


joschi:
@isaki Would it make sense to put this into the Eclipse-specific profile we already have in the parent POM?
https://github.com/dropwizard/dropwizard/blob/c98b009c11bcb7a954e590dcb208dfad1318d39e/pom.xml#L373-L392

isaki:
This was injected by Eclipse automatically. I actually can't take credit for authoring it. I'll add it to the profile and see if it works!

PR #2864, opened 2019-07-31
move call to customizeBuilder() to the bottom

yq314:
###### Problem:
There are bunches of setters called after the call to `customizeBuiler()` which will override whatever is customized in `customizeBuilder()`. For example, if I set a custom requestExecutor in `customizeBuilder()`

```java
  @Override
  protected org.apache.http.impl.client.HttpClientBuilder customizeBuilder(
    org.apache.http.impl.client.HttpClientBuilder builder
  ) {
    builder.setRequestExecutor(customRequestExecutor);
    return builder;
  }
```

It will then be overriden in the next line 
```java
customizeBuilder(builder)
    .setRequestExecutor(new InstrumentedHttpRequestExecutor(metricRegistry, metricNameStrategy, name))
```

###### Solution:
This PR moves the call to `customizeBuilder()` down to the bottom of `createClient()` so the customization happens the last that it won't be affected by operations inside `createClient()`

###### Result:
Allow full customization rather than partially.


yq314:
hmm, there's no easy way to test this actually. `org.apache.http.impl.client.HttpClientBuilder` doesn't expose any get method for me to verify. Any suggestions?

nickbabcock:
> hmm, there's no easy way to test this actually. org.apache.http.impl.client.HttpClientBuilder doesn't expose any get method for me to verify

Yes this is unfortunate, we've worked around this in the past by extracting non public fields. Not great, not terrible.

Checkout `HttpClientBuilderTest.java` to add a test in. For instance if your CustomBuilder hardcodes the user agent to be "MyUserAgent", you'd expect it to hold in spite of any configuration:

```java
@Override
protected org.apache.http.impl.client.HttpClientBuilder customizeBuilder(
    org.apache.http.impl.client.HttpClientBuilder builder
) {
    builder.setUserAgent("MyUserAgent");
    return builder;
}
```

```java
@Test
public void customBuilderHasPriority() throws Exception {
    CustomBuilder builder = new CustomBuilder(new MetricRegistry());
    builder.createClient(apacheBuilder, connectionManager, "test");
    assertThat(FieldUtils.getField(httpClientBuilderClass,"userAgent", true)
        .get(apacheBuilder))
        .isEqualTo("MyUserAgent");
}
```

yq314:
That's very cool, thanks for sharing :) just updated to test to verify that the requestExecutor can be customized.

dropwizard-client/src/test/java/io/dropwizard/client/HttpClientBuilderTest.java
```diff
@@ -50,10 +50,7 @@
 import org.apache.http.impl.conn.SystemDefaultRoutePlanner;
 import org.apache.http.message.BasicHeader;
 import org.apache.http.message.BasicListHeaderIterator;
-import org.apache.http.protocol.BasicHttpContext;
-import org.apache.http.protocol.HTTP;
-import org.apache.http.protocol.HttpContext;
-import org.apache.http.protocol.HttpProcessor;
+import org.apache.http.protocol.*;
```
nickbabcock: I know you probably didn't do this, but we don't use wildcard imports in this project :)

yq314: ah yes it's my intellij 😅 have updated.

joschi:
@yq314 Could you please fix the failing tests so that we can move on with PR?

yq314:
Hi @joschi thanks so much for fixing it for me. Sorry for the late reply, I was on 2 weeks holiday without laptop access 😅

PR #2869, opened 2019-08-06
Slight message change

maffe:
###### Problem:

Missing word

###### Solution:

Added word

###### Result:

Message contains previously missing word


joschi:
@maffe Thanks!

PR #2884, opened 2019-08-18
Add GitHub workflow to close stale issues and PRs

joschi:
Issues and PRs will be marked as stale after 90 days and removed if there wasn't any activity 14 days after that.


joschi:
Looks like we became victim of https://github.com/actions/stale/pull/8.

damccorm:
FYI, ^ that has been fixed so this PR should be good now

Issue #2880, opened 2019-08-16
Dropwizard BOM provides dependency management for modules outside the Dropwizard project


srdo:
The BOM is setting dependency management for modules that aren't part of Dropwizard. This is not ideal, as it can cause conflicts when other projects import the BOM.

A simple example would be if I had a project using the Dropwizard BOM, and also depending on something that depends on `tomcat-jdbc:9.0.8` and `tomcat-dbcp:9.0.8`. In this case I would get `tomcat-jdbc:9.0.22` from Dropwizard, and `tomcat-dbcp:9.0.8`, which might not work together. This will happen even if I don't use a Dropwizard module that requires `tomcat-jdbc`.

See a similar discussion here https://github.com/dropwizard/metrics/issues/1336

Ideally the BOM would only set the versions of io.dropwizard modules, and let the consuming project handle resolving any conflicts between transitive dependencies. This is also how several other modular projects do their BOMs (e.g. JUnit, Jackson, Camel)

joschi:
@dropwizard/committers Any strong opinions on this?

The advantage of having all dependencies in `dropwizard-bom` is that every project automatically can use the pre-configured (transitive) dependencies without having to declare their versions.

In that way it's similar to `org.springframework.boot:spring-boot-dependencies`:
https://github.com/spring-projects/spring-boot/blob/7712f9c0e5a84f04a9c93c195b90d05cfcf50a67/spring-boot-project/spring-boot-dependencies/pom.xml

The difference between JUnit, Jackson, Camel, and Dropwizard (or Spring Boot) is that the latter is a framework that basically rules how to build your application while the others are libraries which are usually added to existing applications.

jplock:
I like having everything self contained

srdo:
Just to make sure we have a common understanding of what removing the transitive dependencies from the BOM would do: Consuming projects get Dropwizard's transitive dependencies on their classpaths regardless of whether the BOM mentions them or not, the only difference is whether the BOM overrides Maven's normal transitive dependency resolution for these artifacts. 

Let me give some examples using `tomcat-jdbc` (these artifacts don't necessarily make sense together, but illustrate the principle):

Example consumer POM:
```
<dependency>
    <groupId>io.dropwizard</groupId>
    <artifactId>dropwizard-bom</artifactId>
    <type>pom</type>
    <scope>import</scope>
</dependency>
<dependency>
    <groupId>io.dropwizard</groupId>
[...]

PR #2897, opened 2019-08-24
Dependency reorganization

joschi:
The purpose of this PR are threefold:

* *Create properties for all dependencies*:
Allow overriding the version of every dependency of Dropwizard by setting a Maven property
* *Split BOM into Dropwizard modules and dependencies*:
The bill of materials for all Dropwizard dependendencies is now named `io.dropwizard:dropwizard-dependencies` and `io.dropwizard:dropwizard-bom` manages the version of all Dropwizard modules, but no other dependencies (refs #2880)
* *Introduce dropwizard-project for common build configuration*:
`io.dropwizard:dropwizard-parent` is still the parent POM for most Dropwizard modules, but automatically includes the dependencies from `io.dropwizard:dropwizard-bom` and `io.dropwizard:dropwizard-dependencies`. The `io.dropwizard:docs` module inherits from `io.dropwizard:dropwizard-project` but doesn't have any overlap with any of the Dropwizard modules carrying code.

TODO:

* [ ] Document changes with regards to `io.dropwizard:dropwizard-bom` in the Dropwizard 2.0.x migration guide


joschi:
@dropwizard/core Any thoughts on this PR?

This would be the last change I want to bring in before releasing Dropwizard 2.0.0.

jplock:
:shipit: 🎉

joschi:
> Shame that it'll be a breaking change for everyone

@nickbabcock I thought about naming the bill of materials differently, i. e. "dropwizard-bom-modules" and keeping "dropwizard-bom" the way it was before.

This would be an option which retains backward-compatibility, but we'll have to come up with a *good* name for the BOM referring to only the Dropwizard modules.

Any suggestions?

nickbabcock:
IMO it's fine in lieu of a better name, and 2.0 is as good as a time we'll get for this breaking change.

PR #2898, opened 2019-08-27
Replace jaxb-api with jakarta.xml.bind-api

maffe:
Change dependency to match the one in `org.glassfish.jaxb:jaxb-runtime` to avoid “overlapping classes” warning

### DIFF
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -117,9 +117,9 @@
                 <version>4.0.1</version>
             </dependency>
             <dependency>
-                <groupId>javax.xml.bind</groupId>
-                <artifactId>jaxb-api</artifactId>
-                <version>2.3.1</version>
+                <groupId>jakarta.xml.bind</groupId>
+                <artifactId>jakarta.xml.bind-api</artifactId>
+                <version>2.3.2</version>
                 <scope>runtime</scope>
             </dependency>
             <dependency>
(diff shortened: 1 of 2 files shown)


maffe:
¯\\\_(ツ)\_/¯

joschi:
There are quite a few dependencies, which depend on both, `javax.xml.bind:jaxb-api` and `org.glassfish.jaxb:jaxb-runtime`, for example Hibernate ORM 5.4.4:
https://github.com/hibernate/hibernate-orm/blob/5.4.4/gradle/libraries.gradle#L80-L81

I guess we'll have to live with the overlapping classes warning, until all dependencies have moved to the `jakarta.xml.bind` prefix.

pkwarren:
@joschi - This still seems like a good fix for dropwizard to move its own dependency to the new jakarta jaxb-api (even if third party dependencies aren't able to migrate yet). Clients will still need to add exclusions to avoid overlapping classes.

joschi:
@pkwarren Would you be able to provide a pull request for this change?

pkwarren:
I think this PR can be re-opened and it will fix the issue with JAXB conflicts. I can submit a PR for avoiding the javax.activation issue if that's of interest (the other conflict in #2908).

pkwarren:
@joschi - I opened a PR which migrates all dropwizard artifacts to use jakarta apis instead of a mix of jakarta and javax.* ones. PR here: #3069.

PR #2899, opened 2019-08-29
Add slf4j marker to dropwizard-json-logging.

motlin:
###### Problem:
json-log-format does not include slf4j marker information in the output.

###### Solution:
Added EventAttribute.MARKER and included it if present in EventJsonLayout.


motlin:
Updated the assertions to include the marker.

joschi:
@motlin LGTM, thanks!

PR #2913, opened 2019-09-09
Nested calls to @UnitOfWork methods cause inconsistent behaviour

matthiasblaesing:
###### Problem:
When a method annotated with @UnitOfWork calls into another method that
is annotated like that, the transaction is prematurely committed when
the inner method returns.

This leads to difficult to debug situations and should be prevented.
Either an exception has to be raised as soon as possible or nesting
needs to do the sane thing.

###### Solution:
The implementation checks if a session is already associated with this
thread. If that is not the case (non-nested), the original code paths
are invoked. If there exists already a session, its state is validated,
if the session is compatible with the requested state, the unit of work
is joined, if not an IllegalStateException is raised.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAspect.java
@@ -52,6 +52,11 @@ public class UnitOfWorkAspect {
     @Nullable
     private SessionFactory sessionFactory;
 
+    // was the session created by this aspect?
+    private boolean sessionCreated;
+    // do we manage the transaction or did we join an existing one?
+    private boolean transactionStarted;
+
     public void beforeStart(@Nullable UnitOfWork unitOfWork) {
         if (unitOfWork == null) {
             return;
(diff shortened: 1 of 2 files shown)


irineuruiz:
Hello,

In version 1.3.12 of Dropwizard we used a class that is created with UnitOfWorkAwareProxyFactory.
The return of one of our resources had a variable that is an instance of this class and that during the creation of the response in JSON saved some data in the database via hibernate.

We are migrating to Dropwizard 2.0.2 and we had a problem with that, UnitOfWork tries to use the same transaction as the resource due to the new code implemented 
and therefore throws the error due to the code 

        if (unitOfWork.transactional ()! = (txn! = null && txn.isActive ())) {
            throw new IllegalStateException (String.format (
                "Existing session transaction state (% s) does not match requested (% b)",
                txn == null? "NULL": Boolean.valueOf (txn.isActive ()),
                unitOfWork.transactional ()
            ));
        }

Because the resource transaction is no longer active and we have the response "Unable to create json". 
This worked perfectly in version 1.3.12

How should I act in such a case?
Is it a bug in the new version?
We are doing something wrong?

PR #2925, opened 2019-09-16
Move ResilientSocketOutputStream into io.dropwizard.logging

joschi:
Using a "reserved" package, such as `ch.qos.logback.core.recovery` from Logback, in an unrelated artifact will cause problems with the Java Platform Module System (JPMS).

In order to prevent a future conflict, the `ResilientSocketOutputStream` and the parent class `ResilientOutputStreamBase` from Logback have been moved into `io.dropwizard.logging`.

While copying code is usually a bad idea, I don't see how we could support Java 9+ with Dropwizard without this change or a change from the Logback project to change the modifier of the required methods in `ResilientOutputStreamBase` to `protected`.


jplock:
Can a comment be added documenting this class is originally from Logback and maybe reference a version or something so we know if we need to keep it in sync with later versions?

PR #2941, opened 2019-09-23
Jackson Afterburner don't use value classloader

### LINKED ISSUE #2909: Jackson Afterburner causes JVM warnings on Java 9 and beyond
When creating ObjectMapper instances with the `Jackson.newObjectMapper` method the Jackson’s Afterburner module is  registered causing the following JVM warning:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.fasterxml.jackson.module.afterburner.util.MyClassLoader (file:.m2/repository/com/fasterxml/jackson/module/jackson-module-afterburner/2.9.9/jackson-module-afterburner-2.9.9.jar) to method java.lang.ClassLoader.findLoadedClass(java.lang.String)
WARNING: Please consider reporting this to the maintainers of com.fasterxml.jackson.module.afterburner.util.MyClassLoader
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

Please see more detailed information here:
https://github.com/FasterXML/jackson-modules-base/issues/37

It would be good to have an option in Dropwizard to disable the Afterburner module, or configure it not `setUseValueClassLoader(false)`, or use alternative Jackson Blackbird module https://github.com/stevenschlansker/jackson-blackbird.

One possible solution I can think of is to add a JackonModuleConfigService service provider interface which would return the Jackson module instances being registered, so the application can disable or reconfigure each individual module.

Alexey1Gavrilov:
Make Afterburner generate bytecode only for public getters/setter
and fields to avoid Java 9+ complains of "Illegal reflective access".

Resolves #2909

### DIFF
--- a/docs/source/manual/upgrade-notes/upgrade-notes-2_0_x.rst
+++ b/docs/source/manual/upgrade-notes/upgrade-notes-2_0_x.rst
@@ -91,3 +91,13 @@ if inclusive is false:
 ::
 
    messageRate must be less than 1 MINUTES
+
+Changed the Jackson's object mapper configuration
+-------------------------------------------------
+
+Note that the Afterburner module is now configured with ``setUseValueClassLoader(false)``
+by default which causes bytecode generation only for public getters/setter and fields to avoid
+Java 9+ complains of ``Illegal reflective access``.
+This might slightly affect performance of your application. See
+` this Jackson issue <https://github.com/FasterXML/jackson-modules-base/issues/37>`_ for more
+information.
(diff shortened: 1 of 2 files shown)


Alexey1Gavrilov:
Oh well, changing the class loading scheme caused test failures and runtime warnings from the Afterburner module. I'm afraid this change will cause quite a lot of breakage of the existing applications. I'm closing it for now.

PR #2943, opened 2019-09-24
Support for requestAttributes in Json access log

gisripa:
Summary
=======
Support for requestAttributes similar to LogbackAccessLayout.
When specified, all non-null values will be included in logs

###### Problem:
We were using servletFilters to populate certain application specific contextual information and pass them along as request attributes, which are in turn used for central logging of access logs. Lack of way to specify request Attributes in json log layout is hindering us to move to json logging instead of file logging. 

###### Solution:
* Added configuration support to specify requestAttribute names to be logged. 
* Retrieving the attribute from `IAccessEvent#getAttribute()` api to add them to json log

### DIFF
--- a/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/AccessJsonLayout.java
+++ b/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/AccessJsonLayout.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
@@ -23,6 +24,7 @@ public class AccessJsonLayout extends AbstractJsonLayout<IAccessEvent> {
 
     private SortedSet<String> requestHeaders = Collections.emptySortedSet();
     private SortedSet<String> responseHeaders = Collections.emptySortedSet();
+    private SortedSet<String> requestAttributes = Collections.emptySortedSet();
 
     @Nullable
     private String jsonProtocolVersion;
@@ -67,11 +69,19 @@ public class AccessJsonLayout extends AbstractJsonLayout<IAccessEvent> {
             .addNumber("status", isIncluded(AccessAttribute.STATUS_CODE), event::getStatusCode)
             .add("userAgent", isIncluded(AccessAttribute.USER_AGENT), () -> event.getRequestHeader(USER_AGENT))
             .add("version", jsonProtocolVersion != null, jsonProtocolVersion)
+            .addMap("requestAttributes", !requestAttributes.isEmpty(),
+                () -> filterRequestAttributes(requestAttributes, event))
             .build();
     }
 
     private boolean isIncluded(AccessAttribute attribute) {
         return includes.contains(attribute);
+    }
+
+    private Map<String, String> filterRequestAttributes(Set<String> requestAttributeNames, IAccessEvent event) {
+        return requestAttributeNames.stream()
+            .filter(name -> event.getAttribute(name) != null)
+            .collect(Collectors.toMap(Function.identity(), event::getAttribute));
     }
 
     private Map<String, String> filterHeaders(Map<String, String> headers, Set<String> filteredHeaderNames) {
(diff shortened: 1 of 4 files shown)


dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/AccessJsonLayout.java
```diff
@@ -11,8 +11,11 @@
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import static ch.qos.logback.access.spi.IAccessEvent.*;
```
jplock: Please don’t use wildcard imports

dropwizard-json-logging/src/test/java/io/dropwizard/logging/json/layout/AccessJsonLayoutTest.java
```diff
@@ -17,8 +17,10 @@
 import java.util.Map;
 import java.util.Set;
 
+import static ch.qos.logback.access.spi.IAccessEvent.*;
```
jplock: Please don’t use wildcard imports

dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/AccessJsonLayout.java
```diff
@@ -67,13 +71,20 @@ public AccessJsonLayout(JsonFormatter jsonFormatter, TimestampFormatter timestam
...
             .build();
     }
 
     private boolean isIncluded(AccessAttribute attribute) {
         return includes.contains(attribute);
     }
 
+    private Map<String, String> includeRequestAttributes(Set<String> requestAttributeNames, IAccessEvent event) {
+        return requestAttributeNames.stream()
+            .collect(Collectors.toMap(Function.identity(), name -> event.getAttribute(name) != null ? event.getAttribute(name) : NA));
```
mattnelson: Is the NA conversion primarily for unstructured logs to provide a parsing location? With json format a key lookup can be done to assert presence. This also introduces a corner case in which a request attribute value of `-` can not be distinguished between unset and set to `-`

The other filterable json attributes are only included if present.
https://github.com/dropwizard/dropwizard/blob/6f8b39458a5758da38dc27ff08cf1e63f7b2f5ff/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/EventJsonLayout.java#L93-L101
https://github.com/dropwizard/dropwizard/blob/6f8b39458a5758da38dc27ff08cf1e63f7b2f5ff/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/AccessJsonLayout.java#L77-L84

dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/AccessJsonLayout.java
```diff
@@ -67,13 +71,20 @@ public AccessJsonLayout(JsonFormatter jsonFormatter, TimestampFormatter timestam
...
+            .addMap("requestAttributes", !requestAttributes.isEmpty(),
+                () -> includeRequestAttributes(requestAttributes, event))
             .build();
     }
 
     private boolean isIncluded(AccessAttribute attribute) {
         return includes.contains(attribute);
     }
 
+    private Map<String, String> includeRequestAttributes(Set<String> requestAttributeNames, IAccessEvent event) {
```
mattnelson: For consistency should be renamed to `filterRequestAttributes` to align with `filterMDC` and `filterHeaders`

dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/AccessJsonLayout.java
gisripa: @mattnelson Didn't have a strong preference but that makes sense, it might have a different meaning in file logging vs key not set in json logging. How does this look, before i add commit ?

```
private Map<String, String> filterRequestAttributes(Set<String> requestAttributeNames, IAccessEvent event) {
        return requestAttributeNames.stream()
            .filter(name -> event.getAttribute(name) != null)
            .collect(Collectors.toMap(Function.identity(), event::getAttribute));
    }
```

dropwizard-json-logging/src/test/java/io/dropwizard/logging/json/layout/AccessJsonLayoutTest.java
gisripa: @jplock IDE auto corrected. Fixed now

jplock:
Looks good to me. Any more feedback @mattnelson ?

PR #2992, opened 2019-10-21
Support comments in DiscoverableSubtypeResolver

jmoney8080:
Add the ability to have comments in the DiscoverableSubtypeResolver class resources file

###### Problem:
Currently if you have any other lines in the discoverable resource files you get a bunch of log lines detailing "cannot load this class".  Example:
```
# this is the foo class
org.github.Foo
```

###### Solution:
Add the ability to allow for comments in the resource file.  Comments are lines that start with `#`.

###### Result:
No more log lines for lines that begin with `#` detailing that it's not a class when parsing the class resource files.

### DIFF
--- a/dropwizard-jackson/src/main/java/io/dropwizard/jackson/DiscoverableSubtypeResolver.java
+++ b/dropwizard-jackson/src/main/java/io/dropwizard/jackson/DiscoverableSubtypeResolver.java
@@ -67,8 +67,6 @@ public class DiscoverableSubtypeResolver extends StdSubtypeResolver {
                             if (loadedClass != null) {
                                 serviceClasses.add(loadedClass);
                             }
-                        } else {
-                            LOGGER.debug("Line began with a comment. Ignoring");
                         }
                     }
                 }
--- a/dropwizard-jackson/src/main/java/io/dropwizard/jackson/DiscoverableSubtypeResolver.java
+++ b/dropwizard-jackson/src/main/java/io/dropwizard/jackson/DiscoverableSubtypeResolver.java
@@ -62,9 +62,13 @@ public class DiscoverableSubtypeResolver extends StdSubtypeResolver {
                      BufferedReader reader = new BufferedReader(streamReader)) {
                     String line;
                     while ((line = reader.readLine()) != null) {
-                        final Class<?> loadedClass = loadClass(line);
-                        if (loadedClass != null) {
-                            serviceClasses.add(loadedClass);
+                        if (!line.startsWith("#")) {
+                            final Class<?> loadedClass = loadClass(line);
+                            if (loadedClass != null) {
+                                serviceClasses.add(loadedClass);
+                            }
+                        } else {
+                            LOGGER.debug("Line began with a comment. Ignoring");
                         }
                     }
                 }


dropwizard-jackson/src/main/java/io/dropwizard/jackson/DiscoverableSubtypeResolver.java
```diff
@@ -62,9 +62,13 @@ protected ClassLoader getClassLoader() {
...
-                        final Class<?> loadedClass = loadClass(line);
-                        if (loadedClass != null) {
-                            serviceClasses.add(loadedClass);
+                        if (!line.startsWith("#")) {
+                            final Class<?> loadedClass = loadClass(line);
+                            if (loadedClass != null) {
+                                serviceClasses.add(loadedClass);
+                            }
+                        } else {
+                            LOGGER.debug("Line began with a comment. Ignoring");
```
joschi: Could we remove that debug log message?

jmoney8080:
Do you have a suggestion on how one would write such a test?  It's mainly ignoring lines that begin with `#`.  Not sure to test for "absence" of the error message being logged at INFO than hijack the logger somehow and count that a message didn't appear?

dropwizard-jackson/src/main/java/io/dropwizard/jackson/DiscoverableSubtypeResolver.java
jmoney8080: done

joschi:
> Do you have a suggestion on how one would write such a test? It's mainly ignoring lines that begin with `#`.

Maybe you could add a file such as `META-INF/services/io.dropwizard.jackson.Discoverable` which contains a comment to the test class path and try loading/discovering it.

mattnelson:
Can this be deferred to the JDK?

https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html#load-java.lang.Class-java.lang.ClassLoader-
https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html#iterator--

jmoney8080:
> > Do you have a suggestion on how one would write such a test? It's mainly ignoring lines that begin with `#`.
> 
> Maybe you could add a file such as `META-INF/services/io.dropwizard.jackson.Discoverable` which contains a comment to the test class path and try loading/discovering it.

I'll get a test added that adds some comments and makes sure it can still load it.

jmoney8080:
> Can this be deferred to the JDK? Which also has support ignore comments.
> 
> > A service provider is identified by placing a provider-configuration file in the resource directory META-INF/services. The file's name is the fully-qualified binary name of the service's type. The file contains a list of fully-qualified binary names of concrete provider classes, one per line. Space and tab characters surrounding each name, as well as blank lines, are ignored. The comment character is '#' ('\u0023', NUMBER SIGN); on each line all characters following the first comment character are ignored. The file must be encoded in UTF-8.
> 
> https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html#load-java.lang.Class-java.lang.ClassLoader-
> https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html#iterator--

this feels like a larger "architectural" change.  I know ServiceLoader can ignore it but this file seems to not be using ServiceLoader at all.  it's a very small `if` statement compared to re-writing the class with ServiceLoader.  So i'd propose we merge this in as is, after a test, and maybe open a github issue for migrating this class to ServiceLoader

joschi:
> Can this be deferred to the JDK? Which also has support ignore comments.

@mattnelson Would you mind creating a separate issue for this which we can schedule for Dropwizard 2.1.x?

mattnelson:
> Would you mind creating a separate issue for this which we can schedule for Dropwizard 2.1.x?

Finally getting around to this and after another round of inspection realized that the iterator on ServiceLoader returns instantiated instances of the class instead of the `Class` object. This PR is IMHO the cleanest way to filter out comments.

joschi:
@jmoney8080 Thanks for your contribution!

PR #3000, opened 2019-10-26
Register HK2 AbstractBinder with Jersey

### LINKED ISSUE #2999: 2.0.0:  Registering org.glassfish.hk2.utilities.binding.AbstractBinder fails with java.lang.IllegalArgumentException: Provider registered to Hk2InjectionManager cannot be process because of incompatible type: class java.lang.Class
Starting with  jersey 2.29 `org.glassfish.hk2.utilities.binding.AbstractBinder` can again be registered with jersey's `ResourceConfig` , see eclipse-ee4j/jersey#4092. However `DropwizardResourceConfig.register`  attempts to create a dynamic proxy for instances of hk2  `AbstractBinder` which result in a failure. 

The `DropwizardResourceConfig.register` should simpy delecate to jersey `ResourceConfig.register` for instances of `org.glassfish.hk2.utilities.binding.AbstractBinder`

aharin:
###### Problem:
Before 2.0.0 version, dropwizard and jeresey allowed to register org.glassfish.hk2.utilities.binding.AbstractBinder. This got broken by jersey and then fixed in 2.29. However dropwizard does not pass hk2 Abstract binder directly to jersey, resulting in a failure.

###### Solution:
This pull request modifies DropwizardResourceConfig, so that instances of org.glassfish.hk2.utilities.binding.AbstractBinder are passed directly to jersey.

###### Result:
As a result of this pull request, 2.0.0 will behave exactly the same as 1.x.x for configuring hk2 dependency injection.

Fixes #2999

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -142,7 +142,7 @@ public class DropwizardResourceConfig extends ResourceConfig {
         // If a class gets passed through as an object, cast to Class and register directly
         if (component instanceof Class<?>) {
             return super.register((Class<?>) component);
-        } else if (Providers.isProvider(clazz) || Binder.class.isAssignableFrom(clazz)) {
+        } else if (Providers.isProvider(clazz) || org.glassfish.hk2.utilities.Binder.class.isAssignableFrom(clazz)) {
             // If Jersey supports this component's class (including Binders), register directly
             return super.register(object);
         } else {
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -14,7 +14,6 @@ import javassist.ClassPool;
 import javassist.CtClass;
 import javassist.LoaderClassPath;
 import org.glassfish.jersey.internal.inject.AbstractBinder;
-import org.glassfish.jersey.internal.inject.Binder;
 import org.glassfish.jersey.internal.inject.Providers;
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.server.ServerProperties;
--- a/dropwizard-jersey/src/test/java/io/dropwizard/jersey/DropwizardResourceConfigTest.java
+++ b/dropwizard-jersey/src/test/java/io/dropwizard/jersey/DropwizardResourceConfigTest.java
@@ -6,6 +6,7 @@ import org.glassfish.jersey.server.model.Resource;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
+import javax.inject.Inject;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
@@ -216,6 +217,23 @@ public class DropwizardResourceConfigTest {
     }
 
     @Test
+    void logResourceWithHK2Binder() {
+        rc.register(ResourceWithInjectedDependency.class);
+        rc.register(new org.glassfish.hk2.utilities.binding.AbstractBinder() {
+            @Override
+            protected void configure() {
+                bindAsContract(Dependency.class);
+            }
+        });
+
+        runJersey();
+        final String expectedLog = String.format("The following paths were found for the configured resources:%n" + "%n"
+            + "    GET     /another (io.dropwizard.jersey.DropwizardResourceConfigTest.ResourceWithInjectedDependency)");
+
+        assertThat(rc.getEndpointsInfo()).contains(expectedLog);
+    }
+
+    @Test
     void logsEndpointsContextPathUrlPattern() {
         rc.setContextPath("/context");
         rc.setUrlPattern("/pattern");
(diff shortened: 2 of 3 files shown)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
```diff
@@ -142,7 +142,7 @@ public ResourceConfig register(final Object component) {
         // If a class gets passed through as an object, cast to Class and register directly
         if (component instanceof Class<?>) {
             return super.register((Class<?>) component);
-        } else if (Providers.isProvider(clazz) || Binder.class.isAssignableFrom(clazz)) {
+        } else if (Providers.isProvider(clazz) || org.glassfish.hk2.utilities.Binder.class.isAssignableFrom(clazz)) {
```
josephlbarnett: probably want to check both `org.glassfish.jersey.internal.inject.Binder` and `org.glassfish.hk2.utilities.Binder` if we want to add pass through for org.glassfish.hk2.utilities.Binder here?

aharin: Can do, but looking at the first condition `Providers.isProvider(clazz)`, it is already checking for `org.glassfish.jersey.internal.inject.Binder`, see `org.glassfish.jersey.internal.inject.Providers.getExternalProviderInterfaces`. `DropwizardResourceConfigTest` also has a test that verifies that `org.glassfish.jersey.internal.inject.Binder` is registered correctly. Do you still want me to add an explicit condition for it?

mattnelson:
Should this effectively be a revert of #2463 to restore 1.x functionality? Which was put in based on the broken(2.27) and now fixed (2.29) bindings.

pstackle:
@mattnelson I don't think so because the final else clause in the overridden register method added in #2463 adds support for registering already instantiated instances of Resource classes without Jersey logging a warning message. As far as I can tell from looking at the issues/PRs related to that topic, it doesn't look like they've added better support for that yet.

aharin:
Indeed, the overridden `DropwizardResourceConfig.register` is still required to stop Jersey complaining when registering resource instances. I suspect that eclipse-ee4j/jersey#3820 addresses that, unfortunately that pull request has been abandoned by its owner, even though Jersey maintainers were happy to merge it. I wonder if we should try and re-submit it.

dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
josephlbarnett: Looking through the jersey code, I agree with you that it `Providers.isProvider` should work for `org.glassfish.jersey.internal.inject.Binder` instances.  Looks like it should also work for `org.glassfish.hk2.utilities.Binder`s since that interface has the `@Contract` annotation?  As long as we can prove that it returns true for implementors of both interfaces (maybe add a test?)...

aharin: Unfortunatly `Providers.isProvider` does not pick up `org.glassfish.hk2.utilities.Binder` .  `Providers.isProvider` is looking for `org.glassfish.jersey.spi.Contract`. However `org.glassfish.hk2.utilities.Binder`  is annotated with `org.jvnet.hk2.annotations.Contract`. 
I am adding a test `ProvidersBinderTest` that demonstrates that.

In my original commit I included  a test `DropwizardResourceConfigTest.logResourceWithHK2Binder`  that fails if I remove second condition. Also there is an existing test, `DropwizardResourceConfigTest.logEndpointBinder` that verifies that `org.glassfish.jersey.internal.inject.Binder` can be registered.

joschi:
I'm a bit hesitant merging this PR because it would tie Dropwizard to HK2 which, at least in theory, is an optional dependency for Jersey.

I'm not aware of a working replacement for `org.glassfish.jersey.inject:jersey-hk2`, though.

Any opinions, @dropwizard/committers?

nickbabcock:
> I'm a bit hesitant merging this PR because it would tie Dropwizard to HK2 

As long as this doesn't preclude other DI routes that others have had success with then I don't mind, as it has always seemed that HK2, Jersey, and Dropwizard have been closely associated.

jhaber:
> I'm a bit hesitant merging this PR because it would tie Dropwizard to HK2 which, at least in theory, is an optional dependency for Jersey.
> 
> I'm not aware of a working replacement for `org.glassfish.jersey.inject:jersey-hk2`, though.

Just wanted to echo this concern. Jersey does have a native CDI-based implementation, and I'm currently working on a Guice-based implementation. However, I can send a PR to make this more pluggable (potentially making the `InjectionManagerFactory` directly configurable)

jhaber:
> > I'm a bit hesitant merging this PR because it would tie Dropwizard to HK2
> 
> As long as this doesn't preclude other DI routes that others have had success with then I don't mind, as it has always seemed that HK2, Jersey, and Dropwizard have been closely associated.

This would prevent users from excluding jersey-hk2 as a dependency because they'll get `NoClassDefFoundErrors`. Also other implementations of `InjectionManager` will presumably have different criteria for what classes can be registered, which this precludes. This is basically hard-coding the HK2 implementation:
https://github.com/eclipse-ee4j/jersey/blob/f2ccfd54bd6a7717fd23d2ea572cf9cf1bfe0ba8/inject/hk2/src/main/java/org/glassfish/jersey/inject/hk2/AbstractHk2InjectionManager.java#L126-L129

However, in the future Dropwizard could be come more aware of the `InjectionManager` in use and just delegate to its `isRegistrable` method

PR #3024, opened 2019-11-12
Fix links to Jersey Client API

maffe:
There are still old URLs in [other files](https://github.com/dropwizard/dropwizard/search?q=%22jersey.github.io%22). I don’t know whether they should point to the latest Jersey version or the one Dropwizard actually uses (which would take some effort to keep in sync).


joschi:
@maffe Thanks for your contribution!

PR #3036, opened 2019-11-23
Remove Stack Trace Printing In Task Servlet To Remove Potential Security Vulnerability

cwilbur12:
###### Problem:
Information from a stack trace propagates to an external user. Stack traces can unintentionally reveal implementation details that are useful to an attacker for developing a subsequent exploit.
###### Solution:
Remove printing of stack trace in `TaskServlet`

### DIFF
--- a/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
+++ b/dropwizard-servlets/src/main/java/io/dropwizard/servlets/tasks/TaskServlet.java
@@ -129,7 +129,6 @@ public class TaskServlet extends HttpServlet {
                 resp.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
                 output.println();
                 output.println(e.getMessage());
-                e.printStackTrace(output);
             } finally {
                 output.close();
             }


joschi:
@cwilbur12 Thanks for your contribution!

The admin interface is not supposed to be exposed to the world. It should be protected so that only you can access it.

I agree that the information in the task servlet can be used for exploration of the system, but if an attacker has access to the admin interface of your Dropwizard application, I think you're already in quite some trouble.

Any further thoughts on this?

j4n1:
> @cwilbur12 Thanks for your contribution!
> 
> The admin interface is not supposed to be exposed to the world. It should be protected so that only you can access it.
> 

Not supposed to be, but reality is that it will for various reasons.

> I agree that the information in the task servlet can be used for exploration of the system, but if an attacker has access to the admin interface of your Dropwizard application, I think you're already in quite some trouble.
> 
> Any further thoughts on this?

Defense in depth and secure by default are good reasons, not exposing information that is being logged just a few lines above is IMHO a good practice. 
I'd go even further and remove  `output.println(e.getMessage());` as well, and print generic message that something went wrong. My 2¢

joschi:
We've made this a configuration setting instead of disabling printing the stack trace completely.

Your input/review on #3037 would be appreciated. 😃

cwilbur12:
> > @cwilbur12 Thanks for your contribution!
> > The admin interface is not supposed to be exposed to the world. It should be protected so that only you can access it.
> 
> Not supposed to be, but reality is that it will for various reasons.
> 
> > I agree that the information in the task servlet can be used for exploration of the system, but if an attacker has access to the admin interface of your Dropwizard application, I think you're already in quite some trouble.
> > Any further thoughts on this?
>

I agree completely here you are probably in real tough shape if they have access to the admin interface. At the same time I just thought it was better practice to just remove it and obfuscate with a more generic logger based message that is already being logged above if there wasn't a very strong reason to keep it. 

> Defense in depth and secure by default are good reasons, not exposing information that is being logged just a few lines above is IMHO a good practice.
> I'd go even further and remove `output.println(e.getMessage());` as well, and print generic message that something went wrong. My 2¢


My original thought was also to remove `output.println(e.getMessage());` as well but that broke test `SslReloadAppTest.badReloadDoesNotChangeTheServerCertificate()` which was beyond my desired scope of digging into the issue so I just added it back in and removed what I thought was a larger vulnerability which was the stack trace itself.


I am completely fine closing if there are better reasons to making it configurable I was just removing from a "good practice" perspective. All valuable points raised for keeping it though. Thanks for the feedback guys! Keep up the amazing work.

PR #3043, opened 2019-12-01
Make tests more stable by using JSONAssert equals

contextshuffling:
###### Problem:

Tests in `io/dropwizard/views/freemarker/MultipleContentTypeTest.java` uses `jackson` to serialize objects into json and assert the serialized results with several hard-coded strings. However, the order of serialized json is not guaranteed so tests may fail if the order is different. So tests may fail or pass without any changes made to the source code and cannot serve as good regression tests.

###### Solution:

This PR proposes to use `JSONAssert` and modify the corresponding json test assertions
###### Result:

Tests become no longer flaky.


joschi:

I've decided to tackle this without the additional dependency in #3056.

Issue #3053, opened 2019-12-07
Length validation in example doesn't work


acsbendi:
I created the simple example app based on the [Getting Started guide](https://www.dropwizard.io/en/stable/getting-started.html), and I run into a problem. The `Saying` class has the `@Length(max = 3)` annotation on its `content` property (see [here](https://github.com/dropwizard/dropwizard/blob/ecb4331543d17ee5bbea6e940ee947a33feb3622/dropwizard-example/src/main/java/com/example/helloworld/api/Saying.java#L9)), The guide states 
> Lastly, the bean leverages validation to ensure the content size is no greater than 3.

This, to my understanding, should prevent any attempt to assign a string longer than 3 characters to the `content` property.

However, this does not seem to work. For example, when running `curl localhost:8080/hello-world`, the `content` will be `"Hello, Stranger!"` which is clearly longer than 3 characters. 

Please let me know if I'm misunderstanding something.

skradel:
I'm inclined to agree with you that the example given in the Getting Started guide is somewhat nonsensical; type "Saying" is only used as a return value type from a resource method, whereas most validation situations are more concerned with validating user input, e.g. validating a POSTed bean or validating URL query parameters.

That said, adding the annotation to the resource method as "public @Valid Saying sayHello" could be used to show a more unusual usage to generate HTTP 5xx errors, where the server found the user's request acceptable but was unable to generate a valid response.

acsbendi:
@skradel Yes, it could be simply removed. I just had the assumption that it's there to show some functionality, disregarding whether it actually makes sense or not. But then again, it didn't seem to have any.

Even that version with the `@Valid` annotation would be better I think, because at least the `@Length(max = 3)` would have some functionality.

joschi:
@acsbendi Thanks for reporting this!

Would you or @skradel be able to provide a PR to improve our example project?

acsbendi:
@joschi You're welcome!

Do you think it's enough to just remove this annotation and the related text? Or do you have an idea for a more useful way of having this annotation?

acsbendi:
Fixed in #3082.

PR #3082, opened 2019-12-25
Example validation annotations

HarryEH:
https://github.com/dropwizard/dropwizard/issues/3053

###### Problem:
As described in the Issue #3053, the Saying validation annotation doesn't entirely make sense. 

###### Solution:
I have removed the ```@Max``` annotation from the ```Saying``` class and removed the ```@Valid``` annotation from the resource. To still show the annotation in use I have modified the ```Person``` class in core and the people resource to reflect that. I have also added a test to show the validation in action on the resource.

###### Result:
I know adding age to a person isn't a particularly sensible thing  to do normally, but it is a good way to show basic use of these annotations in action. Feedback on the test would also be appreciated, without an exception mapper I'm not sure what to test for - other than what I have implemented.

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -349,7 +349,6 @@ To model this representation, we'll create a representation class:
     public class Saying {
         private long id;
 
-        @Length(max = 3)
         private String content;
 
         public Saying() {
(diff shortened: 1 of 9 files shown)


HarryEH:
> Seems good overall, some minor things:
> 
> * if you'd like to make the example a bit more sensible, perhaps you could use a year of birth field instead of age. `@Max` and `@Min` would work with that too.
> * the documentation also needs to be updated, the line "Lastly, the bean leverages validation to ensure the content size is no greater than 3." should be removed from [here](https://github.com/dropwizard/dropwizard/blob/master/docs/source/getting-started.rst).

Thanks!! I have made the changes suggested, and updated the documentation 😄. Also added a test for the ```@Max(value = 9999)```.

PR #3062, opened 2019-12-13
Add junit5 documentation

orevial:
###### Problem:
Dropwizard documentation is not up-to-date: it uses JUnit4 examples whereas JUnit4 rules and such are now deprecated in `dropwizard-testing`.

###### Solution:
Replace JUnit4 examples with newest default JUnit5 examples. This is mostly replacing `@Rule` annotation with `@DropwizardAppExtension`. 

*Note*: I also did a small refactoring on JUnit5 to better match JUnit5 conventions (e.g. no need for `public` modifier anymore), let me know if that's okay and/or if you want me to make another PR for this change.

Relates to #3051

PR #3087, opened 2019-12-28
Replacing ResourceTestRule with ResourceExtension in documentation

WilliamHerbert:
###### Problem:
Dropwizard 2.0 deprecated ResourceTestRule in favor of ResourceExtension. The documentation providing examples of testing resources was not updated to reflect this.

###### Solution:
ResourceTestRule has been removed from the documentation in favor of ResourceExtension.

This work was completed in `auth.rst` and `forms.rst` but not `testing.rst` as that will be handled in #3062 

###### Result:
Documentation is now more up-to-date with reality. :)

### DIFF
--- a/docs/source/manual/auth.rst
+++ b/docs/source/manual/auth.rst
@@ -280,12 +280,12 @@ Add this dependency into your ``pom.xml`` file:
 OAuth Example
 -------------
 
-When you build your ``ResourceTestRule``, add the ``GrizzlyWebTestContainerFactory`` line.
+When you build your ``ResourceExtension``, add the ``GrizzlyWebTestContainerFactory`` line.
 
 .. code-block:: java
 
     @Rule
-    public ResourceTestRule rule = ResourceTestRule
+    public ResourceExtension rule = ResourceExtension
             .builder()
             .setTestContainerFactory(new GrizzlyWebTestContainerFactory())
             .addProvider(new AuthDynamicFeature(new OAuthCredentialAuthFilter.Builder<User>()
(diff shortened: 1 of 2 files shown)


docs/source/manual/auth.rst
```diff
@@ -280,12 +280,12 @@ Add this dependency into your ``pom.xml`` file:
...
 -------------
 
-When you build your ``ResourceTestRule``, add the ``GrizzlyWebTestContainerFactory`` line.
+When you build your ``ResourceExtension``, add the ``GrizzlyWebTestContainerFactory`` line.
 
 .. code-block:: java
 
     @Rule
-    public ResourceTestRule rule = ResourceTestRule
+    public ResourceExtension rule = ResourceExtension
```
joschi: Could we rename this field to `extension` or `resourceExtension`?

WilliamHerbert:
I standardized to `resourceExtension`, per suggestion.

docs/source/manual/auth.rst
```diff
@@ -280,12 +280,12 @@ Add this dependency into your ``pom.xml`` file:
 OAuth Example
 -------------
 
-When you build your ``ResourceTestRule``, add the ``GrizzlyWebTestContainerFactory`` line.
+When you build your ``ResourceExtension``, add the ``GrizzlyWebTestContainerFactory`` line.
 
 .. code-block:: java
 
     @Rule
```
jplock: Do these @Rule and @ClassRule annotations need to be replaced?

WilliamHerbert: Good catch.  I cleaned up the old annotations and added the ExtendsWith annotation to the test class.

cobbce:
Is this documentation update available in the current documentation on dropwizard.io? I'm struggling to keep track of what version the documentation is given the Getting Started guide says "First, add a dropwizard.version property to your POM with the current version of Dropwizard (which is 2.0.1):" but it appears that the documentation itself is not yet updated?

joschi:
@cobbce The changes of this PR are published on https://www.dropwizard.io/.

You can find the specific pages here:

https://www.dropwizard.io/en/latest/manual/auth.html
https://www.dropwizard.io/en/latest/manual/forms.html

PR #3120, opened 2020-01-29
Fix thread safety issue with eTag calculation.

### LINKED ISSUE #3119: Unsynchronized access to shared CRC32 in AssetServlet
As part of migrating from Guava in #2400, the AssetServlet was updated to remove Guava dependencies and migrated from murmur3_128 hash to use CRC32. However, it appears this new logic contains a race condition where invalid hashes can be calculated:

https://github.com/dropwizard/dropwizard/blob/aa1e301bfedbc8b5cbe27ef8cfa42b68f8a47038/dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/AssetServlet.java#L50-L51

In this location, it is using a static CRC32 instance without synchronization, so if multiple assets are cached at the same time this could lead to invalid eTags.

pkwarren:
Previously, a static CRC32 instance was created and shared across
requests. When assets are requested across multiple threads, the
underlying CRC32 could be calculated inconsistently.

Updated the code to create a new CRC32 instance for each request.
Alternatively, a ThreadLocal could be used if this is determined to be a
major performance issue.

Fixes #3119.

### DIFF
--- a/dropwizard-servlets/src/test/java/io/dropwizard/servlets/assets/AssetServletTest.java
+++ b/dropwizard-servlets/src/test/java/io/dropwizard/servlets/assets/AssetServletTest.java
@@ -11,7 +11,16 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import javax.annotation.Nullable;
+
+import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.Future;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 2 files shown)


dropwizard-servlets/src/test/java/io/dropwizard/servlets/assets/AssetServletTest.java
```diff
@@ -145,12 +154,28 @@ public void consistentlyAssignsETags() throws Exception {
...
+            }));
+        }
+        latch.countDown();
+        final Set<String> eTags = new HashSet<>();
+        for (Future<String> future : futures) {
+            eTags.add(future.get());
+        }
+        assertThat(eTags)
+            .describedAs("eTag generation should be consistent with concurrent requests")
+            .hasSize(1);
```
pkwarren: Without this fix, this test failed with:

```
[ERROR] Failures: 
[ERROR]   AssetServletTest.consistentlyAssignsETags:175 [eTag generation should be consistent across threads] 
Expected size:<1> but was:<3> in:
<[""0"", ""1a94de8e"", ""e7bd7e8e""]>
```

joschi:
@pkwarren Thank you very much for your contribution!

PR #3133, opened 2020-02-09
Enable error-prone on Java 9+ builds.

pkwarren:
Newer versions of Java need to run error-prone as a compiler plugin.
Update the profile for Java 9+ builds to run with error-prone plug-in
and same arguments that are used for Java 1.8 build.

### DIFF
--- a/dropwizard-parent/pom.xml
+++ b/dropwizard-parent/pom.xml
@@ -36,11 +36,43 @@
                 <jdk>[9,)</jdk>
             </activation>
             <properties>
-                <maven.compiler.release>8</maven.compiler.release>
-                <maven.compiler.source>8</maven.compiler.source>
-                <maven.compiler.target>8</maven.compiler.target>
                 <argLine>${surefire.argLine} --illegal-access=deny --add-opens java.base/java.net=ALL-UNNAMED</argLine>
             </properties>
+            <build>
+                <plugins>
+                    <plugin>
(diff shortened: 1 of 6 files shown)


dropwizard-parent/pom.xml
```diff
@@ -36,11 +36,43 @@
                 <jdk>[9,)</jdk>
             </activation>
             <properties>
-                <maven.compiler.release>8</maven.compiler.release>
-                <maven.compiler.source>8</maven.compiler.source>
-                <maven.compiler.target>8</maven.compiler.target>
```
pkwarren: Source and target properties already set above. Moved release setting into the configuration for maven-compiler-plugin.

dropwizard-parent/pom.xml
```diff
@@ -36,11 +36,43 @@
...
+                                <path>
+                                    <groupId>com.uber.nullaway</groupId>
+                                    <artifactId>nullaway</artifactId>
+                                    <version>${nullaway.version}</version>
+                                </path>
+                            </annotationProcessorPaths>
+                            <compilerArgs>
+                                <arg>-XDcompilePolicy=simple</arg>
+                                <!-- When run as plugin all args must be passed together -->
+                                <arg>-Xplugin:ErrorProne
```
pkwarren: Unfortunately you can't pass each of these arguments separately - they all need to be included together in the same argument.

PR #3134, opened 2020-02-09
Fix regression in OptionalInt/Long/Double handling.

### LINKED ISSUE #3123: default value not applied to optional parameter type
I'm updating my application from 1.3.16 to dropwizard 2 and have changed the parameter on a method from a `LongParam` to `OptionalLong` as detailed in the deprecation warning for `LongParam`. 

I've also changed my code to call `getAsLong()` now that the parameter type no longer has a `get()` method.  Here's the code following updating to dropwizard 2. 

```
	@ExceptionMetered
	@GET
	@Path("/search")
	@Timed
	@ApiOperation(
			value="search",
			tags="Search",
			notes="Searches all records")
	@ApiResponses(value =
			{ @ApiResponse(code = 400, message = "'fromId' parameter is mandatory") })
	public ResultPage<Result> search(
			@Auth final ServiceUser user,
			@QueryParam("fromId") @DefaultValue("0") final OptionalLong fromId,
			@QueryParam("department") final Set<String> departments,
			@QueryParam("area") final Set<String> areas) {

		final Long idFrom = fromId.getAsLong();
		final List<Result> records = tenancyDao.execute(idFrom, departments, areas);

		return new ResultPageFactory<Result>(pageLimit).build(records);
	}
```

However now my `DefaultValue` annotation on the parameter no longer seems to be getting set when no value is passed to the parameter. This causes the exception from `fromId.getAsLong()` but I would expect it to always have a value due to the default value:
```
! java.util.NoSuchElementException: No value present
! at java.util.OptionalLong.getAsLong(OptionalLong.java:118)
```

Debugging through has led me to [this line in `SingleValueExtractor`](https://github.com/eclipse-ee4j/jersey/blob/2.29.1/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/SingleValueExtractor.java#L61):
```
return fromString((value == null && isDefaultValueRegistered()) ? getDefaultValueString() : value);
```

At this point `value` is `""` (empty string) so although `isDefaultValueRegistered()` is `true`, it uses `value` instead of the default value.

**What do I need to do to get it to use the default value when none is set, as it did with dropwizard 1.3.x?**

Note before I updated my code to use `OptionalLong` I was getting `java.lang.NumberFormatException: For input string: ""` from [this line in `AbstractParam`](https://github.com/dropwizard/dropwizard/blob/release/2.0.x/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/AbstractParam.java#L35) when the `LongParam` constructor was called, but the cause was the same: my param had a value of empty string instead of the default.

pkwarren:
DropWizard 2.x changed the behavior of OptionalInt/Long/Double
parameters in two ways. First, empty values used to result in the
`@DefaultValue` of a parameter being used. In DropWizard 2.x, this changed
to return OptionalXYZ.empty() instead.

Additionally, using an invalid value for a `@DefaultValue` parameter used
to fail fast when the Jersey resource was added. This changed with more
lax parsing in DropWizard 2.x.

Update to preserve the previous behavior of empty string parameter
values and invalid `@DefaultValue` values. This should avoid issues with
adoption of DropWizard 2.x.

Fixes #3123.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -12,6 +12,7 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         java_version: ['openjdk8', 'openjdk11', 'openjdk13']
         os: ['ubuntu-latest', 'windows-latest']
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/DefaultValueUtils.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/DefaultValueUtils.java
@@ -1,0 +1,26 @@
+package io.dropwizard.jersey.optional;
+
+import javax.annotation.Nullable;
+import javax.ws.rs.DefaultValue;
+import java.lang.annotation.Annotation;
+
+final class DefaultValueUtils {
+    private DefaultValueUtils() {}
+
+    /**
+     * Returns the value of the {@link DefaultValue#value()} if found in annotations.
+     * @param annotations Array of annotations (can be null).
+     * @return Value of {@link DefaultValue#value()} if found, otherwise null.
+     */
+    @Nullable
+    static String getDefaultValue(final Annotation[] annotations) {
+        if (annotations != null) {
+            for (Annotation annotation : annotations) {
+                if (DefaultValue.class == annotation.annotationType()) {
+                    return ((DefaultValue) annotation).value();
+                }
+            }
+        }
+        return null;
+    }
+}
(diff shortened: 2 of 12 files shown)


chris-orchard:
@pkwarren I've just updated to `2.0.2` and using `LongParam` as the parameter type I still get
`java.lang.NumberFormatException: For input string: ""`
instead of the default value when the parameter value is empty string.

Looking through your code change in more detail I see the fix only appears to be for the `OptionalLong` et` parameter types.

**Was this only meant to fix the regression for Optional types?** If so will you also be fixing the regression for the deprecated `LongParam` etc types too at some point? Otherwise I'd suggest this warrants a entry in [upgrade guide.](https://github.com/dropwizard/dropwizard/wiki/Upgrade-guide-1.3.x-to-2.0.x) as for me it is still a breaking change.

As I touched on [in my last comment on 3123](https://github.com/dropwizard/dropwizard/issues/3123#issuecomment-581812722) without going into specifics, my ideal goal at this point is just to update to dropwizard 2.x to address security vulnerabilities and library updates.

Although I _could_ change all my apis and test over to the equivalent `Optional` types this won't be a trivial exercise so I was planning on doing this gradually bit by bit.

However It appears my options are now:

1. Stick on dropwizard 1.3.x until I've time to change all parameters over to Optional types, then upgrade.
2. Change all my code and test to use `Optional` types now.
3. Attempt a custom fix myself to maintain the 1.3.x behaviour of empty string values for parameters with default values.

None of these are ideal.

PR #3142, opened 2020-02-12
Executor metrics shouldn't contain fmt string.

### LINKED ISSUE #3139: Jersey Client Is Created Using Format Itself Instead Of Formatted Name
https://github.com/dropwizard/dropwizard/blob/be1b220cfcc92ac26e0d3b1ca32c2a695d0dbbaf/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java#L115

ExecutorServiceManager expects a name, not format. But this line is passing the format itself instead of a name. 

JerseyClientBuilder (line 177) passes the format 'jersey-client-[NAME]-%d' to LifecycleEnvironment, which passes that format to this class as part of constructor. This class then takes the format, but does no formatting using it, rather just passes it wrongfully to ExecutorServiceManager.

Not sure what's the intent of that '%d' is. So can't really provide a pull request for it. Someone familiar with these code need to fix it.

Effect of this is, metric is getting generated with that %d in the name.

Same is occurring with TimeBoundHealthCheck pool also. I haven't yet traced the code responsible for it. It'll be nice to get that fixed too.

pkwarren:
Update ExecutorServiceBuilder to remove the format string from the
metric name. Currently, a thread pool named 'jersey-client-foo-%d' will
have metrics 'jersey-client-foo-%d.created',
'jersey-client-foo-%d.terminated', and 'jersey-client-foo-%d.running'
added to the metrics registry. The '-%d' format string is only meant to
ensure that threads are uniquely named and shouldn't be included in the
name of the metric.

Update ExecutorServiceBuilder to strip the format portion of the
nameFormat from the metric name.

Fixes #3139.

### DIFF
--- a/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
+++ b/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
@@ -6,6 +6,7 @@ import io.dropwizard.util.Duration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nonnull;
 import java.util.Locale;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
@@ -21,6 +22,7 @@ public class ExecutorServiceBuilder {
 
     private static final AtomicLong COUNT = new AtomicLong(0);
     private final LifecycleEnvironment environment;
+    @Nonnull
     private final String nameFormat;
     private int corePoolSize;
     private int maximumPoolSize;
@@ -48,12 +50,12 @@ public class ExecutorServiceBuilder {
         this(environment, nameFormat, buildThreadFactory(nameFormat));
     }
 
+    @SuppressWarnings("ResultOfMethodCallIgnored")
     private static ThreadFactory buildThreadFactory(String nameFormat) {
+        String.format(Locale.ROOT, nameFormat, 0); // Fail fast on invalid name format
         return r -> {
             final Thread thread = Executors.defaultThreadFactory().newThread(r);
-            if (nameFormat != null) {
-                thread.setName(String.format(Locale.ROOT, nameFormat, COUNT.incrementAndGet()));
-            }
+            thread.setName(String.format(Locale.ROOT, nameFormat, COUNT.incrementAndGet()));
             return thread;
         };
     }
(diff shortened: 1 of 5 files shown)


dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
```diff
@@ -48,12 +50,12 @@ public ExecutorServiceBuilder(LifecycleEnvironment environment, String nameForma
         this(environment, nameFormat, buildThreadFactory(nameFormat));
     }
 
+    @SuppressWarnings("ResultOfMethodCallIgnored")
     private static ThreadFactory buildThreadFactory(String nameFormat) {
+        String.format(Locale.ROOT, nameFormat, 0); // Fail fast on invalid name format
         return r -> {
             final Thread thread = Executors.defaultThreadFactory().newThread(r);
-            if (nameFormat != null) {
```
pkwarren: Not sure why this was relaxed in 2.x to allow null nameFormat. Previously, Guava validated the nameFormat was of a valid format and didn't allow null. If this is required, I can make it able to handle a null nameFormat.

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
```diff
@@ -116,6 +119,36 @@ public ExecutorService build() {
         return executor;
     }
 
+    static String getNameWithoutFormat(String nameFormat) {
```
pkwarren: I'm sure there is a better way to determine common prefix + suffix of two strings. This basically formats the string with 0 and then strips off the common prefix/suffix. If it finds a trailing '-' in the common prefix it removes it from the string.

This will take the common pattern 'jersey-client-somename-%d' and return 'jersey-client-somename'.

To get more flexibility, we could change ExecutorServiceBuilder to take a name (without a format string) to be used as the prefix for metrics.

dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
```diff
@@ -45,6 +46,8 @@ public void testGiveAWarningAboutMaximumPoolSizeAndUnboundedQueue() {
             .build();
 
         verify(log).warn(WARNING);
+        assertThat(metricRegistry.getMetrics().keySet())
+            .containsOnly("test.created", "test.terminated", "test.running");
```
pkwarren: Verify metric names.

dropwizard-lifecycle/src/test/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilderTest.java
```diff
@@ -124,6 +127,41 @@ public void shouldUseInstrumentedThreadFactory() {
...
+            { "my-client-%1$d-abc", "my-client-abc" },
+            { "-%d", "" },
+            { "%d", "" },
+            { "%d-abc", "-abc" },
+            { "%10d", "" },
+            { "%10dabc", "abc" },
+            { "%-10d", "" },
+            { "%-10dabc", "abc" },
+            { "%10s", "" },
+            { "%10sabc", "abc" },
```
pkwarren: Try a bunch of different combinations of format strings. Let me know if I've missed combinations that would be useful to test.

pkwarren:
> This should probably go into 2.1.0 in case anyone was relying on this. Can you also please update the release notes to call out this change?

While unlikely that someone was depending on `-%d` being part of metric names, I agree it is a change that may be safer in a minor release.

For release notes, should I create a new `upgrade-nodes-2_1_x.rst` file and add this there? Do we need to allow someone to set a name for metrics independently of the nameFormat so we can give users more control of the naming of metrics?

pkwarren:
@jplock - I've refactored the code some to try and make it easier to read and also updated with release notes for 2.1.0. Could you please take another look? Thanks.

PR #3148, opened 2020-02-15
Fix the wrong assert in the Testing Dropwizard doc

haibin:
###### Problem:

There is a wrong assert in an example in the Testing Dropwizard doc.

###### Solution:

Update the doc.

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -138,8 +138,9 @@ Next, write a test for deserializing a ``Person`` instance from JSON:
         @Test
         public void deserializesFromJSON() throws Exception {
             final Person person = new Person("Luther Blissett", "lb@example.com");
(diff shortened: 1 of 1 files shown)


jplock:
Did the example not work as written if equals() was implemented on Person?

haibin:
No, the example did not work. See below.

```
org.opentest4j.AssertionFailedError: 
Expecting:
 <com.example.helloworld.api.Person@75b25825>
to be equal to:
 <com.example.helloworld.api.Person@18025ced>
but was not.
...
```

joschi:
@haibin Thanks for bringing this up.

The `Person` class is expected to have an implementation of `equals()` and `hashcode()`.

https://github.com/dropwizard/dropwizard/pull/3148/files#diff-610d7574c51413d7ea91d122a34c6bacR62-R64

PR #3149, opened 2020-02-16
Test can fail if map iterates in a different order

testfixer:
###### Problem:

Test `handlesMethodNotAllowedWithHeaders` in `LoggingExceptionMapperTest.java`  depends on the iteration order of  `Response.getHeaders`, and its return type is `MultivaluedMap`. However, it does not guarantee any specific order of entries. Therefore, the assertion in that test cases will fail if the order is different.
###### Solution:

This PR proposes to use `containsAnyOf` instead of `contains`, which verifies that the actual map contains the given entries in any order.
###### Result:

The test becomes no longer flaky.


joschi:
@testfixer Thanks for your PR!

I've decided to address this slightly different in #3152.

PR #3156, opened 2020-02-19
Expose JmxReporter in Bootstrap

### LINKED ISSUE #3154: There is no way to shut down the JMXReporter created by Bootstrap
https://github.com/dropwizard/dropwizard/blob/ce6348e8aa89bdc63314d7563fbc3062fbb2232c/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java#L89

The JMXReporter returned on that line isn't saved so there is no way to clean it up.
That causes a memory leak when the application is restarted in the same JVM.

joschi:
Closes #3154

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
@@ -99,7 +99,7 @@ public class Bootstrap<T extends Configuration> {
     /**
      * Returns the {@link JmxReporter} registered with the bootstrap's {@link MetricRegistry}.
      *
-     * @since 2.0
+     * @since 2.1
      */
     @Nullable
     public JmxReporter getJmxReporter() {
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
@@ -23,6 +23,7 @@ import io.dropwizard.configuration.FileConfigurationSourceProvider;
 import io.dropwizard.jackson.Jackson;
 import io.dropwizard.jersey.validation.Validators;
 
+import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


dropwizard-core/src/main/java/io/dropwizard/setup/Bootstrap.java
```diff
@@ -86,10 +90,22 @@ public void registerMetrics() {
...
+        jmxReporter = JmxReporter.forRegistry(metricRegistry).build();
+        jmxReporter.start();
+
         metricsAreRegistered = true;
     }
 
+    /**
+     * Returns the {@link JmxReporter} registered with the bootstrap's {@link MetricRegistry}.
+     *
+     * @since 2.0
```
pkwarren: Should this be 2.0.2 or 2.1.0 (not sure which release it is planned to go out with)?

joschi: Good point. Since it's an additional feature (API-wise), it should go into Dropwizard 2.1.0. 👍

dropwizard-core/src/test/java/io/dropwizard/setup/BootstrapTest.java
```diff
@@ -93,6 +93,15 @@ public Histogram histogram(String name) {
                         "jvm.attribute.vendor", "jvm.classloader.loaded", "jvm.filedescriptor");
     }
 
+    @Test
+    public void allowsAccessToJmxReporter() {
+        final MetricRegistry newRegistry = new MetricRegistry();
+        bootstrap.setMetricRegistry(newRegistry);
+        assertThat(bootstrap.getJmxReporter()).isNull();
+        bootstrap.registerMetrics();
+        assertThat(bootstrap.getJmxReporter()).isNotNull();
```
pkwarren: Do you want to stop the JmxReporter when this test ends?

PR #3159, opened 2020-02-20
Migrate back to actions/setup-java

joschi:


joschi:
Jetty's ALPN implementation seems to be incompatible with Zulu OpenJDK 8. 😞

So let's postpone this, until we've had the time to dig into the issue.

PR #3191, opened 2020-03-11
Add CachingAuthenticator option to enable negative caches

tagomoris:
###### Problem:

`CachingAuthenticator` isn't caching negative results (`Optional.absent()`) because of #947.

But in other cases, not having a negative cache causes a kind of DoS.
If the underlying authenticator requires a little high cost to verify authentication request and the failure will not be cached, many invalid auth requests will cause resource shortage very easily.

In our case, client agents may send so many requests (possibly with invalid auth token by misconfiguration). If we authenticate those requests without negative caches, our underlying authentication endpoint will meet the unexpected number of requests.

###### Solution:

The change of this pull-requests introduces optional negative caches. CachingAuthenticator will enable negative caches only when it's specified so.


joschi:
@tagomoris Thanks for your contribution!

PR #3197, opened 2020-03-17
Update dead link to jersey documentation on filters and interceptors

todd-toast:
##### Problem:
There was a dead link for jersey filters and interceptors. Instead of leading you to the documentation it would just redirect you to the main jersey page.

###### Solution:
 I have updated this link to be https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/filters-and-interceptors.html.

###### Result:
The docs will have the proper link to jersey filters.


jplock:
Thanks!

PR #3202, opened 2020-03-22
Enable maven-dependency-plugin analysis

pkwarren:
Update dropwizard to run maven-dependency-plugin analyze-only goal as
part of the build. This will automate detection of new dependencies as
well as when dependencies are no longer in use. In some cases there may
be some false positives (seen primarily when only using an import from a
dependency), however overall this should improve the quality of the POM
files for dropwizard.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -39,6 +39,6 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-maven-
     - name: Build with Maven
-      run: ./mvnw -Pjakarta-apis -V -B -ff -s .github/settings.xml install '-Dmaven.test.skip=true' '-Dmaven.javadoc.skip=true'
+      run: ./mvnw -Pjakarta-apis -V -B -ff -s .github/settings.xml install '-DskipTests=true' '-Dmaven.javadoc.skip=true'
     - name: Run tests
       run: ./mvnw -Pjakarta-apis -V -B -ff -s .github/settings.xml verify
--- a/dropwizard-archetypes/src/test/resources/settings.xml
+++ b/dropwizard-archetypes/src/test/resources/settings.xml
@@ -17,6 +17,17 @@
                         <enabled>true</enabled>
                     </snapshots>
                 </repository>
+                <repository>
+                    <id>ossrh</id>
+                    <name>Sonatype Nexus Snapshots</name>
+                    <url>@localRepositoryUrl@</url>
+                    <releases>
+                        <enabled>false</enabled>
+                    </releases>
+                    <snapshots>
+                        <enabled>true</enabled>
+                    </snapshots>
+                </repository>
             </repositories>
             <pluginRepositories>
                 <pluginRepository>
--- a/dropwizard-dependencies/pom.xml
+++ b/dropwizard-dependencies/pom.xml
@@ -26,18 +26,24 @@
         <classmate.version>1.5.1</classmate.version>
         <commons-lang3.version>3.9</commons-lang3.version>
         <commons-text.version>1.8</commons-text.version>
+        <conscrypt-openjdk-uber.version>2.1.0</conscrypt-openjdk-uber.version>
         <error_prone.version>2.3.4</error_prone.version>
         <error_prone_annotations.version>2.3.4</error_prone_annotations.version>
         <freemarker.version>2.3.30</freemarker.version>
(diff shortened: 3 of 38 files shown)


dropwizard-archetypes/src/test/resources/settings.xml
```diff
@@ -17,6 +17,17 @@
                         <enabled>true</enabled>
                     </snapshots>
                 </repository>
+                <repository>
+                    <id>ossrh</id>
+                    <name>Sonatype Nexus Snapshots</name>
+                    <url>@localRepositoryUrl@</url>
```
pkwarren: This took a long time to figure out. Basically because we have the repository entry in the pom to fetch snapshots from ossrh, this also caused archetype builds to not use the dropwizard SNAPSHOT builds from the multi-module build and instead download the last published snapshot from Nexus. This meant that any changes made as part of this PR weren't visible to the archetype build (specifically, the jakarta.validation-api dependency added to dropwizard-dependencies BOM isn't in the published snapshot so the archetype build failed).

pom.xml
```diff
@@ -471,6 +477,22 @@
...
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>analyze</id>
+                        <goals>
+                            <goal>analyze-only</goal>
+                        </goals>
+                        <configuration>
+                            <failOnWarning>true</failOnWarning>
+                            <ignoreNonCompile>true</ignoreNonCompile>
```
pkwarren: I've relaxed warnings here for dependencies not in compile scope - this means that we might pull in unused dependencies at runtime/provided/test scope but that seems like a good compromise.

.github/workflows/maven.yml
```diff
@@ -39,6 +39,6 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-maven-
     - name: Build with Maven
-      run: ./mvnw -Pjakarta-apis -V -B -ff -s .github/settings.xml install '-Dmaven.test.skip=true' '-Dmaven.javadoc.skip=true'
+      run: ./mvnw -Pjakarta-apis -V -B -ff -s .github/settings.xml install '-DskipTests=true' '-Dmaven.javadoc.skip=true'
```
pkwarren: Running with `maven.test.skip` not only skips tests it doesn't even compile them (which means any dependencies required at compile scope for tests will not be detected). maven-dependency-plugin requires all of the code to be compiled as it performs bytecode analysis to determine what is needed.

pkwarren:
Full dependency diff before/after this change: https://gist.github.com/pkwarren/bc32b7b8bf1ebfa31a1e3a509b02b5ec

jplock:
I think @joschi should review this as well since he’s done a lot of work with the dependencies.

pkwarren:
I've updated this PR to pull back a few additional libraries which appear to trigger behavior at runtime. The only remaining dependencies I'm unsure of are `jetty-webapp` and `jetty-xml` - those are no longer pulled in and I'm wondering if we actually need them.

dropwizard-dependencies/pom.xml
```diff
@@ -87,6 +93,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.glassfish.hk2</groupId>
+                <artifactId>hk2-bom</artifactId>
+                <version>${hk2.version}</version>
```
joschi: Does it really make sense to declare a specific version of HK2?
If we do, we'll always have to keep in sync with what version Jersey is using to avoid compatibility issues.

dropwizard-archetypes/src/test/resources/settings.xml
pkwarren: It should work the same during a release build. This repository only allows snapshots and not releases.

dropwizard-dependencies/pom.xml
pkwarren: Yes it will need to stay in sync. We were using hk2 before without being explicit so relying on it being pulled in transitively from other jersey dependencies.

http://maven.apache.org/plugins/maven-dependency-plugin/analyze-only-mojo.html#ignoredUsedUndeclaredDependencies is the only mechanism I know of to keep using hk2 as we were before. This would essentially be disabling maven-dependency-plugin for certain artifact coordinates.

PR #3157, opened 2020-02-19
Escape EL expressions in ViolationCollector

joschi:
Fixes #3153

Issue #3153, opened 2020-02-18
Security issue


pwntester:
What would be a secure email alias to report a security issue?

Note: would be a nice to set a [security policy](https://help.github.com/en/github/managing-security-vulnerabilities/adding-a-security-policy-to-your-repository)

joschi:
@pwntester Thank you very much for raising this issue! 👍

We've added our security policy to this project:
https://github.com/dropwizard/dropwizard/security/policy

Please send your message to dropwizard.committers+security@gmail.com and we'll respond as fast as possible.

pwntester:
Nice, thank you!

PR #3208, opened 2020-03-26
Disable message interpolation in ConstraintViolations by default

joschi:
Disable message interpolation in `ConstraintViolations` by default but allow enabling it explicitly with `SelfValidating#escapeExpressions()`.

Additionally, `ConstraintViolations` now provides a set of methods which take a map of message parameters for interpolation. These message parameters are be escaped by default.

Thanks to @pwntester and @gsmet for their invaluable input!

Refs #3153
Refs #3157


pwntester:
LGTM, thanks!

Issue #3131, opened 2020-02-08
JPMS - split package `io.dropwizard.validation`


willpdp:
`io.dropwizard.validation` is a split package between dropwizard-core and dropwizard-validation. 
 Because of that I wasn't able to compile my app as a JPMS module.

[using 2.0.1] I tested moving `io/dropwizard/validation/InjectValidatorFeature.class` from dropwizard-core.jar into dropwizard-validation.jar and my app can build using JPMS and seems to run ok.

That's probably not the right fix for this but it got rid of the split package issue.  Could `io.dropwizard.validation` have a different package name in dropwizard-core or should that class move into dropwizard-validation?

joschi:
Refs #3233

joschi:
Closed via #3233.

PR #3233, opened 2020-04-06
Clean up package structure

joschi:
In order to properly support the Java Platform Module System (JPMS), the packages in modules must not overlap, or put differently, the packages may not be split into multiple modules.

See also: https://nipafx.dev/java-9-migration-guide/#split-packages

This change set cleans up the overlapping package structure where necessary (and is unfortunately breaking backward compatibility).

Affected packages:

- `dropwizard-core`: `io.dropwizard` -> `io.drowizard.core`
- `dropwizard-views`: `io.dropwizard.views` -> `io.dropwizard.views.common`
- `dropwizard-metrics`: `io.dropwizard.metrics` -> `io.dropwizard.metrics.common`
- `dropwizard-logging`: `io.dropwizard.logging` -> `io.dropwizard.logging.common`

Refs https://github.com/dropwizard/dropwizard/pull/2926
Refs https://github.com/dropwizard/dropwizard/issues/3131
Refs https://github.com/dropwizard/dropwizard/issues/3654


nickbabcock (dismissed):
Yeah looks like the structure needed to be cleaned up to be compliant. Thanks for this and making it backwards compatible.


You also mentioned


> We'll need more tests to see, if everything works with Java 9+ ; when using the module path


I'm not sure how one would test the module path. Were you able to include them in the PR?


> `io.dropwizard.metrics` -> `io.dropwizard.metrics.{common, core}`


I'd say, unless you want to make everything use the `core` suffix, having `io.dropwizard.metrics` match logging and views with the `common` suffix make sense

pkwarren:
I'm worried that the actual changes aren't going to be backwards compatible - if we end up changing the locations of classes it could break consumers.

For example, by changing contract to contract superclass (i.e. from io.dropwizard.Configuration to io.dropwizard.core.Configuration), we may break compatibility. http://wiki.eclipse.org/Evolving_Java-based_APIs_2 is a great reference for these sorts of changes to know the impact. We can also run a tool like https://lvc.github.io/japi-compliance-checker/ to verify the API contract hasn't changed. OkHttp also just updated from Java to Kotlin (https://square.github.io/okhttp/upgrading_to_okhttp_4/) and recommended use of https://github.com/siom79/japicmp to verify no API changes.

If we can't make this backwards compatible, then we might want to consider this for a 3.x release and change groupId, package name, and introduce compat packages if we want to support existing codebases. For example:

* Add io.dropwizard:dropwizard3-core which includes package io.dropwizard3.*
* Add io.dropwizard:dropwizard3-core-compat artifact which includes all classes from dropwizard-core 2.x package which is deprecated and extend dropwizard3-core classes.
* Update io.dropwizard:dropwizard-core to pull in dropwizard3-core and dropwizard3-core-compat artifacts. This way existing dependencies work unchanged.

joschi:
After playing around with this branch a little bit more I couldn't come up with a way to _not_ break backward compatibility in one way or the other.

So I think it makes sense to rip off the bandaid and introduce the breaking change in Dropwizard 2.1.0 to be done with it.

Going forward, this would mean that Dropwizard could be compatible to the JPMS and no other package renaming would have to be done.

The other possibility would be to postpone this change to Dropwizard 3.0.0 and do it together with the move to the Jakarta EE namespace (Jetty 11, Jersey 3, Hibernate Validator 7, etc.) and the Java 11 baseline (or drop this altogether).

@dropwizard/committers Any opinions?

jplock:
@joschi I think this would surprise people in 2.1.0 even if we heavily documented it.  I'd be in favor of starting 3.x with the changes you proposed and maybe at the same time saying 1.x would no longer be maintained/supported (even bug fixes), to encourage people to upgrade to at least 2.x.

joschi:
> and maybe at the same time saying 1.x would no longer be maintained/supported

@jplock Thankfully we already did that with [Dropwizard 1.3.29](https://github.com/dropwizard/dropwizard/releases/tag/v1.3.29). 😃

I don't have a strong opinion on whether to to this change in Dropwizard 2.1.x or 3.0.x.

The only thing which might not be desirable is mixing too many breaking changes into Dropwizard 3.0.x.

pstackle:
I'll put in a vote for holding off the change until a 3.0 release. Having breaking changes is much more expected for a major version bump compared to a minor version bump.

zUniQueX:
> The only thing which might not be desirable is mixing too many breaking changes into Dropwizard 3.0.x.

Thanks to our current upgrade plans, Dropwizard 3.0.x won't introduce that much API changes. Since Jersey will bring JPMS support firstly in 3.1.0 on the `jakarta` namespace, we don't even have the need to bring this to DW 3. But I see no problem bringing this to the 3.0.0 release.

joschi:
@dropwizard/committers I've rebased the branch on `release/3.0.x`.

docs/source/getting-started.rst
```diff
@@ -280,8 +280,8 @@ commands which provide basic functionality. (More on that later.) For now, thoug
     package com.example.helloworld;
 
     import io.dropwizard.Application;
-    import io.dropwizard.setup.Bootstrap;
-    import io.dropwizard.setup.Environment;
+    import Bootstrap;
```
zUniQueX, line 283: Overall looks good, but some doc imports seem to be broken

joschi: @zUniQueX Thanks for noticing! 😄

dropwizard-logging/src/main/java/io/dropwizard/logging/common/filter/FilterFactory.java
```diff
@@ -12,7 +12,7 @@
  * <ol>
  * <li>Create a class which implements {@link FilterFactory}.</li>
  * <li>Annotate it with {@code @JsonTypeName} and give it a unique type name.</li>
- * <li>add a {@code META-INF/services/io.dropwizard.logging.filter.FilterFactory} file with your
+ * <li>add a {@code META-INF/services/FilterFactory} file with your
```
zUniQueX, line 15: @joschi Is this corrent in the new implementation?

zUniQueX (approved):
@joschi This looks very good, thanks for the effort :heart:

I just have a few minor suggestions:
 - `DefaultServerFactoryTest:18-19` multiple imports for `Environment` and `ExceptionMapperBinder`
 - `SimpleServerFactoryTest:17` multiple imports for `Environment`
 - `JdbiTest:8` multiple imports for `Environment`
 - `getting-started.rst:196` broken import
 - `configuration.rst:892` package `logging.common` instead of `logging.filter.common`
 - `FilterFactory:15` broken JavaDoc
 
And one question left: Could it get a problem, that the packages in `dropwizard-e2e` and `dropwizard-example` are named `com.example` and `com.example.helloworld`?

joschi:
@zUniQueX Thanks for the detailed look at this PR! ❤️ 

> And one question left: Could it become a problem, that the packages in `dropwizard-e2e` and `dropwizard-example` are named `com.example` and `com.example.helloworld`?

No, these are not an issue because neither is being published and consumed by other projects and they never end up in the same application and module path.

zUniQueX:
> > And one question left: Could it become a problem, that the packages in `dropwizard-e2e` and `dropwizard-example` are named `com.example` and `com.example.helloworld`?
> 
> No, these are not an issue because neither is being published and consumed by other projects and they never end up in the same application and module path.

Okay, thanks for taking a look at this :+1:

PR #3239, opened 2020-04-09
Replace global dropwizard.version property

joschi:
Many Maven plugins and other tools needed to automate releases don't support the Dropwizard version being used from a global property in POM files.

### DIFF
--- a/dropwizard-example/pom.xml
+++ b/dropwizard-example/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>dropwizard-example</artifactId>
-    <version>2.1.0-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
     <groupId>io.dropwizard</groupId>
     <name>Dropwizard Example Application</name>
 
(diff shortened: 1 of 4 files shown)


dropwizard-example/pom.xml
```diff
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>dropwizard-example</artifactId>
-    <version>2.1.0-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
```
pkwarren, line 6: Why this change? It is going to stay at a SNAPSHOT version?

joschi: Good question! `dropwizard-example` is not being released/deployed like the other artifacts and it's not using `dropwizard-project` or `dropwizard-parent` as parent POM, so there's no use in tracking Dropwizard's version number in this sub-module.

I wanted to make clear, that the application's version doesn't have to be the same as the version of Dropwizard used in it.

The version of `dropwizard-dependencies` is still being updated automatically when running for example `mvn versions:set` or `mvn release:prepare`.

PR #3269, opened 2020-05-02
Allow overriding default media type in AssetServlet/AssetBundle

joschi:
Sometimes the default media type of `text/html` in the `AssetServlet` and `AssetBundle` might not be the best fit for every use case.

This PR exposes a setting to specify the default media type for `AssetBundle` and `AssetServlet` and override the default of `text/html`.

### DIFF
--- a/dropwizard-assets/src/main/java/io/dropwizard/assets/AssetsBundle.java
+++ b/dropwizard-assets/src/main/java/io/dropwizard/assets/AssetsBundle.java
@@ -18,11 +18,13 @@ public class AssetsBundle implements ConfiguredBundle<Configuration> {
     private static final String DEFAULT_ASSETS_NAME = "assets";
     private static final String DEFAULT_INDEX_FILE = "index.htm";
     private static final String DEFAULT_PATH = "/assets";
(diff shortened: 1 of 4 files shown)


dropwizard-assets/src/main/java/io/dropwizard/assets/AssetsBundle.java
```diff
@@ -83,12 +85,31 @@ public AssetsBundle(String resourcePath, String uriPath, String indexFile) {
...
+     *
+     * @param resourcePath     the resource path (in the classpath) of the static asset files
+     * @param uriPath          the uri path for the static asset files
+     * @param indexFile        the name of the index file to use
+     * @param assetsName       the name of servlet mapping used for this assets bundle
+     * @param defaultMediaType the default media type for unknown file extensions
+     * @since 2.0
+     */
+    public AssetsBundle(String resourcePath, String uriPath, String indexFile, String assetsName,
+                        String defaultMediaType) {
```
pkwarren, line 112: This is getting to have quite a few parameters (and of the same type) so might not be that usable if we continue to add new parameters here. It might be nice to investigate using a builder to instantiate the bundle next time.

Issue #3277, opened 2020-05-05
Documentation: Task execute method has changed


barthanssens:
When upgrading from 1.3 to 2.0.9, I've noticed that the execute method signature has slightly changed to `execute(Map<String,List<String>> parameters, PrintWriter output)`, but this is not mentioned in the migration guide, nor in the manual https://www.dropwizard.io/en/latest/manual/core.html#tasks

joschi:
@barthanssens Good catch, thanks a lot!

PR #3279, opened 2020-05-05
Added note on Task.execute method

barthanssens:
###### Problem:

Task.execute method parameter has slightly changed between 1.3 and 2.0.

###### Solution:

Added upgrade note

###### Result:

https://github.com/dropwizard/dropwizard/issues/3277


joschi:
@barthanssens Thanks!

PR #3284, opened 2020-05-09
use JsonUnit to ignore the order of fields in expected JSON string

contextshuffling:
###### Problem:

Test `io.dropwizard.jackson.JacksonTest#objectMapperSerializesNullValues` uses `Jackson` to serialize object `pojo`. However, `Jackson` uses reflection API `getDeclaredFields` (https://github.com/FasterXML/jackson-databind/blob/3c67dc5bd9714635f6999bcb401b132791e9d7d0/src/main/java/com/fasterxml/jackson/databind/introspect/AnnotatedFieldCollector.java#L70) to get the field names, values, and etc. However, API `getDeclaredFields` does not guarantee any specific order of fields. So the serialized result (in this case, a JSON string) can be different depending on the returned result. 

Test `io.dropwizard.jackson.JacksonTest#objectMapperSerializesNullValues` will fail if the serialized string is not in the format as expected. 

An example error message:
```
Expecting:
 <"{"uuid":null,"string":null}">
to be equal to:
 <"{"string":null,"uuid":null}">
but was not.
```

###### Solution:

The test is actually coded with `assertJ`. I found a good test plugin (https://github.com/lukas-krecan/JsonUnit) that integrates well with `assertJ` to compare JSON string without taking the order into consideration.

###### Result:

Test will generate deterministic result (always pass in this case.)

### DIFF
--- a/dropwizard-dependencies/pom.xml
+++ b/dropwizard-dependencies/pom.xml
@@ -73,6 +73,7 @@
         <junit.version>4.13</junit.version>
         <junit5.version>5.6.2</junit5.version>
         <mockito.version>3.3.3</mockito.version>
+        <jsonunit.version>2.17.0</jsonunit.version>
 
         <!-- Plugin versions -->
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
@@ -507,6 +508,12 @@
                 <groupId>org.bouncycastle</groupId>
                 <artifactId>bcprov-jdk15on</artifactId>
                 <version>${bcprov-jdk15on.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>net.javacrumbs.json-unit</groupId>
+                <artifactId>json-unit-assertj</artifactId>
+                <version>${jsonunit.version}</version>
                 <scope>test</scope>
             </dependency>
         </dependencies>
--- a/dropwizard-jackson/pom.xml
+++ b/dropwizard-jackson/pom.xml
@@ -89,6 +89,11 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>net.javacrumbs.json-unit</groupId>
+            <artifactId>json-unit-assertj</artifactId>
+            <scope>test</scope>
+        </dependency>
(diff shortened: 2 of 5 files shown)


joschi:

Any chance doing this without adding an additional dependency, for example with [`@JsonPropertyOrder`](https://fasterxml.github.io/jackson-annotations/javadoc/2.10/com/fasterxml/jackson/annotation/JsonPropertyOrder.html)?

contextshuffling:
@joschi thanks for the comment, I will look into it

contextshuffling:
@joschi Thank you for pointing out that annotation. There're two more tests that cannot be fixed with an annotation. I've updated my PR with those two. Let me know if you still don't want the JsonUnit dependency, and I'll look for other ways to fix those two tests.

joschi:
> There're two more tests that cannot be fixed with an annotation.

Why couldn't they?

contextshuffling:
@joschi applying the annotation for the map field does not seem to have any effect on enforcing the order.

joschi:
@contextshuffling True. We could use a `SortedMap` or `NavigableMap` such as `TreeMap` in that test case.

Did these tests actually fail for you? Because this would be the first time it happened as far as I remember.

contextshuffling:
@joschi I used a tool called NonDex (https://github.com/TestingResearchIllinois/NonDex)

The tests did not fail with the current HashMap implementation in Java 8, 
but using the NonDex tool to check for other implementations make the 
tests fail.
Following is a script to reproduce
```
git clone --depth 1 https://github.com/dropwizard/dropwizard
cd dropwizard
sed -i 's;<plugins>;<plugins> <plugin> <groupId>edu.illinois</groupId> <artifactId>nondex-maven-plugin</artifactId> <version>1.1.2</version> </plugin>;' pom.xml
mvn nondex:nondex -Dtest=io.dropwizard.jackson.JacksonTest#objectMapperSerializesNullValues -pl dropwizard-jackson -DnondexRuns=10
```

Fixing these tests ahead of time will make them not fail when the implementation changes.

joschi:
@contextshuffling While I'm not sure this was a real problem (since the respective tests never failed locally or in CI builds), I've changed the tests accordingly.

Thanks for the hint!

PR #3285, opened 2020-05-09
Reuse DefaultThreadFactory

dennyac:
###### Problem:
The current default thread factory creates a new DefaultThreadFactory instance whenever a new thread is constructed

###### Solution:
Just reusing the same threadFactory per executorService

###### Result:
Will reduce object creation when requesting new threads


jplock:
We’ll need to merge this into the release/2.0.x branch as well

PR #3286, opened 2020-05-09
Add managed instrumented executors

dennyac:
###### Problem:
1. For threads we have a global counter. Would prefer a counter per executor service (Use a global pool and local thread number similar to DefaultThreadFactory). Can move threadFactory logic to its own class. Would've extended threadFactory but its a private static class.
2. nameFormat seems very unintuitive. It's not clear to users that they have to specify one. Also its usage across classes vary. (ScheduledExecutorServiceBuilder doesn't have that requirement).
3. People who aren't familiar with the ThreadPoolExecutor may run into issues like https://github.com/dropwizard/dropwizard/issues/2553 when using the ExecutorServiceBuilder. 

###### Solution:
1. Use logic used in DefaultThreadFactory with the ability to name threads
2. Avoid nameFormat since the NamedThreadFactory deals with this
3. Create helper methods to managed/instrumented executor services from the Executors class

###### Result:
- User passes a prefix and the threadFactory will add pool/thread number. 
- Whatever the user passes is what is used in the metric name. The methods getNameWithoutFormat in this https://github.com/dropwizard/dropwizard/pull/3142 will return a string appended to itself, so it may have to be reverted. @pkwarren - Any feedback/concerns?
- Includes managed and instrumented executor service factory methods. ThreadPoolExecutor Javadocs urges users to use the Executors factory methods that preconfigure settings for the most common usage scenarios. Also the InstrumentedExecutorService has additional metrics like idle/duration timers which InstrumentedThreadFactory (used in ExecutorServiceBuilder) doesn't have

### DIFF
--- a/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
+++ b/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/ExecutorServiceBuilder.java
@@ -7,23 +7,19 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
-import java.util.Locale;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.atomic.AtomicLong;
 
 public class ExecutorServiceBuilder {
     private static Logger log = LoggerFactory.getLogger(ExecutorServiceBuilder.class);
 
-    private static final AtomicLong COUNT = new AtomicLong(0);
     private final LifecycleEnvironment environment;
     @Nonnull
-    private final String nameFormat;
+    private final String namePrefix;
     private int corePoolSize;
     private int maximumPoolSize;
     private boolean allowCoreThreadTimeOut;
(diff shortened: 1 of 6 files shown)


dennyac:
Don't think anything here can be considered as breaking changes. Can update release notes/docs/tests if you think this is a reasonable approach. 

Event if a subset of these changes are acceptable, let me know and I'll make appropriate changes.

pkwarren:
In the previous commit I was mainly hoping to restore the behavior from DropWizard 1.x (which used Guava's ThreadFactoryBuilder and the nameFormat required a format string). This was changed when Guava dependencies were removed and led to incorrect names on the metrics for the executor.

To be fully compatible with what is out there, it feels like we should keep (and deprecate) all usage of nameFormat (including ExecutorServiceBuilder constructors) and add a new `.withMetricPrefix(String)` method to ExecutorServiceBuilder that allows explicit naming of metrics.

Then if nameFormat is not set, we skip wrapping the thread factory (relying on caller to use NamedThreadFactory as you've proposed) and use metricPrefix to name metrics.

dennyac:
I went through the history and totally understand the rationale behind your commit. Let me give this some thought and make some changes locally based on your feedback. Will post back here shortly

dennyac:
Spent some time trying to implement it such that it is backward compatible, but had a few questions. My questions are based on what I expect to be the behavior for end users (Correct me if I'm wrong)
1. Thread name and metric name share prefix, unless user provides their own threadFactory (That's the current behavior and a good way to correlate logs and metrics) 
  - Achieve this by sharing metricPrefix between threadFactory and metrics
2. metric/thread name is always unique 
  - Using a default value with global counter or user provided prefix

Questions - 
1. If users leverage the deprecated constructors and calls withMetricPrefix(), what takes precedence for naming the metrics/threadFactory, nameFormat or metricPrefix?

2. Are we fine with making metricPrefix a constructor argument instead of a builder method. Reasons why - 
- It's harder to share same metricPrefix with threadFactory and metrics, since threadFactory is created in the constructors whereas metricPrefix is set later on in builder method. Can add null check and move threadFactory creation to the build() method, but would like to avoid that. 
- We always require metricPrefix. Can add a null check and provide a default metricPrefix, but would like to avoid that. 

If it's in the constructor, I can create a default metricPrefix (if not provided), share metricPrefix with thread/metric name and avoid null checks. But if we move it to a constructor and I deprecate nameFormat and add namePrefix, you'll have two constructors that look very similar which is weird.

It feels like it may be confusing to the end user if we try to support both nameFormat and namePrefix (this is subjective). Although this PR changes thread/metric name, end users can still get the same thread/metric name they require by passing appropriate parameters (Can document this in release notes?). Code will be a lot cleaner too. Let me know your thoughts and I can go ahead and make the changes

PS - To keep this PR focussed on nameFormat, I can remove the executor factory methods into it's own PR if needed. Let me know

dennyac:
@pkwarren  - Just checking if you had any feedback/concerns and if/how I should proceed.

Tagging other contributors as well - @joschi @jplock

PR #3289, opened 2020-05-11
Minor: Fix testing docs

jan25:
Signed-off-by: Abhilash Gnan <abhilashgnan@gmail.com>

Minor fix in testing docs

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -139,7 +139,7 @@ Next, write a test for deserializing a ``Person`` instance from JSON:
         public void deserializesFromJSON() throws Exception {
             final Person person = new Person("Luther Blissett", "lb@example.com");
             assertThat(MAPPER.readValue(fixture("fixtures/person.json"), Person.class))
-                    .isEqualTo(person);
+                    .isEqualToComparingFieldByField(person);
         }
     }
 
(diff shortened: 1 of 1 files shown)


joschi:
@jan25 Thank you for your contribution!

Could you please describe which issue you are solving with your change?

We imply in the documentation that the `Person` class overrides `equals()` and `hashCode()`, so that `assertThat(...).isEqualTo(...)` would work:
https://github.com/dropwizard/dropwizard/pull/3289/files#diff-610d7574c51413d7ea91d122a34c6bacL62-L64

Maybe we should make that more prominent instead of this particular change? Otherwise, we would have to update a few more test examples in the documentation. 😉

jan25:
Hi @joschi!

I tried to write a test for my project using the examples from the docs, which kept failing due to `assertThat(...).isEqualTo(...)` doing shallow comparison. And, i totally missed the comments of `equals()` etc.

Making commented part prominent sounds a better way than to clutter the docs with long assertJ things. Also, i think `equals()` is rarely used, so other way is split the checks into two `assertThat(...).isEqualTo(...)` checking the fields explicitly. thoughts?

jan25:
Maybe this is not big of a problem, can be fixed if it gets reported widely from other parts of docs. Closing this PR for now.

PR #3298, opened 2020-05-19
Add CodeQL analysis for security scans

joschi:
https://github.com/github/codeql-action


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,44 @@
+name: "Code Scanning - Action"
+
+on:
+  push:
```
joschi, line 4: We should probably remove the `push` event trigger which is covered by the regular CI workflow.

PR #3309, opened 2020-05-26
feat: Introducing severity as a additional field for EventJsonLayout

irvifa:
###### Problem:

As of now, we only have "level" when sometimes the logger is parsed
based on "severity" label.

###### Solution:

Adding a new field in EventJsonLayout


irvifa:
I got:

```
io.dropwizard.configuration.ConfigurationParsingException: C:\Users\RUNNER~1\AppData\Local\Temp\dropwizard-invalid-config9743633157967435864.yml has an error:
  * Configuration at C:\Users\RUNNER~1\AppData\Local\Temp\dropwizard-invalid-config9743633157967435864.yml must not be empty
```

Is this expected?

joschi:
@irvifa Why would we want to log the [event level](http://logback.qos.ch/apidocs/ch/qos/logback/classic/spi/ILoggingEvent.html#getLevel()) twice, only with different keys?

irvifa:
The thing is sometimes logging mechanism/console parse severity as the key of understanding the severity of the event. Although I also understand we can do this by extending the Abstract class and provide our own json formatter.

joschi:
@irvifa This sounds more like you should configure your log shipper accordingly.

Thanks for your contribution, but I don't think that merging this PR makes sense for the majority of Dropwizard users.

Issue #3313, opened 2020-05-31
Logging Application Build Version during StartUp


fazeem84:
I would like to Log Build Version Number along with the Application Name during the server start up.
This can be done in two ways 
  1. By Reading Implementation-Version property from Manifest file
  2. By introducing new file in META-INF(for this to work build plugin needs to be introduced ) to store build info
@joschi I would like to know your thoughts on introducing this feature
I can work on this feature if everybody interested

joschi:
I'm not convinced that this change should be in the core project as the information people usually want to log on startup is very specific to each project.

Maybe you want to log the commit and the branch from which the application was built and a unique identifier to the build in your CI. Maybe you want to log the Maven GAV coordinates.

PR #3326, opened 2020-06-08
Added method to report the Dropwizard Build version

ctasada:
###### Problem:
Logging Application Build Version during StartUp #3313 

###### Solution:
Retrieved by reading Implementation-Version property from Manifest file

###### Result:
Applications needing the Dropwizard can retrieve it directly

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/DropwizardVersion.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/DropwizardVersion.java
@@ -1,0 +1,27 @@
+package io.dropwizard;
+
+/**
+ * Class that exposes the Dropwizard version. Fetches the "Implementation-Version"
+ * manifest attribute from the jar file.
+ * <p>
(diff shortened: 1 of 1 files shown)


joschi:
@ctasada Thank you for your contribution!

I'm not 100% convinced that adding this 2 line implementation to Dropwizard makes sense.

We'd have to maintain it for the foreseeable future and not every Dropwizard application is being deployed in a JAR.

@dropwizard/committers Any opinions on this?

ctasada:
@joschi Even if it's deployed as a WAR this approach should still work. If it's deployed as a fat-jar then it will fail. In that case we need to go for a more complex approach were a property is added on build time to a resource file.

I'm fine taking that other approach if there's consensus

joschi:
I'm not convinced that this change should be in the core project as the information people usually want to log on startup is very specific to each project.

Maybe you want to log the commit and the branch from which the application was built and a unique identifier to the build in your CI. Maybe you want to log the Maven GAV coordinates.

PR #3327, opened 2020-06-08
Start moving example code from docs to runnable code

joschi:
Move example code from documentation from pure markup to real code which is being verified during build.

See also: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -141,31 +141,24 @@ You have three alternatives from here:
 Tutorial
 --------
 
-First, add a ``dropwizard.version`` property to your POM with the current version of Dropwizard
-(which is |release|):
+First, add the ``dropwizard-bom`` bill of materials (BOM) into the ``dependencyManagement`` section of your POM
+with the current version of Dropwizard (which is |release|):
 
-.. code-block:: xml
-
-    <properties>
-        <dropwizard.version>INSERT VERSION HERE</dropwizard.version>
-    </properties>
+.. literalinclude:: /examples/getting-started/pom.xml
+    :language: xml
+    :lines: 14-25
+    :dedent: 4
 
 Add the ``dropwizard-core`` library as a dependency:
 
 .. _gs-pom-dependencies:
 
-.. code-block:: xml
+.. literalinclude:: /examples/getting-started/pom.xml
+    :language: xml
+    :lines: 26-32
+    :dedent: 4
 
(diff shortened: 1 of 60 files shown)


joschi:
@dropwizard/committers This is meant as a start to see if this approach is feasible. We can move the example code one by one in future PRs (or get some help from the community) if it makes sense.

docs/source/getting-started.rst
```diff
@@ -141,31 +141,24 @@ You have three alternatives from here:
...
+with the current version of Dropwizard (which is |release|):
 
-.. code-block:: xml
-
-    <properties>
-        <dropwizard.version>INSERT VERSION HERE</dropwizard.version>
-    </properties>
+.. literalinclude:: /examples/getting-started/pom.xml
+    :language: xml
+    :lines: 14-25
```
pkwarren, line 149: I like this idea quite a bit - having runnable code embedded in the docs is much better than copy/pasting pieces. I wonder though if we should look at having comments in the code and using `start-after`/`end-before` instead of calling out specific line numbers so that the code can be refactored without having to keep track of what line numbers changed.

joschi: Yes, I thought the same, especially since I had to update the line numbers a few times while migrating the example code.

I didn't find any extension for [Sphinx](https://www.sphinx-doc.org) allowing to specify the code snippets with comments as boundaries but would also prefer it very much.

There's a plugin for Jekyll which can include source snippets marked by comments:
https://github.com/tomdalling/jekyll-include_snippet

Maybe there's a corresponding Sphinx extension: https://pypi.org/search/?q=sphinx

pkwarren: This is what sounded promising (but I haven't tried it): https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

> Another way to control which part of the file is included is to use the start-after and end-before options (or only one of them). If start-after is given as a string option, only lines that follow the first line containing that string are included. If end-before is given as a string option, only lines that precede the first lines containing that string are included. The start-at and end-at options behave in a similar way, but the lines containing the matched string are included.

joschi: Yes, I just read about these options too and it seems to work. 😄

I'll adjust the PR in the next few days.

joschi:
Superseded by #5183.

PR #3331, opened 2020-06-13
Logging Application Build Version during Server StartUp #3313

fazeem84:
###### Problem:
For Logging application Build Version

###### Solution:
1. Added version property in Application class it will work similar to application name,if nothings is supplied by the application developer.version will read from application METAINF file
2. Added version property in Environment Object
3. Mofied printBanner() by adding version argument


joschi:
@fazeem84 Thanks for your contribution!

Similar to the comment in #3326 (https://github.com/dropwizard/dropwizard/pull/3326#issuecomment-642536119), I'm not convinced that this change should be in the core project as the information people usually want to log on startup is very specific to each project.

Additionally, the change to the `Environment` class is not backward-compatible and would force all users of Dropwizard to update their code.

fazeem84:
@joschi thank for your valuable feedback..ok I got your point the backward compatibility. what about implementing it in Application class?..we have a --version command which works similar fashion.so why can't we implement as part of Application class.this is just a suggestion from my side.. expecting your thoughts about it

Issue #3311, opened 2020-05-27
Replace com.google.code.findbugs:jsr305 with com.github.spotbugs:spotbugs-annotations


koalalam:
Old project: https://github.com/findbugsproject/findbugs
New project: https://github.com/spotbugs/spotbugs

joschi:
@koalalam Would you be able to provide a PR for this change?

PR #3351, opened 2020-06-27
3311 replace jsr305 annotations with spotbugs-annotations

olegzzz:
###### Problem:
Solution to issue https://github.com/dropwizard/dropwizard/issues/3311

###### Solution:
Replace classes from jsr305 with classes from spotbugs-annotations.

###### Result:
Note that jsr305 dependency is still present at compile time through some 3rd party dependencies. When removed completely, there are warnings like this (although tests pass)
```
[WARNING] unknown enum constant When.UNKNOWN
```
most likely caused by Mockito use of `javax.annotation.meta.When`

### DIFF
--- a/dropwizard-auth/pom.xml
+++ b/dropwizard-auth/pom.xml
@@ -16,32 +16,14 @@
         <dependency>
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-jersey</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>com.google.code.findbugs</groupId>
-                    <artifactId>jsr305</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-logging</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>com.google.code.findbugs</groupId>
-                    <artifactId>jsr305</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-util</artifactId>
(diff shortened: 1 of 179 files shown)


joschi:
@olegzzz Thanks for your contribution!

Is there any particular advantage in using `spotbugs-annotations` over `jsr305`?

olegzzz:
Wasn't the ask was to replace jsr305 annotations with spotbugs annotations?

joschi:
@olegzzz Yes, but do you see any advantage in using one artifact over the other?

brentryan:
The main advantage is to move onto something that is still supported. It appears that jsr305 has gone dormant and everything is moving away from its use: guava, micronaut are just a few...

joschi:
> The main advantage is to move onto something that is still supported.

These are marker annotations, nothing which contains actual logic and it seems that currently more projects (static code analyzers) support the JSR 305 variants of these annotations than the SpotBugs ones.

> It appears that jsr305 has gone dormant and everything is moving away from its use: guava, micronaut are just a few...

At least Guava is only planning to do so but hasn't done so yet: https://github.com/google/guava/issues/2960

As for the choice of replacement, why SpotBugs and not for example [`org.jetbrains:annotations`](https://github.com/JetBrains/java-annotations)?

brentryan:
Here's the example from micronaut PR https://github.com/micronaut-projects/micronaut-core/issues/1633

joschi:
@dropwizard/committers Any opinion on this?
Should we target the change for Dropwizard 2.1.x?

jplock:
I've been struggling with this one myself.  Even though the jsr305 annotations aren't "maintained", they are well supported and not likely to need changing.  I'd probably leave them using `javax.annotation` for now.

olegzzz:
Closing this as not required.

PR #3369, opened 2020-07-16
Update Jetty to the latest secure version, 9.4.30.v20200611

MatthewRyanRead:
###### Problem:

My org is in the process of upgrading Jetty to a recent version in light of [CVE-2019-10247](https://github.com/advisories/GHSA-xc67-hjx6-cgg6) and other vulnerabilities.  We're still using Dropwizard 1.3, which uses `9.4.18.v20190429`.  When we pin a more recent version of Jetty in our projects, services fail to start up with SSL with this error:

```
java.lang.IllegalStateException: KeyStores with multiple certificates are not supported on the base class org.eclipse.jetty.util.ssl.SslContextFactory. (Use org.eclipse.jetty.util.ssl.SslContextFactory$Server or org.eclipse.jetty.util.ssl.SslContextFactory$Client instead)
```

###### Solution:

- Cherry-pick the changes from #2734 to instantiate `SslContextFactory.Server` instead of the deprecated `SslContextFactory`
    -  This might have been meant to be backported to Dropwizard 1.3 anyways, given that DW13 advanced its version of Jetty twice after that (https://github.com/dropwizard/dropwizard/commit/3500399b5b128f5988b6aaf6aa702ae1d1b56c5d & https://github.com/dropwizard/dropwizard/commit/eec03a99e077ce06c053d4e61b64d913d9469fd3).
- Bump the Jetty version to the most recent and secure one ,`9.4.30.v20200611`, which is also in use by Dropwizard 2.0.

###### Result:

I also cherry-picked the changes from #2460, which were also required for the newer Jetty version, and pulled in the latest Jetty by cherry-picking all of the previous Jetty bumps (for maximum traceability).

I modified the cherry-picked commits to avoid the introduction of JUnit 5 and the `dropwizard-dependencies` module, in order to prevent any breaking changes to Dropwizard 1.3.

### DIFF
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -22,7 +22,7 @@
         <guava.version>24.1.1-jre</guava.version>
         <jersey.version>2.25.1</jersey.version>
         <jackson.version>2.9.10.20200621</jackson.version>
-        <jetty.version>9.4.24.v20191120</jetty.version>
+        <jetty.version>9.4.25.v20191220</jetty.version>
         <servlet.version>3.0.0.v201112011016</servlet.version>
         <metrics4.version>4.0.5</metrics4.version>
         <slf4j.version>1.7.26</slf4j.version>
--- a/dropwizard-bom/pom.xml
+++ b/dropwizard-bom/pom.xml
@@ -22,7 +22,7 @@
         <guava.version>24.1.1-jre</guava.version>
         <jersey.version>2.25.1</jersey.version>
         <jackson.version>2.9.10.20200621</jackson.version>
-        <jetty.version>9.4.22.v20191022</jetty.version>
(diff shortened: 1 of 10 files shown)


MatthewRyanRead:
I can't seem to reproduce the `Http2IntegrationTest.testHttp2ManyRequests` failure locally.

joschi:
@MatthewRyanRead Thanks for your contribution!

Are there any reasons you can't upgrade to Dropwizard 2.0.x? Is anything missing?

MatthewRyanRead:
Unfortunately we're just behind.  We upgraded to 1.3 right before 2.0 came out and are still resolving a couple issues with our tooling built around it.

MatthewRyanRead:
Closing in favor of #3411.

PR #2541, opened 2018-11-08
Fix EofException spamming of logs

shail:
##### Problem:

This change is made to address https://github.com/dropwizard/dropwizard/issues/2482. When a client disconnects before the response is written, or the full response has already been written, an `EofException` is thrown. An exception mapper can't be used because this error arises when the response is attempting to be written. This causes a lot of `ERROR` logs which clutters logging.

##### Solution:

The solution that has been implemented is to catch and ignore the `EofException` during the response write. The ideal solution would be to catch the `EofException` and accept configuration to log at different levels. This is a much more involved change and can be implemented if there is want for it. This implementation shouldn't preclude such an effort.

##### Result:

No more `EofException` errors in logs

Issue #2482, opened 2018-09-06
EarlyEofExceptionMapper doesn't seem to work (<=Dropwizard 1.3.5)


winklehner:
As I see so far, dropwizard registers by standard some Exception Mappers, one of the is the EarlyEofExceptionMapper. Unfortunately though this mapper never seems to get called, as I'm seeing lots of those messages when a client aborts the transfer:

! java.io.IOException: Broken pipe
! at java.base/sun.nio.ch.FileDispatcherImpl.writev0(Native Method)
! at java.base/sun.nio.ch.SocketDispatcher.writev(SocketDispatcher.java:51)
! at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:185)
! at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:132)
! at java.base/sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:506)
! at org.eclipse.jetty.io.ChannelEndPoint.flush(ChannelEndPoint.java:266)
! ... 73 common frames omitted
! Causing: org.eclipse.jetty.io.EofException: null
...aso.

But this should be catched right? (org.eclipse.jetty.io.EofException) 

Am I missing something or is this a bug?

winklehner:
Could it be that reason is null and not EOF?

nickbabcock:
We have a test that confirms an eof exception mapper is hit when a client advertises it will send 100000 bytes, but it disconnects after sending a single byte. But there could be an edge case that you discovered. Do you have steps to reproduce?

winklehner:
In my case its always caused when the server sends data to the client, but the client stops receiving or disconnects during transfer. I will try to build up a test case.

nickbabcock:
Ah ok that makes sense. We don't have any tests for that! I can see that functionality slipping through the cracks.

winklehner:
Great, than I can lean back and wait for an update ;)

nickbabcock:
So I investigated this a bit today, and I'm not sure what the best solution is. The only two options are:

- Keep the current behavior
- Log the behavior under a different level

The reason being is that by the time the client has disconnected, the response code (and potentially body) have already been committed (this would change if one used http trailers, but afaik jersey doesn't support that feature). Since the response is already committed it won't hit an exception mapper.

Aside: The good news is that I can replicate that exception by eschewing an http client and closing a `Socket` as soon as the request is sent

winklehner:
[...]

PR #3372, opened 2020-07-17
Swallow EofException when response was incomplete

joschi:
If a client drops the connection before a response was completely sent, Jetty will emit an `org.eclipse.jetty.io.EofException` which is then caught by Jersey and logged.

This can quite noisy in the logs and should typically be swallowed instead of logged including the stack trace.

The introduced `EofExceptionWriterInterceptor` will catch `EofException` which are being thrown while writing and will swallow them (after logging on DEBUG level).

Refs #2482
Refs #2541
Refs eclipse/jetty.project#544

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/EofExceptionWriterInterceptor.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/EofExceptionWriterInterceptor.java
@@ -1,0 +1,45 @@
+package io.dropwizard.jersey.errors;
+
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.MetricRegistry;
+import org.eclipse.jetty.io.EofException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Priority;
+import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.ext.Provider;
+import javax.ws.rs.ext.WriterInterceptor;
+import javax.ws.rs.ext.WriterInterceptorContext;
+import java.io.IOException;
+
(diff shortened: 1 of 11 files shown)


dropwizard-jersey/src/main/java/io/dropwizard/jersey/errors/EofExceptionWriterInterceptor.java
```diff
@@ -0,0 +1,32 @@
...
+@Provider
+public class EofExceptionWriterInterceptor implements WriterInterceptor {
+    private static final Logger LOGGER = LoggerFactory.getLogger(EofExceptionWriterInterceptor.class);
+
+    @Override
+    public void aroundWriteTo(WriterInterceptorContext context) throws IOException, WebApplicationException {
+        try {
+            context.proceed();
+        } catch (EofException e) {
+            LOGGER.debug("Client disconnected while processing and sending response", e);
```
joschi, line 29: We could also add a metric for this if it's regarded as useful.

bripkens: Having this as a metric could be useful indeed. I can imagine that a higher frequency of these exceptions might be an indicator for an overloaded HTTP server.

mattnelson:
@joschi Could this be patched to the 1.3.x line as well?

joschi:
@mattnelson The version of Jersey in Dropwizard 1.3.x is quite different from the one in Dropwizard 2.0.x, so it's not done by just cherry-picking the changes.

I think that it could work with Jersey 2.25.1 (used in Dropwizard 1.3.x) the same way.

Would you be able to prepare a pull request against the `release/1.3.x` branch? I'll gladly review it then.

PR #3379, opened 2020-07-19
Removed some conditional tests (test smell)

leofernandesmo:
###### Problem:
Conditions within the test method will alter the behavior of the test and its expected output, and would lead to situations where the test fails to detect defects in the production method since test statements were not executed as a condition was not met.

###### Solution:
Instead of using a conditional to control the test execution, we use the proper JUnit API with assumeTrue().

###### Result:
Before:
if("en".equals(Locale.getDefault().getLanguage()){...}
After:
assumeTrue("en".equals(Locale.getDefault().getLanguage()),
                "This test executes when the defined language is English ('en'). If not, it is skipped.");

### DIFF
--- a/dropwizard-validation/src/test/java/io/dropwizard/validation/DataSizeValidatorTest.java
+++ b/dropwizard-validation/src/test/java/io/dropwizard/validation/DataSizeValidatorTest.java
@@ -11,6 +11,7 @@ import java.util.List;
 import java.util.Locale;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 class DataSizeValidatorTest {
     @SuppressWarnings("unused")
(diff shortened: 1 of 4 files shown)


joschi:
@leofernandesmo Thank you for your contribution!

I'm wondering if we need to check the locale in the first place since we're running the tests (at least when using Maven Surefire) with an American English locale:
https://github.com/dropwizard/dropwizard/blob/03f77d039f1e74ff7856bb815cd31d1e49cab542/dropwizard-parent/pom.xml#L29

What do you think?

leofernandesmo:
Hi @joschi ,
If we can guarantee the user.language is always "en" in the execution of the tests, then I think we should remove this conditional at all. 
- One question, when I run the tests using any IDE, do I activate the Maven Surefire in the same way?

joschi:
> If we can guarantee the user.language is always "en" in the execution of the tests, then I think we should remove this conditional at all.

These system properties are always set when running the tests via Maven, but if the user's IDE doesn't delegate test execution to Maven, this won't work.

> One question, when I run the tests using any IDE, do I activate the Maven Surefire in the same way?

At least IntelliJ IDEA supports delegating test execution to the build system:
https://www.jetbrains.com/help/idea/delegate-build-and-run-actions-to-maven.html#delegate_to_maven

This being said, I think we cannot reasonably assume that the system properties have been set. 😞

PR #3382, opened 2020-07-21
Swallow EofException when response was incomplete

mattnelson:
Back patch of #3372 to the 1.3.x line.

Creating a draft PR to start a discussion in the context of the proposed changes.

When using the version of jetty currently defined in the dropwizard-bom the test returns with a 500 error. The `jersey-test-framework-provider-jetty` appears to only work with the version of jetty it was originally compiled against[1]. I was unable to get any meaningful logs/diagnostics out of jersey/jetty to further troubleshoot what could be the cause.

Looking for additional guidance on how to proceed.

[1] https://github.com/jersey/jersey/blob/2.25.1/pom.xml#L1919


joschi:
@mattnelson Does it work with the latest version of Jetty 9.2.x (`9.2.30.v20200428`)?

I don't feel very comfortable "downgrading" from Jetty 9.4.x to Jetty 9.2.x when it might break existing user code.

If the problem is only the Jersey testing framework, we might be able to live without a test for this particular feature, given that Dropwizard 1.3.x is (or should be) receiving only critical patches.

joschi:
I've found the underlying issue/exception:
```
java.lang.NoSuchMethodError: 'org.eclipse.jetty.server.Response org.eclipse.jetty.server.Response.getResponse(javax.servlet.http.HttpServletResponse)'
```
The Jersey Test Framework for Jetty is incompatible with Jetty 9.4.x because of this missing method.

As I said before, if the feature works in general I think we can get along without an integration test for it.

mattnelson:
> I don't feel very comfortable "downgrading" from Jetty 9.4.x to Jetty 9.2.x when it might break existing user code.

I agree. I only did that to verify that is was an issue with the dependencies instead of the code being added.

> As I said before, if the feature works in general I think we can get along without an integration test for it.

I have removed the failing test in 4ab31b7

joschi:
Travis CI seems to be failing because of an unrelated issue (Java 8u252 breaking the HTTP/2 integration).

mattnelson:
@joschi Have you had a chance to review?

joschi (approved):
LGTM. Thanks!

PR #3411, opened 2020-08-13
Use SslContextFactory.Server over deprecated SslContextFactory

MatthewRyanRead:
###### Problem:

Same as #3369:

>When we pin a more recent version of Jetty in our projects, services fail to start up with SSL with this error:
>
>```
>java.lang.IllegalStateException: KeyStores with multiple certificates are not supported on the base class org.eclipse.jetty.util.ssl.SslContextFactory. (Use org.eclipse.jetty.util.ssl.SslContextFactory$Server or org.eclipse.jetty.util.ssl.SslContextFactory$Client instead)
>```

###### Solution:

Exactly the title.  Does not bump the Jetty version, but allows users to pin a later version if they so desire without any conflicts.

This is hopefully a more palatable approach than that taken in #3369 &mdash; it's definitely much simpler.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/sslreload/SslReloadTask.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/sslreload/SslReloadTask.java
@@ -21,7 +21,7 @@ public class SslReloadTask extends Task {
     public void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
         // Iterate through all the reloaders first to ensure valid configuration
         for (SslReload reloader : getReloaders()) {
-            reloader.reload(new SslContextFactory());
+            reloader.reload(new SslContextFactory.Server());
         }
 
         // Now we know that configuration is valid, reload for real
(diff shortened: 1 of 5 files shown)


MatthewRyanRead:
```
[ERROR] testHttp2WithCustomCipher(io.dropwizard.http2.Http2WithCustomCipherTest)  Time elapsed: 2.753 s  <<< ERROR!
java.util.concurrent.ExecutionException: java.lang.NoSuchFieldError: nullSession
	at io.dropwizard.http2.Http2WithCustomCipherTest.testHttp2WithCustomCipher(Http2WithCustomCipherTest.java:33)
Caused by: java.lang.NoSuchFieldError: nullSession
```

Can't figure out the cause of this ... once again, tests pass just fine locally.  Something with the travis-ci environment maybe?

MatthewRyanRead:
Ah, Googling `java.lang.NoSuchFieldError: nullSession` is revealing.  It's an SSL issue with the 8u252 JDK.  Apparently newer versions of Jetty work around that, so this PR won't be sufficient.

Looks like we might just need to fork this.  Closing!

black-snow:
Might be related to this change, after upgrading from 1.3.23 to 1.3.25 I receive:

```
java.lang.NoClassDefFoundError: org/eclipse/jetty/server/SecureRequestCustomizer$SslAttributes
```

PR #3442, opened 2020-09-02
Support parsing system property configuration values as arrays when the field does not exist

dhenneke:
###### Problem:

This PR resolves the following limitation when overriding configurations by System Properties:

> The array override facility only handles configuration elements that are arrays of simple strings. *Also, the setting in question must already exist in your configuration file as an array; this mechanism will not work if the configuration key being overridden does not exist in your configuration file. If it does not exist or is not an array setting, it will get added as a simple string setting, including the ‘,’ characters as part of the string.*

We ship our services in Docker Containers to be run in our customer's infrastructure and want to give them the possibility to configure rarely used configurations with System Properties. This includes for example the proxy settings of HTTP Clients. We don't usually configure this in our `config.yml` and don't want them to exchange it with their own. It is usually infrastructure-dependent and a manual exchange will possibly lead to compatibility problems in future updates of a service. But the setting `nonProxyHosts` is not settable when it is not defined in the `config.yml`, since the parser doesn't recognize the field as being an array/collection.

###### Solution:

We propose to add a `ConfigurationMetadata` class that discovers the type of all fields and can be used while parsing the overridden properties. For now, it is only used to check whether a missing node should actually be an array.

This doesn't work well with inheritance in the configuration file (like the `ServerFactory` class in the default `Configuration`). The parser only works when a `defaultImpl` is registered. If it is missing, it can't discover any further properties.

Do you think that would be a valuable addition to the existing logic?

### DIFF
--- a/dropwizard-configuration/pom.xml
+++ b/dropwizard-configuration/pom.xml
@@ -68,6 +68,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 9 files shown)


joschi (approved):
@dhenneke Thank you very much for your contribution!

LGTM.

PR #3453, opened 2020-09-11
Reset Jersey client in tests

huangyq23:
###### Problem:
Junit5 jersey client test is broken in 1.3.x due to reset clients

###### Solution:
Already fixed in master, backport fixes in https://github.com/dropwizard/dropwizard/pull/2764


joschi (approved):
@huangyq23 LGTM, thanks!

PR #3469, opened 2020-09-29
#3468 - Add badge for reproducible build

### LINKED ISSUE #3468: Add badge for reproducible build
Add a new badge for [Reproducible Central project](https://github.com/jvm-repo-rebuild/reproducible-central) to follow the status.

As dropwizard-core 2.0.10 has been verified as reproducible and versions 2.0.11 and 2.0.12 are going to be added in the [Pull Request #37 of jvm-repo-rebuild/reproducible-central](https://github.com/jvm-repo-rebuild/reproducible-central/pull/37)

dialaya:
###### Problem:
Since v2.0.10, [dropwizard-core](https://github.com/jvm-repo-rebuild/reproducible-central#io.dropwizard:dropwizard-core) is been validated as a reproducible build in [Reproducible Central project](https://github.com/jvm-repo-rebuild/reproducible-central)
But who knows?

###### Solution:
A badge in documentation might be good to follow this information. Then users can see directly from the dropwizard repository.

###### Result:
A badge with the "Reproducible build" in status "OK" will be added.

Closes #3468

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ Dropwizard
 [![Javadocs](https://javadoc.io/badge/io.dropwizard/dropwizard-project.svg?color=brightgreen)](https://javadoc.io/doc/io.dropwizard/dropwizard-project)
 [![Documentation Status](https://readthedocs.org/projects/dropwizard/badge/?version=stable)](https://www.dropwizard.io/en/stable/?badge=stable)
 [![Maintainability](https://api.codeclimate.com/v1/badges/11a16ea08c8b5499e2b9/maintainability)](https://codeclimate.com/github/dropwizard/dropwizard/maintainability)
+[![Reproducible Builds](https://img.shields.io/badge/Reproducible_Builds-ok-green?labelColor=blue)](https://github.com/jvm-repo-rebuild/reproducible-central#io.dropwizard:dropwizard-core)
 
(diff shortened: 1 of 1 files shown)


joschi:
@dialaya It looks like this is a static image. Would it ever change from "ok" to something else if Dropwizard didn't provide reproducible builds anymore or if we accidentally introduced a regression which led to snowflake builds again?

bhamail:
@joschi I think the idea is that once a reproducible version has been validated, it is immutable, and will always be available. Therefore the static badge sort of makes sense. If a newer version fails validation, the older ones will still be valid. Also, the link to the particular artifact (e.g. dropwizard-core) will still work, and show users all the versions, even if a newer one hasn't yet passed validation.

dialaya:
@joschi I understand how you would like to proceed... But as @bhamail said, Reproducible is immutable for a version.
The badge is a simple static link to dropwizard known versions in this project.
On [Reproducible Central project](https://github.com/jvm-repo-rebuild/reproducible-central) side they will list validated or certified versions.
If you would like to add new version, you can (may be) add, in your build config, a step to verify reproducibility and create a pull request.
Adding "OK" instead of the list of knows versions look simple for me.

joschi (approved):

joschi:
@dialaya Fair enough. Thanks for your contribution and the explanation!

PR #3471, opened 2020-09-30
Remove deprecated "optimize" compiler flag

### LINKED ISSUE #3466: Remove deprecated "optimize" compiler flag
I just got distracted by a compiler flag in the parent pom, so I thought I'd file an issue here.

"optimize" is a deprecated compiler flag - [see here](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#optimize).

... but it is still used in the parent POM:
* https://github.com/dropwizard/dropwizard/blob/master/dropwizard-parent/pom.xml#L25
* https://github.com/dropwizard/dropwizard/blob/master/dropwizard-parent/pom.xml#L87

matthias-mueller:
Fixes  #3466


arteam (approved):

arteam:
LGTM, thanks for the contribution!

PR #3480, opened 2020-10-03
Disable Checkstyle for Code Climate

pstackle:
Since checkstyle was removed in #2693 , disabling it in the codeclimate config in order to fix the badge


joschi (approved):
Thanks, Peter! 😃

PR #3482, opened 2020-10-04
Hacktoberfest: Simplify conditional experession

Sangaibisi:
###### Problem:
Hard to read and understand if expression.

###### Solution:
Simpfiable conditional expressions. More readable code.


joschi (approved):
@Sangaibisi Thanks!

PR #3489, opened 2020-10-06
Simplify conditional expression

elifnurcevik:
Problem
The conditional expression is not easy to read.

Solution:
Simplify conditional expression

Result:
made it easier to read

This commitment supports to open source community for **Hacktoberfest**.


jplock:
I think the way it was written had a clearer intent.

elifnurcevik:
> I think the way it was written had a clearer intent.

The important thing is that anyone who does maintenance can easily read the code. Nobody can read intent.

elifnurcevik:
Hi @joschi , @arteam what do you think about this pull request.

joschi:
@elifnurcevik Thanks for your contribution.

I would agree with @jplock here that the original version was more straight forward to read.

Issue #3526, opened 2020-10-26
NonBlockingServletHolder still necessary?


josephlbarnett:
Noticed that latest jetty release (9.4.33) was not binary compatible with latest dropwizard release (2.0.14) and filed an issue with jetty to let them know they've broken binary (but not source) compatibility.  They are not going to revert back, but did question if NonBlockingServletHolder is still necessary, more details here:

https://github.com/eclipse/jetty.project/issues/5496#issuecomment-716398885

PR #3527, opened 2020-10-26
Remove obsolete NonblockingServletHolder

joschi:
Remove `NonblockingServletHolder` class which, according to developers of Jetty, isn't required anymore.

Refs https://github.com/eclipse/jetty.project/issues/5496
Refs #3526


joschi:
In order to safely merge this PR, we should add test cases which would trigger the original error otherwise.

joschi:
I'm still thinking about how to make this change in a backward-compatible way.

Maybe just keep `NonblockingServletHolder` as an unused class in `release/2.0.x` in case anyone is using it?
Otherwise, the changes are only affecting internals and could in my opinion go into `release/2.0.x`.

@dropwizard/committers What's your opinion?

pkwarren (approved):

pstackle (approved):
@joschi 

I'm having a hard time coming up with a use case where someone may have been using `NonblockingServletHolder` directly. I think I'd be ok with removing the class completely rather than keep it around as an unused class. If you'd rather keep it as a `@Deprecated` "just in case", I'm ok with that, too.

I agree that this change can go in release/2.0.x.

arteam (approved):

dropwizard-jetty/src/main/java/io/dropwizard/jetty/NonblockingServletHolder.java
```diff
@@ -13,7 +13,12 @@
 /**
  * A {@link ServletHolder} subclass which removes the synchronization around servlet initialization
  * by requiring a pre-initialized servlet holder.
+ *
+ * @deprecated If necessary, use {@link ServletHolder} or {@link org.eclipse.jetty.servlet.FilterHolder} directly.
+ * This class will be removed in Dropwizard 2.1.0.
```
josephlbarnett, line 18: maybe remove the `handle()` override to avoid the jetty incompatibility in anything that's using this directly? 
(maybe remove all the overrides and just have this essentially act as an alias for ServletHolder?)

https://github.com/search?q=NonblockingServletHolder&type=code shows some usages that are not just forks of DW, fwiw

joschi: > maybe remove the handle() override to avoid the jetty incompatibility in anything that's using this directly?

This wouldn't be a problem in Dropwizard 2.0.15 which is being built against Jetty 9.4.33, would it?

josephlbarnett: probably not unless someone tries to run DW 2.0.15 / jetty 9.4.32?  or future breakage in 9.4.34+?  not sure how much work is worth trying to protect against unknowns like that though.

PR #3530, opened 2020-10-27
Fix logFormat configuration inconsistencies

### LINKED ISSUE #3481: logFormat configuration inconveniences
There are two non-obvious limitations in the appender configuration:

1) When using the **layout** option to set a custom `LayoutFactory`, the **logFormat** is always ignored. I think it would be more consistent to apply the **logFormat** to all layouts that derive from `PatternLayoutBase` and otherwise emit a warning if both options are set. That is, changing `AbstractAppenderFactory.buildLayout()` to something like

```java
        if (layout == null) {
            layoutBase = defaultLayoutFactory.build(context, timeZone);
        } else {
            layoutBase = layout.build(context, timeZone);
        }
        if (!Strings.isNullOrEmpty(logFormat)) {
            if (layoutBase instanceof PatternLayoutBase) {
                ((PatternLayoutBase<E>)layoutBase).setPattern(logFormat);
            } else {
                LOGGER.warn("Ignoring 'logFormat', because 'layout' does not extend PatternLayoutBase");
            }
        }
```

2) When setting a **logFormat** that includes time zone information (like the default `LogbackAccessRequestLayout`), the time zone is a fixed part of the string and may contradict the configured **timeZone** option. Maybe the **logFormat** could contain some kind of placeholder (e.g. `<timeZone>`) that gets automatically replaced with the correct value from the **timeZone** option?

cleiner:
###### Problem:
As described in #3481, there are two limitations in the appender configuration:

1. When using the **layout** option to set a custom `LayoutFactory`, the **logFormat** is always ignored.
2. The time zone information in a custom **logFormat** is a fixed part of the string and may contradict the configured **timeZone** option.

###### Solution:
1. `AbstractAppenderFactory` applies the **logFormat** to any `LayoutFactory` that derives from `PatternLayoutBase`, not just to the default factory. If a **logFormat** is set along with a non-pattern **layout**, a warning will be logged.
2. The introduction of the placeholder `%dwTimeZone` allows the time zone information in a custom log format to depend on the value of the **timeZone** option.

Closes #3481

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
@@ -266,7 +266,6 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
         }
     }
 
-    @SuppressWarnings("NullAway")
     protected LayoutBase<E> buildLayout(LoggerContext context, LayoutFactory<E> defaultLayoutFactory) {
         final LayoutBase<E> layoutBase;
         if (layout == null) {
@@ -276,6 +275,7 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
         }
         if (!Strings.isNullOrEmpty(logFormat)) {
             if (layoutBase instanceof PatternLayoutBase) {
+                @SuppressWarnings("NullAway")
                 String logFormatWithTimeZone = logFormat.replace("%dwTimeZone", timeZone.getID());
                 ((PatternLayoutBase<E>)layoutBase).setPattern(logFormatWithTimeZone);
             } else {
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
@@ -19,6 +19,8 @@ import io.dropwizard.util.Strings;
 import io.dropwizard.util.Duration;
 import io.dropwizard.validation.MaxDuration;
 import io.dropwizard.validation.MinDuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import javax.validation.constraints.Max;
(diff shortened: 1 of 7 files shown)


dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
```diff
@@ -263,17 +266,22 @@ public void setLayout(@Nullable DiscoverableLayoutFactory<E> layout) {
         }
     }
 
+    @SuppressWarnings("NullAway")
```
joschi, line 269: I think it would be enough to add this suppression to line 279:
```
                 @SuppressWarnings("NullAway")
                 String logFormatWithTimeZone = logFormat.replace("%dwTimeZone", timeZone.getID());
```

cleiner: Yes, moved it. I was torn on how to fix this one. Thought the method annotation is more readable, even though a bit too broad.

cleiner:
@joschi The Java 11 build repeatedly failed with
```
Error:  Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.7.0.1746:sonar (default-cli) on
project dropwizard-project: You're not authorized to run analysis. Please contact the project administrator. -> [Help 1]
```
This doesn't seem to be something I can fix. Would you take a look?

Also, today's AppVeyor failure looks like a timing dependent problem unrelated to the PR:
```
java.lang.NullPointerException
	at io.dropwizard.logging.json.LayoutIntegrationTests.testLogJsonToConsole(LayoutIntegrationTests.java:156)
```
In the test, there's a `Thread.sleep(100)` on line 152 and then a `objectMapper.readTree(redirectedStream.toString())` -- if the stream is still empty after the sleep, Jackson will happily parse the empty string as a `JsonNode` of type MISSING which then causes the NPE in the call to `isTextual()` on line 156.

joschi:
> This doesn't seem to be something I can fix. Would you take a look?

This has been fixed in #3535. You can either rebase your branch on our latest `master` branch or ignore that particular error for now. 😃

cleiner:
Ah, ok :) Any concerns about squashing the two SuppressWarnings commits while I'm at it?

And another question: Want me to put a loop around that `Thread.sleep()` in `testLogJsonToConsole` to make it less flaky? (in a separate PR)

joschi:
> Any concerns about squashing the two SuppressWarnings commits while I'm at it?

Feel free to do so, if you want. Otherwise we'll squash the commits when merging the PR.

> Want me to put a loop around that `Thread.sleep()` in `testLogJsonToConsole` to make it less flaky? (in a separate PR)

We're already using [Awaitility](http://www.awaitility.org/) in other tests and we could also use it here to wait for the conditions to turn true instead of `Thread#sleep()`.

joschi (approved):
LGTM. Thanks!

PR #3533, opened 2020-10-28
Replace invalid default excludedProtocols in HttpsConnectorFactory

### LINKED ISSUE #3532: Dropwizard reporting incorrectly enabled HTTPS protocols
The deafult Dropwizard behaviour seems to enable TLSv1.1 (despite the documentation stating that TLSv1.1 is in the excludedProtocols by default). It does also not match what is bieng logged by Dropwizard.

Steps to reproduce (using Dropwizard v.2.0.13):
1. Create a simple Dropwizard application. Add HTTPs configuration (do not set any values for supportedProtocols and excludedProtocols)
2. Start up server. Server logs indicate:
`io.dropwizard.jersey.HttpsConnectorFactory: Enabled protocols: [TLSv1.2, TLSv1.3]
io.dropwizard.jersey.HttpsConnectorFactory: Disabled protocols: [SSLv2Hello, SSLv3, TLSv1, TLSv1.1]`
3. `curl --tlsv1.1 https://localhost` - this actually works (using curl version: libcurl/7.29.0, NSS/3.44)

Further proof:
- Add a `ServerLifecycleListener` with the `serverStarted` method as: `server -> LOGGER.info(Arrays.asList(server.getBean(SslContextFactory.class).getSelectedProtocols()))`
- This will print TLSv1.1, TLSv1.2 and TLSv1.3

Notes:
- If I explicitly set excludedProtocols with TLSv1.1, it gets disabled.
- TLS 1.1 may still fail due to the default cipher suites prevent an algorithm agreement - but that is besides the point - the protocol is still enabled.

joschi:
The default list of excluded protocols used in `HttpsConnectorFactory` wasn't working as expected.

Jetty currently doesn't support using regular expressions for supported or excluded protocols.
This is only working for supported and excluded _cipher suites_ as of Jetty 9.4.33.v20201020.

The default list of excluded protocols now only contains valid and complete entries:

SSLv3, TLSv1, and TLSv1.1

Refs https://github.com/eclipse/jetty.project/issues/5531
Fixes #3532

### DIFF
--- a/dropwizard-jetty/src/test/java/io/dropwizard/jetty/HttpsConnectorFactoryTest.java
+++ b/dropwizard-jetty/src/test/java/io/dropwizard/jetty/HttpsConnectorFactoryTest.java
@@ -44,18 +44,18 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assumptions.assumeFalse;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
-public class HttpsConnectorFactoryTest {
+class HttpsConnectorFactoryTest {
     private static final String WINDOWS_MY_KEYSTORE_NAME = "Windows-MY";
     private final Validator validator = BaseValidator.newValidator();
 
     @Test
-    public void isDiscoverable() throws Exception {
+    void isDiscoverable() {
         assertThat(new DiscoverableSubtypeResolver().getDiscoveredSubtypes())
                 .contains(HttpsConnectorFactory.class);
     }
 
     @Test
-    public void testParsingConfiguration() throws Exception {
+    void testParsingConfiguration() throws Exception {
         HttpsConnectorFactory https = new YamlConfigurationFactory<>(HttpsConnectorFactory.class, validator,
                 Jackson.newObjectMapper(), "dw-https")
                 .build(new File(Resources.getResource("yaml/https-connector.yml").toURI()));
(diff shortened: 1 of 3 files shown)


pkwarren (approved):

dropwizard-jetty/src/test/java/io/dropwizard/jetty/HttpsConnectorFactoryTest.java
```diff
@@ -114,15 +114,33 @@ public void testExcludedProtocols() {
...
+
+        SslContextFactory sslContextFactory = factory.configureSslContextFactory(new SslContextFactory.Server());
+        assertThat(sslContextFactory.getExcludeProtocols())
+                .containsExactlyElementsOf(factory.getExcludedProtocols());
+
+        sslContextFactory.start();
+        try {
+            assertThat(sslContextFactory.newSSLEngine().getEnabledProtocols())
+                    .doesNotContainAnyElementsOf(factory.getExcludedProtocols())
+                    .allSatisfy(protocol -> assertThat(protocol).doesNotStartWith("SSL"));
```
pkwarren, line 128: Do we also want to ensure that enabled protocols only includes TLS v1.2/v1.3 and nothing else? Feels like a good test to fail if any older protocols are enabled.

joschi: I also thought about adding this, but then the test will start failing when run on JVMs which support other (newer) protocols in the future.

Or maybe some JVM only supports up to TLSv2. 🤔

arteam (approved):

PR #3537, opened 2020-11-01
Reduce flakiness of LayoutIntegrationTests

cleiner:
###### Problem:
Got random build failures while working on #3530 due to two tests in `LayoutIntegrationTests` waiting for log output using a one-off call to `Thread.sleep()`. In addition to being brittle, the tests will fail with an NPE (while trying to assert on the "timestamp" field) that doesn't point to the cause of the failure directly.

###### Solution:
Using Awaitility instead of `Thread.sleep()` which will periodically check if either the log output is present or a timeout has been reached.


joschi (approved):
LGTM, thanks!

PR #3561, opened 2020-11-09
Add ConfigOverride for random application ports during tests

dhenneke:
###### Problem:

We are doing a lot of integration testing where we need to start the application. We primarily use the config overrides instead of a `test-config.yaml` to do the configuration. Our config files of the test module often only contain the configuration for the random ports:
```yaml
server:
  applicationConnectors:
    - type: http
      port: 0
  adminConnectors:
    - type: http
      port: 0
```

It would be nice to be able to use random ports with the `DropwizardAppRule`/`DropwizardAppExtension` programmatically.

###### Solution:

Add a Config Override that configures the default application and admin connectors to use port `0`.

```java
private static final DropwizardAppExtension<MyConfiguration> EXTENSION =
    new DropwizardAppExtension<>(MyApplication.class,
        // no config file
        null,
        // adds all necessary config overrides for random ports
        randomPorts());
```

I refactored the `ConfigOverride` class to an abstract class and added two implementations: `ConfigOverrideValue` (the existing behavior) and `ConfigOverrideRandomPorts` (the new option). Both are instanciated from static functions in `ConfigOverride`. This change would also allow application developers to build reusable grouped config overrides if needed.

Do you think this is a valid extension? I am happy to discuss the changes and do changes if needed.

### DIFF
--- a/dropwizard-testing/src/test/java/io/dropwizard/testing/junit5/DropwizardAppExtensionRandomPortsConfigOverrideTest.java
+++ b/dropwizard-testing/src/test/java/io/dropwizard/testing/junit5/DropwizardAppExtensionRandomPortsConfigOverrideTest.java
@@ -1,0 +1,39 @@
+package io.dropwizard.testing.junit5;
+
+import static io.dropwizard.testing.ConfigOverride.config;
+import static io.dropwizard.testing.ConfigOverride.randomPorts;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.dropwizard.jetty.HttpConnectorFactory;
+import io.dropwizard.server.DefaultServerFactory;
+import io.dropwizard.testing.app.TestApplication;
+import io.dropwizard.testing.app.TestConfiguration;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(DropwizardExtensionsSupport.class)
+class DropwizardAppExtensionRandomPortsConfigOverrideTest {
+
+    private static final DropwizardAppExtension<TestConfiguration> EXTENSION =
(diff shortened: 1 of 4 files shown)


dropwizard-testing/src/test/java/io/dropwizard/testing/junit5/DropwizardAppExtensionRandomPortsConfigOverrideTest.java
```diff
@@ -0,0 +1,28 @@
...
+            null,
+            "app-rule",
+            randomPorts("app-rule"),
+            config("app-rule", "message", "A new way to say Hooray!"),
+            config("app-rule", "extra", () -> "supplied"));
+
+    @Test
+    void supportsConfigAttributeOverrides() {
+        assertThat(EXTENSION.getLocalPort()).isNotEqualTo(8080);
+        assertThat(EXTENSION.getAdminPort()).isNotEqualTo(8081);
```
joschi, line 26: This is bound to be a flaky test if you run it often enough.

Shouldn't it suffice to check that both methods return an integer in the range of an IP port (0-65535)?

dhenneke: I think it would also be better if I just check if the `Configuration` class has the expected content.

dhenneke: I changed the test to not test the actual ports but just if the `HttpConnectorFactory`s were configured correctly.

joschi:
@dhenneke Thank you for your contribution! 👍 

I've merged the PR with some minor changes.

PR #3567, opened 2020-11-15
Tighten up commons-lang3 dependencies

rhowe:
###### Problem:
I noticed `dropwizard-logging` included a dependency on `commons-lang3` and thought to myself "I bet it doesn't really need that"

###### Solution:
I've discovered `commons-lang3` is mostly only used in tests, so I constrained some of the dependencies to `test` scope where possible.

###### Result:
`dropwizard-client`, `dropwizard-jetty`, `dropwizard-logging` and `dropwizard-migrations` will no longer bundle or require `commons-lang3`

In fact commons-lang3 is only used in the following non-test code:

- dropwizard-benchmarks/src/main/java/io/dropwizard/benchmarks/jersey/ConstraintViolationBenchmark.java
- dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
- dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
- dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintMessage.java
- dropwizard-views/src/main/java/io/dropwizard/views/ViewRenderExceptionMapper.java

Almost entirely for doing reflective things

### DIFF
--- a/dropwizard-client/pom.xml
+++ b/dropwizard-client/pom.xml
@@ -82,10 +82,6 @@
             <artifactId>jersey-common</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-lang3</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
         </dependency>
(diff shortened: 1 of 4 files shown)


jplock (approved):
LGTM

joschi (approved):

PR #3580, opened 2020-11-26
Hibernate: DualSessionFactory

smalleyd2:
###### Problem:
Modern RDBMS systems often provide read-replica clusters with separate access points. With regards to Hibernate, this requires two SessionFactory instances. One factory is for the primary/writer data source. And a separate factory is necessary for the read-replica data source. This requires the usage of the UnitOfWork name property whenever the read-replica is preferred. Ideally, no adjustments to the UnitOfWork properties should be necessary if the readOnly is properly set across endpoints.

###### Solution:
This commit provides a SessionFactory implementation that contains both a writer/primary factory and a reader factory. The DualSessionFactory comes with separate HibernateBundle, UnitOfWorkApplicationListener, and UnitOfWorkAspect classes. Prior to usage, the UnitOfWorkAspect class prepares the DualSessionFactory based on the UnitOfWork annotation readOnly property. When the dual HibernateBundle is dropped into an application, the UnitOfWork requires no changes across endpoints assuming that the readOnly property is already set accordingly.

###### Result:
This commit contains new DualSessionFactory, HibernateBundle, UnitOfWorkApplicationListener, and UnitOfWorkAspect classes.


joschi:
@dropwizard/committers Anybody available to review these changes?

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/dual/DualSessionFactory.java
```diff
@@ -0,0 +1,171 @@
...
+
+import java.sql.Connection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** Represents a wrapper/decorator class for a Hibernate session factory that can manage
+ *  both a primary session factory and a read-only session factory.
+ *
+ * @author smalleyd2
```
jplock, line 35: Please remove the author and date entries. I think @since can specify a version

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/dual/UnitOfWorkApplicationListener.java
```diff
@@ -0,0 +1,109 @@
...
+import javax.ws.rs.ext.Provider;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/** Jersey server listener that wraps dual Hibernate session factories (writer & reader).
+ * 
+ * @author smalleyd
```
jplock, line 21: Remove author and date and change @since to the version

dropwizard-hibernate/src/test/java/io/dropwizard/hibernate/dual/DualSessionFactoryTest.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/** Unit test class that verifies the DualSessionFactory wrapper.
+ * 
+ * @author smalleyd
```
jplock, line 19: Remove author and the date and change @since to the version

jplock (approved):
I think this is a good idea overall, but I’m not set up to test whether it is functional or not.

lakshmansai:
can we check whether we can use MySQL Replication Driver and use unit of work method interceptor to mark connection read only

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/dual/HibernateBundle.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    abstract public PooledDataSourceFactory getReadSourceFactory(T configuration);
+
+    @Override
+    public final void run(T configuration, Environment environment) throws Exception {
+        final PooledDataSourceFactory primaryConfig = getDataSourceFactory(configuration);
+        final SessionFactory primary = requireNonNull(sessionFactoryFactory.build(this, environment, primaryConfig,
+            entities, name()));
+        final PooledDataSourceFactory readerConfig = getDataSourceFactory(configuration);
```
zUniQueX, line 44: Shouldn't this call `getReadSourceFactory(configuration)`?

zUniQueX, line 44: Shouldn't this call `getReadSourceFactory(configuration)`?

smalleyd2: Yes, thank you for the catch. It's now fixed on the feat/3542 branch.

zUniQueX:
I think, that this should work now. But I have a few remarks on the current implementation:
 1. Is it a good idea to override the `name()` method in the `dual.HibernateBundle`? If done so, one has to provide a value to the `@UnitOfWork` annotation every time, even if only one Hibernate bundle is registered. To resolve this, you could revert to `io.dropwizard.hibernate.HibernateBundle.DEFAULT_NAME` in line 55 of `UnitOfWorkAspect`.
 2. It seems to me, that this code is less maintainable, since there is some identical code in the `dual` package and the existing classes. In a seperate test, I managed to use the existing classes with a dynamic proxy. However, dynamic proxies can get ugly fast and I don't know, if this works well in all situations. If you want to look at this approach anyway, you can check it out in [my test](https://github.com/zUniQueX/dropwizard/compare/master...zUniQueX:dual-session-factory).

smalleyd2:
> I think, that this should work now. But I have a few remarks on the current implementation:
> 
> 1. Is it a good idea to override the `name()` method in the `dual.HibernateBundle`? If done so, one has to provide a value to the `@UnitOfWork` annotation every time, even if only one Hibernate bundle is registered. To resolve this, you could revert to `io.dropwizard.hibernate.HibernateBundle.DEFAULT_NAME` in line 55 of `UnitOfWorkAspect`.
> 2. It seems to me, that this code is less maintainable, since there is some identical code in the `dual` package and the existing classes. In a seperate test, I managed to use the existing classes with a dynamic proxy. However, dynamic proxies can get ugly fast and I don't know, if this works well in all situations. If you want to look at this approach anyway, you can check it out in [my test](https://github.com/zUniQueX/dropwizard/compare/master...zUniQueX:dual-session-factory).

@zUniQueX zUniQueX, I've modified the dual.HibernateBundle to properly implement the inherited "name" method. And I made some tweaks to take greater advantage of the existing base HibernateBundle methods to reduce code duplication. Thanks for the continued feedback.

cue108:
Couldn't this feature be removed from milestone 2.1.0 and finally release it without it?

joschi (approved):
@smalleyd2 Thanks for your contribution!

Let's merge it and polish in `release/2.1.x` if necessary.

PR #3582, opened 2020-11-30
fix flaky test

jlee759:
###### Problem:
Fixing Flaky Tests: io.dropwizard.logging.json.layout.MapBuilderTest.testIncludeStringValue
The test running order is not preserved in junit so using containsOnly will make this test flaky if other tests are ran before this.

###### Solution:
Change .containsOnly => .contains

###### Result:
io.dropwizard.logging.json.layout.MapBuilderTest.testIncludeStringValue will not be flaky

### DIFF
--- a/dropwizard-json-logging/src/test/java/io/dropwizard/logging/json/layout/MapBuilderTest.java
+++ b/dropwizard-json-logging/src/test/java/io/dropwizard/logging/json/layout/MapBuilderTest.java
@@ -18,7 +18,7 @@ public class MapBuilderTest {
     @Test
     public void testIncludeStringValue() {
         assertThat(mapBuilder.add("message", true, message).build())
-            .containsOnly(entry("message", message));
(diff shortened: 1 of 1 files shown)


joschi:
> The test running order is not preserved in junit so using containsOnly will make this test flaky if other tests are ran before this.

@jlee759 This is working perfectly fine due to the default test instance lifecycle in JUnit 5:

> In order to allow individual test methods to be executed in isolation and to avoid unexpected side effects due to mutable test instance state, JUnit creates a new instance of each test class before executing each test method (see Test Classes and Methods). This "per-method" test instance lifecycle is the default behavior in JUnit Jupiter and is analogous to all previous versions of JUnit.

https://junit.org/junit5/docs/5.7.0/user-guide/#writing-tests-test-instance-lifecycle

Also please take into account that `contains()` and `containsOnly()` cannot be simply exchanged without changing the semantics of the test.

PR #3601, opened 2020-12-10
Fixed syntax in Testing Database Interactions example

mleegwt:
###### Problem:

Syntax in example was incorrect.

###### Solution:

Fixed syntax.

###### Result:

Example will be closer to a working test.


joschi (approved):
@mleegwt Thanks a lot! 😄

PR #3603, opened 2020-12-13
Remove 'Optionals' from dropwizard-util

rhowe:
###### Problem:
It doesn't look like this has been needed for some time, although I may well be missing something in all the Guava/DI wizardry.

It was introduced when `dropwizard-java8` was merged into the main dropwizard codebase in 3a05b916eff4c8ae0d766827b812a3437a2efb8a but it's not clear to me how it was actually used.

###### Solution:
Remove `io.dropwizard.util.Optionals`, its corresponding test and the now-unused guava dependency from `dropwizard-util`.

###### Result:
`dropwizard-util` will no longer contain `io.dropwizard.util.Optionals` for converting between Java 8 `Optional` and Guava `Optional`.
`dropwizard-util` will no longer include the following dependencies:
```
+- com.google.guava:guava:jar:30.0-jre:compile
|  +- com.google.guava:failureaccess:jar:1.0.1:compile
|  +- com.google.guava:listenablefuture:jar:9999.0-empty-to-avoid-conflict-with-guava:compile
|  +- org.checkerframework:checker-qual:jar:3.8.0:compile
|  +- com.google.errorprone:error_prone_annotations:jar:2.4.0:compile
|  \- com.google.j2objc:j2objc-annotations:jar:1.3:compile
```


joschi:

The `io.dropwizard.util.Optionals` is meant as a utility class to help users migrate from Guava's `Optional<T>` to Java's `Optional<T>` implementation.

Once we drop support for `com.google.common.base.Optional<T>`, we can think about dropping this utility class.

jplock:
@joschi @rhowe should we deprecate in 2.x and remove in 3.0.0?

joschi:
> @joschi @rhowe should we deprecate in 2.x and remove in 3.0.0?

@jplock Makes sense. I would deprecate it in Dropwizard 2.1.x (current `master` branch) and subsequently remove it in Dropwizard 3.0.x.

rhowe:
Gonna close this - it's marked deprecated and I assume there'll be a period in the runup to 3.0.0 where the stuff marked deprecated will get cleaned out

Issue #3615, opened 2020-12-23
Dropwizard emits an error message whenever external logging is configured and Logback is excluded


adriansuarez:
This can be reproduced on 2.0.16 by simply following the instructions in the [documentation](https://www.dropwizard.io/en/latest/manual/core.html#logging) in the _Note_ section beginning with _If you don’t want to use Logback ..._ The only symptom is that the following message is printed to standard error, but the server still starts up:

```
io.dropwizard.configuration.ConfigurationMetadata.$1: Type ch.qos.logback.access.spi.IAccessEvent not present
```

This occurs while Hibernate is validating the deserialized configuration at the field `server.requestLog`, which is configured with `type: external` as instructed in that _Note_ section.

It looks like the root of the problem is that `LogbackAccessRequestLogFactory` pulls in the Logback dependency which is being excluded, and it is also the default implementation of `RequestLogFactory`, as defined with the [`JsonTypeInfo` annotation](https://github.com/dropwizard/dropwizard/blob/05bdecbe59366d4c747aa2f0256dcd768bd6b6e6/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/RequestLogFactory.java#L12). So it is still being loaded when doing Hibernate validation on the config, even if it is not specified in the config.

This is the exception stacktrace that is getting swallowed when the message above is printed:

```
java.lang.TypeNotPresentException: Type ch.qos.logback.access.spi.IAccessEvent not present
	at sun.reflect.generics.factory.CoreReflectionFactory.makeNamedType(CoreReflectionFactory.java:117)
	at sun.reflect.generics.visitor.Reifier.visitClassTypeSignature(Reifier.java:125)
	at sun.reflect.generics.tree.ClassTypeSignature.accept(ClassTypeSignature.java:49)
	at sun.reflect.generics.visitor.Reifier.reifyTypeArguments(Reifier.java:68)
	at sun.reflect.generics.visitor.Reifier.visitClassTypeSignature(Reifier.java:138)
	at sun.reflect.generics.tree.ClassTypeSignature.accept(ClassTypeSignature.java:49)
	at sun.reflect.generics.visitor.Reifier.reifyTypeArguments(Reifier.java:68)
	at sun.reflect.generics.visitor.Reifier.visitClassTypeSignature(Reifier.java:138)
	at sun.reflect.generics.tree.ClassTypeSignature.accept(ClassTypeSignature.java:49)
[...]

PR #3616, opened 2020-12-23
Do not print error message about class not found

adriansuarez:
Only swallow NoClassDefFoundError if the default implementation
generated it, and do not print message to standard output about it.
Currently an error message appears whenever Logback (specifically the
logback-access dependency) is excluded, which users should be able to do
if they configuring logging externally.

###### Problem:

Dropwizard generates error output whenever Logback is excluded according to the instructions in the [documentation](https://www.dropwizard.io/en/latest/manual/core.html#logging). See the linked issue: https://github.com/dropwizard/dropwizard/issues/3615

###### Solution:

Avoid printing to standard error and only the swallow NoClassDefException generated by the default implementation.

###### Result:

Users who configure external logging will not see spurious error messages.

### DIFF
--- a/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationMetadata.java
+++ b/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationMetadata.java
@@ -106,7 +106,6 @@ public class ConfigurationMetadata extends JsonFormatVisitorWrapper.Base {
     }
 
     @Override
-    @SuppressWarnings("java:S106")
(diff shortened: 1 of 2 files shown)


joschi:
@adriansuarez Thanks for your contribution!

Could you please add a few tests for the new behavior?

adriansuarez:
> @adriansuarez Thanks for your contribution!
> 
> Could you please add a few tests for the new behavior?

@joschi: See my latest commit.

joschi (approved):
@adriansuarez Thanks!

PR #3620, opened 2020-12-27
Fix the AssertJ documentation URL

rhowe:
assertj.org is now a domain squatting site. Use the docs link from README.md at
https://github.com/assertj/assertj-core


joschi (approved):
@rhowe Thanks!

PR #3623, opened 2020-12-27
Rename `jdbi3` to `db` packages in example docs

muRn:
###### Problem:
Core page of documentation presents standard project structure incorrectly. Among packages that really exist such as `api`, `health` and `resources` it has `jdbi3` which isn't really a standard structure package judging by [example project](https://github.com/dropwizard/dropwizard/tree/master/dropwizard-example/src/main/java/com/example/helloworld) and also project structure generated from maven archetype.

###### Solution:
Replace `jdbi3` with `db` in the docs.

###### Result:
More accurate docs.


joschi (approved):
@muRn Thank you for your contribution! 😃

PR #3630, opened 2020-12-30
Remove usage of deprecated Mockito#initMocks()

rhowe:
Remove calls to this deprecated method. Neither of the test classes used Mockito annotations, so it wasn't necessary.


joschi (approved):
@rhowe Thanks! 😄

PR #3631, opened 2021-01-01
Replace Mockito#verifyZeroInteractions with verifyNoInteractions

rhowe:
verifyZeroInteractions is deprecated, update to the latest API.


joschi (approved):
@rhowe Thanks a lot!

PR #3634, opened 2021-01-03
Remove usage of deprecated jackson-databind PropertyNamingStrategy.SnakeCaseStrategy

rhowe:
This has been replaced by PropertyNamingStrategies.SnakeCaseStrategy due to a possible classloader deadlock:
https://github.com/FasterXML/jackson-databind/issues/2715

### DIFF
--- a/dropwizard-jackson/src/main/java/io/dropwizard/jackson/AnnotationSensitivePropertyNamingStrategy.java
+++ b/dropwizard-jackson/src/main/java/io/dropwizard/jackson/AnnotationSensitivePropertyNamingStrategy.java
@@ -1,5 +1,6 @@
 package io.dropwizard.jackson;
 
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.cfg.MapperConfig;
 import com.fasterxml.jackson.databind.introspect.AnnotatedField;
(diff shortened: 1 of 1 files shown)


dropwizard-jackson/src/main/java/io/dropwizard/jackson/AnnotationSensitivePropertyNamingStrategy.java
```diff
@@ -9,18 +10,13 @@
...
+ * {@link com.fasterxml.jackson.databind.PropertyNamingStrategies.SnakeCaseStrategy}, and uses
  * the default {@link PropertyNamingStrategy} otherwise.
  */
 public class AnnotationSensitivePropertyNamingStrategy extends PropertyNamingStrategy {
     private static final long serialVersionUID = -1372862028366311230L;
 
-    private final SnakeCaseStrategy snakeCase;
-
-    public AnnotationSensitivePropertyNamingStrategy() {
-        super();
```
jplock, line 21: Do we need to keep the constructor to call super()?

joschi: The default constructor does that.
> The compiler automatically provides a no-argument, default constructor for any class without constructors. This default constructor will call the no-argument constructor of the superclass.

https://docs.oracle.com/javase/tutorial/java/javaOO/constructors.html

joschi (approved):
@rhowe LGTM. Thanks!

dropwizard-jackson/src/main/java/io/dropwizard/jackson/AnnotationSensitivePropertyNamingStrategy.java
rhowe: Also there is no explicit constructor on the superclass (at least in the current version) so `super()` just ended up calling the default constructor AFAIK.

PR #3639, opened 2021-01-05
Support custom StatsCounter in CachingAuthenticator/CachingAuthorization

cyberdelia:
###### Problem:

Caffeine does not allow to [register more than one StatsCounter](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java#L945), raising an `IllegalStateException` if you do so. Since Dropwizard attempt to add its own automatically when using Caffeine, this prevent anyone from using a custom one.

###### Solution:

This adds a new constructor providing a way to supply a custom `StatsCounter` if needed for `CachingAuthorizer` and `CachingAuthorization`, that will replace the dropwizard one if provided.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
@@ -84,7 +84,7 @@ public class CachingAuthorizer<P extends Principal> implements Authorizer<P> {
      *
      * @param metricRegistry the application's registry of metrics
      * @param authorizer     the underlying authorizer
-     * @param builder        a {@link CaffeineSpec}
+     * @param builder        a {@link Caffeine} spec
      * @param supplier       a {@link Supplier<StatsCounter>}
      */
     public CachingAuthorizer(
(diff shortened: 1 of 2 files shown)


dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthorizer.java
```diff
@@ -74,11 +76,28 @@ public CachingAuthorizer(
...
         final Caffeine<Object, Object> builder) {
+        this(metricRegistry, authorizer, builder, () -> new MetricsStatsCounter(metricRegistry, name(CachingAuthorizer.class)));
+    }
+
+    /**
+     * Creates a new cached authorizer.
+     *
+     * @param metricRegistry the application's registry of metrics
+     * @param authorizer     the underlying authorizer
+     * @param builder        a {@link CaffeineSpec}
```
joschi, line 87:
```suggestion
     * @param builder        a {@link Caffeine} spec
```

joschi (approved):

PR #3649, opened 2021-01-11
DropwizardAppExtension support for RegisterExtension

christopher-cudennec:
This PR enables us to use DropwizardAppExtension with JUnit 5's "RegisterExtension" annotation.

###### Problem:
We're adding JUnit 5 capabilities to our open source framework [sda-dropwizard-commons](https://github.com/SDA-SE/sda-dropwizard-commons/). We noticed that JUnit 4's RuleChain is no longer available and are looking for a different solution to specify the order of the execution of the extensions. This is needed to spin up a service (e.g. with Wiremock) beforehand and configure URLs and ports in our application afterwards.

JUnit 5 provides these solutions as far as I know:
- declare the order by enumerating all extensions in your `ExtendWith` annotation
- declare the order by the order of the fields that have a `RegisterExtension` annotation
- you can also add an `Order` annotation to the fields that already have the `RegisterExtension` annotation.

Unfortunately you can't mix these two approaches very well.

You can find more information in the [official documentation](https://junit.org/junit5/docs/current/user-guide/#extensions-registration).

We need to use the programmatic extension registration to be able to configure the test resources in most cases. Declaratice extension is not an alternative.

###### Solution:

We'd like the existing DropwizardAppExtension support the programmatic approach. The only thing that needs to be added are two interfaces. I don't see any disadvantages - I even think using this approach is even easier because you only need one line of code; no need to also add `ExtendWith` annotation.

###### Result:

The changes we made are backwards compatible. The app extension can now be used with JUnit 5's `RegisterExtension`.

If you think the change is valid I'm happy to also add the new feature to the documentation. :)

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardAppExtension.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardAppExtension.java
@@ -20,6 +20,9 @@ import javax.ws.rs.client.Client;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
 
 //@formatter:off
 /**
@@ -33,7 +36,8 @@ import java.util.function.Function;
  * @param <C> the configuration type
  */
 //@formatter:on
-public class DropwizardAppExtension<C extends Configuration> implements DropwizardExtension {
+public class DropwizardAppExtension<C extends Configuration> implements DropwizardExtension,
+    BeforeAllCallback, AfterAllCallback {
 
     private static final int DEFAULT_CONNECT_TIMEOUT_MS = 1000;
     private static final int DEFAULT_READ_TIMEOUT_MS = 5000;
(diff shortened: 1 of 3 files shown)


dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardAppExtension.java
```diff
@@ -175,6 +179,16 @@ public void onRun(C configuration, Environment environment, DropwizardAppExtensi
...
+    @Override
+    public void beforeAll(ExtensionContext extensionContext) throws Exception {
+        this.before();
+    }
+
+    @Override
+    public void afterAll(ExtensionContext extensionContext) {
+        this.after();
+    }
+
```
joschi, line 191: Wouldn't this lead to unexpected results when static and instance extensions are mixed and nested test classes are used?

christopher-cudennec: Regarding your question about static/instance extensions:
https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic-static-fields

Do you have any specific problems in mind?

If we want to be able to distinguish between class level and instance level usage we could provide two extensions:
* one that implements BeforeAll- and AfterAllCallback (like JUnit 4's class rules)
* one that implements BeforeEach- and AfterEachCallback (like an instance rule)

joschi (approved):
LGTM.

christopher-cudennec:
Great, thanks! 🥇

PR #3661, opened 2021-01-18
Use spying instead of reflection in dropwizard-jetty tests

rhowe:
###### Problem:
Some of the `dropwizard-jetty` tests use reflection to examine private fields of other classes. In the case of `ArrayByteBufferPool` there is no API to obtain these values.

This is inherently flaky and not type safe.

###### Solution:
Instead of examining the internals of the object using reflection, add a factory method to proxy the constructor call which we can then spy on using Mockito (I think this is the cleanest way, but maybe someone who knows Mockito better than I do can improve on it).

###### Result:
This improves type safety and reduces the change of a change to the internals of ArrayByteBufferPool breaking the tests.

All we really care about here (I think) is that the pool is created with the parameters we expect.


jplock (approved):

rhowe:
This conflicts with another open PR. I'll merge them together.

PR #3666, opened 2021-01-24
Address SonarCloud issues

rhowe:
###### Problem:
Sonar is grumbling about a few things. Some aren't so valid and others I'm not too sure about, but the ones addressed by this PR do seem to be genuine:

These two:
https://sonarcloud.io/project/issues?id=dropwizard_dropwizard&issues=AXUXdIJdpBosHuY9-PuM&open=AXUXdIJdpBosHuY9-PuM
https://sonarcloud.io/project/issues?id=dropwizard_dropwizard&issues=AXUXdIIWpBosHuY9-Ptj&open=AXUXdIIWpBosHuY9-Ptj
relate to tests in `dropwizard-example` asserting that a `long` is not `null`, which isn't really proving anything.

These two:
https://sonarcloud.io/project/issues?id=dropwizard_dropwizard&issues=AXUXdIWFpBosHuY9-Pyx&open=AXUXdIWFpBosHuY9-Pyx
https://sonarcloud.io/project/issues?id=dropwizard_dropwizard&issues=AXUXdIWFpBosHuY9-Py7&open=AXUXdIWFpBosHuY9-Py7

relate to the `@Nested` annotation being used in `dropwizard-jetty`'s `ZipExceptionHandlingInputStreamTest`. This actually exposes the fact that these tests aren't being run because by default, surefire doesn't run tests in static inner classes!

###### Solution:
Remove the assertions which can't fail in `dropwizard-example`.

Split `ZipExceptionHandlingInputStreamTest` into two test classes. Since one of the inner classes was a parameterised test, it needed a static field and so the alternative was to reconfigure surefire. That seemed like a bit of a heavy-handed option so I just split the test up instead. I changed the `@Test` annotations to `@ParameterizedTest` as suggested by IntelliJ - not 100% sure that was the right thing to do, but it knows Java better than I.

Whilst we're in there, tidy up the assertions to use `assertThatExceptionOfType` instead of open-coding a try-catch.

###### Result:
Fewer Sonar errors, more tests being run, test code is a bit simpler.


joschi (approved):

joschi (approved):
Thanks!

PR #3667, opened 2021-01-24
Use assertThatExceptionOfType in dropwizard-auth

rhowe:
###### Problem:
It is (subjectively) better to specify tests in terms of success rather than terms of failure and also for there to be less code, provided it works and isn't confusing.

With that in mind, some tests could be improved slightly.

###### Solution:
translate

```
try {
    something();
    failBecauseExceptionWasNotThrown();
} catch (Exception e) {
    // assert some things about e
}
```
into
```
assertThatExceptionOfType(Exception.class)
    .isThrownBy(() -> something())
    // assert some more things
```

###### Result:
There is less code and (opinions may differ) the tests become more expressive.

If you agree this is a good thing to do, there are other places where `failBecauseExceptionWasNotThrown` is used and I'm happy to give those the same treatment if they look like suitable candidates.

Also, if these little cleanup PRs are annoying and creating unnecessary code churn and review effort, don't be afraid to say so. I like Dropwizard and have worked with it on many projects (https://github.com/alphagov/verify-hub being the only one which is public) where it has been really helpful, so I thought I'd try and do some odd bits and pieces to keep its quality up.

### DIFF
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/AuthBaseTest.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/AuthBaseTest.java
@@ -18,7 +18,7 @@ import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.HttpHeaders;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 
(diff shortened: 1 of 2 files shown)


joschi (approved):

joschi:
> Also, if these little cleanup PRs are annoying and creating unnecessary code churn and review effort, don't be afraid to say so.

@rhowe Thanks a lot for addressing these refactorings! They're certainly not annoying. 😉 

If you're looking for another thing to contribute, you might want to check out https://github.com/dropwizard/dropwizard/pull/3613.

PR #3671, opened 2021-01-28
Upgrade to Jersey 2.33

joschi:
https://github.com/eclipse-ee4j/jersey/releases/tag/2.33


maffe:
Maybe the release notes for Dropwizard 2.0.19 should mention Jersey 2.33 [no longer includes](https://github.com/eclipse-ee4j/jersey/pull/4634) `org.glassfish.jersey.media:jersey-media-jaxb`. Jersey users are likely to get exceptions like "MessageBodyWriter not found for media type=application/xml" because of this and have to include that library themselves.

joschi:
@maffe Great finding, thanks for bringing this up!

Would be able to create a PR for adding this information to the upgrade notes?

PR #3674, opened 2021-01-29
Register jackson-blackbird on JDK11 and beyond

### LINKED ISSUE #3652: Make implementation-specific Jackson dependencies optional
My team is using:
- Java 11
- Dropwizard 2.0.17
- Jackson 2.12.0

The class `io.dropwizard.jackson.Jackson` has hardcoded dependencies on multiple datatype implementations, including Jackson Afterburner. The Afterburner implementation now [has a replacement](https://cowtowncoder.medium.com/jackson-2-12-features-eee9456fec75) ([Blackbird](https://github.com/FasterXML/jackson-modules-base/tree/master/blackbird)), due to potential compatibility issues with Java 9 and later.

The current implementation requires that consumers have to completely replace the `Bootstrap`-created `ObjectMapper` instance in order to switch to Blackbird, as it's initialized using `Jackson.newObjectMapper()`. Given that Jackson is continuously evolving, it would be helpful if the dependencies could be decoupled.

One possible solution would be to have the dependencies on the [various external implementations be optional](http://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html), and only enabled automatically if present on the classpath. This would allow the `Jackson` implementation to auto-configure based on the consumer needs, while still providing an opinionated configuration capability.

arteam:
Jackson 2.12 shipped a new module `jackson-blackbird` for generating data-binding code dynamically. It is
a `jackson-afterburner` alternative which doesn't use `sun.misc.unsafe` for bytecode manipulation, but instead uses the official `LambdaMetafactory` API.

Closes #3652

### DIFF
--- a/dropwizard-jackson/pom.xml
+++ b/dropwizard-jackson/pom.xml
@@ -58,6 +58,10 @@
             <artifactId>jackson-module-afterburner</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-blackbird</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-joda</artifactId>
         </dependency>
--- a/dropwizard-jackson/src/main/java/io/dropwizard/jackson/Jackson.java
+++ b/dropwizard-jackson/src/main/java/io/dropwizard/jackson/Jackson.java
@@ -1,7 +1,5 @@
 package io.dropwizard.jackson;
 
-import static com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
-
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.guava.GuavaModule;
@@ -9,10 +7,13 @@ import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.fasterxml.jackson.datatype.joda.JodaModule;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.fasterxml.jackson.module.afterburner.AfterburnerModule;
+import com.fasterxml.jackson.module.blackbird.BlackbirdModule;
 import com.fasterxml.jackson.module.paramnames.ParameterNamesModule;
 import io.dropwizard.util.JavaVersion;
 
 import javax.annotation.Nullable;
+
+import static com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
 
 /**
  * A utility class for Jackson.
@@ -63,6 +64,9 @@ public class Jackson {
         mapper.registerModule(new JodaModule());
         if (JavaVersion.isJava8()) {
             mapper.registerModule(new AfterburnerModule());
+        }
+        if (JavaVersion.isJava11OrHigher()) {
+            mapper.registerModule(new BlackbirdModule());
         }
         mapper.registerModule(new FuzzyEnumModule());
         mapper.registerModule(new ParameterNamesModule());
--- a/dropwizard-jackson/src/test/java/io/dropwizard/jackson/JacksonTest.java
+++ b/dropwizard-jackson/src/test/java/io/dropwizard/jackson/JacksonTest.java
@@ -75,6 +75,7 @@ public class JacksonTest {
         Set<Object> registeredModuleIds = Jackson.newObjectMapper().getRegisteredModuleIds();
 
         assertThat(registeredModuleIds).extracting(Object::toString)
+            .isNotEmpty()
             .doesNotContain("com.fasterxml.jackson.module.blackbird.BlackbirdModule");
     }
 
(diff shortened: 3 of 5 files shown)


dropwizard-jackson/src/main/java/io/dropwizard/jackson/Jackson.java
```diff
@@ -64,6 +65,9 @@ private static ObjectMapper configure(ObjectMapper mapper) {
         if (JavaVersion.isJava8()) {
             mapper.registerModule(new AfterburnerModule());
         }
+        if (JavaVersion.isJava11OrHigher()) {
+            mapper.registerModule(new BlackbirdModule());
+        }
```
joschi, line 70: Afterburner and Blackbird are incompatible, so we should only register one of them.
```suggestion
        if (JavaVersion.isJava11OrHigher()) {
            mapper.registerModule(new BlackbirdModule());
        } else if (JavaVersion.isJava8()) {
            mapper.registerModule(new AfterburnerModule());
        }
```

dropwizard-jackson/pom.xml
```diff
@@ -57,6 +57,10 @@
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-afterburner</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-blackbird</artifactId>
+        </dependency>
```
joschi, line 63: Should we make these dependencies optional or rather rely on people to exclude the ones they're not using?

In either case, we'll have to detect the availability of each module in `io.dropwizard.jackson.Jackson`.

mschechter-bellese: I've seen a lot of success with the `optional` model for SDKs I've worked on. It also makes it more difficult for end users to wind up with a conflict :)

Thanks again for all the work on this!

arteam: Agreed, I think going with the optional dependency would be the best way to go. Let's me see what would be the best way to detected Jackson modules in the classpath.

mschechter-bellese: The only way I've seen used is to use `Class.forName` on a fully-qualified class name. It's what's used in `DropwizardKafkaUtils` and `FileDescriptorRatioGauge`.

If there's a centralized place to put this logic (like `dropwizard-util`), that could simplify things.

mattnelson:
This blog indicates that blackbird works on 8+.
https://cowtowncoder.medium.com/jackson-2-12-features-eee9456fec75

> While Blackbird does also work on Java 8, at this point it is mostly recommended for developers running on later JVMs.

Might be able to forego all the optional stuff and switch entirely to blackbird.

mschechter-bellese:
> This blog indicates that blackbird works on 8+.
> [cowtowncoder.medium.com/jackson-2-12-features-eee9456fec75](https://cowtowncoder.medium.com/jackson-2-12-features-eee9456fec75)
> 
> > While Blackbird does also work on Java 8, at this point it is mostly recommended for developers running on later JVMs.
> 
> Might be able to forego all the optional stuff and switch entirely to blackbird.

The `optional` pattern will be useful for more than just the Afterburner/Blackbird switch. All of the following could all be optional so that developers can better control their dependencies:
- Guava
- Joda
- Jackson Afterburner
- Jackson Blackbird
- Jackson ParamNames

For example, the project I'm currently using Dropwizard for does not want to use Guava or Joda, but we can't exclude it.

The ParamNames library can be an optional that defaults to included, so that we can safely exclude it.

mattnelson:
> The `optional` pattern will be useful for more than just the Afterburner/Blackbird switch. All of the following could all be optional so that developers can better control their dependencies:

I'm generally fine with that list no longer being default, but optional dependencies doesn't seem like the correct approach. The bundle pattern already provides that feature. If an additional feature is desired, register the bundle. Introducing the optional pattern would be fragmenting that design pattern. Dropwizard should remain opinionated in the preferred way to integrate official Dropwizard features, that way is currently bundles.
https://www.dropwizard.io/en/latest/manual/core.html?highlight=bundle#bundles

From that list, a few example bundles. Consumers would opt into the feature by including the bundle instead of opting into the implementation detail, which is the dependency.
* dropwizard-jackson-guava
* dropwizard-jackson-joda
* dropwizard-jackson-paramnames

mschechter-bellese:
> > The `optional` pattern will be useful for more than just the Afterburner/Blackbird switch. All of the following could all be optional so that developers can better control their dependencies:
> 
> I'm generally fine with that list no longer being default, but optional dependencies doesn't seem like the correct approach. The bundle pattern already provides that feature. If an additional feature is desired, register the bundle. Introducing the optional pattern would be fragmenting that design pattern. Dropwizard should remain opinionated in the preferred way to integrate official Dropwizard features, that way is currently bundles.
> [dropwizard.io/en/latest/manual/core.html?highlight=bundle#bundles](https://www.dropwizard.io/en/latest/manual/core.html?highlight=bundle#bundles)
> 
> From that list, a few example bundles. Consumers would opt into the feature by including the bundle instead of opting into the implementation detail, which is the dependency.
> 
>     * dropwizard-jackson-guava
> 
>     * dropwizard-jackson-joda
> 
>     * dropwizard-jackson-paramnames

If the bundle pattern would remove the hard-wired Maven dependencies and better suit your long-term direction, I'd be in favor. My major concern is ensuring that Dropwizard consumers have more control over their own classpath, and can prevent conflicts like the Afterburner/Blackbird one.

If I understand the `ConfiguredBundle` workflow, you could even have the bundles themselves check for conflicts and throw relevant `ConfigurationException`s.

Thanks for the detailed clarification!

mschechter-bellese:
Could you please let me know if there's an ETA for completing this (as described by @mattnelson), and if you're looking for contributions?

Thanks!

Issue #3669, opened 2021-01-26
Automatically include file argument (or add documentation) on ConfiguredCommand


pablo-silverback:
`ConfiguredCommand` does not include the configuration file argument automatically but makes you call `addFileArgument` on bootstrap.

Failing to call this method your command will complain that you're passing a file argument and none is expected.

This is not documented anywhere:

<img width="771" alt="Screen Shot 2021-01-26 at 14 33 47" src="https://user-images.githubusercontent.com/77513079/105881638-847e0b80-5fe3-11eb-919d-1135d59662d4.png">

It would be a good idea to either:

- automatically accept the file argument (this is what the docs say, btw)

<img width="1028" alt="Screen Shot 2021-01-26 at 14 37 38" src="https://user-images.githubusercontent.com/77513079/105882166-1ab23180-5fe4-11eb-86ed-f54e1ee47797.png">

- document that the call to `addFileArgument` is needed.

joschi:
@pablo-silverback Do you mean this?
https://github.com/dropwizard/dropwizard/blob/7385a7d7a3a1f0670d5fda376b8cdf516d7c7b30/dropwizard-core/src/main/java/io/dropwizard/cli/ConfiguredCommand.java#L47-L68

That's also how the `check` command (`CheckCommand` class) works:
https://github.com/dropwizard/dropwizard/blob/v2.0.18/dropwizard-core/src/main/java/io/dropwizard/cli/CheckCommand.java

pablo-silverback:
@joschi Exactly that. I believe it should be called always after the `configure` method, or the docs should read something like:

> by default Dropwizard will treat the last argument as the path to the YAML __unless you override the `configure` method. In that case you should call addFileArgument yourself__

With the not so crazy hypothesis that most people will override the `configure` method, the docs are misleading (they sure mislead me)

joschi:
@pablo-silverback I would expect people to call the parent's method if they're overriding a method, i. e. call `super(subparser)` if they're overriding the `configure(Subparser)` method.

If you don't call `super`, you're basically saying that you're not interested in inheriting the behavior of the parent class.

pablo-silverback:
I believe we can agree it doesn't make sense to have a `ConfiguredCommand` without a configuration file. 

Right now not only this is __possible__ (which I personally think it's a bug that needs fixing) but it's also _very likely_ to do it by mistake, if you forget to call `super` or call `addFileArgument` yourself.

[...]

PR #3675, opened 2021-01-29
Clarify the need of calling super from ConfiguredCommand

fernandezpablo85:
###### Problem:

Not calling super from `ConfiguredCommand#configure` skips the declaration of the configuration parameter. Making it mandatory would involve some non-minor API changes so emphasised the need of the super call on the docs (The wording is pretty similar to the [javadoc](https://www.javadoc.io/doc/io.dropwizard/dropwizard-core/1.0.2/io/dropwizard/cli/ConfiguredCommand.html#configure-net.sourceforge.argparse4j.inf.Subparser-))

###### Solution:

Explicitly state the need for the `super` call. 

###### Result:

Hopefully less people will be confused when they configure a `ConfiguredCommand` that fails to accept the configuration file as a parameter.

Closes https://github.com/dropwizard/dropwizard/issues/3669

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -722,6 +722,11 @@ location specified by ``-c``), adapt the ConfiguredCommand_ class as needed.
 
 .. _ConfiguredCommand: https://github.com/dropwizard/dropwizard/blob/master/dropwizard-core/src/main/java/io/dropwizard/cli/ConfiguredCommand.java
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


pablo-silverback:
Here's a preview of how the change looks:
Screen Shot 2021-01-29 at 16 29 18

joschi (approved):
@fernandezpablo85 Thanks!

pablo-silverback:
@joschi anytime!

PR #3688, opened 2021-02-06
Address SonarCloud concerns in dropwizard-auth

rhowe:
Sonar didn't like the accesses of static fields on a derived type, preferring direct access to the supertype instead.

Also, some assertions in the tests can be chained and we can use a few more specialised AssertJ assertions to make the tests more expressive.


arteam (approved):

arteam:
Thank you!

PR #3691, opened 2021-02-07
Remove public modifiers from tests

rhowe:
JUnit 5 no longer requires this.

Generated via:

    sed -i '/@\(Test\|BeforeEach\|AfterEach\)/{n;s/public void/void/}' \
        $(git grep -lF '@Test' */src/test/)

Sadly, Jersey's `JerseyTest` base class declares its `BeforeEach`/`AfterEach` methods public so we have to follow suit there.

### DIFF
--- a/dropwizard-auth/src/test/java/io/dropwizard/auth/AuthBaseTest.java
+++ b/dropwizard-auth/src/test/java/io/dropwizard/auth/AuthBaseTest.java
@@ -51,13 +51,13 @@ public abstract class AuthBaseTest<T extends DropwizardResourceConfig> extends J
 
     @Override
     @BeforeEach
-    public void setUp() throws Exception {
+    void setUp() throws Exception {
         super.setUp();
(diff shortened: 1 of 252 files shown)


rhowe:
Looks like `JdbiHealthCheckTest#testNoTimeoutReturnsHealthy` might be a bit flaky?

jan-molak (approved):
This is a good suggestion, JUnit 5 test methods don't need to be public ([source](https://junit.org/junit5/docs/current/user-guide/#writing-tests-classes-and-methods))

jplock (approved):

rhowe:
Just checked sonar and I think this has reduced the "code smell" count in Sonar from just over 2,000 to just under 1,000 :+1:

rhowe:
Then again, seems I was a bit too keen: https://github.com/dropwizard/dropwizard/pull/3727

PR #3704, opened 2021-02-14
Be stricter when parsing 'Range' headers

rhowe:
###### Problem:
`AssetsServlet` will accept a `Range:` header which isn't RFC-compliant due to the behaviour of `Integer.parseInt` 

`Integer.parseInt()` will gladly accept non-ASCII representations of digits, but RFC7233 specifies that only ASCII 0-9 are valid digits in a `Range` header.

I initially thought this might have some security implications, but I've decided either I'm lacking in creativity or it's just a regular minor bug.

###### Solution:
Treat the header value as ASCII when parsing.

###### Result:
Requests with an invalid `Range` header will be rejected.

### DIFF
--- a/dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/ByteRange.java
+++ b/dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/ByteRange.java
@@ -3,6 +3,7 @@ package io.dropwizard.servlets.assets;
 import io.dropwizard.util.Strings;
 
 import javax.annotation.concurrent.Immutable;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


jplock (approved):

rhowe:
Sorry @jplock I realised that this test wasn't really in line with the refactor of testing thrown exceptions which went in recently, so I've rewritten it to use `assertThatExceptionOfType`. Also I updated the declaration of `asciiString` to be `final`.

rhowe:
It might also be better (and more performant) to do this character set handling in `AssetsServlet#parseRangeHeader` :shrug:

rhowe:
Rebased vs current `master`

PR #3727, opened 2021-02-27
Make jUnit 4 test methods public

rhowe:
This is a regression from f17aea0e7669b6182330151697022b1d44fde8bd - not all tests are jUnit 5. jUnit 4 test methods should be made public.


rhowe:
At least I believe the PR message is true - unless these tests are running in some sort of strange junit4-in-jupiter mode that I didn't know existed?

arteam (approved):

arteam:
Thanks for following up on this!

PR #3731, opened 2021-02-28
Improve dropwizard-migrations tests

rhowe:
###### Problem:
There are some opportunities to improve the `dropwizard-migrations` tests.

- Some fields could be made `final`
- The UTF-8 character set was referenced via a shared string when `StandardCharsets` could be used instead.
- In some cases we were using the ability to create something (e.g. a table) as a proxy for  detecting whether the table existed or not.

###### Solution:
- Make fields `final` which can be.
- Switch to using `StandardCharsets`.
- Test for existence of database entities directly rather than using the ability to create them as a proxy.

###### Result:
Tests should be a bit more precise and Sonar should complain a bit less.


rhowe:
Rebased vs current `master`

arteam (approved):

Issue #3724, opened 2021-02-25
AssetServlet - incorrect cache behavior


bmsq:
Hi,

While investigating caching issues in our app I noticed that [AssetServlet](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/AssetServlet.java#L304) is returning 304 when the `If-None-Match` header matches OR `If-Modified-Since` header is later or equal to the asset modified time.

Looking at [MDN's documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) for these headers, I'm of the understanding that `If-Modified-Since` should be ignored when `If-None-Match` is present.  

As an example of how this causes problems, we have a scenario where two different `index.html` assets are served from the same URL depending on configuration.  Since both these files are packaged in the same jar, the modified timestamps are the same but the ETag is different.  Unfortunately, changing the configuration and restarting the server results in the browser's cache being used because the server is responding 304 (different ETag but same modification date).

PR #3736, opened 2021-03-01
Fixed incorrect cache behaviour (#3724)

twhi21:
###### Problem:
AssetServlet is returning 304 when the If-None-Match header matches OR If-Modified-Since header is later or equal to the asset modified time.
This is incorrect according to [MDN's documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since)

###### Solution:
Based on the documentation if the request contains the If-None-Match header then they are compared instead of checking If-Modified-Since


bmsq:
Fixes #3724

joschi (approved):
@twhi21 Thanks for your contribution!

PR #3788, opened 2021-03-17
Build Dropwizard with Java 16

joschi:
https://mail.openjdk.java.net/pipermail/announce/2021-March/000295.html

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: ['8', '11', '15']
+        java_version: ['8', '11', '16']
         os: ['ubuntu-latest']
     env:
       JAVA_OPTS: "-XX:+TieredCompilation -XX:TieredStopAtLevel=1"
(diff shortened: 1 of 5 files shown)


jplock (approved):

joschi:
The integration tests using MySQL via Docker are failing because MySQL Connector/J is disabling all protocols with TLSv1.2 or higher which leaves none when using Java 16. 😂 

```
com.mysql.cj.jdbc.exceptions.CommunicationsException: 
Communications link failure

The last packet sent successfully to the server was 0 milliseconds ago. The driver has not received any packets from the server.
Caused by: com.mysql.cj.exceptions.CJCommunicationsException: 
Communications link failure

The last packet sent successfully to the server was 0 milliseconds ago. The driver has not received any packets from the server.
Caused by: javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate)
```

Quote from https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-usagenotes-known-issues-limitations.html
> Because Connector/J does not enable connections with TLSv1.2 and higher by default due to compatibility issues, when connecting to servers that restrict connections to use those higher TLS versions, you might encounter com.mysql.cj.exceptions.CJCommunicationsException: javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate). You need to enable connections with TLSv1.2 and higher versions using the enabledTLSProtocols connection property. See Section 6.8, “Connecting Securely Using SSL” for details.

See also: https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-using-ssl.html

PR #3789, opened 2021-03-17
Skip POM signature verification

joschi:
The signature verification of our dependencies is often running into problem, so let's skip POM signature verification for now.

The actual artifacts (JARs) are still verified.


joschi:
Relates to https://github.com/dropwizard/dropwizard/pull/3753.

PR #3973, opened 2021-05-17
AssetServlet: Use MIME type of index file for directory requests

maffe:
With this change the extension of the index file is used do determine the content type of responses of requests to directories.

Background: I have written a filter which uses a servlet extending `AssetServlet` to serve static resources at the same path level as Jersey resources (for example, it’s possible to have dynamic Jersey resources at `/someResource` while also serving static content at paths like `/`, `/robots.txt` and `/favicon.ico`). Currently the servlet replaces the `HttpServletRequest` with a [proxy](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/reflect/Proxy.html) which replaces the return value of `getRequestURI()` with the URI of the index file if the real request URI ends with `/`. This change eliminates the need for the (somewhat ugly) proxy.

If desired, I could provide the code for the filter and servlet in another pull request.


joschi (approved):
@maffe LGTM, thanks!

PR #3978, opened 2021-05-18
Add method to `AbstractDAO` to get `NamedQuery` in a type-safe manner

rbarbey:
###### Problem:
`AbstractDAO` offers a way to easily get named queries but does so in a way that requires casting the returned `Query` to the correct type.

###### Solution:
Added a method to `AbstractDAO` which returns a named query in a type-safe way, `namedTypedQuery(String)`. This methods uses `Session#createNamedQuery(String, Class)` internally where the generic type is retrieved via `AbstractDAO#getEntityClass()`.

I hope I followed all conventions of this project. Please let me know if you would like me to change anything.

### DIFF
--- a/docs/source/manual/hibernate.rst
+++ b/docs/source/manual/hibernate.rst
@@ -123,7 +123,6 @@ contains type-safe wrappers for most of ``SessionFactory``'s common operations:
             return persist(person).getId();
         }
 
-        @SuppressWarnings("unchecked")
         public List<Person> findAll() {
             return list(namedTypedQuery("com.example.helloworld.core.Person.findAll"));
         }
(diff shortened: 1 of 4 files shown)


docs/source/manual/hibernate.rst
```diff
@@ -125,7 +125,7 @@ contains type-safe wrappers for most of ``SessionFactory``'s common operations:
 
         @SuppressWarnings("unchecked")
```
jtvd78, line 126: You should be able to remove this now since you're no longer casting a generic type

rbarbey: Thank you for pointing that out! I've adjusted the comment and removed the `SuppressWarnings` annotation.

joschi (approved):
@rbarbey LGTM, thanks!

@jtvd78 Thanks for your additional review!

PR #4007, opened 2021-05-27
Improved AdminServlet support

### LINKED ISSUE #3876: Add configuration for AdminServlet capabilities
The `AdminServlet` provides a number of features/functions, including thread dumps and CPU profiling. The latter two have the potential to create a denial-of-service attack on a server if a malicious actor were able to access those endpoints somehow. It's theoretically possible even if the endpoints are not exposed externally.

It would be great to be able to disable the various operational endpoints dynamically based on configuration. That would allow developers to leave the functionality on in non-production environments for diagnostic purposes, and reduce the exposed API for production to limit vulnerability.

An example of this type of configuration can be seen in the [Spring Boot Actuator](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-endpoints), which allows each provided endpoint to be controlled independently.

Thanks in advance!

mschechter-bellese:
- Added new configuration structure to provide configuration options
- Updated documentation

Resolves #3876

###### Problem:

The `AdminServlet` has added configuration parameters (see [metrics issue](https://github.com/dropwizard/metrics/pull/1891)). There is no way to currently modify those values via Dropwizard configuration.

###### Solution:

I added a new factory (following the factory patterns elsewhere in the `dropwizard-core` module) to create/configure the `AdminServlet`. The new factory encapsulates all of the configuration options for the servlet. The registration has also moved from `AbstractServerFactory` to the new servlet factory.

Updated documentation has been provided for the new configuration

###### Result:

Dropwizard users can now turn off any or all of the servlets provided by the `AdminServlet`, or change their endpoint URIs. This is useful because the thread dump and profiling servlets can be used to consume application resources and block legitimate access to the application endpoints.

There should be no change for any user who does not use the new configuration settings; everything should continue to operate as before.


mschechter-bellese:
@joschi / @arteam,

Could you please review this when you have a chance?

Thanks!

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
```diff
@@ -0,0 +1,235 @@
...
+
+    private String threadsUri = AdminServlet.DEFAULT_THREADS_URI;
+
+    private boolean healthcheckEnabled = true;
+
+    private String healthcheckUri = AdminServlet.DEFAULT_HEALTHCHECK_URI;
+
+    private boolean cpuProfileEnabled = true;
+
+    private String cpuProfileUri = AdminServlet.DEFAULT_CPU_PROFILE_URI;
```
joschi, line 112: Do you think it might make sense to model this as a map of string -> (boolean, URI)?

That way people could add their own keys and enable certain functionality on their admin connector.

```yaml
adminServlets:
  metrics:
    enabled: true
    uri: /metrics
  ping:
    enabled: true
    uri: /ping
  threads:
    enabled: false
```

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
```diff
@@ -0,0 +1,235 @@
...
+ *     <tr>
+ *         <td>cpuProfileUri</td>
+ *         <td>/pprof</td>
+ *         <td>The relative URI for CPU profiling retrieval.</td>
+ *     </tr>
+ * </table>
+ *
+ * @since 2.0
+ */
+public class AdminServletFactory {
```
joschi, line 92: Could you please add a test for this class?

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
```diff
@@ -0,0 +1,235 @@
...
+ *         </td>
+ *     </tr>
+ *     <tr>
+ *         <td>cpuProfileUri</td>
+ *         <td>/pprof</td>
+ *         <td>The relative URI for CPU profiling retrieval.</td>
+ *     </tr>
+ * </table>
+ *
+ * @since 2.0
```
joschi, line 90:
```suggestion
 * @since 2.1
```

joschi (commented):
@mschechter-bellese Sorry it took a while to get to this.

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
mschechter-bellese: This is interesting; it definitely groups the configuration more clearly. Is it still reasonable that anything not explicitly configured is enabled and available at the default endpoint?

Could you please let me know if there's an example of this type of functionality somewhere else I could use as a model? It seems like creating a simple object that has a `boolean` and `String` properties would be sufficient. I'd also like to model the default value pattern (if they don't provide a URL, for example) on existing code if there's any you could point me to.

Thanks!

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
mschechter-bellese: Yes, as soon as I can get back to this; is there an example I could use for reference?

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
mschechter-bellese: @joschi,

When you say a user would "add their own keys", does that mean you want to be able to have sparse configuration? That's what your example looks like. Also, the `AdminServlet` configuration won't support ad hoc keys, so it seems that's not where you want to go.

Could you please help me understand if I'm missing anything?

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
mschechter-bellese: I've added tests for the factory based on `GzipHandlerFactoryTest`

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
mschechter-bellese: @joschi,

I've pushed a potential update that implements the pattern as I understand it. If this works for you, I'll update the documentation (class and RST), and push another update.

Thanks!

mschechter-bellese: @joschi,

I will address the conflicts as soon as I can; could you please review the direction so far to see if this is how you'd like the configuration to be structured?

mschechter-bellese: @joschi,

Could you please provide feedback on the direction so far? I can definitely pick this back up once I know if I'm going in the right direction.

mschechter-bellese:
@joschi / @arteam,

Could you please provide feedback on the direction I've taken in response to your suggestions (for both configuration and testing)? I will reconcile the documentation and clean up the merge once I have that.

Thanks!

mschechter-bellese:
@joschi,

I've got the code up to date, and am waiting on your input before further changes. 

Also, I noticed that the tests for `ResourceUrl` will fail with Zulu (`azul`) JDKs because the resource type is `jrt`. Switching to a Coretto JDK resolved the issue.

Hope this helps!

mschechter-bellese:
@joschi / @arteam,

Could you please let me know if this pull request is something you're still interested in pursuing?

Thanks!

joschi:
@dropwizard/committers Anyone else wants to take a look at this?

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
```diff
@@ -0,0 +1,294 @@
+package io.dropwizard.server;
+
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.health.HealthCheckRegistry;
+import com.codahale.metrics.servlets.AdminServlet;
+import com.codahale.metrics.servlets.HealthCheckServlet;
+import com.codahale.metrics.servlets.MetricsServlet;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.dropwizard.jetty.MutableServletContextHandler;
+import org.apache.commons.lang3.StringUtils;
```
rhowe, line 10: Can we avoid bringing commons-lang3 back in?

dropwizard-core/src/test/java/io/dropwizard/server/AdminServletFactoryTest.java
```diff
@@ -0,0 +1,152 @@
...
+        assertThat(handler.getServletContext().getAttribute(MetricsServlet.METRICS_REGISTRY)).isEqualTo(metricRegistry);
+        assertThat(handler.getServletContext().getAttribute(HealthCheckServlet.HEALTH_CHECK_REGISTRY))
+            .isEqualTo(healthCheckRegistry);
+    }
+
+    @Test
+    void testBuildPartial() throws URISyntaxException, ConfigurationException, IOException {
+        AdminServletFactory adminServlet = new YamlConfigurationFactory<>(AdminServletFactory.class,
+            BaseValidator.newValidator(), Jackson.newObjectMapper(), "dw")
+            .build(new File(Resources.getResource("yaml/adminservlet_partial.yml").toURI()));
```
rhowe, line 122: This (and other cases where `YamlConfigurationFactory` is loading config from a resource) should use `ResourceConfigurationSourceProvider`

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
zUniQueX: Hi @mschechter-bellese. So far, it looks good to me.

> Also, the `AdminServlet` configuration won't support ad hoc keys, so it seems that's not where you want to go.

This makes sense to me. I'd suggest to add a check, if there are any keys we cannot handle.

dropwizard-core/src/main/java/io/dropwizard/server/AdminServletFactory.java
zUniQueX: Since only `StringUtils.isNotEmpty` is used, I think this can simply be replaced by a small helper method.

PR #4018, opened 2021-06-02
Add @JsonProperty for field layout getter/setter

egginbag:
###### Problem:
Dropwizard allows customized object mapper. If the customization happens to disable AUTO_DETECT_SETTERS (https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/MapperFeature.html#AUTO_DETECT_SETTERS), it will only consider methods explicitly annotated as setters. Thus layout field cannot be recognized. 

Also, it is not consistent with other properties in the same file. The other setters/getters are all annotated.

###### Solution:
Add @JsonProperty annotation for layout property getter and setter

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java
@@ -235,10 +235,12 @@ public abstract class AbstractAppenderFactory<E extends DeferredProcessingAware>
     }
 
     @Nullable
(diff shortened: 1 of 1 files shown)


joschi:
@egginbag Thanks for your contribution!

The `layout` field is not supposed to be settable via the configuration file.
The implementations of `DiscoverableLayoutFactory` are not serializable/deserializable.

joschi (approved):
@egginbag LGTM. Thanks for your contribution! 👍

PR #4024, opened 2021-06-04
[Docs] Add link to xvik/dropwizard-guicey

OneCricketeer:
###### Problem:

Docs reference Guice Bundle for DI, but there are multiple, and unclear which is the one that should be used. 

###### Solution:

Add a link to the docs


joschi:
> Docs reference Guice Bundle for DI, but there are multiple, and unclear which is the one that should be used.

The one working with the latest stable version of Dropwizard is fine (https://github.com/xvik/dropwizard-guicey).

OneCricketeer:
Yup, that's the one I linked

joschi (approved):
@OneCricketeer Thanks for your contribution!

PR #4155, opened 2021-07-22
Adds more robust health checking / scheduling functionality to base D…

natalie-zamani:
…ropwizard framework

https://github.com/dropwizard/dropwizard/issues/4154

This is a first pass at integrating the dropwizard-health functionality into the Dropwizard framework, and adapting the Bundle-based approach into a top-level `Configuration` factory (that is completely optional).

I'm expecting there to be some back and forth on this, and probably a bunch of things I may have missed. That, and I know I'm currently missing more detailed documentation around this change, but I'd like to start getting feedback on the meantime.

Issue #4154, opened 2021-07-22
Integrate dropwizard-health functionality into the Dropwizard framework


natalie-zamani:
We should integrate the https://github.com/dropwizard/dropwizard-health functionality into the Dropwizard framework by default.

It would be valuable to have functionality in the base framework to check dependencies on an ongoing basis, and aggregate the health of those dependencies and other checks to be able to communicate a holistic view of an applications health to infrastructure (whether that be load balancers, platforms like Kubernetes, other services, etc).

natalie-zamani:
Remaining TODOs:

- [ ] Write general usage documentation.
- [ ] Write migration guide from dropwizard-health
- [ ] Evaluate whether to make multiple health factories configurable in a single app
- [ ] Revisit API of `StateChangedCallback`, to possibly provide more info.

PR #4166, opened 2021-07-26
Additional Health functionality & documentation

natalie-zamani:
Enhancements on the base [dropwizard-health](https://github.com/dropwizard/dropwizard-health) functionality, and further integration into the framework. Also, documentation!

Related to #4155. 

#4154


natalie-zamani:
Still a WIP, by the way.

Planning to add:

- [ ] Unit tests for some new code
- [ ] More documenation
- [ ] Migration guide from dropwizard-health

dropwizard-health/src/main/java/io/dropwizard/health/DefaultHealthFactory.java
```diff
@@ -61,8 +66,13 @@
     private List<String> healthCheckUrlPaths = ImmutableList.of("/health-check");
```
bowsersenior, line 66: Would it make sense to add a constant for this path based on `DEFAULT_BASE_NAME`?

dropwizard-health/src/main/java/io/dropwizard/health/response/DetailedJsonHealthResponseProvider.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public HealthResponse minimalHealthResponse(@Nullable String type) {
+        final boolean healthy = healthStatusChecker.isHealthy(type);
+
+        return new HealthResponse(healthy, null, MediaType.APPLICATION_JSON);
+    }
+
+    @Nonnull
+    @Override
+    public HealthResponse partialHealthResponse(@Nullable final String type, @Nonnull final Collection<String> names) {
```
bowsersenior, line 62: What do you think about having sub-classes for `PartialHealthResponse` and `MinimalHealthResponse` to better differentiate them. `PartialHealthResponse` could have a `healthStateViews` member and perhaps even some kind of validation for the view names so that a request for a health check name that is not defined in the config will be detected and a warning can be shown.

dropwizard-health/src/main/java/io/dropwizard/health/response/DetailedJsonHealthResponseProvider.java
```diff
@@ -0,0 +1,82 @@
...
+        final Collection<HealthStateView> healthStateViews = names.stream()
+            .map(healthStateAggregator::healthStateView)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .collect(Collectors.toList());
+        String responseBody = null;
+        try {
```
bowsersenior, line 69: It's not clear to me how invalid health check requests would be handled. For example if invalid values were provided for `type` or `names`. The same concern would apply to `fullHealthResponse()` above.

One idea that comes to mind is that `HealthResponse` would have 3 possible values for `healthy` instead of being a boolean:
* `healthy`
* `unhealthy`
* `invalid_request`

When performing health checks in general there will always be an issue of dealing with indeterminate results. It is probably bad if the health check is invalid due to bad params, but it doesn't mean that the actual service is unhealthy.

dropwizard-health/src/main/java/io/dropwizard/health/DefaultHealthFactory.java
natalie-zamani: Definitely!

dropwizard-health/src/main/java/io/dropwizard/health/response/DetailedJsonHealthResponseProvider.java
natalie-zamani: So with respect to names, we will return an empty optional from `healthStateAggregator::healthStateView` for any name that doesn't exist. This makes it so we allow users to provide names in query params that don't necessarily map to a health check without blowing up (which allows us to handle backwards compatibility more gracefully for when a user adds or removes a check, and the query params don't reflect that change right away.

In terms of the type, that's a good question. We also take the approach of gracefully handling invalid or unrecognized type values. 

In `HealthCheckManager::isHealthy` you can see this. Essentially, if the string provided is `ALIVE` (case-insensitive), then we return the right value for a liveness check, otherwise we just default to a readiness check result.

If a user passed in `STARTUP` (which could correspond to the [relatively new Kubernetes Startup Probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes)), we would want users to start passing in that query param without blowing up. But we might default to still returning the readiness probe result until that code change rolled out.

It's kind of just the philosophy of backwards compatibility and lax validations we're taking with this default response provider, to make it as broadly useful as possible. The important thing is that we allow any user to provide an alternative implementation to this via the usual Dropwizard factory polymorphic config means if they want different behavior though!

dropwizard-health/src/main/java/io/dropwizard/health/response/DetailedJsonHealthResponseProvider.java
natalie-zamani: I made `HealthResponse` abstract, and added two sub-classes `DetailedHealthResponse` and `MinimalHealthResponse`, to make the distinctions between the two clearer, and hopefully make the code a little less error-prone.

dropwizard-health/src/test/java/io/dropwizard/health/response/DetailedJsonHealthResponseProviderTest.java
```diff
@@ -0,0 +1,253 @@
...
+        // then
+        assertThatThrownBy(() -> detailedJsonHealthResponseProvider.partialHealthResponse(null,
+            Collections.singleton(view.getName())))
+            .isInstanceOf(RuntimeException.class)
+            .hasCauseReference(exception);
+        verifyNoInteractions(healthStatusChecker);
+    }
+
+    // Duplicated from dropwizard-testing due to circular deps
+    private String fixture(final String filename) {
```
pstackle, line 245: we can't depend on `dropwizard-testing` even after most of the components got pushed out of `dropwizard-core`?

dropwizard-core/src/main/java/io/dropwizard/Configuration.java
```diff
@@ -8,10 +8,11 @@
 import io.dropwizard.server.DefaultServerFactory;
 import io.dropwizard.server.ServerFactory;
 import io.dropwizard.setup.AdminFactory;
-import java.util.Optional;
+
 import javax.annotation.Nullable;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
+import java.util.Optional;
```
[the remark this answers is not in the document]
natalie-zamani: Yes! My last PR messed up some imports, and this fixes them

dropwizard-health/src/main/java/io/dropwizard/health/response/HealthResponseProvider.java
```diff
@@ -0,0 +1,16 @@
...
+public interface HealthResponseProvider {
+    @Nonnull
+    HealthResponse fullHealthResponse(@Nullable String type);
+
+    @Nonnull
+    HealthResponse minimalHealthResponse(@Nullable String type);
+
+    @Nonnull
+    HealthResponse partialHealthResponse(@Nullable String type, @Nonnull Collection<String> names);
+}
```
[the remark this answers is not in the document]
natalie-zamani: That's totally fair, I felt a little icky with this too. Let me try to make this a little less strange.

dropwizard-health/src/main/java/io/dropwizard/health/check/http/HttpHealthCheck.java
```diff
@@ -1,23 +1,22 @@
-package io.dropwizard.health.http;
+package io.dropwizard.health.check.http;
```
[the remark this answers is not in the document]
natalie-zamani: They can totally live together

dropwizard-health/src/main/java/io/dropwizard/health/response/HealthResponder.java
```diff
@@ -0,0 +1,5 @@
+package io.dropwizard.health.response;
+
+public interface HealthResponder {
+    HealthResponseProvider healthResponseProvider();
```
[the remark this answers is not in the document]
natalie-zamani: Yeah, I have no idea either now that I revisit it. I think I had something in mind, but it's no longer relevant. Will delete.

dropwizard-health/src/main/java/io/dropwizard/health/response/ServletHealthResponder.java
```diff
@@ -0,0 +1,72 @@
...
+                // normalize to lowercase
+                .map(String::toLowerCase)
+                .distinct()
+                .collect(Collectors.toList());
+        }
+
+        final HealthResponse healthResponse;
+        if (names == null) {
+            healthResponse = healthResponseProvider()
+                .minimalHealthResponse(type);
```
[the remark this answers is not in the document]
natalie-zamani: No good reason! 🤷🏼‍♀️

dropwizard-health/src/main/java/io/dropwizard/health/response/ServletHealthResponder.java
```diff
@@ -0,0 +1,72 @@
...
+        healthResponse.getMessage().ifPresent(writer::write);
+        if (!healthResponse.isHealthy()) {
+            response.setStatus(HttpServletResponse.SC_SERVICE_UNAVAILABLE);
+        }
+    }
+
+    @Override
+    public HealthResponseProvider healthResponseProvider() {
+        return healthResponseProvider;
+    }
```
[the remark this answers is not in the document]
natalie-zamani: Yep, removing it 😅

dropwizard-health/src/main/java/io/dropwizard/health/response/ServletHealthResponderFactory.java
```diff
@@ -0,0 +1,53 @@
...
+    public void configure(final String name, final Collection<String> healthCheckUrlPaths,
+                          final HealthResponseProvider healthResponseProvider,
+                          final HealthEnvironment health, final JerseyEnvironment jersey,
+                          final ServletEnvironment servlets, final ObjectMapper mapper) {
+        final ServletHealthResponder servlet = new ServletHealthResponder(healthResponseProvider, cacheControlEnabled,
+            cacheControlValue);
+        servlets
+            .addServlet(name + SERVLET_SUFFIX, servlet)
+            .addMapping(healthCheckUrlPaths.toArray(new String[0]));
+        health.setHealthResponder(servlet);
```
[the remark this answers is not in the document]
natalie-zamani: My reasoning here is because I was trying to rationalize how this would all work if I wanted to plug in something like a `GrpcResponder`. Since gRPC isn't at all plugged into the main Dropwizard framework, I wanted a way to access whatever responder we produced from the `HealthEnvironment` so I could plug it where necessary

dropwizard-health/src/test/java/io/dropwizard/health/response/DetailedJsonHealthResponseProviderTest.java
natalie-zamani: Nope, because there's still that circular dependency.

Core pulls in health, and testing pulls in core.

pstackle: gotcha.

dropwizard-health/src/main/java/io/dropwizard/health/response/ServletHealthResponderFactory.java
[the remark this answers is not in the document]
natalie-zamani: Decided to remove this for now. If there's some value in adding it in the future, I'll circle back

dropwizard-health/src/main/java/io/dropwizard/health/response/HealthResponseProviderFactory.java
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * Configures a health responder for responding to health check requests (e.g. from load balancer).
+     *
+     * @param healthStatusChecker an interface that exposes the ability to check current status of health.
+     * @param healthStateAggregator an interface that exposes the ability to check an aggregate view of all health
+     *                              states.
+     * @param mapper A Jackson object mapper to allow writing JSON responses (if needed).
+     */
+    HealthResponseProvider build(HealthStatusChecker healthStatusChecker, HealthStateAggregator healthStateAggregator,
+                                 ObjectMapper mapper);
```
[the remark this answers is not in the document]
natalie-zamani: It's just about avoiding having to create a separate object mapper for this stuff, and then having to deal with exposing a way to configure it as well as the main object mapper for the app.

dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckScheduler.java
```diff
@@ -44,18 +45,18 @@ public void schedule(final ScheduledHealthCheck check, final boolean healthy) {
     }
 
     private void schedule(final ScheduledHealthCheck check, final Duration initialDelay, final Duration delay) {
-        final ScheduledFuture taskFuture = executorService.scheduleWithFixedDelay(check, initialDelay.toMilliseconds(),
-                delay.toMilliseconds(), TimeUnit.MILLISECONDS);
+        final ScheduledFuture<?> taskFuture = executorService.scheduleWithFixedDelay(check,
```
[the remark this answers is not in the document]
natalie-zamani: `executorService::scheduleWithFixedDelay` returns `ScheduledFuture<?> -- maybe my brain is fried at the moment, but I'm not thinking of a way to coax that type to `ScheduledFuture<ScheduledHealthCheck>` safely

docs/source/manual/core.rst
```diff
@@ -453,6 +453,77 @@ exception was thrown).
...
+
+* An ``alive`` status indicates the application is operating normally and does not need to be restarted to recover from
+  a stuck state. Long-running applications can eventually reach a broken state and cannot recover except by being
+  restarted (e.g. deadlocked threads).
+* A ``ready`` status indicates the application is ready to serve traffic. Applications can temporarily be unable to
+  serve traffic due to a variety of reasons, for example, an application might need to build/compute large caches
+  during startup or can critically depend on an external service.
+
+Long-running applications can eventually reach a broken state and cannot recover except by being restarted
+(e.g. deadlocked threads).
```
pstackle, line 515: this is repeated from the `alive` explanation.

docs/source/manual/core.rst
```diff
@@ -453,6 +453,77 @@ exception was thrown).
...
+
+An example of how you might query the health check, assuming you're using the default responder/responseProvider
+settings in configuration:
+https://<hostname>:<port>/health-check?type=<type>
+(replace `<type>` with `ready` or `alive`; defaults to `ready`)
```
pstackle, line 520:
```
defaults to `ready`
```
should be this:
```
defaults to `ready` if the `type` parameter is not included.
```

docs/source/manual/core.rst
```diff
@@ -453,6 +453,77 @@ exception was thrown).
...
+
+Querying for particular health checks
+https://<hostname>:<port>/health-check?type=<type>&name=<name>
+(replace ``<type>`` with ``ready`` or ``alive``; defaults to ``ready``)
+(replace ``<name>`` with the name of the health check to query. Multiple names can be provided)
```
pstackle, line 525: Are we trying to replace the functionality of the built-in /healthchecks endpoint on the AdminServlet?

natalie-zamani: I'm leaving that as out of scope for this PR honestly. I think that's a bigger change.

I am hoping to provide similar information, maybe in a more opt-in fashion from this endpoint though.

docs/source/manual/core.rst
```diff
@@ -453,6 +453,77 @@ exception was thrown).
...
 Java 1.6's built-in thread deadlock detection to determine if any threads are deadlocked.
 
+.. _man-core-health:
+
+Health
+======
+
+The health checks described in :ref:`man-core-healthchecks` can be configured to create a holistic view of your
+service health, which can then be used to drive decision making by things like `Kubernetes readiness & liveness checks`_,
+or to dictate whether or not a load balancer should route traffic to your service.
```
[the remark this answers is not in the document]
natalie-zamani: Very good point!

natalie-zamani:
Depending on further review comments, the remaining work on this PR is to add a migration guide from dropwizard-health (for release 2.1.x)

docs/source/manual/upgrade-notes/upgrade-notes-2_1_x.rst
```diff
@@ -30,5 +30,24 @@ Details about these changes in Hibernate Validator can be found at:
...
\ No newline at end of file
+================================
+
+Migrating from the HealthBundle
+-------------------------------
+You'll want to remove the ``HealthBundle`` from your `Application`'s ``initialize()`` method, as you no longer need to add a bundle to get the dropwizard-health functionality.
+
+You'll also want to remove a reference to ``HealthConfiguration`` from your `Configuration` class, as it's now part of the base Dropwizard `Configuration`.
+
+Once you do that, and upgrade `2.1.x` you should be able to just rely on the adding the dropwizard-health behavior by adding config under the top-level config ``health``!
```
pstackle, line 41: below is suggested rewording of section. I think a bulleted list with shorter imperative instructions will be clearer and easier to follow.

---
Starting in Dropwizard 2.1, the functionality from the external [dropwizard/dropwizard-health](https://github.com/dropwizard/dropwizard-health) module has been integrated into the main [dropwizard/dropwizard](https://github.com/dropwizard/dropwizard) project. 

Perform the following steps to migrate your application:
- Remove the  ``HealthBundle`` from your `Application`'s ``initialize()`` method.
- Remove the reference to ``HealthConfiguration`` from your application's `Configuration` class.
- Migrate YAML configuration previously associated with your ``HealthConfiguration`` field to instead be nested under a top-level ``health`` field (at the same level as ``server``, ``logging``, and ``metrics``).
  - Review the new and changed configuration options mentioned below.
  - Ensure that you don't already have a custom configuration class member using the ``health`` name or rename it to something else or nest it under another field in order not to conflict with the new top-level field.
- Remove the dependency on the ``io.dropwizard.modules:dropwizard-health`` library from your application's ``pom.xml`` file (or other build dependency definition file).

pstackle:
A question we should consider is what should we do with the admin /healthchecks endpoint and the `healthChecks`/`HealthCheckConfiguration` that is part of the `AdminFactory`. I think having both could end up being a common point of confusion.

docs/source/manual/configuration.rst
```diff
@@ -1274,27 +1274,33 @@ Health
...
+        type: detailed-json
+      responder:
+        type: servlet
 
 
 ============================== =======================  ====================================================================================================
 Name                           Default                  Description
 ============================== =======================  ====================================================================================================
 enabled                        true                     Flag indicating whether to enable health functionality or not.
+name                           (none)                   An identifier used for metric names, thread names, etc.
```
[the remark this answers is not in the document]
natalie-zamani: Yes! The original reason for this was due to metrics naming conflicts. However, we no longer have the issue of multiple bundles being register to account for, so I think that's a good idea to get rid of.

docs/source/manual/configuration.rst
```diff
@@ -1360,81 +1366,51 @@ failureAttempts                3                             The threshold of co
...
 
     health:
-      servlet:
-        type: default
-        cacheControlEnabled: true
-        contentType: "application/json"
-        healthyValue: {"status": "super happy! :)"}
-        unhealthyValue: {"status": "very sad. :("}
+      responseProvider:
+        type: detailed-json
```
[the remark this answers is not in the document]
natalie-zamani: Yep, just `json` makes more sense now.

natalie-zamani:
@pstackle @joschi I'm open to discussing more on the approach for what to do with the health servlet that comes included with the admin stuff, but for now I just added an option to not start the servlet if the user chooses not to. Removing it would be a `3.x` change, if that's what we wanted to do, but I'm not sure that's the best idea either.

Let me know if you think more docs or any other changes are warranted.

joschi (approved):
@natalie-zamani @pstackle @bowsersenior Thanks for the extensive discussion and the review.

The PR looks good to me. Please feel free to merge it. 😄

dropwizard-health/src/main/java/io/dropwizard/health/response/HealthResponse.java
```diff
@@ -1,57 +1,67 @@
...
-    protected final String message;
+public final class HealthResponse {
+    private final boolean healthy;
+    @Nonnull
+    private final String message;
+    @Nonnull
+    private final String contentType;
     @Nonnull
-    protected final String contentType;
+    private final Collection<HealthStateView> views;
```
pstackle, line 15: I don't see why `views` is a member of this. I see `HealthResponse` as an object that is passed from the HealthResponseProvider to a HealthResponder for the purposes of turning it into a materialized response. At this point, the responder should just be worried about how to turn the `message`, `contentType`, and healthy flag into something to send out over the wire (i.e. into http or whatever).

dropwizard-health/src/main/java/io/dropwizard/health/response/HealthResponse.java
```diff
@@ -1,57 +1,67 @@
...
 import java.util.Objects;
-import java.util.Optional;
+import javax.annotation.Nonnull;
 
-public abstract class HealthResponse {
-    protected final boolean healthy;
-    @Nullable
-    protected final String message;
+public final class HealthResponse {
+    private final boolean healthy;
```
pstackle, line 9: Maybe we should move the logic of what the response code is into the HealthResponseProvider? That would more easily open the door if someone wanted to try and make the health response non-binary (healthy vs unhealthy).

natalie-zamani: Great suggestion -- I think a boolean is unnecessarily inflexible here

natalie-zamani: Or at least, **only** a boolean. I'll throw in a status code additionally.

bowsersenior:
Thanks @natalie-zamani , this looks very solid!

PR #4193, opened 2021-08-16
Add anchor to note on config overrides to allow direct links

bowsersenior:
###### Problem:

Provide a link directly to the section in the docs on overriding configuration via system properties.

###### Solution:

Add an anchor to the note in the docs on overriding configs via system properties.


natalie-zamani (approved):

bowsersenior:
Built the docs locally and verified the `#config-override` anchor is generated and works at `manual/core.html#config-override`:

Screen Shot 2021-08-16 at 4 46 33 PM

PR #4198, opened 2021-08-18
Repeatable `@UnitOfWork` annotation

zUniQueX:
###### Problem:
In some situations, there is the need to access multiple databases in a Jersey resource method.
At the moment, only one `@UnitOfWork` annotation can be provided to a Jersey method and therefore only one database can be managed automatically.

###### Solution:
Starting from Java 8, annotations can be repeatable. So make `@UnitOfWork` a repeatable annotation.
If multiple annotations with the same value are provided, the last one gets used.

###### Result:
When a Jersey resource method is annotated with multiple `@UnitOfWork` annotations, for each of these a `UnitOfWorkAspect` gets instantiated and the hibernate `Session`s all get managed automatically.

### DIFF
--- a/docs/source/manual/hibernate.rst
+++ b/docs/source/manual/hibernate.rst
@@ -148,6 +148,21 @@ This will automatically open a session, begin a transaction, call ``findById``, 
 transaction, and finally close the session. If an exception is thrown, the transaction is rolled
 back.
 
+If you are using more than one Hibernate bundle in your application, you can repeat the ``@UnitOfWork`` annotation:
+
+.. code-block:: java
+
+    @GET
+    @Path("...")
+    @Timed
+    @UnitOfWork(value = "hibernate.<db-name-1>")
+    @UnitOfWork(value = "hibernate.<db-name-2>")
+    public ...
+
+In this case, the above specified behaviour applies to all given databases.
+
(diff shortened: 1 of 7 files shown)


zUniQueX:
I just changed some implementation details:
 1. The `@UnitsOfWork` annotation is not public any more. Therefore it is not necessary to evaluate this annotation type in the handling classes.
 2. The `UnitOfWorkApplicationListener` collects all `@UnitOfWork` annotations from the definition and the handling method of a `ResourceMethod`. If an annotation with the same value is provided on the definition and the handling method, then the definition method's annotation overrides the handling method's one, as this is the current behaviour. However, if at any of these methods multiple `@UnitOfWork` annotations with the same value are provided, an `IllegalStateException` is thrown.

I just wonder, if `IllegalStateException` is the correct exception type or if something like an `IllegalAnnotationException` should be thrown. Any opinions on that?

trytuna:
Would be nice if this gets merged. Thanks for the contribution!

jplock (approved):

joschi (approved):
LGTM. Sorry for the long lead time.

zUniQueX:
No problem, that gave me the time to think of a better implementation :wink:

PR #4201, opened 2021-08-19
Using Liquibase Scope for providing objects to CustomChanges

zUniQueX:
###### Problem:
When applying big data migrations with a Liquibase `CustomChange`, one cannot use Hibernate or JDBI at the moment.

###### Solution:
As of Liquibase 4.1.0, the `Scope` class for global configuration got an `enter` method, that takes a `Map<String, Object>` of objects, that are then available through the `get` method of the `Scope` class. Therefore the `MigrationsBundle` can be extended with a method for providing objects, that should be available inside the `CustomChange`.

###### Result:
When inside the `CustomChange`, the new `Scope` is entered and the provided objects are available.


dropwizard-migrations/src/main/java/io/dropwizard/migrations/DbCommand.java
```diff
@@ -50,8 +60,15 @@ public void configure(Subparser subparser) {
...
+        String scopeId = null;
+        if (scopedObjects != null) {
+            scopeId = Scope.enter(scopedObjects);
+        }
         final AbstractLiquibaseCommand<T> subcommand =
             requireNonNull(subcommands.get(namespace.getString(COMMAND_NAME_ATTR)), "Unable find the command");
         subcommand.run(namespace, liquibase);
+        if (scopeId != null) {
+            Scope.exit(scopeId);
+        }
```
joschi, line 72: Would it make sense to put at least the `subcommand.run(...)` statement in a `try` block and exit the scope in the `finally` block?

dropwizard-migrations/src/test/java/io/dropwizard/migrations/LiquibaseScopingTest.java
```diff
@@ -0,0 +1,122 @@
+package io.dropwizard.migrations;
+
+import liquibase.Scope;
+import liquibase.change.custom.CustomTaskChange;
+import liquibase.database.Database;
+import liquibase.database.jvm.JdbcConnection;
+import liquibase.exception.*;
```
joschi, line 7: Could we please list the imports explicitly?

joschi (commented):
@zUniQueX Thanks for your contribution. this looks very interesting! 😃 

I have two minor requests, otherwise LGTM. 👍

dropwizard-migrations/src/main/java/io/dropwizard/migrations/DbCommand.java
zUniQueX: I think, that this is a good idea. Otherwise a further call to the exit method will probably fail, because the scope id wouldn't be matching then

zUniQueX:
@joschi Thanks for the fast review. I pushed the changes as you suggested

joschi (approved):
LGTM. Thanks a lot! 😃

devoto13:
The example with `SessionFactory` does not seem to work, because the `DbMigrateCommand` is `ConfiguredCommand`, not `EnvironmentCommand`. Consequently, `HibernateBundle.run()` is never called and `SessionFactory` is always null.

PR #4211, opened 2021-08-26
Add non-abbreviated Logback layout conversion words to DropwizardLayout.

motlin:
###### Problem:
We use non-abbreviated conversion words in our logFormat.
`%highlight(%-5level) %green(%logger): %message%n%red(%rootException)`

Most abbreviations and longer forms work interchangeably (like `%c` and `%logger`)

However, switching between the abbreviations and longer forms for exception conversion words triggers a change in behavior, because Dropwizard only overrides the converters for the abbreviated forms.

###### Solution:
I changed DropwizardLayout to register additional non-abbreviated conversion words into ch.qos.logback.classic.PatternLayout#DEFAULT_CONVERTER_MAP.

Without this change, the map contains entries like:
"ex" -> "io.dropwizard.logging.PrefixedThrowableProxyConverter"
"exception" -> "ch.qos.logback.classic.pattern.ThrowableProxyConverter"

With this change, any Dropwizard overrides apply consistently to all synonyms.


jplock (dismissed):
LGTM! Thanks for the contribution.

motlin:
One option would be to add additional entries like:
"logbackException" -> "ch.qos.logback.classic.pattern.ThrowableProxyConverter"
"logbackEx" -> "ch.qos.logback.classic.pattern.ThrowableProxyConverter"
for the original logback converters. Another option is to accept this as-is, and leave that work to users. Thoughts?

mattnelson:
My recommendation would be to introduce new dropwizard specific conversion words so that all the standard logback conversion words[1] work as documented, which is where the configuration docs[2] defer to. The core docs[3] do mention the stack prefixing, but they are light on the specifics and/or any details on how to bypass prefixing, but I think that is because the config generally defer to the logback docs.

[1] http://logback.qos.ch/manual/layouts.html#conversionWord
[2] https://www.dropwizard.io/en/latest/manual/configuration.html#console
[3] https://www.dropwizard.io/en/latest/manual/core.html?highlight=prefix#log-format

motlin:
@mattnelson I think that would have been the best way to go in early versions of Dropwizard. Now it feels like a behavior change that could be confusing. I realize my change is also a behavior change, but it would only affect people who set a logFormat and don't use the abbreviations, like yourself. Users who have copied or modified the default log format will still be using abbreviations like rEx. Users who are using the long forms have already run into this behavior and are more likely to understand what's going on.

mattnelson:
> I think that would have been the best way to go in early versions of Dropwizard. Now it feels like a behavior change that could be confusing. I realize my change is also a behavior change, but it would only affect people who set a logFormat and don't use the abbreviations, like yourself. Users who have copied or modified the default log format will still be using abbreviations like rEx. Users who are using the long forms have already run into this behavior and are more likely to understand what's going on.

It wouldn't just impact them, it would take away the option of having non-prefixed exceptions entirely. I don't see an option to configure custom conversion words based on the way that DropwizardLayout used today.

> what are my next steps here?

With the work starting on 3.x, my recommendation would be to introduce new dropwizard/prefixed specific conversion words and make those the new defaults.

zUniQueX:
> With the work starting on 3.x, my recommendation would be to introduce new dropwizard/prefixed specific conversion words and make those the new defaults.

I agree with that. As we refer to the logback docs for the handling of conversion words, we should have those working as described there. We can introduce new ones for 3.x and 4.x and add a note in the migration guide to let users know about it.

zUniQueX:
> @jplock what are my next steps here?

We have several options here IMHO:
 - we can introduce your change, but therefore the docs should be updated
 - we leave the implementation as-is, but we should also update the docs, as the referenced Logback documentation isn't correct because the abbreviated conversion words are overridden
 - we can introduce new conversion words, defer to the Logback documentation for the default ones and add a section in the logging and migration docs about the new words

zUniQueX:
I've now revisited this PR and think that it might be a good idea to introduce new Dropwizard specific conversion words in Dropwizard 3.x. Therefore I've just opened #6207.

@motlin @mattnelson Maybe you can take a look at the new PR, if the change suits your needs.

zUniQueX:
Implemented via #6207

PR #4213, opened 2021-08-31
Add typed exceptions to authentication.

msmerc:
###### Problem:

I want authentication exceptions to have a custom error page - which requires them to be handled by my custom exception mapper. This can be done so long as the `Response` in the `WebApplicationException` doesn't have an entity, so I need to provide a custom `UnauthorizedHandler`. The only issue is that I cannot tell which exceptions are from the Dropwizard Auth package.

[Though to be fair, looking for 401 `UNAUTHORIZED` code is a reasonable workaround]

###### Solution:

Ensure all exceptions thrown by the authentication package extend a specific type.

###### Result:

I'm now able to build custom web pages correctly in my exception mapper.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthorizationException.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthorizationException.java
@@ -1,0 +1,10 @@
+package io.dropwizard.auth;
+
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
+
+public class AuthorizationException extends WebApplicationException {
+    public AuthorizationException(Response response) {
+        super(response);
+    }
+}
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentialAuthFilter.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentialAuthFilter.java
@@ -2,6 +2,7 @@ package io.dropwizard.auth.basic;
 
 import io.dropwizard.auth.AuthFilter;
 import io.dropwizard.auth.Authenticator;
+import io.dropwizard.auth.AuthorizationException;
 
 import javax.annotation.Nullable;
 import javax.annotation.Priority;
@@ -26,7 +27,7 @@ public class BasicCredentialAuthFilter<P extends Principal> extends AuthFilter<B
         final BasicCredentials credentials =
                 getCredentials(requestContext.getHeaders().getFirst(HttpHeaders.AUTHORIZATION));
         if (!authenticate(requestContext, credentials, SecurityContext.BASIC_AUTH)) {
-            throw new WebApplicationException(unauthorizedHandler.buildResponse(prefix, realm));
+            throw new AuthorizationException(unauthorizedHandler.buildResponse(prefix, realm));
         }
     }
 
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/oauth/OAuthCredentialAuthFilter.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/oauth/OAuthCredentialAuthFilter.java
@@ -2,6 +2,7 @@ package io.dropwizard.auth.oauth;
 
 import io.dropwizard.auth.AuthFilter;
 import io.dropwizard.auth.Authenticator;
+import io.dropwizard.auth.AuthorizationException;
 
 import javax.annotation.Nullable;
 import javax.annotation.Priority;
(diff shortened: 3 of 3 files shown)


pstackle:
I think calling it an `AuthorizationException` is confusing/misleading. It's unfortunate that HTTP chose poorly on some names, such as the Authorization header and the name associated with the status code 401, but at this point in the code, _authentication_ is happening, not _authorization_. Note the `authenticate` method name and the javadoc describing that it returns true if the request is _authenticated_. There is also already an `AuthenticationException` defined which is used internally to the `authenticate` method. Authorization is checked later in the flow by Jersey's `RolesAllowedDynamicFeature`.

devoto13:
I wonder what prevents you from building this custom error page response directly inside the custom `UnauthorizedHandler` if you already have one and avoiding the custom exception mapper altogether?

msmerc:
@devoto13 nothing - other than it makes code easier to follow if it's all in one place. It took me a while to figure out where the Unauthorized page was coming from since it's somewhat non-obvious!

msmerc:
@pstackle good call. `AuthenticationException` is used if there's an error during the authentication process and will cause an HTTP 500 response. However I'm happy to rename it. `ClientNotAuthenticatedWebException` maybe?

zUniQueX:
The JAX-RS API already has an [exception type](https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/jakarta/ws/rs/NotAuthorizedException.java) for this. I think throwing this would be a reasonable change.

pstackle:
@zUniQueX I think changing the exception thrown would work against the flexibility being provided by the `UnauthorizedHandler` concept that is included. 

The `NotAuthorizedException` you linked to is specifically associated with an HTTP status code of `401`, which _does_ match the status code used in the `DefaultUnauthorizedHandler`, but you'd be removing the ability for an app to change the associated status code with an authentication failure. In this case, I think the `WebApplicationException` was chosen specifically to allow for the customization.

zUniQueX:
@pstackle I don't think, that this is a big problem. In an `AuthFilter` only the code 401 _should_ be returned.

pstackle:
@zUniQueX I don't disagree that a `401` _should_ be returned as the HTTP status code for an authentication failure. But it seems like the dropwizard-auth `AuthFilter` was specifically designed to leave that detail to the `UnauthorizedHandler` and since that is an interface that is expected to be implemented by users of the library, changing the behavior now would be something that needs to be very well motivated, very well communicated, and likely shouldn't be done until a major version update. That being said, I'm not convinced that it should be changed. The default behavior is already doing the "right" thing by returning a 401, so making the change you are suggesting would just potentially break any users that were relying on the ability to customize the status code. 

In order to achieve a custom error page, which is the original ask, users can provide a custom `UnauthorizedHandler` that returns the desired response or can configure a filter to handle the 401 response code.

zUniQueX:
@pstackle I totally agree not breaking existing implementations with such a change. But we shouln't introduce a new exception for this. I think this also would be very confusing, if a new exception type just mirrors the existing `WebApplicationException` while it would have a misleading name related to the auth process although it won't set the correct status code for an unauthenticated request.

pstackle:
> But we shouln't introduce a new exception for this.

@zUniQueX I agree with you. 

My vote would be that this pull request is closed as it is a change that I don't think we should make. 

There are two different viable methods to achieve the ask for custom auth error pages:
- implement a custom `UnauthorizedHandler` which can be used to provide a custom response
- filter on 401 http status code

zUniQueX:
> My vote would be that this pull request is closed as it is a change that I don't think we should make.

@dropwizard/committers I agree with @pstackle. But as this PR has the improvement label, do you have any other opinion on that?

cue108:
Hi there!
I am not a committer here thus I cannot make the review.
Anyway I have an own artefact waiting for this milestone to complete and release.
If there is anything I can do to accelerate the procedure, please let me know.
cheers!

zUniQueX:
If we really want to integrate this feature, we maybe could use an other approach (haven't tested yet). What about extending the `UnauthorizedHandler` interface with a method
```java
default <T extends WebApplicationException> T buildException(String prefix, String realm) {
    return new WebApplicationException(buildResponse(prefix, realm));
}
```
and throw this exception in the filters?

Then one can override this method for a custom exception type.

zUniQueX:
@msmerc I've revisited this and implemented the aforementioned approach. We could do this, but then we would have to document that users would only profit from overriding it, when they don't set response entities. As you described, exception mappers only process exceptions with no response entities set.

Then we could argue about the need of overriding the `buildResponse` method, as this could be inlined in the `buildException` method. So this could also get a `default` method.

@dropwizard/committers Any opinions on that? Should we extend the interface to allow custom overriding of the exception type?

If there is support for this change, I would suggest to provide a `default` implementation of the `buildResponse` method with the current one from the `DefaultUnauthorizedHandler` and with a `buildException` method like this:
```java
default WebApplicationException buildException(String prefix, String realm) {
    return new WebApplicationException(buildResponse(prefix, realm));
}
```

zUniQueX:
@msmerc @pstackle I played around a bit with this and came up with the implementation in #5026. Maybe you can take a look at that.

zUniQueX:
Implemented via #5026

PR #4256, opened 2021-09-12
Remove unnecessary eq() Mockito references

rhowe:
This isn't needed as equality is the default matcher type. Picked up by Sonar.


joschi (approved):
Thanks!

PR #4261, opened 2021-09-13
Use more specific AssertJ assertions in tests

rhowe:
As suggested by Sonar

This is about 1/3 of the assertion tweaks Sonar would like to see. If you like this, I'll raise a couple more similarly-sized PRs.


jplock (approved):

joschi (approved):
Thanks!

PR #4271, opened 2021-09-14
Mark more test classes as non-public

rhowe:
As identified by Sonar - unnecessary to have these public since JUnit 5


joschi (approved):
Thanks!

PR #4284, opened 2021-09-19
Replace doesNotThrow assertions with more explicit checks

rhowe:
Rather than checking some code doesn't explode, it's better to check that it does something.

### DIFF
--- a/dropwizard-jackson/src/test/java/io/dropwizard/jackson/JacksonTest.java
+++ b/dropwizard-jackson/src/test/java/io/dropwizard/jackson/JacksonTest.java
@@ -1,6 +1,7 @@
 package io.dropwizard.jackson;
 
 import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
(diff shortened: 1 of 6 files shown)


dropwizard-jackson/src/test/java/io/dropwizard/jackson/JacksonTest.java
```diff
@@ -48,11 +48,11 @@ void objectMapperSerializesNullValues() throws IOException {
...
     @Test
-    void objectMapperIgnoresUnknownProperties() {
-        assertThatCode(() ->
-            Jackson.newObjectMapper()
+    void objectMapperIgnoresUnknownProperties() throws JsonProcessingException {
+        assertThat(Jackson.newObjectMapper()
                 .readValue("{\"unknown\": 4711, \"path\": \"/var/log/app/server.log\"}", LogMetadata.class)
-        ).doesNotThrowAnyException();
+                .path)
+            .hasToString("/var/log/app/server.log");
```
joschi, line 55: @rhowe This change breaks the test on Windows: https://ci.appveyor.com/project/dropwizard/dropwizard/builds/40883668#L6261

rhowe: Honestly, these toy OSes. When will we learn? I'll take a look, thanks

joschi (approved):

PR #4304, opened 2021-09-25
Replace dropwizard-utils Throwables with commons-lang3 ExceptionUtils

rhowe:
###### Problem:
The `Throwables` helper has no tests (and I'm not completely convinced that the code is correct).

###### Solution:
The e2e tests were the only consumer of the (untested) Throwables class.

Replace it with commons-lang3's ExceptionUtils, which is already used elsewhere

###### Result:
Increased overall test coverage and hopefully still-passing e2e tests.
One additional consumer of `commons-lang3`, which I had a good go at trying to remove from the codebase, but at least it's not adding it into the dependency tree for users of Dropwizard.


joschi (changes requested):

I'm a bit reluctant to remove a public class/method without deprecating it first and give users a chance to migrate to an alternative.

This is why I'm labeling this issue as "blocked" and scheduled it for Dropwizard 3.0.0.

If you want to change this PR to deprecate `io.dropwizard.util.Throwables` instead of removing it, I'm happy to approve and merge it.

rhowe:
> @rhowe Thanks for your contribution!

Thanks for Dropwizard!

> If you want to change this PR to deprecate `io.dropwizard.util.Throwables` instead of removing it, I'm happy to approve and merge it.

Sounds good to me - I've revised the PR. Did I get the javadoc correct? I'm not exactly a javadocwizard

PR #4326, opened 2021-10-04
Add JsonProperty annotations to constructor parameters of Person in `dropwizard-jackson`

zUniQueX:
###### Problem:
The tests of `dropwizard-jackson` fail after upgrading Jackson to version 2.13.0. Thereby the `Person` class cannot be deserialized.

###### Solution:
The constructor parameters of the `Person` class get annotated with proper `@JsonProperty` annotations.

###### Result:
The test works again.

### DIFF
--- a/dropwizard-jackson/src/test/java/io/dropwizard/jackson/ParanamerModuleTest.java
+++ b/dropwizard-jackson/src/test/java/io/dropwizard/jackson/ParanamerModuleTest.java
@@ -20,7 +20,7 @@ class ParanamerModuleTest {
     }
 
     @Test
-    void deserializePersonWithoutAnnotations() throws IOException {
+    void deserializePerson() throws IOException {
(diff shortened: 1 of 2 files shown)


joschi:
@zUniQueX Thanks for your contribution!

I'll close this pull request because this is a regression in the [Jackson parameter-names](https://github.com/FasterXML/jackson-modules-java8/tree/jackson-modules-java8-2.13.0/parameter-names) module and nothing we should fix on our end.

joschi:
> The tests of `dropwizard-jackson` fail after upgrading Jackson to version 2.13.0. Thereby the `Person` class cannot be deserialized.

I'm unable to reproduce this (and the build with Maven on GitHub Actions is also green).

Does the issue also happen when you're running the respective test locally?
```
./mvnw test -pl dropwizard-jackson -Dtest=ParanamerModuleTest
```

PR #4338, opened 2021-10-08
2.1.0 deprecations

zUniQueX:
Since most dependencies have released 'jakarta versions' (only logback hasn't at the moment), Dropwizard 3.0.0 could be built soon. But therefore some properties have to be removed:
 1. Upgrade to Jetty 11: excluding user agents for GZIP is not supported any more. Therefore remove it.
 1. Upgrade to Httpclient 5 (possible, as only the class `LocalizationMessages` from `jersey-apache-connector` seems to be used):
    - The `RequestConfig.Builder` doesn't support `normalizeURI` any more
    - `ServiceUnavailableRetryStrategy` and `HttpRequestRetryHandler` are combined in `HttpRequestRetryStrategy`. Therefore deprecate using `ServiceUnavailableRetryStrategy`
    - `HttpProcessor` cannot be provided any more
 1. The `DefaultHealthFactory` defines two getter and setter methods for the property `healthChecks`. Therefore deprecate one

The dependency updates will probably introduce more changes, but this is what I recognized so far.

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
@@ -241,7 +241,9 @@ public class HttpClientBuilder {
      *
      * @param httpProcessor a {@link HttpProcessor} instance
      * @return {@code} this
+     * @deprecated
      */
+    @Deprecated
     public HttpClientBuilder using(HttpProcessor httpProcessor) {
         this.httpProcessor = httpProcessor;
         return this;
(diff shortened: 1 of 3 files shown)


javajosh:
Hey there. I came here looking for some idea when jetty would be upgraded past 9  - I for one am looking forward to the upgrade!

joschi:
@javajosh Jetty 10 (`javax.servlet` namespace) requires Java 11 or higher as well as Jetty 11 (`jakarta.servlet` namespace).

Dropwizard 2.x is targeting Java 8 or higher, so Dropwizard 2.x most likely won't upgrade to Jetty 10 or later.

zUniQueX:
Additionally, the upgrade seems to be impossible at the moment. There are conflicts with `logback-access`. See https://github.com/eclipse/jetty.project/issues/5996#issuecomment-784371682 for details

PR #4386, opened 2021-10-29
Add flattening to AccessJsonLayout

### LINKED ISSUE #4396: Add flattening to AccessJsonLayout
Today there are three categories of information that end up as nested objects/dictionaries in the json formatted access logs (when configured):

* requestHeaders
* responseHeaders
* requestAttributes

For simplicity and alignment across other micro-services in out setup, the ability to flatten these nested objects into the root json object would be VERY nice.

The use-case here is to emit access logs aligned across micro-services implemented using different technologies. 
Other options would be to do post-processing of the logs downstream, but  IMO it is much preferred to emit logs the way you want them to look right away:

To get the json logs formatted right from the start, would mean that there is no extra work downstream:
- No need to figure out how to filter/transform externally
- Explain why there are extra post processing steps [to the team]
- In the worst case explain to the "enterprise logging department" somewhere in the org to do an extra transformation.

georgespalding:
Closes  #4396
Make flattening of request-headers, request-attributes and response-headers configurable for AccessJsonLayout

###### Problem:
When configuring f.x RequestAttributes to be emitted in the json access log it is not possible to get a "flat" object where all values are scalars.

###### Solution:
Allow users of AccessJsonLayoutFactory/AccessJsonLayout to configure flattening of the complex fields:
* RequestHeaders
* RequestAttributes
* ResponseHeaders

Note that the caller needs to understand that flattening could cause collisions since there is nothing preventing `RequestHeaders`, `ResponseHeaders` or `RequestAttribute` containing a key that already exists

###### Result:
Given a config like:
```
layout:
  type: access-json
  flattenRequestHeaders: true
  requestHeaders:
    - X-Request-Id
  flattenResponseHeaders: true
  responseHeaders:
    - X-Response-Id
  flattenRequestAttributes: true
  requestAttributes:
    - SomeAttributeName
  customFieldNames:
    SomeAttributeName: some-attribute-name
```

Output Json could look like:
```json
{
"some-attribute-name": "...",
"X-Request-Id": "...",
"X-Response-Id": "..."
}
```

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -951,7 +951,7 @@ includes                 (timestamp, level,
                                                 - ``message``     *true*   Whether to include the formatted message as the ``message`` field.
                                                 - ``exception``   *true*   Whether to log exceptions. If the property enabled and there is an exception, it will be formatted to a string as the ``exception`` field.
                                                 - ``contextName`` *false*  Whether to include the logging context name as the ``context`` field .
-customFieldNames         (empty)                Map of field name replacements . For example ``(requestTime:request_time, userAgent:user_agent)``.
+customFieldNames         (empty)                Map of field name replacements . For example ``(requestTime:request_time, userAgent:user_agent)``. Custom field name collisions will result in one value being picked.
 additionalFields         (empty)                Map of fields to add in the JSON map.
 includesMdcKeys          (empty)                Set of MDC keys which should be included in the JSON map. By default includes everything.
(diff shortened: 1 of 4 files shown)


georgespalding:
@joschi Any chances of bringing this to an approval?

Let me know if there's something more.

georgespalding:
We ended up creating a custom GCPAccessJsonLayout that sets the fields that match https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity 

image

PR #4389, opened 2021-10-30
Move jersey and servlet filters to their own section in the docs

slavadev:
###### Problem:
Currently, in the docs, Jersey filters and Servlet filters are placed in the Representations section, and it seems like they don't belong to it.

###### Solution:
Add a separate section for filters


joschi (approved):
@slavadev Thanks!

PR #4416, opened 2021-11-10
Fix `BadLogTest`

zUniQueX:
###### Problem:
`dropwizard-e2e`'s `BadLogTest` is failing on windows, because a file is not properly closed on application shutdown.

###### Solution:
Stop all logging appenders in the `DropwizardTestSupport.after()` method to free the occupied resources.

###### Result:
Used resources are properly freed and therefore the `BadLogTest` can be re-enabled for Windows.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
@@ -10,6 +10,7 @@ import io.dropwizard.cli.ServerCommand;
 import io.dropwizard.configuration.ConfigurationSourceProvider;
 import io.dropwizard.configuration.YamlConfigurationFactory;
 import io.dropwizard.lifecycle.Managed;
+import io.dropwizard.logging.LoggingUtil;
 import io.dropwizard.setup.Bootstrap;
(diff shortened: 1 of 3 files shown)


dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
```diff
@@ -238,6 +240,7 @@ public void after() {
             stopIfRequired();
```
rhowe, line 240: This calls `configuration.getLoggingFactory().reset()` which does

```
loggerContext.stop();
final Logger logger = loggerContext.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
logger.detachAndStopAllAppenders();
```

so in theory the appenders should already be stopped in `after()`?

zUniQueX: Good point. But this only holds, if the configuration is set. When debugging the test, the configuration is sometimes `null`. The `Application` calls `bootstrapLogging()` anyway. Therefore we should ensure, that the appenders are stopped, if the configuration was not provided.

jplock (approved):

PR #4423, opened 2021-11-15
`SocketTimeoutException` instead of `ConnectException` thrown on Windows

zUniQueX:
When executing the tests on Windows with Java 11, a `SocketTimeoutException` is thrown in a test of `dropwizard-health` instead of a `ConnectException`. However, the test completes successfully when running with Java 17 or when running the test on Linux.

Since the framework will probably support Java 11 for quite some time, the test should succeed on Windows. Therefore the test is modified to allow both exception types.


joschi (approved):
@zUniQueX Thanks!

PR #4424, opened 2021-11-15
Fix build on Windows with Java 17

zUniQueX:
###### Problem:
When building the project with Java 17 on Windows, an error is thrown because of an 'invalid flag'.

###### Solution:
Windows doesn't seem to like new lines in argument definitions, so remove all new lines and place all options on one line.

###### Result:
The build will complete on Windows.


joschi (approved):
@zUniQueX Thanks!

PR #4438, opened 2021-11-20
Improve dropwizard-logging socket-based tests

rhowe:
These tests relied on `TcpServer` and counting the number of log messages
received.

Replace direct `Thread` usage with `ExecutorService` and `Future` so that we
can instead retrieve the log messages actually received and perform assertions
on them.

Do the same for UDP-based tests, introducing `UdpServer`.

Note that the logs in `UdpSocketAppenderFactoryTest` have trailing newlines, unlike those in `TcpSocketAppenderFactoryTest` - is that expected?

### DIFF
--- a/dropwizard-logging/src/test/java/io/dropwizard/logging/ResilientSocketOutputStreamTest.java
+++ b/dropwizard-logging/src/test/java/io/dropwizard/logging/ResilientSocketOutputStreamTest.java
@@ -1,61 +1,26 @@
 package io.dropwizard.logging;
 
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import javax.net.SocketFactory;
 import java.io.IOException;
 import java.net.ServerSocket;
-import java.net.Socket;
 import java.nio.charset.StandardCharsets;
-import java.util.concurrent.CountDownLatch;
+import java.util.List;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
 
 class ResilientSocketOutputStreamTest {
-
-    private ResilientSocketOutputStream resilientSocketOutputStream;
-
-    private Thread thread;
-    private ServerSocket ss;
-    private CountDownLatch latch = new CountDownLatch(1);
-
-    @BeforeEach
-    void setUp() throws Exception {
-        ss = new ServerSocket(0);
-        thread = new Thread(() -> {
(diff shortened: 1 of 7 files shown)


zUniQueX:
One suggestion: maybe you could use a list appender like in the `SelfValidationTest` of the `dropwizard-validation` package. This could probably avoid the need of the extraction from the log message.

rhowe:
> One suggestion: maybe you could use a list appender like in the `SelfValidationTest` of the `dropwizard-validation` package. This could probably avoid the need of the extraction from the log message.

Hmm, the list appender would let us intercept what was submitted to logback, but don't we want to check that the string submitted to the logger is received by the remote log server? I suppose the cleaner thing would be to change the log format to be more deterministic by removing the timestamp.

zUniQueX:
Because the list appender works only locally, this should have the same effect as putting the strings into the list manually (at least, if we assume that logback works correctly). The received messages should then only equal the local log messages, if the remote log server sends all messages back

rhowe:
> Because the list appender works only locally, this should have the same effect as putting the strings into the list manually (at least, if we assume that logback works correctly). The received messages should then only equal the local log messages, if the remote log server sends all messages back

I had a go, but the `ILoggingEvent` is the log event before any formatting gets applied, whereas the remote log server receives a formatted message so we'd have to do some juggling either way. I think in this case comparing the message which gets put into the logging framework with the text received at the far end is probably OK?

zUniQueX:
I just thought it could be a good idea to have the whole messages for comparing logging classes and levels and to simplify the comparisons. But the equality of the log messages should be enough to verify the correctness of the remote logging.

One _could_ modify the list appender to store a list of Strings to be close to your implementation. But this would make this test much more difficult to understand as it would implement basically the same thing in a harder way.

So I think your current implementation is the most readable one.

rhowe:
> So I think your current implementation is the most readable one.

Yeah it would get pretty complicated I think.

I realised that some other tests effectively had an inline `TcpServer` or variant thereof and so I've rewritten them to use the rewritten `TcpServer` and a newly-created `UdpServer`.

rhowe:
Ugh, UDP test is failing on Windows. One for tomorrow, assuming Advent of Code day 1 is easy :joy:

rhowe:
> Ugh, UDP test is failing on Windows. One for tomorrow, assuming Advent of Code day 1 is easy joy

Alright, I think that ought to be fixed now. The logs in `UdpSocketAppenderFactoryTest` had trailing newlines which were platform-dependent. I've dealt with that in the test although it is a difference in behaviour compared to the TCP logs, which don't have newlines. Is that expected?

rhowe:
Looks like `appShouldRecoverOnceCriticalCheckReturnsToHealthyStatus` is a bit flaky

jplock (approved):

Issue #4176, opened 2021-07-28
If request '“Content-Encoding:gzip” with raw string in body instead of binary , the response error code should be 400 instead of 500


wenjin-huang:
If the request has the “Content-Encoding:gzip” in the header, but instead of having binary .gz file as payload, it has a raw string as body payload. The response is 500 error, but it should be 400 BAD request. 
```
o.eclipse.jetty.server.HttpChannelState - unhandled due to prior sendError
! java.util.zip.ZipException: Invalid gzip bytes
! at org.eclipse.jetty.http.GZIPContentDecoder.decodeChunks(GZIPContentDecoder.java:272)
! ... 98 common frames omitted
! Causing: java.lang.RuntimeException: java.util.zip.ZipException: Invalid gzip bytes
! at org.eclipse.jetty.http.GZIPContentDecoder.decodeChunks(GZIPContentDecoder.java:402)
! at org.eclipse.jetty.server.handler.gzip.GzipHttpInputInterceptor$Decoder.decodeChunks(GzipHttpInputInterceptor.java:91)
! at org.eclipse.jetty.server.handler.gzip.GzipHttpInputInterceptor.readFrom(GzipHttpInputInterceptor.java:45)
! at org.eclipse.jetty.server.HttpInput.intercept(HttpInput.java:500)
! ... 95 common frames omitted
! Causing: java.io.IOException: Bad content
! at org.eclipse.jetty.server.HttpInput.intercept(HttpInput.java:504)
! at org.eclipse.jetty.server.HttpInput.nextInterceptedContent(HttpInput.java:464)
! at org.eclipse.jetty.server.HttpInput.addContent(HttpInput.java:629)
! at org.eclipse.jetty.server.HttpChannel.onContent(HttpChannel.java:778)
! at org.eclipse.jetty.server.HttpChannelOverHttp.content(HttpChannelOverHttp.java:245)
! at org.eclipse.jetty.http.HttpParser.parseContent(HttpParser.java:1710)
! at org.eclipse.jetty.http.HttpParser.parseNext(HttpParser.java:1545)
! at org.eclipse.jetty.server.HttpConnection.parseRequestBuffer(HttpConnection.java:376)
! at org.eclipse.jetty.server.HttpConnection.fillAndParseForContent(HttpConnection.java:323)
! at org.eclipse.jetty.server.HttpInputOverHTTP.produceContent(HttpInputOverHTTP.java:33)
[...]

PR #4490, opened 2021-12-12
Adjust status code on invalid gzip input

zUniQueX:
###### Problem:
As denoted in #4176, the status code is wrong if invalid gzip input is parsed. The framework has the class `ZipExceptionHandlingInputStream`, which seems to provide support for this, but this class is never used.

###### Solution:
Provide a custom `GzipHandler` catching `ZipException`s and `EOFException`s to return a status code 400 instead of 500. Additionally, as the unused `ZipExceptionHandlingInputStream` is package private, it should be safe to remove it.

###### Result:
The status code will be correct on invalid gzip input.


jplock (approved):

jplock:
Was `ZipExceptionHandlingInputStream` just leftover code from a previous implementation or are there any references to it?

zUniQueX:
I couldn't find any reference to it. And since it is not `public`, it couldn't be used outside the framework.

Issue #3340, opened 2020-06-19
JUnit 5 test with HttpSession fails with DropwizardExtensionsSupport


brentryan:
We came across a scenario where we wanted to write a unit test for `@Session HttpSession session` being injected into a resource.

We were trying to mimic the behavior of DockerIntegrationTest.java which is included with dropwizard-example project. You can mimic the failure by simply adding `@Session HttpSession session` to `HelloWorldResource` and re-running the test.

It was obvious if this was a limitation in the underlying test container being used here or if there was a way to enable session support in a test.

Here's the output of the error:
```
ERROR [2020-06-19 21:41:38,961] io.dropwizard.jersey.errors.IllegalStateExceptionMapper: Error handling a request: 77e5c65c7de94fa7
! java.lang.IllegalStateException: No SessionManager
! at org.eclipse.jetty.server.Request.getSession(Request.java:1605)
! at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
! at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
! at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
! at java.base/java.lang.reflect.Method.invoke(Method.java:566)
! at org.glassfish.hk2.utilities.reflection.ReflectionHelper.invoke(ReflectionHelper.java:1268)
! at org.jvnet.hk2.internal.MethodInterceptorImpl.internalInvoke(MethodInterceptorImpl.java:85)
! at org.jvnet.hk2.internal.MethodInterceptorImpl.invoke(MethodInterceptorImpl.java:101)
! at org.jvnet.hk2.internal.MethodInterceptorInvocationHandler.invoke(MethodInterceptorInvocationHandler.java:39)
! at com.sun.proxy.$Proxy110.getSession(Unknown Source)
! at io.dropwizard.jersey.sessions.HttpSessionFactory.provide(HttpSessionFactory.java:22)
! at io.dropwizard.jersey.sessions.SessionFactoryProvider.lambda$createValueProvider$0(SessionFactoryProvider.java:40)
! at org.glassfish.jersey.server.spi.internal.ParamValueFactoryWithSource.apply(ParamValueFactoryWithSource.java:50)
[...]

PR #4539, opened 2021-12-23
Session injection

zUniQueX:
###### Problem:
As described in #3340, with the default configuration no session can be injected, although basic support should be provided (see `dropwizard-jersey`s `session` package).

###### Solution:
Register a `SessionHandler` to the application's `MutableServletContextHandler` to support session injection.

###### Result:
When adding a parameter like `@Session HttpSession session` to a resource method, no `IllegalStateException` gets thrown and the session gets injected.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/setup/Environment.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/setup/Environment.java
@@ -17,6 +17,7 @@ import io.dropwizard.jetty.MutableServletContextHandler;
 import io.dropwizard.jetty.setup.ServletEnvironment;
 import io.dropwizard.lifecycle.setup.LifecycleEnvironment;
 import io.dropwizard.validation.InjectValidatorFeature;
+import org.eclipse.jetty.server.session.SessionHandler;
 
 import javax.annotation.Nullable;
 import javax.servlet.Servlet;
@@ -76,6 +77,7 @@ public class Environment {
 
         this.servletContext = new MutableServletContextHandler();
         servletContext.setClassLoader(classLoader);
+        servletContext.setSessionHandler(new SessionHandler());
         this.servletEnvironment = new ServletEnvironment(servletContext);
 
         this.adminContext = new MutableServletContextHandler();
--- a/dropwizard-e2e/src/main/java/com/example/httpsessions/HttpSessionsApp.java
+++ b/dropwizard-e2e/src/main/java/com/example/httpsessions/HttpSessionsApp.java
@@ -1,0 +1,12 @@
+package com.example.httpsessions;
+
+import io.dropwizard.Application;
+import io.dropwizard.Configuration;
+import io.dropwizard.setup.Environment;
+
+public class HttpSessionsApp extends Application<Configuration> {
+    @Override
+    public void run(Configuration configuration, Environment environment) throws Exception {
+        environment.jersey().register(new HttpSessionsResource());
+    }
+}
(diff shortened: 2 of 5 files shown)


mattnelson:
> with the default configuration no session can be injected,

I would venture a guess that the vast majority of dropwizard services are stateless and would not benefit from having a session handler registered by default.

> although basic support should be provided

There is already a method provided for basic support for sessions:

https://github.com/dropwizard/dropwizard/blob/ee494e6704fdf5a2837e21019c64df47496fc493/dropwizard-jetty/src/main/java/io/dropwizard/jetty/setup/ServletEnvironment.java#L181-L189

> As described in #3340

The problem described in #3340 was specifically around a gap in the ability to use sessions in the unit testing extensions. Is there an option to keep the scope of this focused on just the testing flow?

zUniQueX:
I agree, that most users wouldn't use session support. But as the `dropwizard-jersey` module has several classes for session support, users _could_ use sessions.

The default behaviour for this, however, is inconsistent. The `DropwizardResourceConfig` registeres the binder of `SessionFactoryProvider` and the `@Session` annotations therefore get evaluated. If a user annotates a parameter with the `@Session` annotation without this PR, an `IllegalStateException` is thrown because of the absence of a `SessionHandler`.

Therefore I'd either suggest registering the `SessionHandler` by default or removing the `SessionFactoryProvider.Binder` and providing a convenience method for enabling session support.

zUniQueX:
Additionally, the documentation for session support could definitely be improved (created) ;)

zUniQueX:
@mattnelson I revisited this and thought of a better way for registering the `SessionHandler`. I wouldn't like the `IllegalStateException` at the abscence of a `SessionHandler` to be thrown. The solution I now think of (haven't tested yet) providing a `DynamicFeature` evaluating all resource methods and registering a `SessionHandler`, if a `@Session` annotation is found. Then we  would work around having a `SessionHandler` set for all users but also would support sessions in the default configuration.

I would then also suggest adding a log message, if no other `SessionHandler` was set, that a default one is provided.

What do you think of this?

zUniQueX:
After testing the approach with the `DynamicFeature`, I had to realize, that the server is then already started and therefore no `SessionHandler` can be set any more.

The cleanest way I now see for this (and which works in the tests) is to do reflection at application startup. Adding the following code to `EnvironmentCommand#run(Bootstrap, Namespace, T)` would then do the job:
```java
try (ScanResult scanResult = new ClassGraph()
        .acceptPackages(bootstrap.getApplication().getClass().getPackage().getName())
        .enableClassInfo()
        .enableMethodInfo()
        .enableAnnotationInfo()
        .scan()
    ) {
        ClassInfoList classInfos = scanResult.getClassesWithMethodParameterAnnotation(Session.class);
        if (!classInfos.isEmpty()) {
            environment.servlets().setSessionHandler(new SessionHandler());
        }
}
```

But this would therefore slow down the startup, which I think is bad for a Dropwizard application, as speed is a core feature of Dropwizard.

@joschi  @jplock what do you think about this? I think, speed is more important than session support in a REST application, but I'm open for other opinions.

mattnelson:
> I think, speed is more important than session support in a REST application, but I'm open for other opinions.

Could create a new `dropwizard-sessions` maven module which would allow for pre start registration through a bundle.

zUniQueX:
@mattnelson This basically wouldn't change anything, if the current codebase stays unchanged, since users already have an opt-in with the `environment.servlets().setSessionHandler(...)` method. My consideration in this PR was just, if we could have auto configuration for this.

But we could discuss providing a `dropwizard-sessions` module, if the package `io.dropwizard.jersey.sessions` gets moved to that. Then one has to care about setting a `SessionHandler` implicitly, because a dependency must be added for the `@Session` annotation.

zUniQueX:
@joschi Thanks for the feedback. I just wanted to avoid a situation, where users can't use the feature due to the lack of information. Therefore I previously decided on enabling it by default, but I agree most users won't profit from this.

I raised #4681 to update the docs and provide an e2e test for this.

Issue #4238, opened 2021-09-05
Injection into DynamicFeature registered instances


zUniQueX:
Recently, I tried writing an `AuthFilter` instance in which I wanted to use an injected `HttpServletRequest`. As denoted in #1630, injection does not work on instances registered in the DynamicFeature's context.

However, one can provide the injection manually. With the following code I get the `HttpServletRequest` injected into my `AuthFilter` instance:

```java
public class InjectingFeature implements Feature {
    private static class InjectingDynamicFeature implements DynamicFeature {
        private final InjectionManager injectionManager;

        InjectingDynamicFeature(InjectionManager manager) {
            injectionManager = manager;
        }

        @Override
        public void configure(ResourceInfo resourceInfo, FeatureContext context) {
            context.register(new InjectingContainerRequestFilter(injectionManager, context.getConfiguration()));
        }
    }

    @Priority(1)
    private static class InjectingContainerRequestFilter implements ContainerRequestFilter {
        private final InjectionManager injectionManager;
        private final Configuration configuration;

        InjectingContainerRequestFilter(InjectionManager manager, Configuration config) {
            injectionManager = manager;
            configuration = config;
        }

        @Override
        public void filter(ContainerRequestContext requestContext) throws IOException {
            Set<Object> instances = configuration.getInstances();
            for (Object instance : instances) {
                if (!this.getClass().equals(instance.getClass())) {
                    injectionManager.inject(instance);
                }
            }
        }
    }

    @Override
    public boolean configure(FeatureContext context) {
        context.register(new InjectingDynamicFeature(InjectionManagerProvider.getInjectionManager(context)));
        return true;
    }
}
```

I wonder, if the framework could/should provide such an implementation to make injection work on those instances registered in `DynamicFeature`s.

Maybe it would also be reasonable to provide this only for `AuthFilter` instances, as few other components are registered in this way. Then, the injection could also be extended to the `Authenticator` and `Authorizer` instances.

PR #4563, opened 2021-12-30
Provide dependency injection for auth filter objects

zUniQueX:
Raised for consideration regarding issue #4238.


joschi (approved):
LGTM. Thanks! 😄

pstackle:
@zUniQueX Sorry that I didn't get a chance to review this sooner, but I just tried out v2.1.0-beta.8 and this change broke the ability to use dropwizard-auth's DynamicAuthFeature with a non-AuthFilter-based ContainerRequestFilter.

Was that intended?

Looking closer at this change, I'm not sure it is necessary to end up wrapping the user-provided ContainerRequestFilter with another ContainerRequestFilter (InjectingFilter). 

Instead, you should be able to do the injection within the AuthDynamicFeature class before registering the auth filter instance. For example, in the `registerAuthFilter()`  method, you could call `injectionManager.inject(authFilter)` just before calling `context.register(authFilter)` and that should end up populating that instance with the necessary jersey proxy class that will then populate a `@Context HttpServletRequest httpRequest` field with the correct value on each request.

zUniQueX:
> Instead, you should be able to do the injection within the AuthDynamicFeature class before registering the auth filter instance. For example, in the `registerAuthFilter()`  method, you could call `injectionManager.inject(authFilter)` just before calling `context.register(authFilter)` and that should end up populating that instance with the necessary jersey proxy class that will then populate a `@Context HttpServletRequest httpRequest` field with the correct value on each request. 

@pstackle I'm not sure, if this would work. We cannot create a proxy of the user defined auth filter ourselves (the user can define own constructors) and therefore we cannot handle the injection within a proxy.

The `DynamicFeature` configuration process is done once at application startup, so we would be only injecting once into the object. I'm not sure this really is what we want. Or is it enough for every injected object that it's done once to return the current value on every access? Therefore I'm not deep enough into HK2.

We could modify the `InjectingFilter` to inject into those filters too, but to skip the injection into `Authenticator` and `Authorizer` instances. Then we don't have to throw an exception there.

zUniQueX:
@pstackle I looked into the Jersey docs, but I can't find something on the topic when and how often manual DI should be done.

I think the `InjectingFilter` would however provide more guarantees. If we use your approach to do the DI once at configuration of the `DynamicFeature` and the user would however set an injected object to `null`, the object won't get injected on the next request. When using the `InjectingFilter`, a user would get a valid object in this situation.

I can't imagine a case where this is relevant at the moment, but if we can provide this degree of freedom, I think we shouldn't take it out. As both approaches (nearly) do the same things, I would leave the code as-is for now, but I don't have an other good reason for that. If you have more experience with HK2 and this would give us a benefit, I'm happy to raise a PR for that.

pstackle:
@zUniQueX Thanks for digging into this. I know that the Jersey DI stuff is pretty complicated and I usually end up struggling quite a bit with it myself. But, luckily I've done this exact thing before and it works correctly. Since I don't expect you to just take my word for it, I've created the following pull request that showcases the behavior I was describing and adds a couple tests that proves that the injected values are in fact updated with the per-request values. 

PR: https://github.com/dropwizard/dropwizard/pull/5197

> If we use your approach to do the DI once at configuration of the DynamicFeature and the user would however set an injected object to null, the object won't get injected on the next request. 

What is the use case you are thinking of where the user would set an injected field value to null? I think that statement would apply for any DI behavior and I'm not sure it makes sense for us to try and defend against that. By doing the DI setup once, during startup, we avoid completely the overhead of the redundant injection on every request. Given the amount of reflection that is used in HK2's injection code, I'd imagine this is a costly operation and I would not want to start incurring that overhead inline with every authenticated request.

zUniQueX:
> What is the use case you are thinking of where the user would set an injected field value to null? I think that statement would apply for any DI behavior and I'm not sure it makes sense for us to try and defend against that. By doing the DI setup once, during startup, we avoid completely the overhead of the redundant injection on every request. Given the amount of reflection that is used in HK2's injection code, I'd imagine this is a costly operation and I would not want to start incurring that overhead inline with every authenticated request.

As I mentioned, I don't have a use case for it at the moment. I just wanted to get a consistent behavior. When using DI on methods, one gets a new value on every invocation. But actually it shouldn't be necessary on classes. Reducing runtime overhead is a good point.

I was just curious about the case mentioned in your [second link](https://psamsotha.github.io/jersey/2015/12/16/request-scope-into-singleton-scope.html). As the last section describes, there seems to be a bug with the scopes, but that should be good for us.

PR #4565, opened 2022-01-01
Deprecate the `Size` validators in favour of `DataSize` validators

rhowe:
`Size` is deprecated in favour of `DataSize` so it makes sense to also deprecate the corresponding validators.

### DIFF
--- a/dropwizard-validation/src/main/java/io/dropwizard/validation/MaxSizeValidator.java
+++ b/dropwizard-validation/src/main/java/io/dropwizard/validation/MaxSizeValidator.java
@@ -9,7 +9,10 @@ import javax.validation.ConstraintValidatorContext;
 /**
  * Check that a {@link Size} being validated is less than or equal to the
  * minimum value specified.
+ *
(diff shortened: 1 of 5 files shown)


dropwizard-validation/src/main/java/io/dropwizard/validation/MaxSizeValidator.java
```diff
@@ -9,7 +9,10 @@
 /**
  * Check that a {@link Size} being validated is less than or equal to the
  * minimum value specified.
+ *
+ * @deprecated Use {@link MaxSizeValidator} for correct SI and IEC prefixes.
```
jplock, line 13: MaxDataSizeValidator?

rhowe: github doesn't have a facepalm emoji button

jplock (approved):

rhowe:
This reduced the number of sonar code smells by ~4% from 405 to 389

PR #4566, opened 2022-01-01
Update dropwizard-example

rhowe:
Note that I don't actually know how to use Dropwizard, so please check I got this right!

- Remove usage of some deprecated validation classes
- Use `Maps` to prepare for Java 9+
- Annotate unused constructor needed for Jersey

### DIFF
--- a/dropwizard-example/src/main/java/com/example/helloworld/core/Person.java
+++ b/dropwizard-example/src/main/java/com/example/helloworld/core/Person.java
@@ -5,7 +5,6 @@ import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
-import javax.persistence.NamedQueries;
 import javax.persistence.NamedQuery;
 import javax.persistence.Table;
 import javax.validation.constraints.Max;
@@ -14,13 +13,10 @@ import java.util.Objects;
 
 @Entity
 @Table(name = "people")
-@NamedQueries(
-        {
-                @NamedQuery(
-                        name = "com.example.helloworld.core.Person.findAll",
-                        query = "SELECT p FROM Person p"
-                )
-        })
+@NamedQuery(
+    name = "com.example.helloworld.core.Person.findAll",
+    query = "SELECT p FROM Person p"
+)
 public class Person {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
--- a/dropwizard-example/src/main/java/com/example/helloworld/resources/HelloWorldResource.java
+++ b/dropwizard-example/src/main/java/com/example/helloworld/resources/HelloWorldResource.java
@@ -5,7 +5,7 @@ import com.codahale.metrics.annotation.Timed;
 import com.example.helloworld.api.Saying;
 import com.example.helloworld.core.Template;
 import io.dropwizard.jersey.caching.CacheControl;
-import io.dropwizard.jersey.params.DateTimeParam;
+import io.dropwizard.jersey.jsr310.LocalDateParam;
 import org.slf4j.Logger;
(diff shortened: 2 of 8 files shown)


dropwizard-example/src/main/java/com/example/helloworld/core/Person.java
```diff
@@ -14,13 +13,10 @@
 
 @Entity
 @Table(name = "people")
-@NamedQueries(
```
jplock, line 17: @joschi this look ok to you?

jplock (approved):

dropwizard-example/src/main/java/com/example/helloworld/core/Person.java
rhowe: This came up as a linter thing - it said prior to Java 8, you couldn't repeat an annotation multiple times, hence the creation of annotations which simply served as container annotations.
Java 8+ supports specifying multiple annotations of the same type, so `NamedQueries` et al are no longer required.
I believed the linter, but agree it would be good to get someone who actually knows to confirm :)

rhowe: A little more info - here's an Oracle doc introducing repeatable annotations in Java 8: https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html

The JPA 2.2 spec is quite hilariously brief: https://jakarta.ee/specifications/persistence/2.2/persistence_2.2.html

I haven't found the actual JPA announcement supporting them, but here's a blog post talking about it: https://thorben-janssen.com/jpa-2-2-repeatable-annotations/

and here's one about Hibernate 5.3 supporting it: https://in.relation.to/2018/02/07/hibernate-53-repeating-jpa-annotations/

dropwizard-example/src/main/java/com/example/helloworld/resources/HelloWorldResource.java
```diff
@@ -5,7 +5,7 @@
 import com.example.helloworld.api.Saying;
 import com.example.helloworld.core.Template;
 import io.dropwizard.jersey.caching.CacheControl;
-import io.dropwizard.jersey.params.DateTimeParam;
+import org.joda.time.DateTime;
```
joschi, line 8: Should we use `java.time.ZonedDateTime` (or a similar class) here?
We'd like to get rid of Joda Time eventually in favor of the `java.time` API.

rhowe: How about an `Optional<io.dropwizard.jersey.jsr310.ZonedDateTimeParam>`?

joschi: @rhowe Either `*Param` or `Optional<T>` but combining both IMHO doesn't make sense. 😉 

The `*Param` classes are from a time before `Optional<T>` was part of the Java standard library.

rhowe: I see - I was going by the original `Optional<DateTimeParam>` :)

rhowe: OK, I switched it to use `LocalDateParam` and added a little test to `IntegrationTest`.

jplock: Is there a test if the date parameter is missing or the string “null”? Or an empty string?

rhowe: There isn't, but there weren't any tests around this endpoint at all before.
Happy to add some - I didn't know how complete the tests for `dropwizard-example` were intended to be.

PR #4567, opened 2022-01-02
Clean up a few sonar gripes

rhowe:
###### Problem:
Sonar is upset about things which are annotated with `@Deprecated` but which lack the corresponding javadoc `@deprecated` tag.

Sonar is also upset about local variables which have the same name as fields.

The `InstrumentedHttpClientConnectionManager` constructor has been deprecated in favour of a builder.

###### Solution:
Add missing javadoc, rename local variables, use the `InstrumentedHttpClientConnectionManager` builder.

###### Result:
Sonar will grumble less

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/GzipHandlerFactory.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/GzipHandlerFactory.java
@@ -198,13 +198,14 @@ public class GzipHandlerFactory {
     /**
      * @deprecated  gzip handler no longer supports inflate streams
      */
+    @Deprecated
     @JsonProperty
     public boolean isGzipCompatibleInflation() {
         return gzipCompatibleInflation;
     }
 
     /**
-     * @deprecated  gzip handler no longer supports inflate streams
+     * @deprecated gzip handler no longer supports inflate streams
      */
     @Deprecated
     @JsonProperty
@@ -213,7 +214,7 @@ public class GzipHandlerFactory {
     }
 
     /**
-     * @deprecated
+     * @deprecated excluding by user agent will be removed in Jetty 10
      */
     @Deprecated
     public Set<String> getExcludedUserAgentPatterns() {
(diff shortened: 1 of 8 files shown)


jplock (approved):

dropwizard-jetty/src/main/java/io/dropwizard/jetty/GzipHandlerFactory.java
```diff
@@ -213,15 +214,15 @@ public void setGzipCompatibleInflation(boolean gzipCompatibleInflation) {
     }
 
     /**
-     * @deprecated
+     * @deprecated excluding by user agent was removed in Jetty 10
```
joschi, line 217: @rhowe We're still using Jetty 9.4.x because Jetty 10.x requires Java 11.

rhowe: I could've sworn I saw a deprecation warning from a linter somewhere but now I can't find it.

Since we know it's going away, do you think it should be marked as deprecated in order to signal the upcoming change? I have no strong opinion either way.

zUniQueX: @rhowe I marked it as deprecated in a previous PR because I already have an early build of Dropwizard with the Jakarta namespace running. But there are unstable tests at the `dropwizard-http2` module and I didn't want to raise a PR without knowing what's wrong there.

With that said: we're not far away from the ability to upgrade Jersey to 3.0.x, which I think is a good idea, since Jersey 3.1.0 should also be released soon.

jplock (approved):

dropwizard-jetty/src/main/java/io/dropwizard/jetty/GzipHandlerFactory.java
rhowe: Cool, in that case I've amended this PR to add text to the `@deprecated` javadoc tags

PR #4569, opened 2022-01-02
Make `NetUtil` package-private

rhowe:
This class is only used by `HttpConnectorFactory` so it can be package-private.

This is technically a breaking change, but maybe an acceptable one?


zUniQueX:
If it was public before, I'd suggest to deprecate it first. It would be very confusing for users, if they can't access the class any more.

But maybe we could add the deprecation in 2.0.x and remove public access in 2.1.0.

rhowe:
> If it was public before, I'd suggest to deprecate it first. It would be very confusing for users, if they can't access the class any more.
> 
> But maybe we could add the deprecation in 2.0.x and remove public access in 2.1.0.

I think that's fair, I thought I'd try with the bolder option first :)

That then opens up the question of - should this be an internal utility class or one which wants to be exposed to users? I vote internal.

rhowe:
Closing in favour of https://github.com/dropwizard/dropwizard/pull/4584

PR #4570, opened 2022-01-03
Use AssertJ's `Atomic*Assert` support

rhowe:
AssertJ supports asserting on `AtomicInteger` and friends so there's no need to call `.get()` explicitly.


rhowe:
Incorporated into https://github.com/dropwizard/dropwizard/pull/4571

PR #4571, opened 2022-01-03
Small test enhancements

rhowe:
AssertJ 3.22.0 adds `singleElement()` for `ObjectArrayAssert`. Use it in `GzipHandlerTest`.

Use parameterized tests in `ExecutorServiceBuilderTest`

AssertJ supports asserting on `AtomicInteger` and friends so there's no need to call .get() explicitly.

PR #4572, opened 2022-01-03
Use parameterized tests in `ExecutorServiceBuilderTest`

rhowe:
jUnit can do the hard work for us here


rhowe:
Incorporated into https://github.com/dropwizard/dropwizard/pull/4571

PR #4573, opened 2022-01-05
add configuration for maxIdle

abdelmouheimen:
add configuration for maxIdle in dataSourceFactory.java

###### Problem:
We have used dropwiard in our microservices application and we have encountred a problem related to the number of inactif connections that still opened and the configuation of maxIdle resolve the problem 

###### Solution:
add the possibility to configure the maxIdle Variable in DataSourceFactory 

###### Result:
 the configuration of  maxIdle resolve our problem

### DIFF
--- a/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
+++ b/dropwizard-db/src/main/java/io/dropwizard/db/DataSourceFactory.java
@@ -167,6 +167,13 @@ import java.util.concurrent.TimeUnit;
  *         </td>
  *     </tr>
  *     <tr>
+ *         <td>{@code maxIdle}</td>
+ *         <td>1</td>
+ *         <td>
+ *             The maximum size of the inactif connection pool.
+ *         </td>
+ *     </tr>
+ *     <tr>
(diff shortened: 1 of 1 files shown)


zUniQueX:
I think, it's not that easy to implement this. Technically, it would be a breaking change to users who expect `maxIdle` to be equal to `maxActive`. One could however resolve this by implementing the new property as a nullable reference type and setting it to `maxSize`, if the value isn't provided.

Additionally, I'd suggest to provide validators, which ensure that the value is between `minSize` and `maxSize`.

With these two things implemented, it shouldn't be needed to update `asSingleConnectionPool()`. However, one could discuss to set the value there.

Lastly, the updated docs provide a default value of 1, which should be 10 for the current implementation. When the type would be changed to `Integer`, this should be `null` in my opinion.

zUniQueX:
@abdelmouheimen Any update on this? The current implementation introduces breaking changes and therefore won't be merged. Could you please update this PR accordingly, if you're still interested in this?

zUniQueX:
Closing this due to inactivity of the author.

PR #4585, opened 2022-01-06
Update HK2's repackaged jakarta.inject to jakarta.inject-api

zUniQueX:
###### Problem:
Jersey 2.35 uses a repackaged `javax.inject` OSGi bundle from HK2, which seems to get no further updates. So especially the package would not support the `jakarta.inject` namespace.

###### Solution:
Update `org.glassfish.hk2.external:jakarta.inject` to `jakarta.inject:jakarta.inject-api`.

###### Result:
The migration of Dropwizard to the Jakarta namespace will get easier.

### DIFF
--- a/dropwizard-auth/pom.xml
+++ b/dropwizard-auth/pom.xml
@@ -58,8 +58,8 @@
             <artifactId>jakarta.ws.rs-api</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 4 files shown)


joschi (approved):
@zUniQueX Looks good to me, thanks! 😃

PR #4587, opened 2022-01-07
Register resources in the same way for Java 8 and 11

rhowe:
Is "it passes tests" enough validation?

Use `CtClass#toClass(MethodHandles.Lookup)` to register resources. Note that I'm not really familiar with this API - am I doing it right?

The javaassist docs suggest this will work under both Java 8 and 11.

Deprecate `JavaVersion` as it is now unused.

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java
@@ -9,7 +9,6 @@ import io.dropwizard.jersey.caching.CacheControlledResponseFeature;
 import io.dropwizard.jersey.params.AbstractParamConverterProvider;
 import io.dropwizard.jersey.sessions.SessionFactoryProvider;
 import io.dropwizard.jersey.validation.FuzzyEnumParamConverterProvider;
-import io.dropwizard.util.JavaVersion;
 import javassist.ClassPool;
 import javassist.CtClass;
 import javassist.LoaderClassPath;
@@ -29,6 +28,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 import javax.validation.constraints.NotNull;
 import java.io.Serializable;
+import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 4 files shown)


zUniQueX:
Actually, the [JavaDoc](https://www.javassist.org/html/javassist/CtClass.html#toClass(java.lang.Class)) of the ```toClass(java.lang.Class)``` method suggests to use the method with ```MethodHandles.Lookup```.

I just looked briefly at that JavaDoc and was wondering, if we should prefer ```publicLookup```, since we would otherwise provide access to private fields of ```DropwizardResourceConfig```, if I'm understanding this correctly.

zUniQueX:
Seems like one can [specify an alternative class](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/MethodHandles.Lookup.html#in-java.lang.Class-). Maybe it would be a reasonable to change the class to ```SpecificBinder```. This would then be closer to the old implementation.

rhowe:
```
final Object binderProxy = cc.toClass(MethodHandles.lookup().in(SpecificBinder.class))
    .getConstructor(Object.class, Class.class)
    .newInstance(object, clazz);
```

?

zUniQueX:
That's how I'm understanding the docs. As I said, I have no experience with this, so I'm not sure about this either.

rhowe:
The java != 8 path currently ends up here: https://github.com/jboss-javassist/javassist/blob/master/src/main/javassist/util/proxy/DefineClassHelper.java#L288

I guess a naiive use of `MethodHandles.Lookup` as I'm proposing would break if you're doing cross-module stuff?

zUniQueX:
I don't know, if this would be a big problem. In your linked method, no external `Lookup` object is used, so the naive argument should be enough. Or do you mean the method in line 310?

zUniQueX:
I think, I got your comment wrong. That call should just ensure, that the calling module (Javassist in this case) can read the neighbor class of the class to be defined. As we use this for `SpecificBinder`, which lies in Jersey's module, which we should import when using JPMS, our provided ```MethodHandles.Lookup``` already has the read permission. As the ```Lookup``` then defines the class, our provided object should give Javassist enough permissions.

zUniQueX:
I came back to this one and don't think providing an other `MethodHandles.Lookup` would have a better effect. So this implementation should be the best and cleanest we can do.

zUniQueX:
@rhowe @jplock The test reports of newer PRs often show errors on Java 8 related to `MethodHandles.Lookup`. In my opinion this is related to this PR. Should we do anything about this?

Issue #4568, opened 2022-01-02
Error about filename collision in dropwizard-logging tests


rhowe:
I was going to raise a PR, but I'm not really sure what the fix should be, so I'm raising an issue instead.

During the dropwizard-logging tests, this error is logged:

```
13:19:50,448 |-ERROR in ch.qos.logback.core.rolling.RollingFileAppender[file-appender] - 'File' option has the same value "./logs/max-file-size-example.log" as that given for appender [file-appender] defined earlier.
13:19:50,448 |-ERROR in ch.qos.logback.core.rolling.RollingFileAppender[file-appender] - Collisions detected with FileAppender/RollingAppender instances defined earlier. Aborting.
13:19:50,448 |-ERROR in ch.qos.logback.core.rolling.RollingFileAppender[file-appender] - For more information, please visit http://logback.qos.ch/codes.html#earlier_fa_collision
```

The cause looks to date back all the way to 2015 in commit cc53136ae8754c20df268b6ea1a0cf3528618450 which adds an extra logfile to the test logging config, but with the same name as one which already exists.

What's the right thing to do? Remove one of them or change the names so they don't clash?

PR #4636, opened 2022-01-22
Remove duplicate logfile from dropwizard-logging tests (#4568)

rhowe:
This was added in cc53136ae8754c20df268b6ea1a0cf3528618450 but isn't used by the tests.

It causes errors to be logged during tests as the `currentLogFilename` is identical to another log configuration in the same file.

`FileAppenderFactoryTest` tests the behaviour of `FileAppenderFactory` directly without roundtripping through the YAML configuration parser.

Remove the duplicate log file definition.


joschi (approved):
@rhowe Thanks!

PR #4666, opened 2022-01-30
Increase test coverage for `HealthCheckManager`

rhowe:
Test the behaviour of `HealthCheckManager#healthStateView()` and
`HealthCheckManager#healthStateViews()`

Question: Is it correct that when a healthcheck is removed via `HealthCheckManager#onHealthCheckRemoved()`, `HealthCheckManager#healthStateViews()` returns a collection containing `null`? Should `onHealthCheckRemoved()` remove it from the `checks` collection?

`onHealthCheckRemoved()` is never called by Dropwizard.


joschi (approved):
@rhowe Thanks!

PR #4680, opened 2022-02-03
Extend DAOTestExtension with junit5 callbacks

eikenilsknopp:
###### Problem:
The DAOTestExtension currently does not implement the Junit5 `BeforeAll` and `AfterAll` callbacks, therefore its not possible to register it as a proper extension during testing.

###### Solution:
Implement the `BeforeAllCallback` and `AfterAllCallback` interfaces.

###### Result:
DAOTestExtension can be registered as a proper extension during testing using the `@RegisterExtension` annotation.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DAOTestExtension.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DAOTestExtension.java
@@ -1,7 +1,11 @@
 package io.dropwizard.testing.junit5;
 
 import io.dropwizard.testing.common.DAOTest;
+import org.hibernate.HibernateException;
 import org.hibernate.SessionFactory;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
 
 import java.util.concurrent.Callable;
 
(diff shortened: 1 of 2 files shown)


dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DAOTestExtension.java
```diff
@@ -70,6 +74,20 @@ public void after() {
         daoTest.after();
     }
 
+    @Override
+    public void beforeAll(ExtensionContext extensionContext) throws Exception {
+        try {
+            daoTest.before();
```
joschi, line 80: Would it make sense to call `DAOTestExtension#before()` here so that the behavior is guaranteed to be the same?

dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DAOTestExtension.java
```diff
@@ -70,6 +74,20 @@ public void after() {
...
+        } catch (Throwable e) {
+            throw new HibernateException(e);
+        }
+    }
+
+    @Override
+    public void afterAll(ExtensionContext extensionContext) {
+        daoTest.after();
```
joschi, line 88: Would it make sense to call `DAOTestExtension#after()` here so that the behavior is guaranteed to be the same?

joschi:
@eikenilsknopp Could you please also add a test for the new behavior?

joschi (approved):
@eikenilsknopp LGTM, thanks a lot for your contribution!

PR #4681, opened 2022-02-03
Provide documentation for session support

zUniQueX:
###### Problem:
Dropwizard generally provides the ability to use HTTP sesssions, but does not provide documentation on this.

###### Solution:
Add a documentation section on session support and provide an e2e test explaining how to use it.


joschi (approved):
@zUniQueX Awesome, thanks for your contribution! ❤️

PR #4689, opened 2022-02-06
improve assertion in DockerIntegrationTest testLogFileWritten

sullis:
###### Problem:

DockerIntegrationTest checks the content of application.log
The assertions are not adequate.

###### Solution:

Improved assertion DockerIntegrationTest

###### Result:

Unit test passes.


joschi (approved):
@sullis Thanks for your contribution!

PR #4691, opened 2022-02-09
Further linter appeasements

rhowe:
Address a few linter concerns

- Add missing `@Deprecated` tag to FreemarkerViewRenderer constructor
- Simplify dropwizard-jersey CustomDerialization (test helper)
- Fix javadoc reference in dropwizard-core Bundle
- Adjust test class name in LazyLoadingTest
- Remove unused imports from DataSourceFactoryTest
- Remove unnecessary cast from DefaultServerFactory
- Fix markdown in SECURITY.md
- Remove unnecessary character escape
- Remove stray `@JsonProperty` in MetricsFactory

### DIFF
--- a/dropwizard-metrics/src/main/java/io/dropwizard/metrics/MetricsFactory.java
+++ b/dropwizard-metrics/src/main/java/io/dropwizard/metrics/MetricsFactory.java
@@ -93,12 +93,6 @@ public class MetricsFactory {
     }
 
     /**
-     * @since 2.0
-     */
-    @JsonProperty
-
-
-    /**
      * Configures the given lifecycle with the {@link com.codahale.metrics.ScheduledReporter
      * reporters} configured for the given registry.
      * <p />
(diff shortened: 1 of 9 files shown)


dropwizard-metrics/src/main/java/io/dropwizard/metrics/MetricsFactory.java
```diff
@@ -109,7 +103,10 @@ public void setReportOnStop(boolean reportOnStop) {
      *
      * @param environment the lifecycle to manage the reporters.
      * @param registry the metric registry to report metrics from.
+     *
+     * @since 2.0
      */
+    @JsonProperty
```
jplock, line 109: I don’t think this should have an annotation.

rhowe: Right you are - looks like an extra `@JsonProperty` snuck in in eedb84c27cd3efb073eac959710ec59147e17a8b

jplock (approved):

PR #4716, opened 2022-02-15
Add a test for `SslReloadTask`

rhowe:
Validate the behaviour of `SslReloadTask`, i.e. that it performs a dry-run of all `SslReload`ers and only performs the reload if they all succeed their dry run.

Move the dry run logic into `SslReload` in order to simplify testing.

Deprecate `SslReload#reload(SslContextFactory)` as it is no longer required


joschi (approved):

rhowe:
This ticked SonarQube test coverage over 75%

PR #4830, opened 2022-02-27
Add default methods to dropwizard-lifecycle's `Managed` interface

rhowe:
Not everything needs both a start and stop method. Provide default no-op implementations.

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -136,10 +136,6 @@ create a new ``MessageQueueFactory`` class:
             MessageQueueClient client = new MessageQueueClient(getHost(), getPort());
             environment.lifecycle().manage(new Managed() {
                 @Override
-                public void start() {
-                }
-
-                @Override
                 public void stop() {
                     client.close();
                 }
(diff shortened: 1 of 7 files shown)


zUniQueX:
According to https://wiki.eclipse.org/Evolving_Java-based_APIs_2 this change should be binary compatible, so we could bring this to 2.1.0.

zUniQueX:
@rhowe Thanks for your contribution! The changes look good, but I think the documentation could be improved a bit. [Our docs](https://www.dropwizard.io/en/latest/manual/core.html#managed-objects) sound more like one has to implement both methods, which then wouldn't be the case any more.

rhowe:
> @rhowe Thanks for your contribution! The changes look good, but I think the documentation could be improved a bit. [Our docs](https://www.dropwizard.io/en/latest/manual/core.html#managed-objects) sound more like one has to implement both methods, which then wouldn't be the case any more.

Good shout. I'll add some doc updates to this PR and let you review

jplock (approved):

rhowe:
I changed the `and` to `and/or` in the docs. I was going to write a bit extra, but less is more?

zUniQueX:
It's hard to answer stylistic questions, but I think that would be enough. I just wanted to avoid confusion, if users just want to implement one of the two methods.

zUniQueX:
@rhowe I've just seen, that the JavaDoc for `Managed` has a similar formulation to the docs. Maybe you provide an update for that as well. The change to `and/or` should be enough, but the JavaDoc for `Managed` will probably get confusing then.

rhowe:
@zUniQueX good catch. I've updated the javadoc - see what you make of it.

zUniQueX:
Thanks, that looks good

zUniQueX (approved):

PR #4831, opened 2022-02-27
Explain/remove empty methods

rhowe:
Sonar gets upset when a method does nothing without explaining why.

In some cases, the empty methods can just fall back to a default no-op implementation in the corresponding `interface` - remove such methods.

For the rest, add a comment explaining why the method is empty, or at least acknowledging that it does nothing intentionally.


zUniQueX (approved):

zUniQueX:
According to https://wiki.eclipse.org/Evolving_Java-based_APIs_2 moving methods up the class hierarchy shouldn't break API compatibility, if the client doesn't need to reimplement the method.

The JavaDocs of the interfaces with default methods specify no-op implementations for both the `javax` and the `jakarta` packages, so we shouldn't break anything with this for the future.

@dropwizard/committers I think this could be merged to the current `master` branch. Any opinions for holding this back to 3.0.0?

jplock (approved):

PR #4836, opened 2022-03-01
Ignore code coverage metrics for dropwizard-benchmarks

rhowe:
This module's code is already test code. Don't report on test coverage for it.

Note: I'm not sure if this is quite the way to do it - the Sonar docs aren't very clear to me.

https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/#header-3

### DIFF
--- a/dropwizard-benchmarks/pom.xml
+++ b/dropwizard-benchmarks/pom.xml
@@ -21,6 +21,9 @@
         <maven.site.skip>true</maven.site.skip>
         <maven.site.deploy.skip>true</maven.site.deploy.skip>
         <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
+		  <!-- Ignore test coverage metrics for this module  -->
+		  <jacoco.skip>true</jacoco.skip>
+        <sonar.coverage.exclusions>**/*.java</sonar.coverage.exclusions>
     </properties>
 
     <artifactId>dropwizard-benchmarks</artifactId>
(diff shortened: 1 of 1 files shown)


dropwizard-benchmarks/pom.xml
```diff
@@ -21,6 +21,8 @@
         <maven.site.skip>true</maven.site.skip>
         <maven.site.deploy.skip>true</maven.site.deploy.skip>
         <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
+        <!-- Ignore test coverage metrics for this module  -->
+        <sonar.coverage.exclusions>**/*.java</sonar.coverage.exclusions>
```
joschi, line 25: @rhowe Thanks for your contribution!

I think setting `sonar.skip` to `true` should do the trick: https://docs.sonarqube.org/9.3/analysis/scan/sonarscanner-for-maven/#header-6

```suggestion
        <sonar.skip>true</sonar.skip>
```

rhowe: I found that as well, but that will skip all Sonar analysis including code smells etc which seems a _bit_ strong?

joschi: @rhowe Ah, I see. You really only wanted to disable coverage metrics. Nevermind then… 😅

rhowe: Hm, I like that approach. Are you able to trigger a sonar run against this branch somehow to see if the change actually does the right thing?

joschi (approved):

rhowe:
I'm not sure it really worked - Sonar still complains about code coverage in dropwizard-benchmarks: https://sonarcloud.io/component_measures?id=dropwizard_dropwizard&metric=new_coverage&view=list

PR #4837, opened 2022-03-01
Update Apache HttpClient to version 5

zUniQueX:
This PR provides dependency updates for `httpclient` and `httpcore`. Following the migration guide, all classes are reimported and conflicts are resolved.

Support for deprecated fields was dropped, but I've seen that `HttpProcessor` can be re-added with providing `HttpRequestInterceptor` and `HttpResponseInterceptor` objects to the `HttpClientBuilder`. But therefore we have to decide whether to use `add...InterceptorFirst(...)` or `add...InterceptorLast(...)`. Any opinions on that?


zUniQueX:
I've implemented registering the `HttpRequestIntercepror` first and the `HttpResponseInterceptor` last. This shouldn't make any difference as we only have one `HttpProcessor`. But maybe we could allow usage of more objects in upcoming versions and then the new `HttpProcessor` does outermost evaluation.

dropwizard-dependencies/pom.xml
```diff
@@ -477,6 +471,11 @@
                 <artifactId>mockito-junit-jupiter</artifactId>
                 <version>${mockito.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-inline</artifactId>
+                <version>${mockito.version}</version>
+            </dependency>
```
joschi, line 478: Do we need this new dependency? I don't see any changes in the usage of Mockito in our tests.

joschi, line 478: Do we need this new dependency? I don't see any changes in the usage of Mockito in our tests.

joschi (approved):
Looks good to me overall. Thanks for the effort!

One question about `mockito-inline` remaining.

dropwizard-dependencies/pom.xml
zUniQueX: Unfortunately, `CloseableHttpResponse` is a `final` class now and we cannot mock this any more. To simplify the migration, I wanted to reuse the existing tests. `mockito-inline` allows to mock `final` classes, so the tests work again after adding this dependency.

joschi: @zUniQueX Okay, got it. Thanks for the explanation!

Feel free to merge the PR. :smile:

zUniQueX: Thanks for the fast review. There probably is a more elegant way than using `mockito-inline`. But in this PR I just wanted to get the dependency upgrade done. So I'll leave this as-is for now.

PR #4890, opened 2022-03-05
Deprecate `FixtureHelpers`

rhowe:
_Question_ This is a bit speculative - I don't _think_ these helpers seem to offer much value, but maybe you know differently?

Jackson can read directly from `URL`s and `InputStream`s so these helpers don't especially useful.

In addition, Java 9+ has `InputStream#readAllBytes()` which can be combined with `String(byte[])` to parse resources to strings.

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -19,8 +19,8 @@ Testing Representations
 
 While Jackson's JSON support is powerful and fairly easy-to-use, you shouldn't just rely on
 eyeballing your representation classes to ensure you're producing the API you think you
-are. By using the helper methods in `FixtureHelpers`, you can add unit tests for serializing and
-deserializing your representation classes to and from JSON.
+are. You can add unit tests for serializing and deserializing your representation classes to and
+from JSON.
 
 Let's assume we have a ``Person`` class which your API uses as both a request entity (e.g., when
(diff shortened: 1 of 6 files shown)


joschi (approved):
@rhowe LGTM, thanks for your contribution! ❤️

joschi:
> _Question_ This is a bit speculative - I don't _think_ these helpers seem to offer much value, but maybe you know differently?

They provided more value when Jackson and the Java standard library didn't support all the functionality they do today. I'm fine with deprecating the class and eventually removing it if nobody objects.

rhowe:
> > _Question_ This is a bit speculative - I don't _think_ these helpers seem to offer much value, but maybe you know differently?
> 
> They provided more value when Jackson and the Java standard library didn't support all the functionality they do today. I'm fine with deprecating the class and eventually removing it if nobody objects.

Cool. The helpers around loading of resources are now fairly minimal in Dropwizard and once support for Java 8 is dropped at some point in the future they can become even more minimal I think.

zUniQueX:
> Cool. The helpers around loading of resources are now fairly minimal in Dropwizard and once support for Java 8 is dropped at some point in the future they can become even more minimal I think.

@rhowe if you've got concrete ideas of reducing the code further on Java 9+ please feel free to provide PRs against the `release/3.0.x` branch :wink:

PR #4914, opened 2022-03-07
Replace dropwizard-util collection wrappers with JCL

rhowe:
We can use the JCL collections helpers to replace these utility classes.

Remove:

- io.dropwizard.util.Lists
- io.dropwizard.util.Maps
- io.dropwizard.util.Sets

This also allows us to remove dropwizard-util as a dependency of

- dropwizard-auth
- dropwizard-configuration
- dropwizard-json-logging
- dropwizard-migrations
- dropwizard-views

Remove it from dropwizard-request-logging too, as it wasn't used there.


rhowe:
Is this going to make maintaining the `release/3.0.x` branch too painful? It was quick enough to put together and would be easy enough to remake in future.

zUniQueX:
Thanks for the contribution!

I think one should prefer JCL methods whenever possible because they could be implemented more efficiently. So this should be worth the effort.

The change of using the JCL `of` creator methods instead of our custom ones should be unproblematic. But I'm not convinced that the parameterless `of` methods are more readable than the helper methods of the `Collections` class. Did you have a specific reason to do that change?

rhowe:
The rationale was pretty basic - `Set.of()` is more concise than `Collections.emptySet()` or even `emptySet()` with a static import and (at least to me) having collections constructed the same way throughout rather than one method for empty collections and another for non-empty ones is a bit jarring. Also it allowed the import of `Collections` to be removed in a few cases.

I don't have particularly strong opinions either way though, and the two approaches are functionally identical as far as I know. Happy to do whatever.

zUniQueX:
This was just my stylistic suggestion. When other developers contribute to the project, we won't have the consistency for a long time as they will use their own code preferences on that.

I don't mind, if we change this for now or not.

I just recognized a more critical point in the review of this PR: the `Map` constructed from the `Maps` class is not unmodifyable, as is the one from the JCL. Users relying on a mutable `Map` would have strange to debug errors. Since the `Maps` class is deprecated, users have to migrate anyway. But maybe we should add a notice to the JavaDoc?

rhowe:
Good spot although I think the question really is whether we want to document a migration path from the existing classes in dropwizard-util or not - removal isn't quite the same as a behaviour change.

My impression is that they were really aimed at being used internally within Dropwizard as a stand-in for the Guava collections helpers and were only begrudgingly made public so they could be shared around the codebase? How well publicised do we think they are?

zUniQueX:
That would be nice. If the maps get exposed in any way, this would be a behaviour change.

I just recognized I looked into the wrong file. For example in the `Lists` class you added a `@deprecated` JavaDoc tag with suggesting to replace the methods with JCL methods. If this was in the `Maps` class too, this could sound like replacing the class with the JCL `Map` class. Therefore I suggested this. But you are right, we should have nothing to do there.

rhowe:
I checked, the only non-test, non-example code which used `Maps` was `SimpleServerFactory` which passed it to the constructor for `ContextRoutingHandler` and that just iterates through the map and extracts the entries.

zUniQueX:
Okay, many thanks for the lookup

zUniQueX:
Looks good now. But I've just seen that `Sets` isn't deprecated in 2.1.0 yet. Maybe you could kindly run your code deprecator for that :wink:

rhowe:
See https://github.com/dropwizard/dropwizard/pull/4947 for deprecation of `Sets` in 2.1.x

zUniQueX (approved):

PR #4916, opened 2022-03-08
Remove deprecated parts of dropwizard-util

rhowe:
These are all deprecated in 2.1.x and unused by Dropwizard itself.

Replace `ByteStream#toByteArray()` with `InputStream#readAllBytes()`

I've left `Size` and `SizeUnit` because removing those is a bit less trivial and worth doing separately.


zUniQueX (approved):
Thanks, that looks good. But I think we should then add the deprecation for `ByteStreams` in 2.1.0 to let users know it'll get removed.

zUniQueX:
@rhowe I have created #4917 with the deprecation and would merge this PR afterwards.

PR #4948, opened 2022-03-10
Merge master to release/3.0.x

rhowe:


rhowe:
It was a straightforward merge and it will lead to tidier history if someone with commit access does the merge directly and pushes to release/3.0.x instead of merging this PR

joschi:
We're trying to get rid of the `master` branch, so I won't merge this PR but rather `master` -> `release/2.1.x` and then `release/2.1.x` -> `release/3.0.x` (as before).

rhowe:
Should PRs be raised against those release/* branches rather than against master then?

joschi:
> Should PRs be raised against those release/* branches rather than against master then?

@rhowe Yes, please. I think I'll change the default branch to `release/2.1.x` soon so that this will automatically be done for new PRs.

PR #4975, opened 2022-03-12
Simplify `Optional` handling in dropwizard-jersey

rhowe:
We can use `Optional#orElseThrow(Supplier<? extends Throwable>)` instead of explicitly checking for `Optional#isPresent()` here.


rhowe:
`EmptyOptionalException` looks like something we could deprecate and remove in dropwizard 3.x? Does it make sense to switch to Java 11's `Optional#orElseThrow()` and the `NoSuchElementException` (exists since Java 8) it throws?

joschi (approved):

PR #4978, opened 2022-03-13
Remove commons-lang3 as a dependency

rhowe:
commons-lang3 was only used for `ExceptionUtils` in order to navigate exception chains.

Add a helper method to dropwizard-util `Throwables` to perform the same task.

### DIFF
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/ZipExceptionHandlingGzipHandler.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/ZipExceptionHandlingGzipHandler.java
@@ -1,6 +1,5 @@
 package io.dropwizard.jetty;
 
-import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.eclipse.jetty.http.BadMessageException;
 import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.server.Request;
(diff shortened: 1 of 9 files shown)


rhowe:
This un-deprecates `Throwables`, which has been removed in `release/3.0.x` but if you think adding this functionality back in is OK, then we can always reintroduce the class to 3.0.x

dropwizard-jetty/src/main/java/io/dropwizard/jetty/ZipExceptionHandlingGzipHandler.java
```diff
@@ -22,12 +24,15 @@
     public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException {
         try {
             super.handle(target, baseRequest, request, response);
-        } catch (Exception ex) {
-            Throwable rootCause = ExceptionUtils.getRootCause(ex);
-            if (rootCause instanceof ZipException || rootCause instanceof EOFException) {
-                throw new BadMessageException(HttpStatus.BAD_REQUEST_400, rootCause.getMessage(), rootCause);
+        } catch (IOException | ServletException ex) {
+            Optional<BadMessageException> badMessageException = findThrowableInChain(t -> t instanceof ZipException || t instanceof EOFException, ex)
```
rhowe, line 28: This is technically a change in behaviour as it finds the first matching exception from the tip of the chain whereas `ExceptionUtils#getRootCause(Throwable)` returns the deepest exception in the chain.

I think in this case the behaviour is now arguably more correct? Is it possible for a ZipException or an EOFException to be in the chain without us wanting to return a 400?

joschi (approved):
LGTM, thanks!

PR #5009, opened 2022-03-21
Make better use of AssertJ's string matchers

rhowe:
Where we are using `String#format(...)` to deal with platform-specific line endings, we can instead use `isEqualToNormalizingNewlines`.

We can also replace some calls to `String#format(String...)` with AssertJ's built-in support for format strings.

### DIFF
--- a/dropwizard-migrations/src/test/java/io/dropwizard/migrations/DbClearChecksumsCommandTest.java
+++ b/dropwizard-migrations/src/test/java/io/dropwizard/migrations/DbClearChecksumsCommandTest.java
@@ -31,23 +31,23 @@ class DbClearChecksumsCommandTest {
     void testHelpPage() throws Exception {
         final ByteArrayOutputStream out = new ByteArrayOutputStream();
         MigrationTestSupport.createSubparser(clearChecksums).printHelp(new PrintWriter(new OutputStreamWriter(out, UTF_8), true));
-        assertThat(out.toString(UTF_8.name())).isEqualTo(String.format(
-            "usage: db clear-checksums [-h] [--migrations MIGRATIONS-FILE]%n" +
-                "          [--catalog CATALOG] [--schema SCHEMA] [file]%n" +
-                "%n" +
-                "Removes all saved checksums from the database log%n" +
-                "%n" +
-                "positional arguments:%n" +
-                "  file                   application configuration file%n" +
-                "%n" +
-                "named arguments:%n" +
-                "  -h, --help             show this help message and exit%n" +
-                "  --migrations MIGRATIONS-FILE%n" +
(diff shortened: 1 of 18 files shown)


rhowe:
For the migration command help text output tests, what do you think about moving them to files in `src/test/resources/`, at least until Java 17...

dropwizard-migrations/src/test/java/io/dropwizard/migrations/DbMigrateCommandTest.java
```diff
@@ -61,10 +61,10 @@ void testDryRun() throws Exception {
...
         migrateCommand.setOutputStream(new PrintStream(baos));
         migrateCommand.run(null, new Namespace(Collections.singletonMap("dry-run", true)), conf);
-        assertThat(baos.toString(UTF_8.name())).startsWith(String.format(
-                "-- *********************************************************************%n" +
-                "-- Update Database Script%n" +
-                "-- *********************************************************************%n"));
+        assertThat(baos.toString(UTF_8.name())).isEqualToNormalizingNewlines(
+                "-- *********************************************************************\n" +
+                "-- Update Database Script\n" +
+                "-- *********************************************************************\n");
```
joschi: Incorrect replacement (starts with vs. equals).

rhowe: Right you are - fun that the test still passed!

dropwizard-migrations/src/test/java/io/dropwizard/migrations/DbClearChecksumsCommandTest.java
```diff
@@ -31,23 +31,23 @@ void testRun() throws Exception {
...
-                "named arguments:%n" +
-                "  -h, --help             show this help message and exit%n" +
-                "  --migrations MIGRATIONS-FILE%n" +
-                "                         the file containing  the  Liquibase migrations for%n" +
-                "                         the application%n" +
-                "  --catalog CATALOG      Specify  the   database   catalog   (use  database%n" +
-                "                         default if omitted)%n" +
-                "  --schema SCHEMA        Specify the database schema  (use database default%n" +
-                "                         if omitted)%n"));
+        assertThat(out.toString(UTF_8.name())).isEqualTo(
```
zUniQueX: Looks like there is a NormalizingNewlines missing

rhowe: Right you are. I think I was flipping between `hasToString()` and this and ended up broken in the middle somewhere!

joschi (approved):

PR #5010, opened 2022-03-21
Jetty 10

zUniQueX:
This PR provides a dependency update to Jetty major version 10.

As discussed earlier in various issues and discussions, Jetty 10 will break request logging with Logback 1.2.x. After playing around a bit with this branch, I fixed it with overriding the `log` method of the `LogbackAccessRequestLog` class and providing a custom `ServerAdapter` implementation.

This means we would have to adjust those classes ourselves when Logback provides changes, but I think being blocked for our 3.0.x release is worse then doing this. @dropwizard/committers if anyone has an other opinion on this, please feel free to close this PR.


jplock (approved):

zUniQueX:
@joschi If we have this one out, would we directly want to provide alpha versions of 3.0.0 and 4.0.0? Thanks to the work of @rhowe I think we have most deprecations addressed and therefore could give users an early build to test their applications against.

rhowe:
Do we want to keep jUnit 4.x support around? I'm Happy to take the axe to that if not.

The other things I was looking at is whether dropwizard-jackson and dropwizard-jersey could have guava split out of them somehow, perhaps into a separate pom? Also I was wondering if we could incorporate java.net.http into dropwizard-client as an alternative to apache-httpclient.

joschi:
> Do we want to keep jUnit 4.x support around?

I was thinking about that too, and I wouldn't be totally opposed to removing it in Dropwizard 3+.

@dropwizard/committers Any opinions on this?

joschi:
> If we have this one out, would we directly want to provide alpha versions of 3.0.0 and 4.0.0?

Yes, I think so.

zUniQueX:
> > Do we want to keep jUnit 4.x support around?
> 
> I was thinking about that too, and I wouldn't be totally opposed to removing it in Dropwizard 3+.
> 
> @dropwizard/committers Any opinions on this?

I don't have a strong opinion on this. For example Jersey still uses jUnit 4.x and I don't see the need of removing support for it. However, having less code to maintain would make things easier.

joschi (approved):
@zUniQueX LGTM, thanks for the great work!

zUniQueX:
> Also I was wondering if we could incorporate java.net.http into dropwizard-client as an alternative to apache-httpclient.

I would want to wait with that. I've raised a PR on this at Jersey recently and this could make it to 3.1.0. If we need anything more for customization than Jersey provides so far, maybe it could be fixed there and we don't need a custom connector.

PR #5023, opened 2022-03-22
Docs example code refactor

zUniQueX:
Test using `:start-after:` and `:end-before:`


zUniQueX:
@joschi I tested refactoring the examples with the `start-after` and `end-before` directives in https://github.com/dropwizard/dropwizard/pull/5023/commits/39d61ddf7b70d54585fb7e4779fc69f27dba45b4. So far, this looks quite feasible. What do you think about this?

PR #5045, opened 2022-03-28
Remove deprecated `Throwables#getRootCause(Throwable)`

rhowe:
This was deprecated in 2.1.x and is not used by Dropwizard.


joschi (approved):
LGTM, thanks Russell!

PR #5050, opened 2022-03-28
Remove guava as a dependency of dropwizard-auth

rhowe:
This was only here in order to annotate a field as `@VisibleForTesting`

### DIFF
--- a/dropwizard-auth/pom.xml
+++ b/dropwizard-auth/pom.xml
@@ -46,10 +46,6 @@
             <artifactId>jsr305</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-        </dependency>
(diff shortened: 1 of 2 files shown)


rhowe:
What's the take on guava for 3.x? Is it still useful in a Java 11 world?

joschi (approved):

joschi:
> What's the take on guava for 3.x? Is it still useful in a Java 11 world?

Yes, I think Guava is still quite useful in general but I'd like to reduce the _default_ dependency on it if it makes sense.

rhowe:
> > What's the take on guava for 3.x? Is it still useful in a Java 11 world?
> 
> Yes, I think Guava is still quite useful in general but I'd like to reduce the _default_ dependency on it if it makes sense.

So e.g. a dropwizard-jersey-guava module?

PR #5092, opened 2022-04-02
Change log messages for unhealthy initialized health checks

### LINKED ISSUE #4944: dropwizard-health: improve logging behavior when check's state is initially set to false
Dropwizard version: 2.1.0-beta.7

A health check can be configured to have an `initialState` of `false` so that an app won't prematurely advertise itself as being healthy before it has a chance to perform a check. When that is configured, an `ERROR` level statement is logged during startup.

Instead, since this is a supported configuration, this case should not be considered an error.

> ERROR [2022-03-09 20:47:23,946] io.dropwizard.health.HealthCheckManager [thread=main]: A critical dependency is now unhealthy: name=example-app-db, type=READY

zUniQueX:
Closes #4944 

This PR maintains a flag in a `SchaduledHealthCheck`, if the check already had recovered. This provides the ability to display a better log message, if the check was initialized as unhealthy.

### DIFF
--- a/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java
+++ b/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java
@@ -174,7 +174,14 @@ class HealthCheckManager implements HealthCheckRegistryListener, HealthStatusChe
                     return;
             }
         } else {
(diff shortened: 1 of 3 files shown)


jplock (approved):

joschi:
Let's merge the PR and release another 2.1.0-beta*. We can polish it after that if necessary.

@pstackle Feedback appreciated. 😉

PR #5138, opened 2022-04-08
Remove jUnit 4.x support

rhowe:
Raised more as a "we could" rather than necessarily a "we should".

Is it worth moving it to a new module rather than outright removing it? `dropwizard-testing-junit4`?


zUniQueX:
> Is it worth moving it to a new module rather than outright removing it? `dropwizard-testing-junit4`?

I've already thought about that. This could be problematic, as it would depend on the `io.dropwizard.testing.common` package. If the jUnit 4.x module would depend on the jUnit 5.x one, I think this would be surprising.

However, we could split the package into three modules and extract the common classes to a seperate one.

rhowe:
> > Is it worth moving it to a new module rather than outright removing it? `dropwizard-testing-junit4`?
> 
> I've already thought about that. This could be problematic, as it would depend on the `io.dropwizard.testing.common` package. If the jUnit 4.x module would depend on the jUnit 5.x one, I think this would be surprising.
> 
> However, we could split the package into three modules and extract the common classes to a seperate one.

I think that with the appropriate marketing, it could be fine - `dropwizard-testing-junit4-compat` or something, at least a name that hinted it was an addon to dropwizard-testing whose days were numbered.

rhowe:
An alternative could be to mark the jUnit 4 support classes as deprecated, although that doesn't remove it from the classpath (yes I know you can exclude dependencies in your build tool, but I don't think that kind of behaviour should be encouraged)

zUniQueX:
> An alternative could be to mark the jUnit 4 support classes as deprecated, although that doesn't remove it from the classpath (yes I know you can exclude dependencies in your build tool, but I don't think that kind of behaviour should be encouraged)

This wouldn't give us any advantage IMO. If we won't remove the jUnit 4.x classes, we would have to provide support for it.

joschi (approved):
I've extracted the JUnit 4.x classes into a separate module: https://github.com/dropwizard/dropwizard-testing-junit4

With this in mind, I'm okay with removing JUnit 4.x support in the `release/3.0.x` branch.

@rhowe Could you please add a note to use https://github.com/dropwizard/dropwizard-testing-junit4 to the Dropwizard 3.0.x upgrade notes?
https://github.com/rhowe/dropwizard/blob/remove-junit4/docs/source/manual/upgrade-notes/upgrade-notes-3_0_x.rst

rhowe:
> @rhowe Could you please add a note to use https://github.com/dropwizard/dropwizard-testing-junit4 to the Dropwizard 3.0.x upgrade notes? https://github.com/rhowe/dropwizard/blob/remove-junit4/docs/source/manual/upgrade-notes/upgrade-notes-3_0_x.rst

Done, although do check my wording

zUniQueX:
Looks good :+1: 

Just one small thing: Markdown uses an other link notation than RST. So your link in the upgrade document is not correctly displayed.

rhowe:
> Just one small thing: Markdown uses an other link notation than RST. So your link in the upgrade document is not correctly displayed.

So it does. Fixed & rebased.

PR #5141, opened 2022-04-09
Make ConfigurationFactoryFactory#create public

fuukasumi:
###### Problem:
In its current state, the `ConfigurationFactoryFactory`'s `create` method cannot be inherited outside the package (e.g. by a third-party plugin). This makes it impossible to build and configure a custom ConfigurationFactory, as the method to create a configuration factory can't be overridden.

###### Solution:
Make `create` on `ConfigurationFactoryFactory` public, giving external libraries the ability to inherit/override if necessary.

###### Result:
All unit tests still pass, and increasing the visibility on the method does not cause any regression issues. This will simply increase the extensibility/inheritability of Dropwizard as a whole.

### DIFF
--- a/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationFactoryFactory.java
+++ b/dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationFactoryFactory.java
@@ -5,7 +5,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import javax.validation.Validator;
 
 public interface ConfigurationFactoryFactory<T> {
-    ConfigurationFactory<T> create(Class<T> klass,
+    public ConfigurationFactory<T> create(Class<T> klass,
             Validator validator,
             ObjectMapper objectMapper,
             String propertyPrefix);


dropwizard-configuration/src/main/java/io/dropwizard/configuration/ConfigurationFactoryFactory.java
```diff
@@ -5,7 +5,7 @@
 import javax.validation.Validator;
 
 public interface ConfigurationFactoryFactory<T> {
-    ConfigurationFactory<T> create(Class<T> klass,
+    public ConfigurationFactory<T> create(Class<T> klass,
```
joschi, line 8: Methods declared in Java interfaces are always public. 😉

> Every method declaration in the body of an interface is implicitly public.

https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.4

fuukasumi: Do you have some idea then, as to why I might be getting this error when I attempt to override the DefaultConfigurationFactoryFactory with my own?

```
Exception in thread "main" java.lang.IncompatibleClassChangeError: Implementing class
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:473)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:355)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
```

I assumed it was due to some incompatibility between the ConfigurationFactoryFactory jars, and that said incompatibility was the visibility on the `create` method (that being the only change that was made to the `ConfigurationFactoryFactory` class since it was first created).

Fyi, it's declared (roughly) like this:
```
public class MyConfigurationFactoryFactory<T> extends DefaultConfigurationFactoryFactory<T>{

public MyConfigurationFactoryFactory(final Class<T> configurationClass) {
	    this.configurationClass = configurationClass;
}

@Override
	  public ConfigurationFactory<T> create(
	          Class<T>     klass,
	          Validator    validator,
	          ObjectMapper objectMapper,
	          String       propertyPrefix) {
                 return new MyConfigurationFactory<T>(super.create(
	          klass,
	          validator,
	          configureObjectMapper(objectMapper.copy()),
	          propertyPrefix), configurationClass);
}
```

joschi: @fuukasumi Which version of Dropwizard are you using?
Are you using a version mix of different Dropwizard artifacts?
Are you using the Dropwizard BOM in your project?

PR #5149, opened 2022-04-12
Replace `@NotThreadSafe` with jUnit 5 `Execution(SAME_THREAD)`

rhowe:
This allows us to drop a dependency on jcip-annnotations


joschi (approved):
LGTM, thanks!

PR #5169, opened 2022-04-14
Migrations file handling

### LINKED ISSUE #5150: `dropwizard-migrations` doesn't find migrations file in `2.1.0-beta.8`
As the docs state, Liquibase [changed the file handling](https://docs.liquibase.com/concepts/changelogs/how-liquibase-finds-files.html) in 4.x.x. Therefore migration files from the file system aren't found any more.

```
liquibase.exception.ChangeLogParseException: The file db/changelog/migrations.xml was not found in
Specifying files by absolute path was removed in Liquibase 4.0. Please use a relative path or add '/' to the classpath parameter.
	at liquibase.parser.core.xml.XMLChangeLogSAXParser.parseToNode(XMLChangeLogSAXParser.java:99)
	at liquibase.parser.core.xml.AbstractChangeLogParser.parse(AbstractChangeLogParser.java:15)
	at liquibase.Liquibase.getDatabaseChangeLog(Liquibase.java:369)
	at liquibase.Liquibase.lambda$update$1(Liquibase.java:224)
	at liquibase.Scope.lambda$child$0(Scope.java:180)
	at liquibase.Scope.child(Scope.java:189)
	at liquibase.Scope.child(Scope.java:179)
	at liquibase.Scope.child(Scope.java:158)
	at liquibase.Liquibase.runInScope(Liquibase.java:2405)
	at liquibase.Liquibase.update(Liquibase.java:211)
	at liquibase.Liquibase.update(Liquibase.java:197)
	at liquibase.Liquibase.update(Liquibase.java:193)
	at liquibase.Liquibase.update(Liquibase.java:185)
	at io.dropwizard.migrations.DbMigrateCommand.run(DbMigrateCommand.java:65)
	at io.dropwizard.migrations.DbCommand.run(DbCommand.java:70)
	at io.dropwizard.migrations.AbstractLiquibaseCommand.run(AbstractLiquibaseCommand.java:68)
	at io.dropwizard.cli.ConfiguredCommand.run(ConfiguredCommand.java:98)
	at io.dropwizard.cli.Cli.run(Cli.java:78)
	at io.dropwizard.Application.run(Application.java:94)
```

This should be caused from [not setting any root path to the `FileSystemResourceAccessor`](https://github.com/dropwizard/dropwizard/blob/release/2.1.x/dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibaseWithFileSystemMigrationsFile.java#L20).

zUniQueX:
###### Problem:
Liquibase 4.x.x changed its file finding and handling. Therefore migration files in the `dropwizard-migrations` module were not found.

###### Solution:
Add root paths to the `FileSystemResourceAccessor`. Liquibase suggests adding the path `'/'` as a fix when migrating to 4.x.x. This seems to work on Linux, but not on Windows. So this PR adds the operating system specific root path (`C:\` for Windows and `/` otherwise). Additionally, the location of the code source is added (acquired through the `CloseableLiquibase` class).

This allows finding files with an absolute path and files located under `src/main/resources`, that are packaged into the jar as resources. Previously, it was possible to provide paths relative to the root folder of the project, which now isn't possible any more. However, the sources are unlikely to be deployed, so this should be enough in my opinion, when adding this behavior change to the migration guide. This also includes changes to be made to `.xml` files including other ones, as the include mechanism uses the same resolving strategy.

###### Result:
Users can specify custom files to the `dropwizard-migrations` module instead of using the default `migrations.xml`.

Resolves #5150

### DIFF
--- a/dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibaseWithFileSystemMigrationsFile.java
+++ b/dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibaseWithFileSystemMigrationsFile.java
@@ -8,35 +8,42 @@ import liquibase.exception.LiquibaseException;
 import liquibase.resource.FileSystemResourceAccessor;
 
 import java.io.File;
-import java.net.URISyntaxException;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
 
 public class CloseableLiquibaseWithFileSystemMigrationsFile extends CloseableLiquibase implements AutoCloseable {
-
     CloseableLiquibaseWithFileSystemMigrationsFile(
         ManagedDataSource dataSource,
         Database database,
         String file
     ) throws LiquibaseException, SQLException {
+        this(dataSource,
+            database,
+            file,
+            FileSystems.getDefault());
+    }
+
+    CloseableLiquibaseWithFileSystemMigrationsFile(
+        ManagedDataSource dataSource,
+        Database database,
+        String file,
+        FileSystem fileSystem
+    ) throws LiquibaseException, SQLException {
         super(file,
-            new FileSystemResourceAccessor(getRootPaths().toArray(new File[0])),
+            new FileSystemResourceAccessor(getRootPaths(fileSystem).toArray(new File[0])),
             database,
             dataSource);
     }
 
-    private static List<File> getRootPaths() {
+    private static List<File> getRootPaths(FileSystem fileSystem) {
         List<File> rootPaths = new ArrayList<>();
-        boolean isWindows = System.getProperty("os.name", "").toLowerCase().contains("win");
-        if (isWindows) {
-            rootPaths.add(new File("C:\\"));
-        } else {
-            rootPaths.add(new File("/"));
-        }
+        fileSystem.getRootDirectories().forEach(path -> rootPaths.add(path.toFile()));
         try {
             rootPaths.add(new File(CloseableLiquibase.class.getProtectionDomain().getCodeSource().getLocation().toURI()));
-        } catch (URISyntaxException ignored) {
+        } catch (Exception ignored) {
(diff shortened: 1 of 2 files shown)


dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibaseWithFileSystemMigrationsFile.java
```diff
@@ -17,9 +21,25 @@ public class CloseableLiquibaseWithFileSystemMigrationsFile extends CloseableLiq
...
+    }
+
+    private static List<File> getRootPaths() {
+        List<File> rootPaths = new ArrayList<>();
+        boolean isWindows = System.getProperty("os.name", "").toLowerCase().contains("win");
+        if (isWindows) {
+            rootPaths.add(new File("C:\\"));
+        } else {
+            rootPaths.add(new File("/"));
+        }
```
joschi: Should we let Java NIO do the hard work and do something like this?
```suggestion
        rootPaths.add(Paths.get("/").toFile());
```

https://docs.oracle.com/javase/tutorial/essential/io/pathOps.html#create

zUniQueX: Thanks for the suggestion. But in the linked docs I can't read that using `/` would work on Windows too.

After looking deeper into the docs about that, I would change the code to use [`FileSystems.getDefault().getRootDirectories()`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystem.html#getRootDirectories()). This way we would avoid to deal with OS specific paths and would additionally cover situations with multiple drives.

joschi: Yes, let's use `getRootDirectories()`. 👍

joschi (approved):
Thanks!

zUniQueX:
Looks good! :+1: Should we keep the issue open to have a reminder for the migration guide?

PR #5170, opened 2022-04-14
Replace `assumeThat` with jUnit 5 `DisabledIf` et al

rhowe:
jUnit 5 introduces `DisabledIf` and `EnabledIf`, allowing tests or entire test classes to be disabled based on arbitrary conditions.

Replace use of `assumeThat` with this newer mechanism.

I'm not completely convinced, personally - it's not as type-safe (although IntelliJ IDEA knows about it).

For the tests where we did `assumeThat` in a `BeforeEach`, this should be more performant I suppose, but the rest are much of a muchness I think.

See what you think...

For `dropwizard-jersey/src/test/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapperTest.java` I just removed the `assumeThat` since the test explicitly changes the locale anyway (which is probably bad?!).

### DIFF
--- a/dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationValidationExceptionTest.java
+++ b/dropwizard-configuration/src/test/java/io/dropwizard/configuration/ConfigurationValidationExceptionTest.java
@@ -3,6 +3,7 @@ package io.dropwizard.configuration;
 import io.dropwizard.validation.BaseValidator;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIf;
 
 import javax.annotation.Nullable;
 import javax.validation.ConstraintViolation;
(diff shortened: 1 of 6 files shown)


jplock (approved):
I think it is more explicit

dropwizard-configuration/src/test/java/io/dropwizard/configuration/EnvironmentVariableLookupTest.java
```diff
@@ -1,14 +1,14 @@
...
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIf;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assumptions.assumeThat;
 
 class EnvironmentVariableLookupTest {
     @Test
+    @DisabledIf("isNopeEnvVarSet")
```
zUniQueX: Could we use [`@DisabledIfEnvironmentVariable`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/condition/DisabledIfEnvironmentVariable.html) for this? If the value is `null`, the annotation should have no effect, so this should be what we want.

rhowe: I think we can actually deprecate and remove `EnvironmentVariableLookup`... PR coming right up..

rhowe: See what you think of https://github.com/dropwizard/dropwizard/pull/5188

joschi (approved):
Thanks!

Issue #5136, opened 2022-04-07
maven-javadoc-plugin fails due to Automatic-Module-Name


joschi:
`3.0.0-beta.1`: https://github.com/dropwizard/dropwizard/runs/5875633807?check_suite_focus=true#step%3A4%3A3444=
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.3.2:jar (attach-javadocs) on project dropwizard-jersey: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1 - /home/runner/work/dropwizard/dropwizard/dropwizard-jersey/src/main/java/io/dropwizard/jersey/DropwizardResourceConfig.java:29: error: cannot find symbol
Error:  import javax.annotation.Nullable;
```

`4.0.0-beta.1`: https://github.com/dropwizard/dropwizard/runs/5875544966?check_suite_focus=true#step:4:10598
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.3.2:jar (attach-javadocs) on project dropwizard-testing: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1 - /home/runner/work/dropwizard/dropwizard/dropwizard-testing/src/main/java/io/dropwizard/testing/common/DropwizardTestResourceConfig.java:22: error: cannot access ResourceConfig
Error:  class DropwizardTestResourceConfig extends DropwizardResourceConfig {
Error:  ^
Error:    class file for org.glassfish.jersey.server.ResourceConfig not found
Error:  /home/runner/work/dropwizard/dropwizard/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardExtensionsSupport.java:8: error: package org.junit.platform.commons.util is not visible
Error:  import org.junit.platform.commons.util.ReflectionUtils;
Error:                                   ^
Error:    (package org.junit.platform.commons.util is declared in module org.junit.platform.commons, which does not export it to module io.dropwizard.testing)
Error:  
Error:  Command line was: /usr/lib/jvm/temurin-11-jdk-amd64/bin/javadoc @options @packages
```

🔗 https://issues.apache.org/jira/projects/MJAVADOC/issues/MJAVADOC-707

joschi:
> ```
[...]

PR #5176, opened 2022-04-15
Migrate away from JSR305

zUniQueX:
As denoted in #5136, JSR305 causes problems on Java 9+ due to reusing the `javax.annotation` package.

But the annotations from JSR305 can be replaced with annotations from other packages. NullAway states in the [configuration docs](https://github.com/uber/NullAway/wiki/Configuration#custom-nullability-annotations), that any annotations with the names `@Nullable`, `@Nonnull`, `@NonNull` and `@NotNull` are considered and not only those from JSR305. Therefore the annotations from `org.jetbrains.annotations` should be fine. (I've used Jetbrains' annotations here, since the package is much smaller than `checker-qual` or similar packages.)

The annotations of the package `javax.annotation.concurrent` can be replaced by annotations from errorprone. The [errorprone docs](https://errorprone.info/bugpattern/Immutable) additionally state, that the JSR305 `@Immutable` annotation is not evaluated, but the errorprone one is. So we should have an additional advantage there.

The one `@NotThreadSafe` annotation is replaced with jUnit's `@Execution(SAME_THREAD)`, as it was done in #5149.


jplock (approved):

joschi:
> (I've used Jetbrains' annotations here, since the package is much smaller than `checker-qual` or similar packages.)

The Checker framework is being pulled in by Caffeine and by Google Guava, so it's a dependency in almost all of our modules anyway.

Would it make sense to use the annotations from the Checker framework instead of replacing `jsr305` with the JetBrains annotations? That way we at least would get rid of another dependency. 😉

- https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/Nullable.html
- https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/NonNull.html

zUniQueX:
> Would it make sense to use the annotations from the Checker framework instead of replacing `jsr305` with the JetBrains annotations? That way we at least would get rid of another dependency. :wink:

I thought about this too. If we have a transitive dependency on that through guava and caffeine anyway, I don't see a big problem using those annotations. I just wanted to avoid pulling in a heavyweight dependency like `checker-qual`, if we just use a few annotations from it. Unfortunately, the Checker framework seems to have no extra package for the annotations.

I'll update this PR with replacing the Jetbrains annotations with the ones from the Checker framework in a separate commit, then we can revert it, if we change our mind.

zUniQueX:
@joschi I've now used `checker-qual` annotations and we can actually get rid of a few dependency declarations in addition to not introducing the `org.jetbrains.annotations` dependency :wink:

However, the use of the annotations isn't as straightforward as using other nullness annotations, since the `@Target` meta annotation is different for the `checker-qual` ones.

joschi (approved):
Thanks!

PR #5183, opened 2022-04-15
Include runnable code snippets into docs

zUniQueX:
This PR adapts the current changes of `release/3.0.x` to the great work @joschi did in #3327 and integrates `:start-after:` and `:end-before:`.

The naming conventions I used in the comments can definitely be improved, but this is just a test to examine the maintainability of this approach.


jplock (approved):

zUniQueX:
@joschi What do you think about this? I think the JavaDoc-like notation for `.java` files is good, but I'm not sure about the notations for referencing code in `.xml` and `.yml` files and for referencing parts inside methods.

zUniQueX:
Since I think this is really cool at all, users won't recognize the underlying source code. I think it would be really helpful, if we could link to the sources from the docs. Any opinion on that?

The literalinclude directive doesn't support this, but it seems adding a custom literalinclude directive [is doable](https://github.com/wpilibsuite/sphinxext-remoteliteralinclude).

zUniQueX:
I've now created a custom `dropwizard_literalinclude` directive that provides a link to the source lines in the caption of the code block.

I've never used Python or Sphinx before, so the code could definitely be improved. But I've tested common use cases and so far it seems to work without problems.

joschi (approved):
LGTM, thanks for the effort of making my half-assed PR work. 😅

zUniQueX:
Thanks for the review. I just wanted to test, if this is possible overall, if the pipelines are stable and if the styling of the captions is good enough. But I'm happy you think it is ready for a merge :wink:

PR #5197, opened 2022-04-20
rework auth filter injection behavior to only setup injection once at startup

### LINKED ISSUE #5191: v2.1.0-beta.8 broke use of AuthDynamicFeature with non-AuthFilter authentication filter
Starting with v2.1.0-beta.8, our use of `dropwizard-auth`'s `AuthDynamicFeature` to handle registering our custom authentication filter (read: not based on `io.dropwizard.auth.AuthFilter`) for annotated resource methods, broke.

```
ERROR [2022-04-18 18:49:12,871] org.glassfish.grizzly.servlet.ServletHandler: service exception:
! java.lang.IllegalArgumentException: The filter must be an instance of AuthFilter
! at io.dropwizard.auth.InjectingFilter.<init>(InjectingFilter.java:34)
! at io.dropwizard.auth.AuthDynamicFeature.registerAuthFilter(AuthDynamicFeature.java:102)
! at io.dropwizard.auth.AuthDynamicFeature.registerAuthFilter(AuthDynamicFeature.java:95)
! at io.dropwizard.auth.AuthDynamicFeature.configure(AuthDynamicFeature.java:67)
```

Because we need additional behavior beyond what the `AuthFilter` provides, we create our own `ContainerRequestFilter` and register that using `io.dropwizard.auth.AuthDynamicFeature` from the dropwizard-auth module so that we can use the various JAX-RS authz annotations (e.g. `RolesAllowed`, `PermitAll`) as well as the `@Auth` annotation from Dropwizard in order to control which resource methods the authentication filter is applied to. 

This has worked fine with previous versions of Dropwizard and has been working fine even with v2.1.0-beta.7. 

It looks like the change in behavior was introduced as part of this PR: https://github.com/dropwizard/dropwizard/pull/4563
The change looks like it is now wrapping the user's provided ContainerRequestFilter instance inside another ContainerRequestFilter named `InjectingFilter`. The `InjectingFilter` is throwing the error because my ContainerRequestFilter is not an instanceof `AuthFilter` (line 34). 

CC: @zUniQueX

pstackle:
This change reworks the DI setup for auth filters to inject only once during setup and not on every request.

When injection is performed, each injectable field is assigned a special proxy class that is automatically updated to the per-request values. The `InjectionManager.inject()` method does not need to be invoked for each request.

The following links provide helpful explanations of the behavior:
- https://stackoverflow.com/a/35953104/901236
- https://psamsotha.github.io/jersey/2015/12/16/request-scope-into-singleton-scope.html

Resolves: #5191
Closes: #5196


dropwizard-auth/src/main/java/io/dropwizard/auth/AuthInjectionHelper.java
```diff
@@ -0,0 +1,29 @@
+package io.dropwizard.auth;
+
+import org.glassfish.jersey.internal.inject.InjectionManager;
+
+import javax.ws.rs.container.ContainerRequestFilter;
+
+/**
+ * Utility class to help with injection into auth filters.
+ */
+public class AuthInjectionHelper {
```
zUniQueX: As this is more an internal utility, should we make it package-private?

zUniQueX (commented):
Looks good!

dropwizard-auth/src/main/java/io/dropwizard/auth/AuthInjectionHelper.java
pstackle: Yes, good catch. I've fixed up visibility on that class as well as on the tests

zUniQueX: Thanks :+1:

zUniQueX (approved):

PR #5279, opened 2022-05-10
Health check configurable to run on admin servlet

raymondchen625:
###### Problem:
The health check under the `health` section is hard-coded to run always on the application servlet. We don't always wish to expose this endpoint to the outside world. Although we can configure the load balancer to block the access from external, having it running on the admin servlet and exposing only the application port would be easier.

###### Solution:
Make it configurable by adding a new `onAdminServlet` option. When set to `true`, the `/health-check' endpoint will be on the admin servlet, just like the `/healthcheck`.

###### Result:
The default behavior is not changed, the health check still runs on the application servlet if not configured otherwise.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -1293,6 +1293,7 @@ Health
         type: json
       responder:
         type: servlet
+      onAdminServlet: false
 
 
 ============================== =======================  ====================================================================================================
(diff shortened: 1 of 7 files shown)


joschi:
@raymondchen625 Have you taken a look at the new `dropwizard-health` module?

If you're on Dropwizard 2.0.x, see https://github.com/dropwizard/dropwizard-health.

raymondchen625:
> @raymondchen625 Have you taken a look at the new `dropwizard-health` module?
> 
> If you're on Dropwizard 2.0.x, see https://github.com/dropwizard/dropwizard-health.

I have. My plan is to upgrade to 2.1 first, which has the health module merged. Then add the change for the health check used in K8s.

PR #5305, opened 2022-05-20
Fix logging of query parameters for classic request logging

wilbertpol:
###### Problem:
We are using classic type request logging through syslog. After upgrading to 2.x the query parameters are no longer getting logged in the access logs.

###### Solution:
Updated the pattern in io.dropwizard.request.logging.old.ClassicLogFormat.

### DIFF
--- a/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/old/ClassicLogFormat.java
+++ b/dropwizard-request-logging/src/main/java/io/dropwizard/request/logging/old/ClassicLogFormat.java
@@ -16,6 +16,6 @@ public class ClassicLogFormat {
         // `AbstractNCSARequestLog`, and this is the best approximation
         // of that class
         return "%{client}a - %u %{dd/MMM/yyyy:HH:mm:ss Z|" + tz.getID()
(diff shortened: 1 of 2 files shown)


zUniQueX:
Hi @wilbertpol, thanks for your contribution!

Unfortunately, I'm not quite sure, if we can integrate this PR into Dropwizard 2.x. Users _could_ have sensitive information in their query parameters that shouldn't be logged. If we would change the log format, this would lead to an information exposure security issue.

But I totally understand the need of making the log format configurable. Jetty declares the [`EXTENDED_NCSA_FORMAT`](https://www.eclipse.org/jetty/javadoc/jetty-9/constant-values.html#org.eclipse.jetty.server.CustomRequestLog.EXTENDED_NCSA_FORMAT), which is similar to this format but is a standard. The most general solution I can now imagine is to extend the `ClassicLogFormat` with a static field and setter for a `java.util.function.Function<TimeZone, String>` supplying the current pattern. This would allow users to specify a custom lambda expression with their own format and even with the NCSA format at initialization time of the application.

For Dropwizard 3.x we could use `%r` instead of `%m %U %H` in the pattern.

I would appreciate an updated PR on this topic and maybe a follow-up one for Dropwizard 3.x.

PR #5361, opened 2022-06-01
Spotless with Palantir Java format

joschi:
Introduce [Spotless](https://github.com/diffplug/spotless/tree/maven/2.22.5/plugin-maven) for automatic code formatting.

<a href="https://gitpod.io/#https://github.com/dropwizard/dropwizard/pull/5361"></a>


zUniQueX:
Any reason to switch the formatting now after we recently introduced formatting with IntelliJ defaults (e.g. in #3233)? Do we simply want to be independent of IntelliJ?

joschi:
> Do we simply want to be independent of IntelliJ?

This kind of was the idea, but I also don't see enough advantages of using this right now.

zUniQueX:
> This kind of was the idea, but I also don't see enough advantages of using this right now.

Thanks for clarification. I just wondered why we now would want to switch to an other formatter. I like the idea of format enforcement at build time so the code style is consistent throughout the whole project. But I think it will make development harder, if the IDE uses other formatting rules.

PR #5413, opened 2022-06-18
Update maven-enforcer-plugin to v3.1.0

### LINKED ISSUE #4196: dropwizard-example maven-enforcer-plugin Dependency Convergence outputs invalid results
I was checking the `dropwizard-example` project and I saw that it uses very old `1.4.1` version of `maven-enforcer-plugin`. When I updated the version to `3.0.0` (the latest one) - the DependencyConvergency task started to fail and indeed the dependency convergency fails. I think this should be updated and either `<DependencyConvergence />` rule should be removed or `dropwizard-dependencies` should point to the correct artifacts.

zUniQueX:
`maven-enforcer-plugin` had an issue in v3.0.0 with the `DependencyConvergence` rule. v3.1.0 seems to have that fixed.

Refs #4196


joschi (approved):
Thanks! 😄

PR #5385, opened 2022-06-06
Generate SBOM for Dropwizard using CycloneDX Maven plugin

joschi:
https://cyclonedx.org/
https://github.com/CycloneDX/cyclonedx-maven-plugin

PR #5419, opened 2022-06-19
Fix CycloneDX execution

zUniQueX:
Seems like the CycloneDX execution was skipped for the release of 2.0.30. I've tested this locally and somehow the plugin only gets invoked, if the `executions` section is specified.


joschi (approved):
@zUniQueX Thanks!

joschi:
Refs https://github.com/dropwizard/dropwizard/pull/5385

PR #5430, opened 2022-06-20
Use Visual Studio 2022 image and Java 17 on AppVeyor

joschi:
Update to latest runtime image and Java 17 on AppVeyor.


zUniQueX:
Good to see the build directly succeeds :+1:

PR #5437, opened 2022-06-21
Remove code source path in favor of `CloseableLiquibaseWithClassPathMigrationsFile`

### LINKED ISSUE #5428: `dropwizard-migrations` uses wrong class for resolving files on the file system
For resolving files on the file system in the `dropwizard-migrations` module, the root paths are provided as well as a code source path ([see here](https://github.com/dropwizard/dropwizard/blob/release/2.1.x/dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibaseWithFileSystemMigrationsFile.java#L43-L45)). The code source is thereby the `dropwizard-migrations` jar instead of the currently executing jar (when not using a shaded jar).

As we have the option to use the `CloseableLiquibaseWithClassPathMigrationsFile` for resolving classes in the current jar, it should be enough to provide the root paths from the file system for the `CloseableLiquibaseWithFileSystemMigrationsFile`. Therefore the code source path can be removed.

zUniQueX:
Closes #5428

### DIFF
--- a/dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibaseWithFileSystemMigrationsFile.java
+++ b/dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibaseWithFileSystemMigrationsFile.java
@@ -41,11 +41,6 @@ public class CloseableLiquibaseWithFileSystemMigrationsFile extends CloseableLiq
     private static List<File> getRootPaths(FileSystem fileSystem) {
         List<File> rootPaths = new ArrayList<>();
         fileSystem.getRootDirectories().forEach(path -> rootPaths.add(path.toFile()));
-        try {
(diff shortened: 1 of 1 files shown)


dropwizard-migrations/src/main/java/io/dropwizard/migrations/CloseableLiquibaseWithFileSystemMigrationsFile.java
```diff
@@ -41,11 +41,6 @@ public class CloseableLiquibaseWithFileSystemMigrationsFile extends CloseableLiq
     private static List<File> getRootPaths(FileSystem fileSystem) {
```
rhowe, line 41: Should this method be inlined now?

zUniQueX: I don't see a big advantage either way. Therefore I'd leave this as-is.

joschi (approved):
Thanks! 😄

PR #5464, opened 2022-06-28
Change access modifier of `@UnitsOfWork` to public

zUniQueX:
###### Problem:
When using multiple `@UnitOfWork` annotations, the compiler complains that the `@UnitsOfWork` container annotation cannot be found.

###### Solution:
Change the access modifier of the `@UnitsOfWork` annotation to public.


joschi (approved):
Duh!

Thanks for the fix!

zUniQueX:
I should have tested this directly in a standalone application :facepalm:

The documentation on repeatable annotations unfortunately doesn't leave a word on this topic.

PR #5472, opened 2022-06-28
Support for customizing the application's ObjectMapper

joschi:
Some users of Dropwizard 2.x and Java 8 want to be able to customize the application's `ObjectMapper` instance in a more granular way than it was possible before.

This change set adds the `ObjectMapperFactory` interface and a default implementation which replaces the existing `Jackson` class which offered only static methods to construct a preconfigured `ObjectMapper` instance.

Refs https://github.com/dropwizard/dropwizard/discussions/5319

### DIFF
--- a/docs/source/manual/internals.rst
+++ b/docs/source/manual/internals.rst
@@ -40,7 +40,7 @@ Startup Sequence
 
 	Bootstrap(application: Application<T>) {
 	  this.application = application;
-	  this.objectMapper = Jackson.newObjectMapper();
+	  this.objectMapper = new DefaultObjectMapperFactory().newObjectMapper();
 	  this.bundles = new ArrayList<>();
 	  this.configuredBundles = new ArrayList<>();
 	  this.commands = new ArrayList<>();
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -568,7 +568,7 @@ assert the expected widget is deserialized based on the ``type`` field.
 
     public class WidgetFactoryTest {
 
-        private final ObjectMapper objectMapper = Jackson.newObjectMapper();
+        private final ObjectMapper objectMapper = new DefaultObjectMapperFactory().newObjectMapper();
         private final Validator validator = Validators.newValidator();
         private final YamlConfigurationFactory<WidgetFactory> factory =
                 new YamlConfigurationFactory<>(WidgetFactory.class, validator, objectMapper, "dw");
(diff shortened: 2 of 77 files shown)


zUniQueX:
@joschi I think this is a good idea overall, but that seems a bit too complicated to me.

What about extending the `Application` class with a `Consumer<ObjectMapper>`, that does the configuration which currently is provided by the `Jackson#configure(ObjectMapper)` method?

Then we would add a method `Jackson.newObjectMapper(Consumer<ObjectMapper> configurer)` and let `Jackson.newObjectMapper()` default to `Jackson.newObjectMapper(objectMapper -> configure(objectMapper))`. This change should preserve ABI compatibility and let users configure the `ObjectMapper` through the `Application`.

zUniQueX:
I've opened an other draft PR with my described approach. Feel free to close it, if it's not flexible enough.

joschi:
> What about extending the `Application` class with a `Consumer<ObjectMapper>`, that does the configuration which currently is provided by the `Jackson#configure(ObjectMapper)` method?

@zUniQueX I don't think this is improving the situation over what we have right now.

Users can configure and set the `ObjectMapper` for the application (and bootstrap) with [`Bootstrap#setObjectMapper(ObjectMapper)`](https://javadoc.io/static/io.dropwizard/dropwizard-core/2.1.0/io/dropwizard/setup/Bootstrap.html#setObjectMapper-com.fasterxml.jackson.databind.ObjectMapper-) in the [`Application#initialize(Bootstrap)`](https://javadoc.io/static/io.dropwizard/dropwizard-core/2.1.0/io/dropwizard/Application.html#initialize-io.dropwizard.setup.Bootstrap-) method.

If I understood correctly, the users in https://github.com/dropwizard/dropwizard/discussions/5319 want to have the possibility to get a configured `ObjectMapper` without any JDK-specific modules (i. e. Jackson Afterburner or Blackbird) and then be able to customize this further.

With the approach you've suggested, they would still need to copy the code of the default `Consumer<ObjectMapper>` which we deliver with Dropwizard.

To minimize changes, we could offer a new static factory method in the `Jackson` class which returns an `ObjectMapper` instance without Jackson Afterburner or Blackbird, but that would also not make customizing these factory any easier without copying their code.

zUniQueX:
@joschi Sounds reasonable to me. The referenced PR however has one advantage over the current codebase: the modules used in between the Bootstrap construction and the initialize method previously were not configurable and therefore the Blackbird warning gets issued. The small change lets users completely disable Blackbird and therefore stop registration on Bootstrap construction.

As you stated, this will not be a big advantage for most users. It was just a thought to implement the desired behaviour requested in the discussion with minimal changes.

joschi:
With https://github.com/dropwizard/dropwizard/pull/5477 resolving the immediate issue, I'll close this PR.

PR #5500, opened 2022-07-07
Wrap exceptions thrown by beforeStart with MappableException

andref:
###### Problem:

We have a custom `HibernateBundle` that provides multi-tenancy and must support the case where tenant database instances are offline. We'd like to handle the errors caused by an offline tenant using `ExceptionMapper` as well, but that's impossible because exceptions thrown during the `beforeStart` phase are not wrapped in `MappableException`. (See #5501.)

This is similar to issue https://github.com/dropwizard/dropwizard/issues/949. Back in those good old times, it was impossible to create `ExceptionMapper`s to handle persistence exceptions thrown during the `afterEnd` phase of the unit of work. The fix in PR https://github.com/dropwizard/dropwizard/pull/958 wraps the exception with Jersey's `MappableException`before rethrowing.

[This gist](https://gist.github.com/andref/9869a3c483b57eb8bf1f145740fe5c71) is a minimal test case.

###### Solution:

* Exceptions thrown by `UnitOfWorkAspect.beforeStart` are wrapped in `MappableException`. 
* I'm unsure about the test: I had to config the H2 TCP server, so that I could turn it off and trigger an error during beforeStart. That might not be appropriate in the project's build infrastructure.
* It might be sensible to wrap any exception thrown by the `UnitOfWorkEventListener`. I didn't do that though.

###### Result:

If an exception is thrown during the `beforeStart` stage, the user can create an `ExceptionMapper` to produce a web response in a way similar to what is already possible during the `afterEnd` stage.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
@@ -73,13 +73,17 @@ public class UnitOfWorkApplicationListener implements ApplicationEventListener {
         public void onEvent(RequestEvent event) {
             final RequestEvent.Type eventType = event.getType();
             if (eventType == RequestEvent.Type.RESOURCE_METHOD_START) {
-                methodMap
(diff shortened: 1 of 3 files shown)


andref:
Do I need to create an issue as well?

andref:
I forgot to create a separate branch, sorry.

Issue #5501, opened 2022-07-07
[dropwizard-hibernate] Custom ExceptionMapper not invoked when Hibernate session is opened


andref:
### Overview

This is similar to issue #949. Back in those good old times, it was impossible to create `ExceptionMapper`s to handle persistence exceptions thrown during the `afterEnd` phase of the unit of work. The fix in PR #958 wraps the exception with Jersey's `MappableException` before rethrowing.

We have a custom `HibernateBundle` that provides multi-tenancy and must support the case where tenant database instances are offline. We'd like to handle the errors caused by an offline tenant using `ExceptionMapper`s as well, but that's impossible because exceptions thrown during the `beforeStart` phase are not wrapped in `MappableException`.

### Steps to reproduce

Using the provided [minimal test case](https://gist.github.com/andref/9869a3c483b57eb8bf1f145740fe5c71), first turn off the database server (using httpie here, `curl -X` at will):

    http POST localhost:3000/db/stop

Request the resource method that has a unit of work:

    http GET localhost:3000/db

This normally returns "hello, world", but if the database is turned off, an exception will be thrown during the `beforeStart` phase. The app has an exception mapper that can handle the situation (see stack trace) and respond with a nice HTTP 503, but it is never invoked and we get an ugly 500 instead:

```http
HTTP/1.1 500 Server Error
Cache-Control: must-revalidate,no-cache,no-store
Connection: close
Content-Length: 382
Content-Type: text/html;charset=iso-8859-1

<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 500 Request failed.</title>
</head>
<body><h2>HTTP ERROR 500 Request failed.</h2>
<table>
<tr><th>URI:</th><td>/db</td></tr>
<tr><th>STATUS:</th><td>500</td></tr>
<tr><th>MESSAGE:</th><td>Request failed.</td></tr>
<tr><th>SERVLET:</th><td>jersey</td></tr>
</table>

</body>
</html>
```

<details>
<summary>Stack trace</summary>
<pre>
WARN  [2022-07-07 18:33:38,211] org.eclipse.jetty.server.HttpChannelState: unhandled due to prior sendError
! java.net.SocketTimeoutException: Connect timed out
! at java.base/sun.nio.ch.NioSocketImpl.timedFinishConnect(NioSocketImpl.java:546)
[...]

PR #5506, opened 2022-07-08
Wrap exceptions thrown by beforeStart with MappableException

andref:
###### Problem:

We have a custom `HibernateBundle` that provides multi-tenancy and must support the case where tenant database instances are offline. We'd like to handle the errors caused by an offline tenant using `ExceptionMapper`, but that's impossible because exceptions thrown during the `beforeStart` phase are not wrapped in `MappableException`. (See #5501.)

This is similar to issue #949. Back in those good old times, it was impossible to create `ExceptionMapper`s to handle persistence exceptions thrown during the `afterEnd` phase of the unit of work. The fix in PR #958 wraps the exception with Jersey's `MappableException`before rethrowing.

[This gist](https://gist.github.com/andref/9869a3c483b57eb8bf1f145740fe5c71) is a minimal test case.

###### Solution:

* `HibernateException`s thrown by `UnitOfWorkAspect.beforeStart` are wrapped in `MappableException`. This is stricter than what is done during `afterEnd`, where any Exception is wrapped.
* I'm unsure about the test: I had to config the H2 TCP server, so that I could turn it off and trigger an error during beforeStart. That might not be appropriate in the project's build infrastructure.
* It might be sensible to wrap any exception thrown by the `UnitOfWorkEventListener`. I didn't do that though.

###### Result:

If an exception is thrown during the `beforeStart` stage, the user can create an `ExceptionMapper` to produce a web response in a way similar to what is already possible during the `afterEnd` stage.


dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
```diff
@@ -6,8 +6,10 @@
 import org.glassfish.jersey.server.monitoring.ApplicationEventListener;
 import org.glassfish.jersey.server.monitoring.RequestEvent;
 import org.glassfish.jersey.server.monitoring.RequestEventListener;
+import org.hibernate.HibernateException;
 import org.hibernate.SessionFactory;
 
+import javax.persistence.PersistenceException;
```
zUniQueX: This import isn't used and should be removed.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
```diff
@@ -73,13 +75,17 @@ private static class UnitOfWorkEventListener implements RequestEventListener {
...
-                );
+                try {
+                    methodMap
+                        .computeIfAbsent(event.getUriInfo().getMatchedResourceMethod(), UnitOfWorkEventListener::registerUnitOfWorkAnnotations)
+                        .forEach(unitOfWork ->
+                            unitOfWorkAspects
+                                .computeIfAbsent(unitOfWork.value(), hibernateName -> new UnitOfWorkAspect(sessionFactories))
+                                .beforeStart(unitOfWork)
+                    );
+                } catch (HibernateException e) {
```
zUniQueX: I think we should stay consistent here with the handling around `afterEnd` and catch all `Exception`s and not only `HibernateException`s.

dropwizard-hibernate/src/test/java/io/dropwizard/hibernate/SQLNonTransientConnectionExceptionMapper.java
```diff
@@ -0,0 +1,19 @@
...
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+import java.sql.SQLNonTransientConnectionException;
+
+@Provider
+public class SQLNonTransientConnectionExceptionMapper implements ExceptionMapper<SQLNonTransientConnectionException> {
+    @Override
+    public Response toResponse(SQLNonTransientConnectionException e) {
+        return Response.status(Response.Status.SERVICE_UNAVAILABLE)
+                       .entity(new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(), "Connection not available"))
```
zUniQueX: Although this status isn't used, I'd suggest to change it to 503 too.

zUniQueX (changes requested):
Thanks for the contribution!

zUniQueX:
Hi @andref. I've now checked, if we can simplify the test and stop using the server.

The following code closes the database's `Session` and produces a `JdbcSQLNonTransientConnectionException` too:

```java
try (Session session = hibernate.getSessionFactory().openSession()) {
    session.doWork(connection -> connection.unwrap(JdbcConnection.class).getSession().close());
}
```

Maybe we can utilize this to avoid introducing the H2 server.

andref:
Hi @zUniQueX! Unfortunately, by the time we have a current `Session` from which to extract the `JdbcConnection`, the `beforeStart` phase has already run and the error must happen inside the `UnitOfWorkAspect.beforeStart`.

I'm fixing the other items in your review, thanks!

zUniQueX:
@andref To obtain the `JdbcConnection`, the `beforeStart` method doesn't have to be invoked. You can obtain the `SessionFactory` from your application to the executing test and instantiate the `Session` manually without the need to use the `UnitOfWorkAspect`. Or am I missing something here?

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkApplicationListener.java
andref: I originally did just that. However, `beforeStart` throws a diagnostic `IllegalArgumentException` if the `@UnitOfWork` being set up specifies a Hibernate bundle name that is not registered during application start up. I reasoned that errors like those are actually bugs in the application and wouldn't usually be handled through an `ExceptionMapper`. Should I wrap the `IAE` with `MapableException` anyway or should special-case it?

zUniQueX: We shouldn't change the current behaviour. As the `IllegalArgumentException` is the only one intentionally thrown, we can handle it as a special case.

andref:
> Or am I missing something here?

@zUniQueX Only because I didn't explain myself properly. 😅

The integration test case I wrote is verifying specifically whether an exception thrown _during the `beforeStart` phase_ can be handled by an `ExceptionMapper` — that is the behavior this PR implements. As such, how to obtain a `Session` and close its connection is beside the point if the exception that would cause can't happen inside `UnitOfWorkAspect.beforeStart`.

I'm wondering that maybe setting up the `DataSourceFactory` with a bogus validation query and `checkConnectonOnBorrow = true` might do the trick.

andref:
@zUniQueX I replaced the integration test case with a simple unit test. The change I'm proposing is just too small.

zUniQueX:
@andref As we've now discussed how to write an integration test, I added a simple one :wink:

zUniQueX (approved):

andref:
> @andref As we've now discussed how to write an integration test, I added a simple one 😉

Oh... OH! Now I understand.

Issue #5518, opened 2022-07-12
add ability to configure custom alive endpoints


richeyh:
Currently after the addition of the health-module into Dropwizard core it allows users to pass a list of [custom endpoints](https://github.com/dropwizard/dropwizard/blob/release/2.1.x/dropwizard-health/src/main/java/io/dropwizard/health/DefaultHealthFactory.java#L61) that it will also listen on. However, these endpoints all default to the readiness of the application unless you pass the `?type=alive` query param. It would be nice if you could configure another list of liveness endpoints that defaulted to the above behavior without the query param present. 

This would bring this project in line with [spring boots probe settings](https://spring.io/blog/2020/03/25/liveness-and-readiness-probes-with-spring-boot) and better in line with [k8s probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

I'm happy to contribute the changes for this if the maintainers agree.

richeyh:
Also needed as part of this and would be a follow on ask is support for Regex in the [name field of the HealthCheckConfiguration](https://github.com/dropwizard/dropwizard/blob/release/2.1.x/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckConfiguration.java#L14). Without that if you have a library that registers multiple checks for things like Kafka for example you could not mark all of the checks critical if you have multiple connections or have to enumerate every single connection for the application twice.

pstackle:
Related to the original issue, I asked a question on the PR. 

For the follow-on ask:
Right now there is a simple 1:1 mapping between a registered `HealthCheck` and the configuration that determines if a given check is critical or not and whether the overall health of the app is healthy or not. Supporting a regex for the name seems like it would add a lot of complexity to the health evaluation. 

If you have a specific case where you want a whole set of HealthChecks to effectively be treated as a single one (i.e. all are marked critical and have the same schedule config), you could create a sort of aggregator `HealthCheck` that you register and configure once. That aggregator `HealthCheck` would have a collection of the relevant `HealthCheck`s and its `check()` method would just iterate over the collection of checks, call `check()` on each one, and then return a `Result` based on whether they were all healthy or if one or more of them were unhealthy. I think that would address your use case and wouldn't need to overcomplicate the existing health evaluation logic.

richeyh:
[...]

PR #5549, opened 2022-07-18
add feature to allow adding custom liveness paths to dropwizard health

richeyh:
###### Problem:
This is written up in https://github.com/dropwizard/dropwizard/issues/5518

TLDR:
Need a way to break up alive and readiness checks on separate URL's

###### Solution:
This solution modifies the `ServletHealthResponderFactory.java` to have a new optional configurable field `livenessCheckUrlPaths`. This field is a list of string URLS that should only return the results of the alive checks. This makes it a non breaking change as the probably better solution would be to add the field to the `DefaultHealthFactory` and rename the `healthCheckUrlPaths` to be more clear. However, this change would be breaking and probably best served for the 3.x or 4.x release. 

###### Result:
There is now a new option to give consumers further control of their health checks.

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -491,6 +491,37 @@ Define the following health check configurations in your `config.yml` file:
         - name: user-cache
           critical: false
 
+For a full list of configuration options see the following snippet
+
+.. code-block: yaml
+    health:
+      enabled: true
+      delayedShutdownHandlerEnabled: true
+      shutdownWaitPeriod: 10s
+      initialOverallState: true
+      healthCheckUrlPaths: ["/alive/health", "/alive/health/readiness"]
+      responseProvider:
+        type: json
+      responder:
+        type: servlet
+        cacheControlEnabled: true
+        cacheControlValue: "no-store"
+        livenessCheckUrlPaths: ["/alive/health/liveness"]
+      healthChecks:
+        - name: ready-check
+          critical: false
+          initialState: true
+          type: ready
+          schedule:
+            checkInterval: 2500ms
+            downtimeInterval: 10s
+            failureAttempts: 2
+            successAttempts: 1
+        - name: alive-check
+          critical: false
+          initialState: true
+          type: alive
+
 .. note::
 
     This behavior was integrated from the `Dropwizard Health module`_. If you are migrating from that module
--- a/dropwizard-health/src/main/java/io/dropwizard/health/response/ServletHealthResponder.java
+++ b/dropwizard-health/src/main/java/io/dropwizard/health/response/ServletHealthResponder.java
@@ -10,10 +10,16 @@ import javax.ws.rs.core.HttpHeaders;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import io.dropwizard.health.HealthCheckType;
+
 import static java.util.Objects.requireNonNull;
+
+import static io.dropwizard.health.response.JsonHealthResponseProvider.CHECK_TYPE_QUERY_PARAM;
 
 public class ServletHealthResponder extends HttpServlet {
     private static final Logger LOGGER = LoggerFactory.getLogger(ServletHealthResponder.class);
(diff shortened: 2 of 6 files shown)


pstackle:
@richeyh Can you explain in more detail your use case such that the current behavior doesn't support it? 

The current approach does allow you to configure separate liveness and readiness probes for a Kubernetes pod via the query string parameter you mentioned:
```
# use for readiness probe
https://<hostname>:<port>/health-check?type=ready

# use for liveness probe
https://<hostname>:<port>/health-check?type=alive
```

This approach has the benefit of making it easier to support additional types of health probes that come up in the future as well. 

The purpose of the `healthCheckUrlPaths` was not to have different paths return different results or be tied to a specific type of health check, but rather to support an app that needed to migrate the health check URL used from one location to another and allowing to configure multiple enables that to happen seamlessly. In most cases, there is no need to ever configure anything beyond the default value and even in the case of a migration, one wouldn't typically need more than two.

The reason that the health check call returns the readiness result by default is that it was initially the only one that was supported, the `alive` support came later, and it is also the most commonly used one.

richeyh:
Yep, so right now with the default being liveness it essentially means unless you use query params you can't get two distinct urls. Our use case is we have our health check library that currently supports this and we want to move to the health check library that's part of core. However, for us, it's important to keep the urls the same which would mean the ability to configure a new url for liveness alone(not just health as it currently stands). As this seemed like a good use case for teams vs using a query param.  Also when you look at other frameworks like spring boot, which allows url config per type, it seemed like this would be a welcome addition.

pstackle:
Can you explain this in more detail?
> As this seemed like a good use case for teams vs using a query param.

I think I'm still missing something with respect to what the problem is with query parameters.

richeyh:
> Can you explain this in more detail?

Yep so right now you can only configure `healthCheckUrlPaths`  this corresponds to endpoints that only function for readiness of the application. There is no config currently to add a distinct endpoint for only livenesss. The use of the query param is a workaround but it still does not give applications a way to have two endpoints that serve different purposes, by default every endpoint I configure would have to be for readiness. Now if we go look at other frameworks we see in [spring boot](https://www.baeldung.com/spring-boot-actuators#6-health-groups) they are distinct configurable endpoints to further clarify the difference in readiness. 

 My full use case to match the current behavior and runtime constraints of my environment is I need my app to serve readiness on `/management/health/readiness` and liveness on `/management/health/liveness`. Yes I can probably achieve this with a redirect filter, however it seems worthwhile that other users would want the pattern of configurability for the endpoints that are determining the liveness and readiness of their application.

pstackle:
> My full use case to match the current behavior and runtime constraints of my environment is I need my app to serve readiness on /management/health/readiness and liveness on /management/health/liveness.

If your app is deployed in Kubernetes, what prevents you from configuring the liveness and readiness probe urls as part of your pod spec?

```
    livenessProbe:
      httpGet:
        path: /health-check?type=alive
    readinessProbe:
      httpGet:
        path: /health-check?type=ready
```

As I mentioned earlier, trying to support a list of URLs per status type doesn't seem like that would scale very well to me. 

Is the hangup related to the use of query params or the fact that you don't have control of the status type value? 

Hypothetically, if the following worked, would that somehow work better for you or would you still be blocked?
```
alive: /health-check/alive
ready: /health-check/ready

general pattern: /[configured url]/[status type]
```

richeyh:
> If your app is deployed in Kubernetes, what prevents you from configuring the liveness and readiness probe urls as part of your pod spec?

We have internal contracts for all our applications to match. Right now our node apps and our spring boot apps can conform to the contract. Our Dropwizard apps with our bundle can also conform. However, without a change in the core health check bundle, it will not be able to conform. 

> As I mentioned earlier, trying to support a list of URLs per status type doesn't seem like that would scale very well to me.

Do you mean there would be more than just the two status types? Are you looking to adopt something like health check groups? My understanding is that there would only be two groups/types of health checks alive and ready. 

> Hypothetically, if the following worked, would that somehow work better for you or would you still be blocked?

It's closer, we would also want to be able to customize the name of the groups or status type from your example.

pstackle:
To close the loop on this:
It sounds like your requirements are pretty specific to the exact strings you need for your probe urls, so doing your own redirect filter is probably the better approach. 

Trying to support a list of URLs per status type wouldn't scale very well as Kubernetes expands the types of probes that it supports.

Given the above, I'm going to close this PR. If you strongly disagree, please open a Discussion item to allow for further discussion about the use case.

PR #5638, opened 2022-08-02
Replace `org.glassfish.hk2.external:jakarta.inject` with `jakarta.inject:jakarta.inject-api`

### LINKED ISSUE #5527: overlapping use of jakarta.inject and jakarta.inject-api
jakarta.inject and jakarta.inject-api both include javax.inject classes. 

dropwizard-core depends on jersey-bean-validation and jersey-common which both depend on jakarta.inject. core also depends on dropwizard-jersey which depends on jakarta.inject-api. 

dropwizard-testing also uses jakarta-inject, while dropwizard-auth uses jakarta.inject-api.

I can use exclusions in my pom file to work around this, but dropwizard should be consistent on which one it uses across projects.

zUniQueX:
`org.glassfish.hk2.external:jakarta.inject` is a repackaged OSGi version of `javax.inject` and can be replaced by `jakarta.inject:jakarta.inject-api`.

Closes #5527

### DIFF
--- a/dropwizard-auth/pom.xml
+++ b/dropwizard-auth/pom.xml
@@ -60,14 +60,32 @@
         <dependency>
             <groupId>org.glassfish.jersey.containers</groupId>
             <artifactId>jersey-container-servlet-core</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.glassfish.hk2.external</groupId>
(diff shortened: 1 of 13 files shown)


joschi (approved):
LGTM, thanks! 😄

PR #5689, opened 2022-08-07
Properly handle `null` in PortRangeValidator

### LINKED ISSUE #5684: [dropwizard-validation] @PortRange validation throws NullPointerException if value is null
#### Issue description:
Configuration validation of an `Integer` property annotated with `@PortRange` throws a NullPointerException if the property value is unset (that is, `null`):

```
javax.validation.ValidationException: HV000028: Unexpected exception during isValid call.
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateSingleConstraint(ConstraintTree.java:186)
	at org.hibernate.validator.internal.engine.constraintvalidation.SimpleConstraintTree.validateConstraints(SimpleConstraintTree.java:62)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:75)
	at org.hibernate.validator.internal.metadata.core.MetaConstraint.doValidateConstraint(MetaConstraint.java:130)
	at org.hibernate.validator.internal.metadata.core.MetaConstraint.validateConstraint(MetaConstraint.java:123)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateMetaConstraint(ValidatorImpl.java:555)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForSingleDefaultGroupElement(ValidatorImpl.java:518)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForDefaultGroup(ValidatorImpl.java:488)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForCurrentGroup(ValidatorImpl.java:450)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateInContext(ValidatorImpl.java:400)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validate(ValidatorImpl.java:172)
	at io.dropwizard.configuration.BaseConfigurationFactory.validate(BaseConfigurationFactory.java:244)
	at io.dropwizard.configuration.BaseConfigurationFactory.build(BaseConfigurationFactory.java:129)
	at io.dropwizard.configuration.BaseConfigurationFactory.build(BaseConfigurationFactory.java:110)
	at io.dropwizard.cli.ConfiguredCommand.parseConfiguration(ConfiguredCommand.java:139)
	at io.dropwizard.cli.ConfiguredCommand.run(ConfiguredCommand.java:85)
	at io.dropwizard.cli.Cli.run(Cli.java:78)
	at io.dropwizard.Application.run(Application.java:94)
	…
Caused by: java.lang.NullPointerException
	at io.dropwizard.validation.PortRangeValidator.isValid(PortRangeValidator.java:22)
	at io.dropwizard.validation.PortRangeValidator.isValid(PortRangeValidator.java:10)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateSingleConstraint(ConstraintTree.java:180)
	... 18 more
```
The exception will be thrown even if the property is also annotated with `@NotNull`.

I think the issue is caused by a missing `null` check in `PortRangeValidator.isValid`: https://github.com/dropwizard/dropwizard/blob/d9a937ea62718b80c03bc7d8ffd1d3b444b1ef40/dropwizard-validation/src/main/java/io/dropwizard/validation/PortRangeValidator.java#L22

#### Expected behaviour

The `@PortRange` validator should silently accept a `null` value.

#### Code example
```java
import io.dropwizard.Configuration;
import io.dropwizard.validation.PortRange;
import javax.validation.constraints.NotNull;

public class AppConfiguration extends Configuration {

    @NotNull
    @PortRange
    private Integer port = null;
}
```

joschi:
The `PortRangeValidator` will now ignore `null` values of `Integer` fields and treat them as valid, similar to other validators are working.

If a port range should be mandatory consider using a primitive `int` or mark the `Integer` field as `@NotNull`.

Fixes #5684

### DIFF
--- a/dropwizard-validation/src/main/java/io/dropwizard/validation/PortRangeValidator.java
+++ b/dropwizard-validation/src/main/java/io/dropwizard/validation/PortRangeValidator.java
@@ -19,6 +19,6 @@ public class PortRangeValidator implements ConstraintValidator<PortRange, Intege
 
     @Override
     public boolean isValid(Integer value, ConstraintValidatorContext context) {
-        return value == 0 || (value >= min && value <= max);
+        return value == null || value == 0 || (value >= min && value <= max);
     }
 }
--- a/dropwizard-validation/src/test/java/io/dropwizard/validation/PortRangeValidatorTest.java
+++ b/dropwizard-validation/src/test/java/io/dropwizard/validation/PortRangeValidatorTest.java
@@ -86,7 +86,7 @@ public class PortRangeValidatorTest {
 
     @Test
     @SuppressWarnings("NullAway")
-    void rejectsNull() {
+    void acceptsNull() {
         example.nullablePort = null;
         assertThat(ConstraintViolations.format(validator.validate(example)))
             .isEmpty();
--- a/dropwizard-validation/src/test/java/io/dropwizard/validation/PortRangeValidatorTest.java
+++ b/dropwizard-validation/src/test/java/io/dropwizard/validation/PortRangeValidatorTest.java
@@ -21,6 +21,9 @@ public class PortRangeValidatorTest {
         @PortRange(min = 10000, max = 15000)
         public int otherPort = 10001;
 
+        @PortRange
+        public Integer nullablePort = 1;
+
         @Valid
         List<@PortRange Integer> ports = Collections.emptyList();
     }
@@ -79,5 +82,13 @@ public class PortRangeValidatorTest {
         example.ports = Collections.singletonList(-1);
         assertThat(ConstraintViolations.format(validator.validate(example)))
             .containsOnly("ports[0].<list element> must be between 1 and 65535");
+    }
+
+    @Test
+    @SuppressWarnings("NullAway")
+    void rejectsNull() {
+        example.nullablePort = null;
+        assertThat(ConstraintViolations.format(validator.validate(example)))
+            .isEmpty();
     }
 }


ro-kr:
The change causes the PortRangeValidator to _reject_ a null value. Is that intentional? It would be inconsistent with the other validators as they _accept_ a null value. I think rejecting a null value should be handled by `@NotNull`.

dropwizard-validation/src/main/java/io/dropwizard/validation/PortRangeValidator.java
```diff
@@ -19,6 +19,6 @@ public void initialize(PortRange constraintAnnotation) {
 
     @Override
     public boolean isValid(Integer value, ConstraintValidatorContext context) {
-        return value == 0 || (value >= min && value <= max);
+        return value != null && (value == 0 || (value >= min && value <= max));
```
ro-kr: The change causes the PortRangeValidator to _reject_ a null value. Is that intentional? It would be inconsistent with the other validators as they _accept_ a null value. I think rejecting a null value should be handled by `@NotNull`.

joschi: @ro-kr Good catch! That was indeed not intentional. Thanks for catching this! 😄

dropwizard-validation/src/test/java/io/dropwizard/validation/PortRangeValidatorTest.java
```diff
@@ -80,4 +83,12 @@ void rejectsInvalidPortsInList() {
         assertThat(ConstraintViolations.format(validator.validate(example)))
             .containsOnly("ports[0].<list element> must be between 1 and 65535");
     }
+
+    @Test
+    @SuppressWarnings("NullAway")
+    void rejectsNull() {
```
zUniQueX: Shouldn't this method be named `acceptsNull`?

joschi: That was a leftover of https://github.com/dropwizard/dropwizard/pull/5689#discussion_r939708200. 😓

zUniQueX: Okay, thanks :+1:

PR #5691, opened 2022-08-09
Allow parameter conversion exceptions to be mapped to responses by exception mappers

### LINKED ISSUE #5381: FuzzyEnumParamConverter does not allow some error responses to be remapped using exception mappers
If `FuzzyEnumParamConverter` calls the `fromStringMethod` but is unable to convert the string to an enum value, it will throw a `WebApplicationException` constructed using a `Response` parameter instead of constructed using a `String` message. This will cause Jersey to return the `Response` as-is, bypassing any exception mappers that may be set up. This prevents exception mappers from being used to return error messages in a consistent format if an invalid enum value is provided as a path parameter. `FuzzyEnumParamConverter` should construct the `WebApplicationException` using a `String` message parameter instead of a `Response` so that it can be remapped into a response by the exception mappers.

I'm planning on submitting a PR to fix this issue.

tmelmoth:
###### Problem:

This PR fixes #5381

If a client sends a request to a resource that has a parameter of an enum type or a type that extends `AbstractParam`, and the parameter cannot be parsed, an error is returned to the client in a fixed format of `{"code":code,"message":"message"}` instead of having an opportunity of being mapped to a response by the exception mappers.

Most errors in Dropwizard result in throwing an exception which can be mapped to a response by exception mappers. This allows exception mappers to be implemented to return error responses in a consistent format for most types of errors.

The `FuzzyEnumParamConverterProvider` and `AbstractParam` parameter converters are used for enum parameters and parameters of types that extend `AbstractParam`. If the parameters cannot be parsed, they will throw a `WebApplicationException` with a `Response` containing an entity. Jersey will return this entity directly, instead of mapping the exception to a response using the exception mappers. See `org.glassfish.jersey.server.ServerRuntime.Responder#mapException` for the logic that directly returns the entity if the `WebApplicationException` contains one.

###### Solution:

This PR fixes `FuzzyEnumParamConverterProvider` and `AbstractParam` so that they throw a `WebApplicationException` with a status code and message that can be mapped to a response by exception mappers, instead of a `WebApplicationException` containing an entity.

###### Result:

As long as the default exception mappers provided by Dropwizard are registered (`registerDefaultExceptionMappers` option is on), then the response sent to the client will be identical to before this change for enum types and all subclasses of `AbstractParam` that exist in this repo.

If the default exception mappers are not registered (`registerDefaultExceptionMappers` option is off), then the response may change. This is probably the desired behavior if the exception mappers are overridden to return error responses in a consistent format, but could result in unexpected results if no replacement exception mappers are registered, or if the app expects the responses for parameter parsing errors to be in a different form than other error responses.

This change attempts to keep backwards compatibility for any classes that extend `AbstractParam` by keeping the `protected Response error(String input, Exception e)` method fingerprint the same. Subclasses which override `AbstractParam.error(String input, Exception e)` can still return a `Response` with an entity that will get returned directly instead of being mapped using exception mappers.


pstackle (approved):
Thanks for the contribution @tmelmoth !

This fixes a long-time pain point I've had with respect to getting consistently formatted error responses. I'm glad you were able to track down the source of the problem and provide a solution. 

Since there is a scenario that you mentioned where the behavior can change when someone has disabled the default exception mappers and/or has extended/overridden `AbstractParam` or any of its children, I think it would be helpful to include a mention/explanation in the upgrade notes. 
https://github.com/dropwizard/dropwizard/blob/release/2.1.x/docs/source/manual/upgrade-notes/upgrade-notes-2_1_x.rst

docs/source/manual/upgrade-notes/upgrade-notes-2_1_x.rst
```diff
@@ -146,3 +146,27 @@ However, if you are experiencing issues with the client modifications, you can e
...
+
+There are two types of parameters that may result in different error responses in the above situations if there is a validation error:
+
+* Enum parameters
+* Parameters that are any subclass of ``AbstractParam`` (Dropwizard provides the subclasses ``InstantParam``, ``InstantSecondParam``, ``LocalDateParam``, ``LocalDateTimeParam``, ``LocalTimeParam``, ``OffsetDateTimeParam``, ``YearMonthParam``, ``YearParam``, ``ZonedDateTimeParam``, ``ZoneIdParam``, ``BooleanParam``, ``DateTimeParam``, ``DurationParam``, ``InstantParam``, ``IntParam``, ``LocalDateParam``, ``LongParam``, ``NonEmptyStringParam``, ``SizeParam``, and ``UUIDParam``)
+
+If this code change causes an undesired change in the error responses:
+
+* If the error responses are not being mapped to a response by any exception mapper, then either register the default exception mappers (by enabling the ``registerDefaultExceptionMappers`` option) or register Dropwizard's ``ThrowableMapper`` to restore the previous behavior, or register your own custom exception mapper that maps ``WebApplicationException`` instances.
```
pstackle: I think you meant `LoggingExceptionMapper` here.
```suggestion
* If the error responses are not being mapped to a response by any exception mapper, then either register the default exception mappers (by enabling the ``registerDefaultExceptionMappers`` option) or register Dropwizard's ``LoggingExceptionMapper`` to restore the previous behavior, or register your own custom exception mapper that maps ``WebApplicationException`` instances.
```

pstackle (approved):
@tmelmoth Thanks for adding the release notes. They look good to me except for the one mapper name. Would you mind fixing that and then squashing and rebasing against the latest? 

@zUniQueX @joschi any objections for this change to get merged in for the next 2.1 release?

zUniQueX:
@pstackle Normally, I'd suggest to hold feature changes back for new minor releases. But as our release process will change with versions 3.x and 4.x, we must bring major updates for the 2.x series in minor releases. Therefore I have no preference for this change to be released either in 2.1.x or 2.2.x.

docs/source/manual/upgrade-notes/upgrade-notes-2_1_x.rst
tmelmoth: @pstackle Thanks for finding my typo, I've squashed the release note changes into one commit and rebased.

pstackle:
@zUniQueX I've been thinking of this as a bug fix and not a new feature because the fact that the framework already supports exception mappers to allow a developer to modify the error responses means that these particular corner cases are not intended/expected behavior. It just wasn't immediately noticeable because the bad behavior is only triggered in very specific corner cases and if you just used the default set of exception mappers, the fixed response format for these errors are indistinguishable from what the default mappers would create.

zUniQueX:
@pstackle If you say it's more like a bugfix, then this should be released in 2.1.x IMO. I personally try to deliver bugfixes as soon as possible

PR #5692, opened 2022-08-11
Add toJavaDuration to Duration

### LINKED ISSUE #5534: Rename dropwizard Duration
It would be nice if the DropWizard `Duration` class had a different name like `ConfigurationDuration` or `YamlReadableDuration` or something like that. I find that a lot of the APIs I use need a regular `java.time.Duration` and I ended up having to fully qualify one or the other. Most of us only use the DropWizard Duration as a simpler string to write in a config file. 

Another solution would be to add a method `toJavaDuration()` so I can just take one of those and turn it into the `java.time.Duration` I need. 

Minor for sure, just thought I'd document it while I was running into it.

sleberknight:
###### Problem:
As discussed in #5534 it is sometimes necessary to convert a Dropwizard `Duration` to a Java `java.time.Duration`.

###### Solution:
This PR adds the `toJavaDuration()` method to the Dropwizard `io.dropwizard.util.Duration` class.

###### Result:
It's now really easy to convert a Dropwizard `Duration` to a `java.time.Duration`

Closes #5534

### DIFF
--- a/dropwizard-util/src/test/java/io/dropwizard/util/DurationTest.java
+++ b/dropwizard-util/src/test/java/io/dropwizard/util/DurationTest.java
@@ -2,12 +2,15 @@ package io.dropwizard.util;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.time.temporal.ChronoUnit;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


dropwizard-util/src/test/java/io/dropwizard/util/DurationTest.java
```diff
@@ -888,4 +891,35 @@ void testSerialization() throws IOException, ClassNotFoundException {
...
+        "1 day, 1, DAYS",
+        "252 days, 252, DAYS",
+        "36500 days, 36500, DAYS"
+    })
+    void convertsToJavaDuration(String durationSpec, long amount, ChronoUnit unit) {
+        Duration duration = Duration.parse(durationSpec);
+
+        assertThat(duration.toJavaDuration())
+            .isEqualTo(java.time.Duration.of(amount, unit));
+    }
```
joschi, line 924: Thanks for writing the extensive test! ❤️

joschi (approved):
@sleberknight Thank you for your contribution! 😄

PR #5716, opened 2022-08-14
Suppress logging during dropwizard-validations tests

rhowe:
Suppress console logging during these tests

### DIFF
--- a/dropwizard-validation/src/test/java/io/dropwizard/validation/SelfValidationTest.java
+++ b/dropwizard-validation/src/test/java/io/dropwizard/validation/SelfValidationTest.java
@@ -2,14 +2,15 @@ package io.dropwizard.validation;
 
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
-import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.classic.spi.LoggingEvent;
 import ch.qos.logback.core.AppenderBase;
 import io.dropwizard.util.Maps;
 import io.dropwizard.validation.selfvalidating.SelfValidating;
+import io.dropwizard.validation.selfvalidating.SelfValidatingValidator;
 import io.dropwizard.validation.selfvalidating.SelfValidation;
 import io.dropwizard.validation.selfvalidating.ViolationCollector;
+import org.hibernate.validator.messageinterpolation.AbstractMessageInterpolator;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
@@ -32,6 +33,7 @@ class SelfValidationTest {
 
     private static class ListAppender extends AppenderBase<ILoggingEvent> {
         private final List<ILoggingEvent> events = new ArrayList<>();
+        private final Level level = Level.INFO;
 
         public List<ILoggingEvent> getAllLoggingEvents() {
             return events;
@@ -43,7 +45,7 @@ class SelfValidationTest {
 
         @Override
         protected void append(ILoggingEvent iLoggingEvent) {
-            if (iLoggingEvent.getLevel() != Level.DEBUG) {
+            if (iLoggingEvent.getLevel().isGreaterOrEqual(level)) {
                 events.add(iLoggingEvent);
             }
         }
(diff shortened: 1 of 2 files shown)


dropwizard-validation/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,11 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <outputPatternAsHeader>false</outputPatternAsHeader>
```
zUniQueX: This feature is [documented](https://logback.qos.ch/manual/encoders.html#outputPatternAsHeader) as disabled by default. Any reason this line must be added?

zUniQueX:
Thanks for the contribution!

I don't think it's a good idea to have different levels in the `logback-test.xml` and the filtering of the `ListAppender`. This PR would change the semantics of the `isEmpty()` check on the list of the `ListAppender`. Wouldn't it be a better idea to call `setAdditive(false)` for the logger of `SelfValidationTest` and use only the `ListAppender` and not the console appender?

As we do not rely on logging in the other tests, I don't have a strong opinion, if it's a good idea to set the log event globally or locally. But I'd want to avoid a situation, where the semantics of the appender would confuse developers.

dropwizard-validation/src/test/resources/logback-test.xml
joschi: Probably copy & pasted from some other source. Feel free to remove this line.

rhowe:
@zUniQueX I agree with everything you say, I just didn't see a nice alternative. I'll have a play around and see.

rhowe:
> @zUniQueX I agree with everything you say, I just didn't see a nice alternative. I'll have a play around and see.

Alright, I think I've managed to do something not too ugly...

dropwizard-validation/src/test/java/io/dropwizard/validation/SelfValidationTest.java
```diff
@@ -55,10 +58,12 @@ protected void append(ILoggingEvent iLoggingEvent) {
     @BeforeAll
     static void setUp() {
         LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
-        Logger root = loggerContext.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
-        listAppender.setContext(root.getLoggerContext());
+        Logger selfValidatingValidatorLogger = loggerContext.getLogger(SelfValidatingValidator.class);
+        selfValidatingValidatorLogger.setAdditive(false);
+        selfValidatingValidatorLogger.setLevel(Level.DEBUG);
```
zUniQueX: Logback defines the `DEBUG` level by default, but our appender can only process events with level `INFO` or higher. We _could_ change the level, but I don't have a strong opinion on either version.

dropwizard-validation/src/test/java/io/dropwizard/validation/SelfValidationTest.java
```diff
@@ -360,6 +368,13 @@ void violationMessagesAreEscapedByDefault() {
 
     @Test
     void messageParametersExample() {
+        Logger hibernateLogger = (Logger)LoggerFactory.getLogger(AbstractMessageInterpolator.class);
+        hibernateLogger.setAdditive(false);
+        hibernateLogger.setLevel(Level.DEBUG);
```
zUniQueX: Same as above with the `INFO` level.

dropwizard-validation/src/test/java/io/dropwizard/validation/SelfValidationTest.java
```diff
@@ -55,10 +58,12 @@ protected void append(ILoggingEvent iLoggingEvent) {
     @BeforeAll
     static void setUp() {
         LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
```
zUniQueX: We don't need the `LoggerContext` here. As you already did in the `messageParametersExample()` method, we can get Logback's `Logger` object by casting the logger obtained from the `LoggerFactory.getLogger(...)` call.

dropwizard-validation/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,10 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+        </encoder>
+    </appender>
+    <root level="off">
+        <appender-ref ref="STDOUT"/>
```
zUniQueX, line 8: I wonder, if we need to define an `<appender-ref />` here. If we turn off logging, we don't need an appender IMO. The tests work, if I remove the appender so maybe we don't need it.

zUniQueX (approved):
Thanks, LGTM!

I've just marked a few minor things you _could_ change, but don't have to. Feel free to change or ignore them.

dropwizard-validation/src/test/resources/logback-test.xml
rhowe: Hm, good question. This is the same `logback-test.xml` we have everywhere else. I'll raise a separate PR removing that from all of them and we can discuss the merits there.

dropwizard-validation/src/test/java/io/dropwizard/validation/SelfValidationTest.java
rhowe: No strong opinions either - as you deduced I was picking up on the logback default, but maybe we can shave some microseconds off the test run by upping the log level to INFO. I'll do that.

dropwizard-validation/src/test/java/io/dropwizard/validation/SelfValidationTest.java
rhowe: Right you are, it's gone.

zUniQueX:
Thanks for the changes :+1:

PR #5725, opened 2022-08-16
Disable check for static keyword on SLF4J loggers

zUniQueX:
As described in the [README](https://github.com/KengoTODA/findbugs-slf4j#slf4j_logger_should_be_non_static) of the `findbugs-slf4j` project and the linked SLF4J FAQ, there is no preference, if a logger should be declared `static` or not.

Since most of our loggers are declared static, we can disable this check to have a cleaner compile output.


joschi:
> Since most of our loggers are declared static, we can disable this check to have a cleaner compile output.

Would it make sense to explicitly disable the check with `@SuppressWarnings` on the respective statements instead of disabling this check completely?

zUniQueX:
> Would it make sense to explicitly disable the check with `@SupressWarnings` on the respective statements instead of disabling this check completely?

I don't see a benefit from that. We would then have to do so for all future usages of loggers to avoid the warnings. As the SLF4J project doesn't claim, if static or instance loggers are the 'better' alternative, I think we can disable the check. Because our README states we provide opinionated code, I'd say the choice to use static loggers is our opinion on that topic :wink:

joschi:
> We would then have to do so for all future usages of loggers to avoid the warnings.

Only if they are non-static. 😉

zUniQueX:
> Only if they are non-static.

The check favors instance loggers, so if they are static, right?

As we [use a static logger in the example](https://github.com/dropwizard/dropwizard/blob/release/2.1.x/dropwizard-example/src/main/java/com/example/helloworld/resources/HelloWorldResource.java#L25) I think most of our contributors would also follow this pattern and use static loggers.

But I may even get you wrong. This PR isn't that critical, so if we are unsure about it, feel free to close it.

zUniQueX:
I just wondered, how many loggers would be affected, if we would use `@SuppressWarnings`. So I wrote a simple lookup script using classgraph and added it to the startup of `dropwizard-example`, as there are used many of our internal classes. This won't be correct as a reference, but is okay for a quick overview:

```java
long nonStaticLoggers = 0;
long staticLoggers = 0;
try (ScanResult result = new ClassGraph().enableAllInfo().acceptPackages("io.dropwizard").scan()) {
    ClassInfoList classInfoList = result.getAllClasses();
    for (ClassInfo classInfo : classInfoList) {
        FieldInfoList fieldInfoList = classInfo.getFieldInfo();
        for (FieldInfo fieldInfo : fieldInfoList) {
            if (fieldInfo.getTypeDescriptor() instanceof ClassRefTypeSignature) {
                String fieldClassName = ((ClassRefTypeSignature) fieldInfo.getTypeDescriptor()).getBaseClassName();
                if (fieldClassName.equals("org.slf4j.Logger")) {
                    if (fieldInfo.isStatic()) {
                        staticLoggers++;
                    } else {
                        nonStaticLoggers++;
                    }
                }
            }
        }
    }
}
System.out.println();
System.out.printf("Non-static loggers: %d%n", nonStaticLoggers);
System.out.printf("Static loggers: %d%n", staticLoggers);
System.out.println();
```

The output shows the following result:

```
Non-static loggers: 11
Static loggers: 68
```

So we're using more static loggers than instance ones. I don't know, if we then would want to suppress warnings on all static loggers to make errorprone happy.

pstackle (approved):
If I'm reading the slf4j docs correctly, it sounds like there is a slight benefit by means of less CPU overhead for using a static logger and the benefit of using non-static loggers mainly comes into play when deploying multiple applications within an application server. 

My vote would be to disable the check.

joschi (approved):
I've been convinced. Let's go! 😄 

@zUniQueX Thanks for the PR!

PR #5748, opened 2022-08-23
Use Byte Buddy instead of Javassist  in UnitOfWorkAwareProxyFactory

sjthomason:
As Hibernate uses Byte Buddy internally to create proxies, this allows removal of the dependency on Javassist.

### DIFF
--- a/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
+++ b/dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
@@ -1,14 +1,21 @@
 package io.dropwizard.hibernate;
 
-import javassist.util.proxy.Proxy;
-import javassist.util.proxy.ProxyFactory;
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.dynamic.loading.ClassLoadingStrategy;
+import net.bytebuddy.implementation.MethodDelegation;
+import net.bytebuddy.implementation.bind.annotation.Origin;
+import net.bytebuddy.implementation.bind.annotation.RuntimeType;
+import net.bytebuddy.implementation.bind.annotation.SuperCall;
+import net.bytebuddy.matcher.ElementMatchers;
 import org.hibernate.SessionFactory;
 
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.Callable;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


zUniQueX:
Hi @sjthomason, thanks for the contribution!

I wonder, if we can extend this change to remove Javassist completely from the project. Except in `dropwizard-hibernate`, Javassist seems to be only used in `dropwizard-jersey`. Could you take a look, if we can use ByteBuddy there too?

The second question I have in mind is, if we'd want to have a compile time dependency on ByteBuddy. In Hibernate 6.x it got a runtime dependency. As the proxy seems to be fairly simple, maybe we could even use a `java.reflect.Proxy` to remove the 'proxy dependency' completely?

sjthomason:
Thanks for the review!

I took a look at `dropwizard-jersey` and after replacing Javassist with ByteBuddy it is used by `jersey-hk2` so that doesn't help much and actually adds a dependency.  While I do have working code that passes the unit tests, I suggest not touching `dropwizard-jersey` at this time.

Regarding the compile time vs runtime scope, I looked at the hibernate source and they have some byte code manipulation code in tree so perhaps that's why they are using the runtime scope in 6.x.

AFAIK, using `java.reflect.Proxy` would require a breaking change as the wrapped class would need to implement an interface that could be implemented by `java.reflect.Proxy`.

zUniQueX:
Thanks for the detailed answer.

Actually, the breaking change wouldn't be a problem, since Hibernate 6.x could firstly be integrated into Dropwizard 4.x. I just wanted to ask, if we'd want to bring the change of this PR, if we'd change this again with the Hibernate 6.x upgrade.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
```diff
@@ -69,39 +78,19 @@ public <T> T create(Class<T> clazz, Class<?> constructorParamType, Object constr
      * @param <T>                   the type of the class
      * @return a new proxy
      */
-    @SuppressWarnings("unchecked")
     public <T> T create(Class<T> clazz, Class<?>[] constructorParamTypes, Object[] constructorArguments) {
-        final ProxyFactory factory = new ProxyFactory();
-        factory.setSuperclass(clazz);
+        final DynamicType.Builder<T> builder = new ByteBuddy().subclass(clazz);
```
zUniQueX: Do we need to save the builder into a variable here?

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
```diff
@@ -122,4 +111,39 @@ public UnitOfWorkAspect newAspect() {
...
+    class MethodInterceptor {
+
+        private final Map<String, SessionFactory> sessionFactories;
+
+        public MethodInterceptor(Map<String, SessionFactory> sessionFactories) {
+          this.sessionFactories = sessionFactories;
+        }
+
+        @RuntimeType
+        Object invoke(@This Object self, @Origin Method overridden, @SuperMethod Method proceed,
```
zUniQueX: Byte-buddy also defines the `@SuperCall` annotation to use a parameter of type `@SuperCall Callable<?>`. As we use the `@This` and `@AllArguments` annotated parameters only for the invocation of the super method, maybe we could simplify this here.

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
sjthomason: No I'll change it to:
```java
        final Class<? extends T> proxied = new ByteBuddy().subclass(clazz).method(ElementMatchers.any())
            .intercept(MethodDelegation.to(new MethodInterceptor(sessionFactories)))
            .make()
            .load(clazz.getClassLoader(), ClassLoadingStrategy.Default.INJECTION)
            .getLoaded();
```

dropwizard-hibernate/src/main/java/io/dropwizard/hibernate/UnitOfWorkAwareProxyFactory.java
sjthomason: I was trying to emulate the existing code but that's definitely cleaner.  I've revised the PR.

sjthomason:
After looking at this for a while, an optimization can be made if the original method is unannotated or only has a single `@UnitOfWork` annotation.  What are your thoughts on something like this:
```java
if (unitsOfWork.length == 0) {
    return proxy.call();
}
final Map<UnitOfWork, UnitOfWorkAspect> unitOfWorkAspectMap;
if (unitsOfWork.length == 1) {
    unitOfWorkAspectMap = Collections.singletonMap(unitsOfWork[0], newAspect(sessionFactories));
} else {
    unitOfWorkAspectMap = new HashMap<>();
    Arrays
        .stream(unitsOfWork)
        .collect(Collectors.toMap(UnitOfWork::value, Function.identity(), (first, second) -> second))
        .values()
        .forEach(unitOfWork -> unitOfWorkAspectMap.put(unitOfWork, newAspect(sessionFactories)));
}
```

zUniQueX:
> What are your thoughts on something like this

Makes totally sense to me. Thanks for updating the PR! This will probably give us a performance benefit for methods, where no `@UnitOfWork` annotation is present.

zUniQueX:
@dropwizard/committers Any objections to bringing this change to 2.1.2?

joschi (approved):
@sjthomason LGTM, thanks!

@zUniQueX Thanks for reviewing!

PR #5749, opened 2022-08-24
JavaDoc enhancements

zUniQueX:
This PR is meant as a starting point to fix JavaDoc errors during the run of the compiler and during javadoc generation via the `maven-javadoc-plugin`.

@dropwizard/committers I think up-to-date JavaDoc is always a good thing, although this might take some time to accomplish. I've now corrected the errors in the `dropwizard-util` module. Do you think it's worth doing so for all other modules?

I've experienced, that the invocation of the `javadoc:jar` target on a JDK 17 causes errors, even with our `<doclint>none</doclint>` settings. The errors vanish when passing `-Ddoclint=none` on the command line. Maybe I'm missing something here, but we should be able to build the project on a JDK 17.

### DIFF
--- a/dropwizard-configuration/src/main/java/io/dropwizard/configuration/BaseConfigurationFactory.java
+++ b/dropwizard-configuration/src/main/java/io/dropwizard/configuration/BaseConfigurationFactory.java
@@ -43,6 +43,9 @@ public abstract class BaseConfigurationFactory<T> implements ConfigurationFactor
 
     private final Class<T> klass;
     private final String propertyPrefix;
+    /**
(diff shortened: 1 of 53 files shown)


zUniQueX:
When getting to `dropwizard-logging`, I noticed that JavaDoc for our factory objects isn't really needed/wanted. I could add some general comments, but getters and setters are self-explanatory there. Any opinions on that point?

For the modules I've processed so far a few JavaDoc comments could actually help users. See what you think about it.

joschi (approved):

PR #5770, opened 2022-08-27
Remove redundant outputPatternAsHeader logging setting from tests

rhowe:
This defaults to false - we don't need to set it to false explicitly.

See https://logback.qos.ch/manual/encoders.html#outputPatternAsHeader

If you'd rather merge into 2.0.x, close this and look at https://github.com/dropwizard/dropwizard/pull/5769 instead


joschi (approved):
@rhowe Thanks a lot for the cleanup. Solid work as always. ❤️

rhowe:
> @rhowe Thanks a lot for the cleanup. Solid work as always. heart

This one was @zUniQueX really, I just pressed the buttons

Issue #5479, opened 2022-06-29
Integration tests overwrite global log config


ericdevries:
I have a bunch of integration tests and unit tests, all setup according to the way the documentation specifies. However, when executing `mvn clean package` it runs the tests, and as soon as it has hit an integration test, it seems to completely overwrite the config that is supposed to be in logback.xml.

The sample output is this

```
[INFO] Running nl.knaw.dans.sword2.core.service.ZipServiceImplTest
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.909 s - in nl.knaw.dans.sword2.core.service.ZipServiceImplTest
[INFO] Running nl.knaw.dans.sword2.core.service.BagExtractorImplTest
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.197 s - in nl.knaw.dans.sword2.core.service.BagExtractorImplTest
[INFO] Running nl.knaw.dans.sword2.core.service.BagItManagerImplTest
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.017 s - in nl.knaw.dans.sword2.core.service.BagItManagerImplTest
[INFO] Running nl.knaw.dans.sword2.core.service.FilesystemSpaceVerifierImplTest
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.013 s - in nl.knaw.dans.sword2.core.service.FilesystemSpaceVerifierImplTest
[INFO] Running nl.knaw.dans.sword2.core.service.DepositHandlerImplTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.291 s - in nl.knaw.dans.sword2.core.service.DepositHandlerImplTest
[INFO] Running nl.knaw.dans.sword2.auth.SwordAuthenticatorTest
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.34 s - in nl.knaw.dans.sword2.auth.SwordAuthenticatorTest
[INFO] Running nl.knaw.dans.sword2.health.DiskSpaceHealthCheckTest
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.021 s - in nl.knaw.dans.sword2.health.DiskSpaceHealthCheckTest
[...]

PR #5771, opened 2022-08-27
Minimise test logging configuration

rhowe:
If we're turning logging off, we don't really need to configure an appender.

Is this what you were thinking, @zUniQueX?

### DIFF
--- a/dropwizard-jetty/src/test/resources/logback-test.xml
+++ b/dropwizard-jetty/src/test/resources/logback-test.xml
@@ -1,13 +1,3 @@
 <configuration>
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
-        </encoder>
-    </appender>
-
-    <logger level="info" name="io.dropwizard.jetty.HttpsConnectorFactory"/>
-
-    <root level="off">
-        <appender-ref ref="STDOUT"/>
-    </root>
+    <root level="off"/>
 </configuration>
--- a/dropwizard-testing/src/test/resources/logback-test.xml
+++ b/dropwizard-testing/src/test/resources/logback-test.xml
@@ -1,10 +1,3 @@
 <configuration>
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
-        </encoder>
-    </appender>
-    <root level="info">
-        <appender-ref ref="STDOUT"/>
-    </root>
+    <root level="off"/>
 </configuration>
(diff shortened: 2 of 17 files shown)


zUniQueX:
That's what I thought of, thank you very much.

dropwizard-jetty/src/test/resources/logback-test.xml
```diff
@@ -1,13 +1,3 @@
 <configuration>
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
-        </encoder>
-    </appender>
-
-    <logger level="info" name="io.dropwizard.jetty.HttpsConnectorFactory"/>
```
zUniQueX, line 8: I think this would cause changed behaviour. The appender would be inherited from the root logger, but this logger would actually log, wouldn't it?

dropwizard-testing/src/test/resources/logback-test.xml
```diff
@@ -1,10 +1,3 @@
 <configuration>
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
-        </encoder>
-    </appender>
-    <root level="info">
-        <appender-ref ref="STDOUT"/>
-    </root>
+    <root level="off"/>
```
zUniQueX, line 2: Logging would be disabled here, although it was enabled previously.

zUniQueX (commented):
I think we have some messages in the logs, which we curently won't need/want to have in the build output.

I'll investigate, if the behaviour changes would be desirable or if we should keep the current configurations.

dropwizard-jetty/src/test/resources/logback-test.xml
rhowe: It was added in aabb36660401e32b2a7aa60a3320f12ad59ece1b which added a test for building the HTTPS connector.

The only thing really of interest which gets logged when doing that are the SSL parameters, but we don't need those logs for the test - it checks what the HTTPS connector actually gets configured with.

dropwizard-testing/src/test/resources/logback-test.xml
rhowe: Introduced 7 years ago in 14bbdada40fb680095797d8ee822c56eb1fa2ed9 which enabled running tests in parallel. I suspect this was done to make it easier to see what was going on with the logging acquisition spinlock etc.

dropwizard-jetty/src/test/resources/logback-test.xml
zUniQueX: Agreed

dropwizard-testing/src/test/resources/logback-test.xml
zUniQueX: Yes, seems to be enabled for debugging purposes.

At the moment, the logging configuration for `dropwizard-testing` will probably get overridden fast as described in #5479. So it won't be that important if we have logging enabled or disabled, since the `BootstrapLogging.bootstrap()` method will enable the Dropwizard defaults at the first integration test.

zUniQueX (approved):

PR #5772, opened 2022-08-28
Suppress warnings for static loggers

zUniQueX:
Don't display a warning, if a `Logger` is declared `static`.

Raised as a followup for #5725. @joschi Did you mean doing this instead of disabling the check?


pstackle (commented):
I'm not a fan of this approach. This seems like it just makes the code messier for little value when we can just disable the check globally.

zUniQueX:
Thanks for the feedback.

That's exactly my opinion. We tend to use static loggers and are totally fine with it. If now a new logger is created, a warning appears all of a sudden although there is nothing wrong with it.

PR #5778, opened 2022-08-29
Address errorprone warnings in dropwizard-logging

rhowe:
###### Problem:
errorprone warns about javadoc and locking in `dropwizard-logging`

###### Solution:
Add javadoc summaries for public methods which have javadoc comments already.

Remove javadoc from `FileAppenderFactory#isTotalSizeCapValid()` - it's  a validation method so the `@since` tag is pretty superfluous.

Take the `CHANGE_LOGGER_CONTEXT_LOCK` immediately before the `try` in `DefaultLoggingFactory`.

###### Result:
errorprone warnings in `dropwizard-logging` are reduced to just one, relating to the static logger (there are separate PRs dealing with that).

### DIFF
--- a/dropwizard-logging/src/main/java/io/dropwizard/logging/FileAppenderFactory.java
+++ b/dropwizard-logging/src/main/java/io/dropwizard/logging/FileAppenderFactory.java
@@ -208,6 +208,8 @@ public class FileAppenderFactory<E extends DeferredProcessingAware> extends Abst
     }
 
     /**
+     * Returns the total size threshold at which archived log files will be
+     * removed. A zero value means there is limit.
      * @since 2.0
      */
     @JsonProperty
@@ -217,6 +219,8 @@ public class FileAppenderFactory<E extends DeferredProcessingAware> extends Abst
     }
 
     /**
+     * Sets the total size threshold at which archived log files will be
+     * removed. A zero value means there is limit.
      * @since 2.0
      */
     @JsonProperty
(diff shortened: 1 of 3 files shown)


dropwizard-logging/src/main/java/io/dropwizard/logging/FileAppenderFactory.java
```diff
@@ -243,9 +247,6 @@ public void setImmediateFlush(boolean immediateFlush) {
         this.immediateFlush = immediateFlush;
     }
 
-    /**
-     * @since 2.0
-     */
```
joschi: question: Why should this Javadoc comment be removed?

If nothing else, it documents since when this method is available.

rhowe: I thought it wasn't all that valuable to have on a validator method which people weren't likely to be calling directly or depending on in their software.

zUniQueX: I think, this doesn't necessarily have to be there. But since it is already, this provides additional information to our users. The method is `public`, so the JavaDoc would display that the method is available since version 2.0.

joschi: Yes, it would be great if we could restore that Javadoc comment and then IMHO we're good to go. 🐎

rhowe: Restored, along with a summary line to keep errorprone quiet.

zUniQueX: Thanks :+1:

zUniQueX (approved):

PR #5781, opened 2022-08-31
Update logback to version 1.3.0

zUniQueX:
Updates Logback to version 1.3.0 and therefore SLF4J to version 2.0.0.

Additionally drops support for JMX because Logback 1.3.0 dropped support for it.


joschi (approved):

zUniQueX:
This PR contains changes similar to #5777. The commit for removing the request log workaround is not ported back, since Logback 1.3.0 is compiled with Jetty 9.4.48.v20220622 and the method return type of `Response.getHttpFields()` has changed since this version and 10.0.11.

PR #5825, opened 2022-09-08
Only configure JMX, if the JMXConfigurator is present

zUniQueX:
This PR allows using SLF4J 2.0.0 and Logback 1.3.0 with Dropwizard 2.1.x.


pstackle (approved):
LGTM

PR #5855, opened 2022-09-12
Drop support for Java 8 in dropwizard-jersey

rhowe:
Dropwizard 3.x targets Java 11, so we can remove support for Java 8 here.

This removes the only usage of dropwizard-util's `JavaVersion` so that can go too.


zUniQueX:
Thanks very much!

In #4587 you proposed a further change to that implementation. Anything we should additionally change here?

rhowe:
That was just a different attempt to remove `JavaVersion` in a way that should work on Java 8 and 11, but I didn't really know what I was doing and I think someone reported issues so it got reverted.
This ought to be safe enough, since it just drops the Java 8 codepath. If what's left is broken, it was hopefully broken before I touched it.

zUniQueX:
IIRC the Java 8 tests broke in the other PR, so I wanted to ask, if there were any other concerns like performance improvements or something.

As you said, this change should be safe, so never change running code :wink:

zUniQueX (approved):

rhowe:
I see what you mean. I think if we're dropping Java 8 we can probably do this with the JCL using the method I proposed there. I'll raise a PR and we can see what it looks like.

PR #5871, opened 2022-09-14
Upgrade to hibernate-core 6.1.3

### LINKED ISSUE #5168: Upgrade to Hibernate ORM 6.0.x
- https://hibernate.org/orm/releases/6.0/
- https://in.relation.to/2022/03/31/orm-60-final/

zUniQueX:
Refs #5168 

After looking at the Hibernate 6.0.x and 6.1.x migration guides and the Jackson hibernate module, I think we can upgrade Hibernate to 6.1.x while using the Jackson module for the 5.x series using the `jakarta` namespace.

This introduces the following changes:
 - drop support for `jadira.usertype`. The latest version (7.0.0.CR1) is from 2018 and the library hasn't provided Hibernate 6 support yet
 - drop Hibernate `Criteria`: The methods to provide `Criteria` instances were deprecated and are removed in Hibernate 6
 - removal of `Serializable` as id restriction: Hibernate 6 supports any `Object` as id, so the restriction for an id object to implement `Serializable` got removed. Therefore the method signatures have to be changed to support all `Object`s
 - `uniqueElement`: The `uniqueElement` method was moved to `AbstractSelectionQuery` and the access modifier was changed to `protected`. Therefore add a small utility method which mirrors the behaviour
 - removal of `AvailableSettings.USE_NEW_ID_GENERATOR_MAPPINGS`: The `AvailableSettings.USE_NEW_ID_GENERATOR_MAPPINGS` property was removed. As the default value was `true` before, simply remove setting the property

### DIFF
--- a/dropwizard-dependencies/pom.xml
+++ b/dropwizard-dependencies/pom.xml
@@ -31,7 +31,7 @@
         <error_prone.version>2.15.0</error_prone.version>
         <freemarker.version>2.3.31</freemarker.version>
         <guava.version>31.1-jre</guava.version>
-        <hibernate-core.version>5.6.11.Final</hibernate-core.version>
+        <hibernate-core.version>6.1.3.Final</hibernate-core.version>
         <hibernate-validator.version>7.0.5.Final</hibernate-validator.version>
         <hk2.version>3.0.3</hk2.version>
         <httpclient.version>5.1.3</httpclient.version>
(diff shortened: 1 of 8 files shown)


joschi (approved):
LGTM, thank you!

The Jadira Usertypes library should probably have been removed from Dropwizard much earlier. Thanks for finally doing it!

joschi:
@zUniQueX Given that [Hibernate ORM 6.1.x](https://hibernate.org/orm/releases/6.1/) supports Jakarta Persistence 3.0, should we target this PR for Dropwizard 3.0.x instead for 4.0.x?

zUniQueX:
Jakarta Persistence API 3.0 is included in Jakarta EE 9 and therefore already includes the "big bang change". Am I missing something at compatibility to Java EE 8?

joschi:
@zUniQueX You're right. I've mixed up JPA 2.2 and Jakarta Persistence 3.0, both of which are supported by Hibernate ORM 5.6.x.

zUniQueX:
> The Jadira Usertypes library should probably have been removed from Dropwizard much earlier. Thanks for finally doing it!

@joschi Should we drop the Jadira library for 3.x too to simplify the migration path?

PR #6134, opened 2022-10-27
Fixed Possible Json Ordering Permutations Problem in Tests

BoyuLi4:
# Summary

Following the problem of flakiness that occurred in the test file
`dropwizard-health io.dropwizard.health.response.JsonHealthResponseProviderTest`
with below three tests
```
shouldHandleMultipleHealthStateViewsCorrectly
shouldHandleSingleHealthStateViewCorrectly
```
The test flakiness is due to comparisons between Json Strings and outputs from
`response.getMessage()` which is an Object in JSON format that compare with a JSON file locates in the `resources `file.

However, JsonObject does not guarantee entry orders, its object is an unordered set of name/value pairs, and internal permutations may occur in the output as JSON String.

## Brief changelog

Since JSON library `com.fasterxml.jackson.databind.ObjectMapper` is used in the current project, I use `mapper.readTree(JSON String)` when comparing them, thus the equals() method ignores the order and treats them as equal.This will make the test more stable.

## Verifying this change

This test avoids nested or different orders of JSON strings that cause flaky errors.
Since I didn't touch any Source code, it should be no impact on the project.
And it passed the local test with.`mvn clean install -P contrib-check`

### DIFF
--- a/dropwizard-health/src/test/java/io/dropwizard/health/response/JsonHealthResponseProviderTest.java
+++ b/dropwizard-health/src/test/java/io/dropwizard/health/response/JsonHealthResponseProviderTest.java
@@ -25,6 +25,7 @@ import java.util.Optional;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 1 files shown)


zUniQueX:
Hi @BoyuLi4. Thanks for your contribution!

The changes look good. Just a stylistic suggestion: we mostly use AssertJ assertions and not the assertions from JUnit. Would you mind converting your assertions to the AssertJ equivalents?

BoyuLi4:
Hi,
I deleted the JUnit test and changed it to AssertJ. Hope it looks better now.

joschi (approved):

PR #6177, opened 2022-11-10
Update dependency com.fasterxml.jackson:jackson-bom to v2.14.0 (release/2.1.x)

tingstad:
Jackson 2.13 -> 2.14.0 to avoid CVE-2022-40153

The currently used jackson-bom [references](https://github.com/FasterXML/jackson-bom/blob/jackson-bom-2.13.4.20221013/pom.xml#L63) jackson-dataformat-xml 2.13.4 which [depends](https://github.com/FasterXML/jackson-dataformat-xml/blob/jackson-dataformat-xml-2.13.4/pom.xml#L90) on `woodstox-core:6.3.1`.

woodstox-core >= 6.0.0, < 6.4.0 is affected by a high severity vulnerability alert,  CVE-2022-40153.

jackson-dataformat 2.14.0, [via](https://github.com/FasterXML/jackson-bom/blob/jackson-bom-2.14.0/pom.xml#L63) jackson-bom 2.14.0, has updated the [dependency](https://github.com/FasterXML/jackson-dataformat-xml/blob/jackson-dataformat-xml-2.14.0/pom.xml#L90) to 6.4.0 which does not have the vulnerability.

### DIFF
--- a/dropwizard-dependencies/pom.xml
+++ b/dropwizard-dependencies/pom.xml
@@ -36,7 +36,7 @@
         <hk2.version>2.6.1</hk2.version>
         <httpclient.version>4.5.13</httpclient.version>
         <httpcore.version>4.4.15</httpcore.version>
-        <jackson.version>2.13.4.20221013</jackson.version>
+        <jackson.version>2.14.0</jackson.version>
         <jakarta.annotation-api.version>1.3.5</jakarta.annotation-api.version>
         <jakarta.el.version>3.0.4</jakarta.el.version>
         <jakarta.inject-api.version>1.0.5</jakarta.inject-api.version>


joschi:
@tingstad Thanks for your contribution.

Dropwizard 2.1.x will not update to Jackson 2.14.x for compatibility reasons and the 3.0.x (#6178) and 4.0.x (#6179) branches have already been updated.

Dropwizard is neither using `jackson-dataformat-xml` nor Woodstox, so I would classify the risk as minimal.

This being said, you should be able to override the version of Jackson brought in by Dropwizard by importing the Jackson BOM in the desired version into your project.

tingstad:
Thank you for the info. Are the compatibility reasons something I should be worried about when overriding the Jackson BOM in combination with Dropwizard 2.1 in my project?

joschi:
> Are the compatibility reasons something I should be worried about when overriding the Jackson BOM in combination with Dropwizard 2.1 in my project?

None that I know of. The tests of Dropwizard 2.1.x are also green with Jackson 2.14.0 (see #6171), so I wouldn't expect any very obvious breaking changes.

mattnelson:
> Dropwizard 2.1.x will not update to Jackson 2.14.x for compatibility reasons and the 3.0.x (#6178) and 4.0.x (#6179) branches have already been updated.

Can you provide more details on the compatibility issues? Jackson is a frequent source of CVEs. If 2.1.x is no longer compatible with future jackson releases that creates a situation similar to the 2.1.x/slf4j-2.x discussion where your options would be vulnerable or upgrade to a beta version.
https://github.com/dropwizard/dropwizard/discussions/5821#discussioncomment-3588434

joschi:
> Can you provide more details on the compatibility issues?

As mentioned before, the tests of Dropwizard 2.1.x are all green with [Jackson 2.14.0](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.14), so at least from that perspective it's compatible.

But this doesn't mean that none of the users of Dropwizard 2.1.x will run into any issues, so we follow a conservative update strategy in this case. Again, if you want to use Jackson 2.14.0 in your projects, that's perfectly possible by importing the Jackson 2.14.0 BOM into it.

There are currently no open CVEs against Jackson 2.13.4, so there's also no immediate pressure to update or carve out a Dropwizard 2.2.x release (which I'd like to avoid in favor of Dropwizard 3.0.x).

> where your options would be vulnerable or upgrade to a beta version.

To be honest, Dropwizard 3.0.0-beta.X and 4.0.0-beta.X are currently just as well maintained as Dropwizard 2.1.x _and_ any feedback on those versions will increase the chance of them being released. 😉

mattnelson:
> Again, if you want to use Jackson 2.14.0 in your projects, that's perfectly possible by importing the Jackson 2.14.0 BOM into it.

This is exactly what we did and did not see any issues with dropwizard itself and I can provide one more data point to that effect.
We did discover an issue with 1 service in https://github.com/tariqkb/dropwizard-patch/issues/15 which caused us to roll that one back to 2.13.x.

> To be honest, Dropwizard 3.0.0-beta.X and 4.0.0-beta.X are currently just as well maintained as Dropwizard 2.1.x _and_ any feedback on those versions will increase the chance of them being released. 😉

I have a branch taking 3.x for a spin locally. A few import updates and some changes due to the version bump in caffeine. So far so good.

PR #6267, opened 2022-11-29
Cache proxy classes

### LINKED ISSUE #6263: UnitOfWorkAwareFactory can lead to java.lang.OutOfMemoryError: Metaspace
A new class is added to the classpath each time the UnitOfWorkAwareFactory create method is called. If multiple proxies of the same class are needed, there will be as many classes added to the classpath as the number of proxies created. Over time, the jvm's metaspace can fill up and in our case can lead to `java.lang.OutOfMemoryError: Metaspace`. This is a regression since https://github.com/dropwizard/dropwizard/pull/5748.

zUniQueX:
Closes #6263 

This PR introduces caching for proxy classes. Javassist automatically did this while ByteBuddy doesn't use a cache internally.

This change will provide small benefit in comparison to the maintenance IMHO. See what you think about it.

### DIFF
--- a/dropwizard-hibernate/pom.xml
+++ b/dropwizard-hibernate/pom.xml
@@ -68,6 +68,14 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.checkerframework</groupId>
+            <artifactId>checker-qual</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>jsr305</artifactId>
         </dependency>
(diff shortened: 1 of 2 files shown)


dropwizard-hibernate/pom.xml
```diff
@@ -67,6 +67,14 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
```
DanyBv, line 72: Hi @zUniQueX, if you don't want to add an additional dependency, there is a [cache](https://javadoc.io/static/net.bytebuddy/byte-buddy/1.10.2/net/bytebuddy/TypeCache.html) provided by ByteBuddy.

DanyBv (approved):

dropwizard-hibernate/pom.xml
zUniQueX: I've been thinking about this too but the classes are always loaded by the classloader of the class being proxied, so we can use a normal cache. Caffeine is already pulled in by `dropwizard-jackson`, so this dependency is already present.

With this solution we additionally let users modify the cache (e.g. its size), which isn't possible with the cache of ByteBuddy.

joschi (approved):
@zUniQueX LGTM, thanks!

PR #6621, opened 2023-02-01
Switch back to subclass mock maker in `dropwizard-servlets`

zUniQueX:
###### Problem:
The update to Mockito 5.x brought the new inlining mockmaker as default, which currently breaks our tests in the `dropwizard-servlets` module.

###### Solution:
Use the older subclassing mockmaker, which was the default in Mockito up to 5.x.


joschi (approved):
@zUniQueX Thanks! 😄

PR #6642, opened 2023-02-06
Timestamp field comes first in generated JSON log output

darcy-harrison:
###### Problem:
json-logging outputs JSON objects with undefined field order, but some log processing services (ex. Splunk) require the timestamp field to be within the first N characters of an object.

###### Solution:
The `MapBuilder` class now maintains insertion order (via LinkedHashMap) except that the "timestamp" field is always written first.

###### Result:
The output JSON from json-logging always has the timestamp field appear first within the object

### DIFF
--- a/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/MapBuilder.java
+++ b/dropwizard-json-logging/src/main/java/io/dropwizard/logging/json/layout/MapBuilder.java
@@ -1,7 +1,7 @@
 package io.dropwizard.logging.json.layout;
 
 import javax.annotation.Nullable;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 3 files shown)


pstackle (approved):
Thanks @darcy-harrison! This definitely addresses an issue that people in my org have encountered in the past.

pstackle:
Thanks @darcy-harrison! Would you mind cherry picking this change to the `release/3.0.x` and `release/4.0.x` branches as well, so that the change can be included for those upcoming major versions?

darcy-harrison:
> Thanks @darcy-harrison! Would you mind cherry picking this change to the `release/3.0.x` and `release/4.0.x` branches as well, so that the change can be included for those upcoming major versions?

Done. See #6668, and #6669

PR #6668, opened 2023-02-09
3.0.x Timestamp field comes first in generated JSON log output

darcy-harrison:
Duplicate of #6642 except targeting release/3.0.x

###### Problem:
json-logging outputs JSON objects with undefined field order, but some log processing services (ex. Splunk) require the timestamp field to be within the first N characters of an object.

###### Solution:
The MapBuilder class now maintains insertion order (via LinkedHashMap) except that the "timestamp" field is always written first.

###### Result:
The output JSON from json-logging always has the timestamp field appear first within the object


joschi (approved):
@darcy-harrison Thank you for your contribution!

PR #6692, opened 2023-02-14
Add configurable response code meters

dennyac:
###### Problem:
Currently we only have top level 1xx/2xx/3xx/4xx/5xx meters at the jetty level. We have many use cases where we care about specific response codes like 401/503 and resort to manually instrumenting responses from every resource. This PR allows users to configure what response code meters to include.

###### Solution:
With the new release of metrics (since 4.2.16), we can specify what response code meters to include. Exposed the same via ```AbstractServerFactory```. Additionally allowed specifying the ```metricPrefix``` to include for jetty metrics

###### Result:
To maintain backward compatibility the default behavior remains the same with this change. Users will have to explicitly specify the DETAILED/ALL responseMeteredLevel if they want more response code meters.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -36,6 +36,8 @@ maxThreads                          1024                                        
 minThreads                          8                                                The minimum number of threads to keep alive in the thread pool. Note that each Jetty connector consumes threads from the pool. See :ref:`HTTP connector <man-configuration-http>` how the thread counts are calculated.
 maxQueuedRequests                   1024                                             The maximum number of requests to queue before blocking
                                                                                      the acceptors.
+responseMeteredLevel                COARSE                                           The response metered level to decide what response code meters are included
+metricPrefix                        (none)                                           The metricPrefix to use in the metric name for jetty metrics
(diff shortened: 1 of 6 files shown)


dennyac:
FYI - @joschi 

Hoping this approach is acceptable and can be included in the upcoming release 🙂

joschi (approved):
@dennyac Thanks for your contribution in Dropwizard Metrics and here!

dennyac:
@joschi  - Thanks for the review and accepting the changes. Should we be concerned about this - https://github.com/dropwizard/metrics/pull/3179 (Fix potential NPE in InstrumentedResourceMethodApplicationListener)

dennyac:
This PR specifically that bumps the metric version to 4.2.16 which contains the bug - https://github.com/dropwizard/dropwizard/pull/6693

Wanted your thoughts on releasing a new version of the metrics library which should include the fix.

joschi:
@dennyac We'll wait with a release of Dropwizard 2.1.5 until Dropwizard Metrics has been bumped to https://github.com/dropwizard/metrics/releases/tag/v4.2.17.

dennyac:
Sounds good. Thanks. Was just checking.

PR #6696, opened 2023-02-14
Add step for dependency tree submission

joschi:
- https://github.com/marketplace/actions/maven-dependency-tree-dependency-submission
- https://github.com/github/dependency-submission-toolkit

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -36,8 +36,10 @@ jobs:
         path: ~/.sonar/cache
         key: ${{ runner.os }}-sonar
         restore-keys: ${{ runner.os }}-sonar
+    - name: Submit Dependency Snapshot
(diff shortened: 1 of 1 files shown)


pstackle:
Unfortunately, it looks like there are a couple issues that might make this not work yet.

I see a [documented limitation](https://github.com/advanced-security/maven-dependency-submission-action#limitations):
> Currently the action is limited to single module Maven projects, with a future update that will add support for multi-module based projects.

and I see this [issue about Maven Wrapper support](https://github.com/advanced-security/maven-dependency-submission-action/issues/6) which we'd have to workaround, i think.

joschi:
@pstackle Yep, this was merely an experiment. 😅 

The missing support for Maven wrapper isn't that important. Dropwizard can be built with the preinstalled version of Maven in the GitHub runners.

The limitation on single module projects on the other hand is a deal breaker for now. 🤷

PR #6866, opened 2023-03-23
Use Mockito BOM in dropwizard-dependencies

JKomoroski:
###### Problem:
Keeping Mockito versions in sync with dropwizard can be a little tedious.

###### Solution:
Mockito publishes a bom, use that bom in the dropwizard-dependencies bom so importers of the dropwizard-dependencies bom can enjoy some small ergonomic benefit by no longer needing to re-declare versions for various mockito jars.

###### Result:
My dropwizard apps can now depend on several more jars without having to manage them:
- mockito-errorprone
- mockito-inline
- mockito-proxy
- mockito-subclass

### DIFF
--- a/dropwizard-dependencies/pom.xml
+++ b/dropwizard-dependencies/pom.xml
@@ -457,13 +457,10 @@
             </dependency>
             <dependency>
                 <groupId>org.mockito</groupId>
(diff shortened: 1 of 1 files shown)


joschi (approved):
@JKomoroski Thanks a lot for your contribution!

PR #6907, opened 2023-03-26
Set application state healthy if there are no unhealthy critical checks

### LINKED ISSUE #6794: When initialOverallState=false, the HTTP response code to readiness check is always 503
When I have `initialOverallState: false` set in my `health` configuration, the HTTP request to the health check endpoint `/health-check` always returns `503` response code, no matter what the actual statuses of my health checks are. For example, it will give a `503` response code with a `healthy:true` in the response body like this:
```
curl -LSi 'localhost:8080/health-check?type=ready&name=mycheck'
HTTP/1.1 503 Service Unavailable
Date: Sat, 11 Mar 2023 00:00:07 GMT
Cache-Control: no-store
Content-Type: application/json
Content-Length: 74

[{"name":"mycheck","healthy":true,"type":"READY","critical":false}]
```

I read the code and the problems seems to be related to the class `io.dropwizard.health.HealthCheckManager`:
- When `initialOverallState` is set to `false`, the state `isAppHealthy` [is initialized as `false`](https://github.com/dropwizard/dropwizard/blob/a6af1e088b53e0a9a7c670949a5dcd448cbd889f/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java#L30-L30).
- When the last failed health check becomes healthy, it's supposed to set both `isAppAlive` and `isAppHealthy` to true. But:
  - If it's an `alive` check, the method [handleCriticalHealthChange()](https://github.com/dropwizard/dropwizard/blob/a6af1e088b53e0a9a7c670949a5dcd448cbd889f/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java#L166-L166) is called. Because its type is `ALIVE`, only the state [`isAppAlive` is updated](https://github.com/dropwizard/dropwizard/blob/a6af1e088b53e0a9a7c670949a5dcd448cbd889f/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java#L166-L166).
  - If it's a `ready` check, the method [handleNonCriticalHealthChange()](https://github.com/dropwizard/dropwizard/blob/a6af1e088b53e0a9a7c670949a5dcd448cbd889f/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java#L202-L202) is called. It [only logs a message](https://github.com/dropwizard/dropwizard/blob/a6af1e088b53e0a9a7c670949a5dcd448cbd889f/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java#L204) and doesn't change either `isAppAlive` or `isAppHealthy` state.
So in any of these cases, the `isAppHealthy` is never set to `true`
- Because of this, the method [isHealthy()](https://github.com/dropwizard/dropwizard/blob/a6af1e088b53e0a9a7c670949a5dcd448cbd889f/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java#L227-L227) never returns `true`. So [the code in JsonHealthResponseProvider](https://github.com/dropwizard/dropwizard/blob/a6af1e088b53e0a9a7c670949a5dcd448cbd889f/dropwizard-health/src/main/java/io/dropwizard/health/response/JsonHealthResponseProvider.java#L63-L72) always gets `healthy=false` and returns `503`.

Dropwizard version `2.1.5`

joschi:
An application with an initial health state of unhealthy would never get healthy if there were no critical (or `ALIVE`) checks.

The changes in this MR will set the application health state to healthy if an uncritical health check changes to healthy and there are no critical health checks or all of them are healthy 

Fixes #6794

@natalie-zamani @pstackle It would be great if one of you could take a look at this and verify that the changes make sense.

### DIFF
--- a/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java
+++ b/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckManager.java
@@ -99,7 +99,7 @@ class HealthCheckManager implements HealthCheckRegistryListener, HealthStatusChe
 
         // handle initial state of 'false' to ensure counts line up
         if (!initialState && critical) {
-            handleCriticalHealthChange(name, type, false);
+            handleCriticalHealthChange(check, false);
         }
 
         scheduler.scheduleInitial(check);
@@ -121,9 +121,9 @@ class HealthCheckManager implements HealthCheckRegistryListener, HealthStatusChe
         }
 
         if (check.isCritical()) {
-            handleCriticalHealthChange(check.getName(), check.getType(), isNowHealthy);
+            handleCriticalHealthChange(check, isNowHealthy);
         } else {
-            handleNonCriticalHealthChange(check.getName(), check.getType(), isNowHealthy);
+            handleNonCriticalHealthChange(check, isNowHealthy);
         }
 
         scheduler.schedule(check, isNowHealthy);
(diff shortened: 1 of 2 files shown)


pstackle (approved):
Nice! Changes look good to me.

PR #6916, opened 2023-03-28
Fix mustache docs link in views.rst

Querela:
Fix mustache link in dropwizard-views

###### Problem:
Exiting link points to: https://mustache.github.com/mustache.5.html

###### Solution:
Link should point to: https://mustache.github.io/mustache.5.html

###### Result:
Working link.


joschi (approved):
@Querela Thanks for fixing the link! ❤️

PR #6973, opened 2023-04-06
Update Step By Step Example with JDK 11 Reference

JKomoroski:
###### Problem:
Release 3/4 both use JDK 11 Now

###### Solution:
Small Documentation fix

###### Result:
Documentation is accurate for the new releases.

I assume this Patch will need to go into both the 3.x and the 4.x branch, but it  wasn't clear to me how I should go about making pull requests for both, maybe some updates to the contributing doc is warranted?

### DIFF
--- a/docs/source/manual/example.rst
+++ b/docs/source/manual/example.rst
@@ -11,7 +11,7 @@ Dropwizard Example, Step by Step
 * Preconditions
 
   * Make sure you have Maven_ installed
-  * Make sure ``JAVA_HOME`` points at JDK 8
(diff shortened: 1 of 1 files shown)


joschi (approved):

joschi:
@JKomoroski Thank you very much for your contribution! ❤️

joschi:
> I assume this Patch will need to go into both the 3.x and the 4.x branch, but it wasn't clear to me how I should go about making pull requests for both, maybe some updates to the contributing doc is warranted?

I'll take care of merging it into the `release/4.0.x` branch too.

PR #7027, opened 2023-04-22
Issue#7026: ConnectionKeepAliveStrategy didn't use a keepAlive value from HttpClientConfiguration

### LINKED ISSUE #7026: ConnectionKeepAliveStrategy didn't use a keepAlive value from HttpClientConfiguration
Since default value of connectionKeepAlive value in `RequestConfig` is 3 minutes, the following logic will always have 3 minutes for `TimeValue duration` if server didn't return `keep-alive` in HTTP response.

```java
public TimeValue getKeepAliveDuration(HttpResponse response, HttpContext context) {
                    final TimeValue duration = super.getKeepAliveDuration(response, context);
                    return (duration.getDuration() == -1) ? TimeValue.ofMilliseconds(keepAlive) : duration;
                }
```
RequestConfig in Apache HTTP client.

https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java#L46

It may miss this logic `.setConnectionKeepAlive(keepAlive > 0 ? TimeValue. ofMilliseconds(keepAlive) : null)` in below code?

https://github.com/dropwizard/dropwizard/blob/release/4.0.x/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java#L370

```java
        final long keepAlive = configuration.getKeepAlive().toMilliseconds();

        final RequestConfig requestConfig
                = RequestConfig.custom().setCookieSpec(cookiePolicy)
                .setResponseTimeout(timeout, TimeUnit.MILLISECONDS)
                .setConnectionKeepAlive(keepAlive > 0 ? TimeValue. ofMilliseconds(keepAlive) : null)
                .setConnectTimeout(connectionTimeout, TimeUnit.MILLISECONDS)
                .setConnectionRequestTimeout(connectionRequestTimeout, TimeUnit.MILLISECONDS)
                .build();
```

This is causing a HTTP call error when Dropwizard HTTP client is trying to communicate with Dropwizard HTTP server since Dropwizard HTTP server's default idle timeout is 30 seconds. After 30 seconds, subsequent HTTP requests will fail from Drowizard HTTP client.

nsclass:
Resolve the issue #7026 
###### Problem:
ConnectionKeepAliveStrategy didn't use the keepAlive value from HttpClientConfiguration. It always returns 3 minutes which is a default value in RequestConfig class.

This is causing failure of subsequent HTTP requests after 30 seconds to Dropwizard HTTP server because the default idle timeout value in Dropwizard is 30 seconds.

###### Solution:
Set connectionKeepAlive value on creating a `RequestConfig` if `keepAlive` value is greater than 0.
```java
        final RequestConfig requestConfig
                = RequestConfig.custom().setCookieSpec(cookiePolicy)
                .setResponseTimeout(timeout, TimeUnit.MILLISECONDS)
                .setConnectTimeout(connectionTimeout, TimeUnit.MILLISECONDS)
                .setConnectionKeepAlive(keepAlive > 0 ? TimeValue.of(-1, TimeUnit.MILLISECONDS) : null)
                .setConnectionRequestTimeout(connectionRequestTimeout, TimeUnit.MILLISECONDS)
                .build();
```

###### Result:
Now, unit test `usesKeepAliveForPersistentConnections` is correctly verifying that `getKeepAliveDuration`  is the same value in `HttpClientConfiguration`


dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
```diff
@@ -371,6 +371,7 @@ protected ConfiguredCloseableHttpClient createClient(
                 = RequestConfig.custom().setCookieSpec(cookiePolicy)
                 .setResponseTimeout(timeout, TimeUnit.MILLISECONDS)
                 .setConnectTimeout(connectionTimeout, TimeUnit.MILLISECONDS)
+                .setConnectionKeepAlive(keepAlive > 0 ? TimeValue.of(-1, TimeUnit.MILLISECONDS) : null)
```
zUniQueX: If we return `null` here, this can lead to a NPE. I'd rather suggest to always set a fixed value of `-1`.

From the JavaDoc:
> A value of 0 or less may be returned to indicate that there is no suitable suggestion.

This wouldn't apply a 'magic' timeout value of three minutes for our users but rather let the behavior unspecified.

If a user actually specifies a value, this one would be applied through the custom `DefaultConnectionKeepAliveStrategy`.

dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
```diff
@@ -355,9 +355,9 @@ protected ConfiguredCloseableHttpClient createClient(
             final InstrumentedHttpClientConnectionManager manager,
             final String name) {
         final String cookiePolicy = configuration.isCookiesEnabled() ? StandardCookieSpec.RELAXED : StandardCookieSpec.IGNORE;
-        final Integer timeout = (int) configuration.getTimeout().toMilliseconds();
-        final Integer connectionTimeout = (int) configuration.getConnectionTimeout().toMilliseconds();
-        final Integer connectionRequestTimeout = (int) configuration.getConnectionRequestTimeout().toMilliseconds();
+        final int timeout = (int) configuration.getTimeout().toMilliseconds();
```
zUniQueX, line 358: Why should be use primitives here now?

zUniQueX (changes requested):

dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
nsclass: I think that it did unnecessary boxing.
```java
final Integer timeout = (int) configuration.getTimeout().toMilliseconds();
```

dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
nsclass: ok got it. updated.

dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java
zUniQueX: Sorry. I thought the builder methods would require boxed types. Then it won't made any difference, where the boxing is performed.

zUniQueX (approved):

PR #7033, opened 2023-04-23
Add CI job for linting YAML files

joschi:
https://github.com/frenck/action-yamllint


.circleci/config.yml
```diff
@@ -10,7 +11,8 @@ jobs:
       - checkout
       - restore_cache:
           key: dropwizard-{{ checksum "pom.xml" }}
-      - run: ./mvnw --no-transfer-progress -V -B -ff -Dsurefire.rerunFailingTestsCount=1 install
+      - run: ./mvnw --no-transfer-progress -V -B -ff \
```
nsclass, line 14: it looks like that this change is causing a failure in workflow by adding extra `\`? 
```bash
./mvnw --no-transfer-progress -V -B -ff \ -Dsurefire.rerunFailingTestsCount=1 install
```

Issue #1824, opened 2016-11-16
RFC7230 compliance cannot be disabled in Jetty


dren-dk:
I'm terminating TLS in nginx, but I still want access to the full client certificate in my application, until I upgraded dropwizard I did that by putting the client certificate in a header in nginx by using this config line:
proxy_set_header X-SSL-CERT $ssl_client_cert;

The result is that nginx inserts a multi-line header value containing the PEM encoded client certifcate, something that has been deprecated by RFC7230 and although the issue has been known for years by nginx, no alternative has surfaced, see: https://bugs.eclipse.org/bugs/show_bug.cgi?id=483795

I would like to be able to turn off RFC7230 compliance in Jetty until I can get a compliant nginx deployed.

Would it be possible to add a httpCompliance field of the type org.eclipse.jetty.http.HttpCompliance to HttpConnectorFactory that will cause the org.eclipse.jetty.server.HttpConnectionFactory to be created with the specificed org.eclipse.jetty.http.HttpCompliance parameter?

arteam:
Hi @dren-dk,

Yes, that's totally possible. Shouldn't be really hard to implement.

dren-dk:
I've just created a pull request with the change as it did seem to be very straight forward.

arteam:
Closed via #1825

PR #7042, opened 2023-04-25
expose uri compliance

denis-yuen:
###### Problem:
This is modeled on https://github.com/dropwizard/dropwizard/issues/1824 
Like with HttpCompliance, we ran into an issue where breaking changes with UriCompliance necessitated a workaround by creating a CustomHttpConnectorFactory.  This allowed us to add AMBIGUOUS_EMPTY_SEGMENT as noted at https://stackoverflow.com/questions/74395500/jetty-returns-400-when-uri-has-2-slashes-in-a-row-while-tomcat-doesnt-care

###### Solution:
This would have been accomplished in a more simple way by being able to configure uricompliance.  

###### Result:
New optional uriCompliance setting in the dropwizard configuration file.


dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
```diff
@@ -658,6 +683,43 @@ protected HttpComplianceDeserializer(@Nullable Class<?> vc) {
...
+            this(null);
+        }
+
+        protected UriComplianceSerializer(@Nullable Class<UriCompliance> t) {
+            super(t);
+        }
+
+        @Override
+        public void serialize(UriCompliance httpCompliance, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
+            if (httpCompliance == null) {
```
pstackle: copy/paste?
```suggestion
        public void serialize(UriCompliance uriCompliance, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
            if (uriCompliance == null) {
```

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
```diff
@@ -226,6 +227,16 @@
  *         </td>
  *     </tr>
  *     <tr>
+ *         <td>{@code uriCompliance}</td>
+ *         <td>RFC7230</td>
```
zUniQueX: The default value should be `DEFAULT` here.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
```diff
@@ -226,6 +227,16 @@
...
+ *         <td>
+ *             This sets the uri compliance level used by Jetty when parsing http, this can be useful when
+ *             attempting to avoid breaking changes with Jetty 10 and onward;
+ *
+ *             Possible values are set forth in the org.eclipse.jetty.http.UriCompliance enum.
+ *         </td>
+ *     </tr> *
```
zUniQueX: The additional star at the line end should be removed to have a cleaner javadoc.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
```diff
@@ -658,6 +683,43 @@ protected HttpComplianceDeserializer(@Nullable Class<?> vc) {
         }
     }
 
+    private static class UriComplianceSerializer extends StdSerializer<UriCompliance> {
```
zUniQueX, line 691: As this is the third pair of similar (de)serializers, maybe we should provide utility classes for such 'simple' enum (de)serializers in the `dropwizard-util` or `dropwizard-jackson` module. @dropwizard/committers Any opinions on that?

zUniQueX (changes requested):

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
joschi: They indeed look all quite similar.

I'm wondering if they could be replaced by [`com.fasterxml.jackson.databind.deser.std.EnumDeserializer`](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.15.2/com/fasterxml/jackson/databind/deser/std/EnumDeserializer.html#withResolved-java.lang.Boolean-java.lang.Boolean-java.lang.Boolean-).

zUniQueX: @denis-yuen Did you already take a look at this one?

denis-yuen: Sorry, have not had a chance.

denis-yuen: Taking a quick look, one issue with using the enum serializer/deserializer is that while `UriCompliance`, `CookieCompliance` and `HttpCompliance` all contain `Violation` enums and look a lot like enums, they aren't actually enums.

zUniQueX: @denis-yuen Thanks for the input. But I think the classes could use an other common (de)serializer implementation because of their similatity. If Jackson provides no such implementation, we could implement one ourselves into the `dropwizard-util` module with public access or into the `dropwizard-jetty` module with package-private access.

zUniQueX:
@denis-yuen I've tried to add a common Jackson (De)Serializer base class for this use case in #7687. Then I've ported your changes to use the new serializer classes. That looks cleaner IMHO and allows us to easier add more properties like these.

denis-yuen:
(closed in favour of https://github.com/dropwizard/dropwizard/pull/7687 )

PR #7100, opened 2023-05-10
Add Codespaces/devcontainer support

joschi:
- https://containers.dev/
- https://docs.github.com/en/codespaces/overview


zUniQueX (approved):
LGTM

zUniQueX:
@joschi I've re-investigated this and there seems to be a reactor ordering problem regarding `dropwizard-archetypes` because the SNAPSHOT versions of the other modules don't have to be installed yet when building the archetypes. Maybe we should default to a single-threaded build.

zUniQueX:
@joschi Looks good now after the archetypes have been externalized :+1:

PR #7218, opened 2023-06-11
Add Maven build cache

joschi:
https://maven.apache.org/extensions/maven-build-cache-extension/


zUniQueX:
@joschi When excluding the test db from `dropwizard-migrations` all modules are cached successfully. I've pushed the changes only to the branch for 4.0.x for testing. Locally it runs fine but Sonar seems to have problems with the binaries.

https://sonarsource.atlassian.net/browse/MSONAR-196

joschi:
@zUniQueX I'll close these PRs. I don't think using the Maven cache plugin makes sense here.

We can take another look at it when we get complaints about the build times. 😅

PR #7269, opened 2023-06-27
Added Valid annotation to DiscoverableLayoutFactory in  AbstractAppenderFactory (release/2.1.x)

chhatbarnikita:
###### Problem:
The layout field in `AbstractAppenderFactory` does not have `@Valid` annotation that is required to validate the fields. This change is necessary because when extending and creating custom logging layouts, the annotations based validations for the newly added fields does not work.

###### Solution:
Added `@Valid` for `DiscoverableLayoutFactory` layout in `AbstractAppenderFactory`.

###### Result:
See solution.


joschi (approved):
@chhatbarnikita Thank you very much for your contribution!

PR #7317, opened 2023-07-16
Generate H2 db file

zUniQueX:
If a major H2 version update has to be performed, the `test-db.mv.db` file may have an incompatible storage format for the new H2 version. This PR generates the H2 db file during build to prevent storage format incompatibilities.


joschi (approved):
LGTM, thanks! 😄

PR #7389, opened 2023-08-08
Clean test db

zUniQueX:
Appveyor doesn't clean and re-initialize the repo in a build. Re-running the H2 Script tool against the existing database fails with an error. This PR cleans the H2 database in the default build lifecycle to force re-generation in the test cycle.


joschi (approved):
LGTM, thanks!

PR #7532, opened 2023-09-18
Fix NullPointerException in DropwizardApacheConnector

### LINKED ISSUE #7319: Content-Encoding header is null but not expected
`DropwizardApacheConnector` has the private class BufferedJerseyRequestHttpEntity that expects a `Content-Encoding` header that is not suppose to be mandatory.

Mahoney:
###### Problem:
If `JerseyClientConfiguration.setGzipEnabledForRequests(false)` then there is no `CONTENT_ENCODING` header
on the `org.glassfish.jersey.client.ClientRequest`.

`org.glassfish.jersey.client.ClientRequest` returns null from `getRequestHeader` if there is no header, which causes
this code to fail.

###### Solution:
Check for null and handle it correctly

Fixes #7319

### DIFF
--- a/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
+++ b/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java
@@ -30,6 +30,7 @@ import java.util.ArrayList;
 import java.util.Optional;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.List;
 
 /**
  * Dropwizard Apache Connector.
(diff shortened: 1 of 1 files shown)


Mahoney:
Don't have time to write a failing test right now, might be able to come back to it in a few days.

zUniQueX (changes requested):
@Mahoney The change looks good. But the behavior is now inconsistent between the buffered and the chunked entity. Could you extract the new method to the enclosing class and call it in the buffered entity too?

Mahoney:
@zUniQueX updated with the requested change

zUniQueX (approved):
Thanks, LGTM!

Mahoney:
What further needs to happen to get this merged?

zUniQueX:
@Mahoney We currently investigate some infrastructure issues. This PR will probably be merged afterwards.

joschi:
@Mahoney Thanks a lot for your contribution! ❤️

PR #1893, opened 2017-01-18
[WIP] Implement Dropwizard Unix socket connector 

arteam:
This pull requests some basic work for supporting a Unix Socket connector in Dropwizard. The implementation seems to be rather simple, but I stumled on some problems: 
* I wasn't able to run a server in the default mode because of some admin thread pool management issues. 
* The server doesn't shut down gracefully. It waits 30 seconds and then killed forcefully after a SIGTERM signal is sent.

The original issue is #1889.

PR #7614, opened 2023-10-06
Unix domain socket connector

natan-abolafya:
###### Problem:
Since Java 16, Unix Domain Sockets are supported natively. And since Jetty 10, there is a ready made connector for it and it is not considered experimental anymore.
The typical use case is a reverse proxy connecting to unix socket instead of TCP which has various advantages (which I can't really provide a source right now :)).

###### Solution:
Created a new connector type "unix-socket".
I've reimplemented #1893 but it needed some adjustments considering how much dropwizard changed since then.
The test was also a little tricky (and not very pretty) considering there is no unix socket support in Apache HTTP client. Or at least I couldn't find any but please correct me otherwise.

###### Result:
It's pretty simple to listen on unix sockets now. That was something we wanted in our product for a long time.
Also this works on Windows since Windows now also supports unix domain sockets.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -492,8 +492,6 @@ Name                             Default                          Description
 path                             /tmp/dropwizard.sock             The path to the unix domain socket file.
 ================================ ================================ ======================================================================================
 
-.. _sslyze: https://github.com/nabla-c0d3/sslyze
-
 .. _man-configuration-http2:
 
 HTTP/2 over TLS
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -472,7 +472,7 @@ disableSniHostCheck              false                            Whether to dis
 
 .. _man-configuration-unix-socket:
 
-HTTPS
+Unix Domain Socket
 -----
 
 Extends the attributes that are available to the :ref:`HTTP connector <man-configuration-http>` but does not require port.
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -470,6 +470,30 @@ disableSniHostCheck              false                            Whether to dis
 
 .. _sslyze: https://github.com/nabla-c0d3/sslyze
 
+.. _man-configuration-unix-socket:
+
+HTTPS
+-----
+
+Extends the attributes that are available to the :ref:`HTTP connector <man-configuration-http>` but does not require port.
+
+.. code-block:: yaml
+
+    # Extending from the default server configuration
+    server:
+      applicationConnectors:
+        - type: unix-socket
+          ...
+          path: /path/to/file
+
+================================ ================================ ======================================================================================
+Name                             Default                          Description
+================================ ================================ ======================================================================================
+path                             /tmp/dropwizard.sock             The path to the unix domain socket file.
+================================ ================================ ======================================================================================
+
+.. _sslyze: https://github.com/nabla-c0d3/sslyze
+
 .. _man-configuration-http2:
 
 HTTP/2 over TLS
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -473,7 +473,7 @@ disableSniHostCheck              false                            Whether to dis
 .. _man-configuration-unix-socket:
 
 Unix Domain Socket
------
+------------------
 
 Extends the attributes that are available to the :ref:`HTTP connector <man-configuration-http>` but does not require port.
 
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
@@ -703,7 +703,7 @@ public class HttpConnectorFactory implements ConnectorFactory {
         return name(HttpConnectionFactory.class,  bindHost, Integer.toString(port), "connections");
     }
 
-    protected Connector buildConnector(Server server,
+    protected ServerConnector buildConnector(Server server,
                                              Scheduler scheduler,
                                              ByteBufferPool bufferPool,
                                              String name,
--- a/dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
+++ b/dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
@@ -703,7 +703,7 @@ public class HttpConnectorFactory implements ConnectorFactory {
         return name(HttpConnectionFactory.class,  bindHost, Integer.toString(port), "connections");
     }
 
-    protected ServerConnector buildConnector(Server server,
+    protected Connector buildConnector(Server server,
                                              Scheduler scheduler,
                                              ByteBufferPool bufferPool,
                                              String name,
--- a/dropwizard-unix-socket/pom.xml
+++ b/dropwizard-unix-socket/pom.xml
@@ -21,17 +21,45 @@
     <dependencies>
         <dependency>
             <groupId>io.dropwizard</groupId>
-            <artifactId>dropwizard-core</artifactId>
+            <artifactId>dropwizard-util</artifactId>
         </dependency>
-
         <dependency>
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-jetty</artifactId>
         </dependency>
-
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-unixdomain-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-server</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.eclipse.jetty.toolchain</groupId>
+                    <artifactId>jetty-jakarta-servlet-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-io</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-util</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.checkerframework</groupId>
+            <artifactId>checker-qual</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
         </dependency>
 
         <dependency>
@@ -52,6 +80,21 @@
         <dependency>
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-testing</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.dropwizard</groupId>
+            <artifactId>dropwizard-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.dropwizard</groupId>
+            <artifactId>dropwizard-jersey</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.ws.rs</groupId>
+            <artifactId>jakarta.ws.rs-api</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/dropwizard-unix-socket/pom.xml
+++ b/dropwizard-unix-socket/pom.xml
@@ -15,7 +15,7 @@
     <name>Dropwizard Unix Socket Support</name>
 
     <properties>
-        <module.name>io.dropwizard.dropwizard-unix-socket</module.name>
+        <module.name>io.dropwizard.dropwizard.unixsocket</module.name>
         <maven.compiler.release>16</maven.compiler.release>
     </properties>
 
--- a/dropwizard-unix-socket/pom.xml
+++ b/dropwizard-unix-socket/pom.xml
@@ -16,7 +16,6 @@
 
     <properties>
         <module.name>io.dropwizard.dropwizard.unixsocket</module.name>
-        <maven.compiler.release>16</maven.compiler.release>
     </properties>
 
     <dependencies>
--- a/dropwizard-unix-socket/pom.xml
+++ b/dropwizard-unix-socket/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>io.dropwizard</groupId>
         <artifactId>dropwizard-parent</artifactId>
-        <version>4.0.3-SNAPSHOT</version>
+        <version>5.0.0-SNAPSHOT</version>
         <relativePath>../dropwizard-parent</relativePath>
     </parent>
 
--- a/dropwizard-unix-socket/pom.xml
+++ b/dropwizard-unix-socket/pom.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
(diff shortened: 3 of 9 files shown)


dropwizard-unix-socket/src/main/java/io/dropwizard/unixsocket/UnixSocketConnectorFactory.java
```diff
@@ -0,0 +1,77 @@
...
+ *             The path to the unix domain socket file.
+ *         </td>
+ *     </tr>
+ * </table>
+ * <p/>
+ */
+@JsonTypeName("unix-socket")
+public class UnixSocketConnectorFactory extends HttpConnectorFactory {
+
+    private String path = "/tmp/dropwizard.sock";
```
natan-abolafya, line 42: I've changed this from `socketFilename` to `path` thinking it's implied when `type: unix-socket`. Like how it's just `port` instead of `tpcServerPort`.

natan-abolafya:
Should I add documentation before/after https://www.dropwizard.io/en/stable/manual/core.html#ssl? Or is Configuration Reference enough?

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
```diff
@@ -720,7 +720,7 @@ protected String httpConnections() {
         return name(HttpConnectionFactory.class,  bindHost, Integer.toString(port), "connections");
     }
 
-    protected ServerConnector buildConnector(Server server,
+    protected Connector buildConnector(Server server,
```
zUniQueX: This change modifies the public API and therefore is a breaking change.

pom.xml
```diff
@@ -47,6 +47,7 @@
         <module>dropwizard-request-logging</module>
         <module>dropwizard-json-logging</module>
         <module>dropwizard-health</module>
+        <module>dropwizard-unix-socket</module>
```
zUniQueX, line 50: Is there a good reason to introduce a separate module for this change instead of integrating the class into `dropwizard-jetty`?

docs/source/manual/configuration.rst
```diff
@@ -495,6 +495,30 @@ disableSniHostCheck              false                            Whether to dis
 
 .. _sslyze: https://github.com/nabla-c0d3/sslyze
 
+.. _man-configuration-unix-socket:
+
+HTTPS
```
zUniQueX: Shouldn't this be unix-socket?

dropwizard-unix-socket/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        <version>4.0.3-SNAPSHOT</version>
+        <relativePath>../dropwizard-parent</relativePath>
+    </parent>
+
+    <artifactId>dropwizard-unix-socket</artifactId>
+    <name>Dropwizard Unix Socket Support</name>
+
+    <properties>
+        <module.name>io.dropwizard.dropwizard.unixsocket</module.name>
+        <maven.compiler.release>16</maven.compiler.release>
```
zUniQueX: That one's not good. If one builds the project with Java 11 - which currently is the minimum required version - this will cause problems.

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,121 @@
...
+import io.dropwizard.testing.junit5.DropwizardExtensionsSupport;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.io.*;
```
zUniQueX: Our codebase prefers single class imports. So please don't use wildcard imports.

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.net.UnixDomainSocketAddress;
+import java.nio.channels.Channels;
+import java.nio.channels.SocketChannel;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+@ExtendWith(DropwizardExtensionsSupport.class)
+public class UnixSocketConnectorFactoryTest {
```
zUniQueX: JUnit 5 test classes can and should be package-private

zUniQueX (changes requested):
Hi @natan-abolafya. Most changes look reasonable. But with the breaking changes related to the public API and the minimum Java version we can't integrate this PR into the Dropwizard versions 3.x and 4.x.
Dropwizard 5.x will use Jetty 12 and therefore require a Java 17 baseline. For the major release we can additionally integrate API breaking changes.

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
natan-abolafya: It isn't actually necessary. I did it because the PR I was using as a baseline #1893 was doing it and it led to less code duplication. I can remove this commit if it helps https://github.com/dropwizard/dropwizard/pull/7614/commits/22cc888d7af0792698c32002b4bf897ef80044eb.
Though if java 16 is a blocker for this PR, then might as well keep it unless you think it's not worth the breaking change.

pom.xml
natan-abolafya: mainly because the other PR did the same. If it doesn't help with anything, I can move it to jetty.

dropwizard-unix-socket/pom.xml
natan-abolafya: I'm not very experienced with maintaining libraries or even maven really. Any chance we can make this package optional and require java 16 without breaking dropwizard-core's java 11 support?

natan-abolafya:
Hi @zUniQueX. Great review, thank you. Unless we can find a way to do this for 4.0 (by making it an optional package that requires java 16 for example), should I just rebase this for master then?

zUniQueX:
Hi @natan-abolafya. The `master` branch is retired and is therefore not in use. We currently only use the `release/x.x.x` branches for active development.

I'd target this change to the `release/5.0.x` branch which will be created for an upcoming Jakarta EE 10 release.

> by making it an optional package that requires java 16 for example

There are other (and better) options than just releasing it with a specific Java version. If we'd release the artifact with Java 16 and a user pulls this artifact in a project using Java 11, the compiler will throw an error. The better option would be to create a multirelease JAR. Thereby different code for specific Java versions is packaged into one JAR file. E.g. Jersey uses this for some components which require newer Java versions (for an example see the [Jetty 12 upgrade](https://github.com/eclipse-ee4j/jersey/pull/5372)). This doesn't require a specific Java version for the artifact to be used, but rather requires a specific runtime.

natan-abolafya:
ok, thanks. then I'll wait for 5.x release branch. I'll keep this PR open just in case unless you think otherwise.

natan-abolafya:
Hi @zUniQueX. I've noticed that release-5.0.x is there now and rebased it.

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+import java.net.StandardProtocolFamily;
+import java.net.UnixDomainSocketAddress;
+import java.nio.channels.Channels;
+import java.nio.channels.SocketChannel;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+@DisabledOnOs(OS.WINDOWS) // Windows has Unix Socket support but not on all versions.
```
zUniQueX, line 28: The test is disabled for windows, but the normal connector build doesn't care about the OS. Should we add a warning, that running on Windows may cause issues?

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @BeforeEach
+    public void setUp() {
+        final File socket = new File("/tmp/dropwizard.sock");
+        socket.deleteOnExit();
+
+        socketAddress = UnixDomainSocketAddress.of(socket.getPath());
+    }
+
+    @Test
+    public void testClient() throws Exception {
```
zUniQueX: JUnit 5 test methods should be package-private

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+                writer.print(httpRequest);
+                writer.flush();
+
+                verifyHelloWorldReceived(reader);
+            }
+        }
+    }
+
+    @Test
+    public void testManyCalls() throws Exception {
```
zUniQueX: JUnit 5 test methods should be package-private

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+        try (SocketChannel channel = SocketChannel.open(StandardProtocolFamily.UNIX)) {
+            channel.connect(socketAddress);
+            OutputStream os = Channels.newOutputStream(channel);
+            OutputStreamWriter osw = new OutputStreamWriter(os, UTF_8);
```
zUniQueX: The Writer object should be closed properly

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+        socketAddress = UnixDomainSocketAddress.of(socket.getPath());
+    }
+
+    @Test
+    public void testClient() throws Exception {
+        try (SocketChannel channel = SocketChannel.open(StandardProtocolFamily.UNIX)) {
+            channel.connect(socketAddress);
+            OutputStream os = Channels.newOutputStream(channel);
+            OutputStreamWriter osw = new OutputStreamWriter(os, UTF_8);
```
zUniQueX: The Writer object should be closed properly

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+            PrintWriter writer = new PrintWriter(osw);
```
zUniQueX: The Writer object should be closed properly

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    @Test
+    public void testManyCalls() throws Exception {
+        try (SocketChannel channel = SocketChannel.open(StandardProtocolFamily.UNIX)) {
+            channel.connect(socketAddress);
+            OutputStream os = Channels.newOutputStream(channel);
+            OutputStreamWriter osw = new OutputStreamWriter(os, UTF_8);
+            PrintWriter writer = new PrintWriter(osw);
```
zUniQueX: The Writer object should be closed properly

dropwizard-unix-socket/src/test/resources/yaml/usock-server.yml
```diff
@@ -0,0 +1,7 @@
+server:
```
zUniQueX, line 2: Please start a YAML file with '---'

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+                for (int i = 0; i < 1000; i++) {
+                    writer.print(httpRequest);
+                    writer.flush();
+
+                    verifyHelloWorldReceived(reader);
+                }
+            }
+        }
+    }
+
```
zUniQueX, line 56: The tests look reasonable but also rather complicated. I'm personally missing a client call. The Jetty client supports UNIX domain sockets, so we could add a test like this one:

```suggestion
    @Test
    void testHttpClientCall() throws Exception {
        ClientConnector clientConnector = ClientConnector.forUnixDomain(Paths.get("/tmp", "dropwizard.sock"));
        HttpClient httpClient = new HttpClient(new HttpClientTransportOverHTTP(clientConnector));
        httpClient.start();
        ContentResponse contentResponse = httpClient.GET("http://localhost:0/app/hello");
        assertThat(contentResponse).isNotNull().satisfies(response -> {
            assertThat(response.getContentAsString()).isEqualTo("{\"hello\": \"World\"}");
        });
        httpClient.stop();
    }

```

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+    private static void verifyHelloWorldReceived(BufferedReader reader) throws IOException {
+        for (int i = 0; i < 5; i++) {
+            String line = reader.readLine();
+            if (line.isEmpty()) {
+                // here comes body. Need to be specific otherwise it hangs for a long time.
+                char[] buffer = new char[18];
+                int read = reader.read(buffer, 0, 18);
```
zUniQueX: Here you'll always only read 18 bytes. If the response is longer, this wouldn't be detected. I'd like to have a check, if the size of the response is equal to 18 bytes or if some bytes will get swallowed.

zUniQueX (dismissed):
The test looks good overall, I've just added a few minor comments

dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java
zUniQueX: If the change isn't necessary, we shouldn't make it.

pom.xml
zUniQueX: I'm fine with that. The HTTP2 connector also resides in a separate module.

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
natan-abolafya: Well, the test is working on my Windows 10 but not on AppVeyor which didn't look very out of date but for some reason the test kept failing. Microsoft is a bit vague about at what version the unix socket support came to Windows (at least with my googling skills), so not sure what kind of warning to add here.
I guess you mean in the documentation, right?

Another angle is: `unix` in **unix** domain socket should raise an eyebrow on someone who is working on windows :).

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
natan-abolafya: image

Should I configure my Intellij for this?

dropwizard-unix-socket/src/test/java/io/dropwizard/unixsocket/UnixSocketConnectorFactoryTest.java
natan-abolafya: ah, I couldn't find jersey client support earlier and didn't think of using jetty client instead. I prefer that over the current one tbh. I'll change the existing ones instead.

zUniQueX:
LGTM

@dropwizard/committers Anyone else wants to take a look at this one?

joschi (approved):
@natan-abolafya LGTM, thanks!

PR #7617, opened 2023-10-08
Restrict execution of the scacap/action-surefire-report

zUniQueX:
The `scacap/action-surefire-report` action needs write permissions to work properly. Actions triggered from forked repositories cannot acquire those permissions and the action will fail.

Restrict execution of this action to be run in the main dropwizard repo only.


joschi (approved):
@zUniQueX LGTM, thanks! ❤️

PR #7629, opened 2023-10-10
Ensure lexicographical order in messages from HealthCheckConfigValidator

dserfe:
###### Problem:
1. This PR is to fix a flaky test `io.dropwizard.health.HealthCheckConfigValidatorTest.startValidationsShouldFailIfAHealthCheckConfiguredButNotRegistered` in module `dropwizard-health`. We found this when using the latest version of dropwizard.
2. Steps to reproduce the test failure:
- Run the following commands:
```
mvn edu.illinois:nondex-maven-plugin:2.1.1:nondex -pl dropwizard-health -Dtest=io.dropwizard.health.HealthCheckConfigValidatorTest#startValidationsShouldFailIfAHealthCheckConfiguredButNotRegistered
```
- Then you will get the following test failures:
```
[ERROR]   HealthCheckConfigValidatorTest.startValidationsShouldFailIfAHealthCheckConfiguredButNotRegistered:151 
Expecting actual:
  "The following configured health checks were not registered: [check-2, check-3]"
to contain:
  "[check-3, check-2]" 
```
###### Solution:
The flakiness is caused by lines 150-151: `assertThat(e.getMessage()).contains("[check-3, check-2]");`, the assertion asserts the orders of these two elements must be `check-3,check-2`, but actually the order is non-deterministic, which can lead the test fail. To fix the flakiness, the assertion can be changed to ` assertThat(e.getMessage()).contains("check-2").contains("check-3");`, which does not assume an order.

### DIFF
--- a/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckConfigValidator.java
+++ b/dropwizard-health/src/main/java/io/dropwizard/health/HealthCheckConfigValidator.java
@@ -32,9 +32,10 @@ class HealthCheckConfigValidator implements Managed {
             .collect(Collectors.toSet());
 
         // find health checks that are registered but do not have a configured schedule
-        final Set<String> notConfiguredHealthCheckNames = registeredHealthCheckNames.stream()
+        final List<String> notConfiguredHealthCheckNames = registeredHealthCheckNames.stream()
             .filter(healthCheckName -> !configuredHealthCheckNames.contains(healthCheckName))
-            .collect(Collectors.toSet());
+            .sorted()
+            .collect(Collectors.toList());
 
         if (!notConfiguredHealthCheckNames.isEmpty()) {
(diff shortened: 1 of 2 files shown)


zUniQueX:
Hi @dserfe. Thanks for your contribution!

Although we haven't noticed this behavior in many CI runs, this sounds like we should fix this before it'll cause problems for other users.

Unfortunately, your new assertions change the logic of the test. The old assertion additionally validates, that there are only two elements. Now the assertions doesn't check, if `check-1` is in the exception message. We should either extract the stringified list and use a list assertion or add a string assertion, that `check-1` isn't contained.

joschi (approved):
@dserfe Thanks for reporting this! I've decided to take a slightly different approach and ensure the output of `HealthCheckConfigValidator` is ordered so that we don't have to rely on the (arbitrary) order of the items in a `Set`.

PR #7692, opened 2023-10-16
Add JReleaser to simplify release announcements

zUniQueX:
The 'Trigger Release' workflow allows to easily publish a new release. But afterwards the announcements have to be done manually.

This PR adds a basic [JReleaser](https://jreleaser.org/) config to automate the announcement process.


zUniQueX:
@joschi Thanks for adding the missing announcer :heart: But I don't think it's a good idea to use the twitter template name in the mastodon announcer. If we only want to have one template for now, we should rename it to `announcement.tpl` or something.

Any other steps I'm missing in this PR? Testing was successful in my forked repo so far, so it should be good to go otherwise.

joschi (approved):
> Any other steps I'm missing in this PR?

Nope, all ready to go. 👍

PR #7944, opened 2023-11-28
Minor code quality improvements

JadejaJayraj:
###### Problem:

There is a variable 'prince' in the authenticate method of AuthFilter.java instead of principal, maybe because principal variable name is already taken. So, as it is used to for principal.get(), i changed the variable to getPrincipal to make it more meaningful. I also noticed that authenticate method can be made more modular. I then noticed that POJOConfigurtionFactory class has very low test coverage.

###### Solution:

changed authenticate method in AuthFilter.java to be more concise by extracting creation of security context to separate method and changed the variable prince to getPrincipal to make it more meaningful. I also created POJOConfigurtionFactoryTest class to write test cases for it in junit 5.


zUniQueX (changes requested):
@JadejaJayraj Please open a PR for only one problem at a time. This makes it easier to review.

Additionally, please don't commit generated files. We won't review simple cleanup PRs with over 15.000 changes.

joschi:
Closed due to inactivity of author.

PR #7962, opened 2023-12-01
Upgrade to Jakarta EE 10 components

zUniQueX:
This PR updates the dependencies to Jakarta EE 10 components and does necessary migrations.

The main change of this PR is the upgrade to Jetty 12, which removes the servlet classes from the Jetty core.


zUniQueX:
@dropwizard/committers Supporting the logback request log gets really tricky with Jetty 12. I've introduced a workaround here but that is neither clean nor efficient. Should we think of dropping support for the logback request log in favor of the Jetty `CustomRequestLog`?

joschi:
> Should we think of dropping support for the logback request log in favor of the Jetty `CustomRequestLog`?

Yes, let's see how this would look like.

Is there anything we could ask the Logback maintainer for which would enable us to avoid the migration?

joschi (approved):
LGTM. ❤️ 

I've resolved the merge conflicts in #8038 ~but feel free to update this PR and merge it if auto-merge of the new PR fails~. 😅

zUniQueX:
Thanks for the review :+1:

> Is there anything we could ask the Logback maintainer for which would enable us to avoid the migration?

I don't think so. Logback depends on the servlet api and the servlet classes aren't available in the request log stage (hence the `LogbackAccessRequestLogAwareHandler`). My workaround probably cannot be ported to logback and won't be introduced.

Logback access is broken for a few versions now and I don't think it'll be maintained well in the future either. This would be a good moment to drop support IMHO.

PR #8014, opened 2023-12-11
only create the requestLog once to speed up the test

TestBoost:
###### Problem:

In the test class `LogbackClassicRequestLogFactoryTest`, there are three tests but all of them are not modifying `requestLog` or other related classes and they are just accessing `requestLog`. Repeatly creating the `requestLog` object from the `yaml` file will increase the test runtime.

###### Solution:

The solution is to just create the `requestLog` once before running all tests in the test class `LogbackClassicRequestLogFactoryTest`.

###### Result:

There will be no repeated cost when creating the `requestLog` by justing creating it once. The test runtime can also be improved.


joschi (approved):
@TestBoost Thanks for your contribution!

PR #8066, opened 2023-12-18
Managed virtual executor service

### LINKED ISSUE #8004: Managed virtual thread pool
Would it be possible to have this in dropwizard? Or does it need to wait until dropwizard uses java 19?
Something like
```
environment.lifecycle().virtualExecutorService()
```

I'm using it like this in the meantime:
```
    var factory = new InstrumentedThreadFactory(Thread.ofVirtual().factory(), environment.lifecycle().getMetricRegistry(), "virtual-pool");
    var virtualThreadExecutor = Executors.newThreadPerTaskExecutor(factory);
    environment.lifecycle().manage(new ExecutorServiceManager(virtualThreadExecutor, Duration.seconds(5), "virtual-pool"));
```
have not confirmed it's managed as a regular thread pool yet though.

natan-abolafya:
###### Problem:
Java 21 has virtual/lightweight thread support now which is a really powerful tool for performance. We use thread pools quite a lot in our application and were looking forward to replace them with virtual threads instead. But also want them to be managed by dropwizard so we get nice metrics and all.
https://github.com/dropwizard/dropwizard/issues/8004

###### Solution:
Used reflection in order to support older java.
I didn't create/reuse any builders as all the fields in the ExecutorServiceBuilders are obsolete for virtual threads. We let JVM handle it all instead.
I am not sure if this really covers all the aspects of "being managed by dropwizard". Would be good if someone more familiar with the code makes sure that it does.

### DIFF
--- a/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
+++ b/dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
@@ -1,17 +1,23 @@
 package io.dropwizard.lifecycle.setup;
 
+import com.codahale.metrics.InstrumentedThreadFactory;
 import com.codahale.metrics.MetricRegistry;
+import io.dropwizard.lifecycle.ExecutorServiceManager;
 import io.dropwizard.lifecycle.JettyManaged;
 import io.dropwizard.lifecycle.Managed;
 import io.dropwizard.lifecycle.ServerLifecycleListener;
+import io.dropwizard.util.Duration;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.util.component.ContainerLifeCycle;
 import org.eclipse.jetty.util.component.LifeCycle;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 
 import static java.util.Objects.requireNonNull;
@@ -59,6 +65,25 @@ public class LifecycleEnvironment {
 
     public ExecutorServiceBuilder executorService(String nameFormat, ThreadFactory factory) {
         return new ExecutorServiceBuilder(this, nameFormat, factory);
+    }
+
+    public ExecutorService virtualExecutorService(String nameFormat) {
+        try {
+            Object virtualThreadBuilder = Thread.class.getDeclaredMethod("ofVirtual").invoke(null);
+            Object virtualThreadFactory = Class.forName("java.lang.Thread$Builder")
+                .getDeclaredMethod("factory")
+                .invoke(virtualThreadBuilder);
+            InstrumentedThreadFactory factory = new InstrumentedThreadFactory((ThreadFactory) virtualThreadFactory, getMetricRegistry(), nameFormat);
+            ExecutorService virtualThreadExecutor = (ExecutorService) Executors.class
+                .getDeclaredMethod("newThreadPerTaskExecutor", ThreadFactory.class)
+                .invoke(null, factory);
+            manage(new ExecutorServiceManager(virtualThreadExecutor, Duration.seconds(5), nameFormat));
+            return virtualThreadExecutor;
+        } catch (InvocationTargetException invocationTargetException) {
+            throw new IllegalStateException("Error while creating virtual executor service", invocationTargetException.getCause());
+        } catch (Exception exception) {
+            throw new IllegalStateException("Error while creating virtual executor service", exception);
+        }
     }
 
     public ScheduledExecutorServiceBuilder scheduledExecutorService(String nameFormat) {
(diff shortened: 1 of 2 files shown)


dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
```diff
@@ -61,6 +67,25 @@ public ExecutorServiceBuilder executorService(String nameFormat, ThreadFactory f
...
+        try {
+            Object virtualThreadBuilder = Thread.class.getDeclaredMethod("ofVirtual").invoke(null);
+            Object virtualThreadFactory = Class.forName("java.lang.Thread$Builder")
+                .getDeclaredMethod("factory")
+                .invoke(virtualThreadBuilder);
+            InstrumentedThreadFactory factory = new InstrumentedThreadFactory((ThreadFactory) virtualThreadFactory, getMetricRegistry(), nameFormat);
+            ExecutorService virtualThreadExecutor = (ExecutorService) Executors.class
+                .getDeclaredMethod("newThreadPerTaskExecutor", ThreadFactory.class)
+                .invoke(null, factory);
+            manage(new ExecutorServiceManager(virtualThreadExecutor, Duration.seconds(5), nameFormat));
```
zUniQueX: I'd suggest to return an `ExecutorServiceBuilder`. This will allow customizations of the `ExecutorService` and automatically manage it in the `build()` method.

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
```diff
@@ -61,6 +67,25 @@ public ExecutorServiceBuilder executorService(String nameFormat, ThreadFactory f
         return new ExecutorServiceBuilder(this, nameFormat, factory);
     }
 
+    public ExecutorService virtualExecutorService(String nameFormat) {
```
zUniQueX, line 70: Maybe we should add overloads for the `ScheduledExecutorService`s too.

zUniQueX (dismissed):
Hi @natan-abolafya. Thanks for your contribution!

I've added a few comments we should care about before merging.

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
natan-abolafya: > I didn't create/reuse any builders as all the fields in the ExecutorServiceBuilders are obsolete for virtual threads. We let JVM handle it all instead.

If the concern is consistency, it will be rather complicated. You can't really set min/max threads with virtual threads for example. Or maybe you can by manipulating the ThreadFactory but there wouldn't be much point to it really conceptually. Java `Executors` itself does not really provide that kind of option. It's only `newThreadPerTaskExecutor` which is basically "let the virtual thread pool of JVM handle everything".

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
natan-abolafya: That's a good idea. I can fix that after our discussion about builders are complete because kind of depends on that. `poolSize` in the builder also would be irrelevant in this case. Potentially `RejectedExecutionHandler` too, though I haven't read up on that much.

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
zUniQueX: I'm relatively new to virtual threads and just wanted to ask, if we can modify the underlying executor of platform threads. If there is no such option, we indeed can't use the `ExecutorServiceBuilder` for this purpose.

I just wanted to avoid the 'magic' number with 5 seconds as the shutdown time. I'd suggest to provide an overload to allow customizations of that parameter then.

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
zUniQueX: After further looking into this, there's no such factory method for virtual `ScheduledExecutorService` objects in the `Executors` class. We _could_ provide a single threaded dispatching `ScheduledExecutorService` that forwards to a `newVirtualThreadPerTaskExecutor()`, but I don't think that will be very convenient.

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
natan-abolafya: yeah, from what I can see there isn't much we can change when it comes to virtual thread pool management.

I will add an overload for the shutdown time.

dropwizard-lifecycle/src/main/java/io/dropwizard/lifecycle/setup/LifecycleEnvironment.java
natan-abolafya: It's already kind of possible to do it by passing `Thread.ofVirtual().factory()` to 
```
ScheduledExecutorServiceBuilder scheduledExecutorService(String nameFormat, ThreadFactory factory)
```

I can make a new method `scheduledVirtualExecutorService` for convenience but maybe unnecessary?

zUniQueX (approved):

PR #8073, opened 2023-12-19
Remove `logback-access` from the core framework

zUniQueX:
As mentioned in [this comment](https://github.com/dropwizard/dropwizard/pull/7962#issuecomment-1857352214) we could/should think of dropping support for `logback-access`. Logback doesn't support Jetty 12 yet and the [related issue](https://github.com/qos-ch/logback/issues/719) hasn't received comments for about two months.

This PR is meant to show the consequences of removing the `logback-access` dependency from the core framework modules.
The `LogbackAccessRequestLogFactory` will be removed and `LogbackClassicRequestLogFactory` will be the new default.

The JSON logging related `access-json` layout will be removed too.

I've further evaluated what it would look like supporting `logback-access` without the hard dependency in `dropwizard-request-logging` and `dropwizard-json-logging`. We currently have a workaround to get the `LogbackAccessRequestLog` working again, so we could lower the migration costs with supporting the workaround for some time (then preferably in a separate repo which will be deprecated directly). A necessary precondition for that is introducing the `ServerCustomizer` interface to allow customizations of the Jetty handler chain.


zUniQueX:
@dropwizard/committers I've come back to this one and don't have a strong opinion on distributing the separate artifact or not. Personally, I don't like introducing the `ServerCustomizer`.

Feel free to share your opinions, I'll update the PR accordingly.

zUniQueX:
@dropwizard/committers The logback team recently pushed a release of `logback-access` with Jetty 12 compatibility in a [new repo](https://github.com/qos-ch/logback-access). Shall I close this PR if we get further updates or should we move away from `logback-access` anyway?

pstackle:
I would vote in favor of keeping support for `logback-access` as I see the logging integration/configuration as a pretty important component of the Dropwizard framework. It is unfortunate about how long the Logback project took to provide support for Jetty 12, but since it seems like there is at least some form of solution there now, it'd be nice to continue supporting it. 

I'm somewhat biased since we've recently started migrating some of our apps to use JSON logging across the board so we get requests and service log content all in JSON format and were able to achieve that while keeping compatibility with the field names that Splunk's `access_combined` sourcetype uses, which minimizes the disruption to existing splunk queries.

zUniQueX:
Okay @pstackle. I'm fine with that.

Then we'll additionally have to discuss, if we should continue to ship with our own Jetty adapter with full servlet support or rely on Logback's base implementation.

PR #8093, opened 2023-12-20
extra info for enableOCSP and enableCRLDP

Azsaell:
###### Problem:
I've spent a little too much time trying to understand why setting **enableOCSP** to true does nothing. Only after searching the source code, did I discover that I also need **validatePeers**, and without it **enableOCSP** will be just ignored. 

###### Solution:
Adding short info that for **enableCRLDP** or **enableOCSP** to work any of the **validate*** options needs to be set to true.

###### Result:
It will save the time of someone similar to me in the future.


joschi (approved):
@Azsaell Awesome, thanks for your contribution!

PR #8138, opened 2023-12-29
Improve virtual thread handling

### LINKED ISSUE #8137: Server starts and stops immediately when both enableVirtualThreads and enableAdminVirtualThreads are set to true
To reproduce, use  branch release/5.0.x and a Java 21 runtime:
- edit `example.yml` and add `enableVirtualThreads: true` and `enableAdminVirtualThreads: true` to the `server` section.
- start the server as described in the [readme](https://github.com/dropwizard/dropwizard/tree/release/5.0.x/dropwizard-example)

Note: this works fine with only `enableVirtualThreads: true` or `enableAdminVirtualThreads: true` ; the issue appears only when both options are set.

zUniQueX:
Fixes #8137

If the Jetty thread pool(s) only contain virtual threads, there is no application platform thread that will keep the JVM alive and hence a shutdown will be initiated right after the server startup. This PR starts a new platform thread that calls `Server.join()` to properly keep the application alive:

```
"dw-awaiter" id=97 state=WAITING
    - waiting on <0x2197f7df> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
    - locked <0x2197f7df> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
    at java.base@21.0.1/jdk.internal.misc.Unsafe.park(Native Method)
    at java.base@21.0.1/java.util.concurrent.locks.LockSupport.park(LockSupport.java:371)
    at java.base@21.0.1/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionNode.block(AbstractQueuedSynchronizer.java:519)
    at java.base@21.0.1/java.util.concurrent.ForkJoinPool.unmanagedBlock(ForkJoinPool.java:3780)
    at java.base@21.0.1/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3725)
    at java.base@21.0.1/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:1707)
    at app//org.eclipse.jetty.util.thread.AutoLock$WithCondition.await(AutoLock.java:126)
    at app//org.eclipse.jetty.util.thread.QueuedThreadPool.join(QueuedThreadPool.java:822)
    at app//org.eclipse.jetty.server.Server.join(Server.java:662)
    at app//io.dropwizard.core.cli.ServerCommand.lambda$run$0(ServerCommand.java:55)
    at app//io.dropwizard.core.cli.ServerCommand$$Lambda/0x00007ff17c3ad300.run(Unknown Source)
    at java.base@21.0.1/java.lang.Thread.runWith(Thread.java:1596)
    at java.base@21.0.1/java.lang.Thread.run(Thread.java:1583)
```

Additionally, the virtual threads executor is set for Jetty's `QueuedThreadPool`s for tasks not dispatched to the thread pool (see [`AdaptiveExecutionStrategy`](https://github.com/jetty/jetty.project/blob/jetty-11.0.x/jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java#L474-L477)).

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/core/cli/ServerCommand.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/core/cli/ServerCommand.java
@@ -50,6 +50,13 @@ public class ServerCommand<T extends Configuration> extends EnvironmentCommand<T
             server.addEventListener(new LifeCycleListener());
             cleanupAsynchronously();
             server.start();
+            new Thread(() -> {
+                try {
+                    server.join();
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
+            }, "dw-awaiter").start();
         } catch (Exception e) {
             LOGGER.error("Unable to start server, shutting down", e);
             try {
--- a/dropwizard-core/src/main/java/io/dropwizard/core/server/AbstractServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/core/server/AbstractServerFactory.java
@@ -57,6 +57,7 @@ import java.util.EnumSet;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.stream.Collectors;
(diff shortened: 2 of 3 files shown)


dropwizard-core/src/main/java/io/dropwizard/core/cli/ServerCommand.java
```diff
@@ -50,6 +50,13 @@ protected void run(Environment environment, Namespace namespace, T configuration
             server.addEventListener(new LifeCycleListener());
             cleanupAsynchronously();
             server.start();
+            new Thread(() -> {
+                try {
+                    server.join();
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
+            }, "dw-awaiter").start();
```
cty123, line 59: What about only doing this when virtual thread flag is enabled for both admin and app connectors? And maybe have a new function called something like `runBlocking` such that the user can have better control?

zUniQueX: @cty123 IMHO joining the server had to be done in all earlier versions, so I'd want to integrate this without an extra check.

The `runBlocking` method could be integrated. Do you have a special use case why someone would want to override this method?

cty123: I think it would be helpful if the user is able to know when the server has shutdown, such that the user can perform some cleanups right after runBlocking function exits. Currently this is also achievable via shutdown hook, so it's not absolutely necessary.

zUniQueX: That's already possible with `ConfiguredCommand#cleanup()`. If that's the only concern, we shouldn't duplicate that method.

joschi (approved):

PR #8287, opened 2024-01-26
Unix domain socket connector

joschi:
Since Java 16, Unix Domain Sockets are supported natively. And since Jetty 10, there is a ready made connector for it and it is not considered experimental anymore.

The typical use case is a reverse proxy connecting to UNIX socket instead of TCP which has various advantages.

Created a new connector type `unix-socket`.

I've reimplemented #1893 but it needed some adjustments considering how much Dropwizard changed since then.

The test was also a little tricky (and not very pretty) considering there is no UNIX socket support in Apache HTTP client. Or at least I couldn't find any but please correct me otherwise.

It's pretty simple to listen on UNIX sockets now. That was something we wanted in our product for a long time. Also this works on Windows since Windows now also supports UNIX domain sockets.

Refs #1893
Refs #7614
(cherry picked from commit 1de3eef39871859804dd03b61a3cb1c1488b308c)


zUniQueX:
@joschi When this is only supported for Java 16+ should we throw an exception when building the connector for a lower Java version or just keep that up to a user?

On the `release/5.0.x` branch this isn't a problem because we target Java 17+ there.

PR #8364, opened 2024-02-12
Avoid autoboxing in dropwizard-servlets ByteRange

rhowe:
Courtesy of spotbugs


rhowe:
I raised this against 3.0.x - should I instead raise it against 4.0.x?

joschi:
@rhowe Targeting the 3.0.x release branch is fine. I've forward-ported the change into 4.0.x and 5.0.x.

Thanks! ❤️

PR #8446, opened 2024-03-01
Unix Socket Connector: Attempt to delete the preexisting socket file

natan-abolafya:
###### Problem:
We have noticed that if the dropwizard application dies (or `kill -9`ed) abruptly, it leaves the socket file behind. And when the application starts again, it fails with
```
ERROR[ServerCommand] Unable to start server, shutting down
java.lang.RuntimeException: java.io.IOException: Could not bind UnixDomainServerConnector to <path_to_socket>
```

###### Solution:
UnixSocketConnectorFactory tries to delete the socket file on startup.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -485,11 +485,14 @@ Extends the attributes that are available to the :ref:`HTTP connector <man-confi
         - type: unix-socket
           ...
           path: /path/to/file
+          deleteSocketFileOnStartup: false
 
 ================================ ================================ ======================================================================================
 Name                             Default                          Description
 ================================ ================================ ======================================================================================
 path                             /tmp/dropwizard.sock             The path to the unix domain socket file.
+deleteSocketFileOnStartup        false                            Whether to delete socket file on path (if it exists) on startup
+                                                                  instead of failing to start.
 ================================ ================================ ======================================================================================
 
 .. _man-configuration-http2:
--- a/dropwizard-unix-socket/pom.xml
+++ b/dropwizard-unix-socket/pom.xml
@@ -65,6 +65,10 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
 
         <dependency>
             <groupId>org.junit.jupiter</groupId>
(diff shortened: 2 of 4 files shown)


zUniQueX:
Hi @natan-abolafya. I'm not sure, if we'd want to use an automatic delete here.

I wonder, what would happen, if someone deploys two applications on the same server and forgets to override the default path. This would cause a clash between the two applications and the second application to start would try to delete the socket of the first one.

I think a meaningful exception should be a good alternative here.

natan-abolafya:
Hi @zUniQueX.
Yeah, if I remember correctly it throws an error not being able to create file for the unix socket, so that's not a problem. I think not being able to recover itself in certain cases is also bad. Perhaps introduce a flag `deleteSocketFileOnStartup` or something like that?

zUniQueX:
> Perhaps introduce a flag `deleteSocketFileOnStartup` or something like that?

Sounds good. I'd prefer having that flag turned off by default. Then a user would see the error the first time such a crash happens and will need to explicitly handle it. That should ensure that a user provides different socket files for all deployed applications.

natan-abolafya:
Alright, done. Not super happy with the description in the documentation but couldn't rephrase it either. I'm open to suggestions there.

zUniQueX:
@natan-abolafya Thanks, LGTM.

I wonder, if we now provide exception handling on the provided path, if we should validate the path length since there [is a limit](https://github.com/jetty/jetty.project/blob/jetty-12.0.x/jetty-core/jetty-unixdomain-server/src/main/java/org/eclipse/jetty/unixdomain/server/UnixDomainServerConnector.java#L54-L55). I'd also be happy, if this gets addressed in a separate PR.

natan-abolafya:
I personally would argue that we should leave that to the OS as it already provides a meaningful message and that message is propagated to the exception message on startup. So if any OS has a different behavior or it changes in the future, dropwizard wouldn't need to update it in the future.

```
Caused by: java.net.SocketException: Unix domain path too long
        at java.base/sun.nio.ch.UnixDomainSockets.bind0(Native Method)
        at java.base/sun.nio.ch.UnixDomainSockets.bind(UnixDomainSockets.java:115)
        at java.base/sun.nio.ch.ServerSocketChannelImpl.unixBind(ServerSocketChannelImpl.java:326)
        at java.base/sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:299)
        at org.eclipse.jetty.unixdomain.server.UnixDomainServerConnector.bindServerSocketChannel(UnixDomainServerConnector.java:261)
        ... 13 more
```

zUniQueX (approved):

PR #8493, opened 2024-03-14
Feat/add cache support for stale while revalidate

### LINKED ISSUE #8492: Add cache support for stale-while-revalidate
### Problem

`CacheControl` annotation is missing support for `stale-while-revalidate`

khaledosama999:
###### Problem:
No support for stale while revalidate header in cache control #8492 

###### Solution:
Add it as part of the `CacheControl` annotation

Closes #8492

### DIFF
--- a/dropwizard-jersey/src/main/java/io/dropwizard/jersey/caching/CacheControl.java
+++ b/dropwizard-jersey/src/main/java/io/dropwizard/jersey/caching/CacheControl.java
@@ -162,13 +162,35 @@ public @interface CacheControl {
      *         fresh
      */
     int maxAge() default -1;
-
     /**
      * The time unit of {@link #maxAge()}.
      *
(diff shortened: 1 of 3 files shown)


joschi (approved):
@khaledosama999 LGTM, thanks for your contribution! ❤️

joschi:
@khaledosama999 Would you please take a look at the failing test?

zUniQueX:
Hi @khaledosama999. Seems like you're just missing the implementation in the `CacheControlledResponseFeature`. As suggested from the [spec](https://www.rfc-editor.org/rfc/rfc9111#field.cache-control), the `stale-while-revalidate` directive isn't a default directive, but rather an extension. So this should probably set via [`getCacheExtension()`](https://jakarta.ee/specifications/restful-ws/3.0/apidocs/jakarta/ws/rs/core/cachecontrol#getCacheExtension()).

PR #8732, opened 2024-05-06
Update package name to correct path in 3.0 docs

wheatevo:
Updates the package name to the correct path in the 3.0 upgrade documentation.

```
io.drowizard.core -> io.dropwizard.core
```

###### Problem:
I noticed a small package typo while updating a legacy 2.0 app and reviewing the 3.0 documentation.

###### Solution:
Updated the package to the correct name.

###### Result:
Documentation for the 3.0 upgrade will be updated the reflect the correct dropwizard-core package name.


zUniQueX (approved):

joschi (approved):
@wheatevo Thanks for your contribution! 👍

PR #8755, opened 2024-05-10
Update documentation of gzip bufferSize property

tbcardoso:
###### Problem:
The documentation for the gzip `bufferSize` property mentions it is used for compressing. It is actually used for decompressing requests.

The `bufferSize` property in [GzipHandlerFactory](https://github.com/dropwizard/dropwizard/blob/d3dd6fc3677cd2b84c5018f1e8cdef3a969db2c3/dropwizard-jetty/src/main/java/io/dropwizard/jetty/GzipHandlerFactory.java#L247) is passed to the `setInflateBufferSize` method of GzipHandler.
Per the documentation of the [setInflateBufferSize](https://eclipse.dev/jetty/javadoc/jetty-11/org/eclipse/jetty/server/handler/gzip/GzipHandler.html#setInflateBufferSize(int)):
> Set the size (in bytes) of the Inflater buffer used to inflate comrpessed requests.

###### Solution:

Just fix the documentation of the bufferSize property to mention `decompressing` instead of `compressing`.


zUniQueX (approved):
LGTM. Thanks for your contribution.

PR #9111, opened 2024-08-07
Update getting-started.rst

zed1025:


joschi:
Thank you for your contribution!

"Sane" is the correct word in this case. 😉

PR #9157, opened 2024-08-15
Use JUnit's ReflectionSupport rather than its internal ReflectionUtils

### LINKED ISSUE #9191: DropwizardExtensionsSupport does not work with the Junit5 1.11
This commit in Junit 5.11: https://github.com/junit-team/junit5/commit/abb5ed16be3a9ce552f4a45c11264ded608ae9da breaks the usage in https://github.com/dropwizard/dropwizard/blob/81d6b6ef8b92f8a67ead232a15ad799602a2e9b1/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardExtensionsSupport.java#L149

I am not sure how it would be possible to have some code that works for pre-5.11 and 5.11 but as it is right now it makes impossible to update the applications to use Junit 5.11 and keep the usage of the Junit Dropwizard extension

marcphilipp:
###### Problem:

`DropwizardExtensionsSupport` is using JUnit-internal API that was removed in JUnit 5.11.0.

Related issue: junit-team/junit5#3925

###### Solution:

This change replace the call to internal API with a call to a public, supported one that was introduced in JUnit 5.4.

###### Result:

Future versions of Dropwizard won't break when JUnit changes its internal API.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardExtensionsSupport.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardExtensionsSupport.java
@@ -6,7 +6,7 @@ import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
-import org.junit.platform.commons.util.ReflectionUtils;
+import org.junit.platform.commons.support.ReflectionSupport;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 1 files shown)


nerdkid93:
Hello, I noticed this bug as well today when trying to use the new `5.11.0` version of JUnit. However, I am still on dropwizard 3, so can this fix get backfilled to v3 as well?

joschi:
@marcphilipp Thanks a lot for your contribution! ❤️

tmysik:
Will this fix be merged also into DW release 2?

pstackle:
@tmysik Likely not as Dropwizard v2 has been EOL since January earlier this year. 

https://github.com/dropwizard/dropwizard/discussions/7880

tmysik:
OK, thank you.

YSavitski:
Hi @joschi 
Is this fix avaialble for Dropwizard 3.x?

PR #9257, opened 2024-08-30
Use the SLF4J BOM to Manage SLF4J Versions

JKomoroski:
###### Problem:
Sometimes, downstream projects use slf4j dependencies that are not managed by dropwizard. (ie, slf4j-simple for test scope) Projects have to then keep their managed version of slf4j in sync with upstream. 

###### Solution:
Slf4j now has a bom. Use it!

Since the bom only exists for version 2.0.x, I don't think this can be back ported to the DW 2.1.x branch.

###### Result:
Simplified dependency management for dropwizard projects.

### DIFF
--- a/dropwizard-dependencies/pom.xml
+++ b/dropwizard-dependencies/pom.xml
@@ -352,23 +352,10 @@
             <!-- Logging dependencies -->
             <dependency>
                 <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-api</artifactId>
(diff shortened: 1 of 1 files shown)


joschi (approved):
@JKomoroski Thank you for your contribution!

> Since the bom only exists for version 2.0.x, I don't think this can be back ported to the DW 2.1.x branch.

Dropwizard 2.1.x is EOL anyway and won't receive any updates. 😉

PR #9296, opened 2024-09-14
Immediately release http client resources

### LINKED ISSUE #9246: Test hangs indefinitely
See the discussion [here](https://github.com/dropwizard/dropwizard/discussions/8176#discussioncomment-8083810).

> This test hangs after completing the first ~6 requests. The same happens on the release/3.0.x branch, but the release/2.0.x branch does not show this problem:

```java
package com.graphhopper.application.resources;

import io.dropwizard.core.Application;
import io.dropwizard.core.Configuration;
import io.dropwizard.core.setup.Environment;
import io.dropwizard.testing.junit5.DropwizardAppExtension;
import io.dropwizard.testing.junit5.DropwizardExtensionsSupport;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;

import javax.ws.rs.POST;
import javax.ws.rs.Path;
import javax.ws.rs.client.Entity;
import javax.ws.rs.core.Response;

@ExtendWith(DropwizardExtensionsSupport.class)
public class Dropwizard3Test {

    @Path("/")
    public static class MyResource {
        @POST
        public Response doPost() {
            return Response.ok().build();
        }
    }

    public static class MyApp extends Application<Configuration> {

        @Override
        public void run(Configuration configuration, Environment environment) throws Exception {
            environment.jersey().register(new MyResource());
        }
    }

    private static final DropwizardAppExtension<Configuration> app = new DropwizardAppExtension<>(MyApp.class, new Configuration());

    @Test
    public void myTest() {
        for (int i = 0; i < 10; i++) {
            post();
            System.out.println("post " + i + " complete");
        }
    }

    private static void post() {
        Response response = app.client().target("http://localhost:8080/not_found").request().post(Entity.json("{}"));
        System.out.println("success: " + response.getStatus());
    }
}
```

From my investigations it seems this is a client issue, which is "fixed" with:

```java
app.client().property(Apache5ClientProperties.REUSE_STRATEGY, (ConnectionReuseStrategy) (request, response, context) -> false);
```

But maybe the proper fix should be different? 

@otbutz pointed out in the discussion that maybe `response.close();` could be used as a fix? Or maybe the issue is related to #8884?

zUniQueX:
Fixes #9246

When retrieving responses from Jersey clients in the `dropwizard-testing` module to `Response` objects instead of reading them directly, a user has to ensure the `Response` gets closed correctly to release resources of the underlying http client.

In earlier versions of dropwizard this wasn't a problem since there an other http client connector was used.

This PR adds a response interceptor reading the whole entity into a byte array to allow releasing http client resources immediately.

### DIFF
--- a/dropwizard-testing/pom.xml
+++ b/dropwizard-testing/pom.xml
@@ -170,6 +170,10 @@
             <artifactId>httpclient5</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents.core5</groupId>
+            <artifactId>httpcore5</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.glassfish.jersey.core</groupId>
             <artifactId>jersey-client</artifactId>
             <exclusions>
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardAppExtension.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardAppExtension.java
@@ -12,7 +12,15 @@ import io.dropwizard.lifecycle.Managed;
 import io.dropwizard.testing.ConfigOverride;
 import io.dropwizard.testing.DropwizardTestSupport;
 import jakarta.ws.rs.client.Client;
-import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
+import org.apache.hc.core5.http.ContentType;
+import org.apache.hc.core5.http.EntityDetails;
+import org.apache.hc.core5.http.HttpEntity;
+import org.apache.hc.core5.http.HttpEntityContainer;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpResponse;
+import org.apache.hc.core5.http.HttpResponseInterceptor;
+import org.apache.hc.core5.http.io.entity.ByteArrayEntity;
+import org.apache.hc.core5.http.protocol.HttpContext;
 import org.checkerframework.checker.nullness.qual.Nullable;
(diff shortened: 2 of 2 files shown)


zUniQueX:
@dropwizard/committers I'm not very happy with this implementation as it may cause performance overheads. Should we allow this behavior only when explicitly requested by a user through a property?

joschi:
> I'm not very happy with this implementation as it may cause performance overheads.

Since it's only for HTTP-based tests I'd swallow the pill right now and revisit the issue when there are complaints.

joschi (approved):

JoergSiebahn:
Hi there,

I just stumbled over an edge case, likely introduced with this change, that causes a failure when testing for responses with chunked encoding.
We have a service that streams huge amount of data directly from the database as Json to the output stream while still reading from the database. That requires chunked encoding because the content length is unknown when writing the header. That also implies that there is no `Content-length` header or `Content-length` being `-1`.

We verify also the headers to make sure, we are streaming from this endpoint directly. I assume, rewriting to the `ByteArrayEntity` in this PR causes that the response has a length and (to my surprise) not only a change for `getContentLength()` of the entity but also for `getHeaderString("Content-length")` of the `Response`. We now have a `Content-length` header matching the actual response size.

If someone stumbles over the same or similar issues when testing responses in detail, you can create your own client with the minimum you need for your test. In our case we only need Json deserialization and changed `DW_APP_EXTENSION.client().…` to `new JerseyClientBuilder().register(newJacksonFeature(DW.getEnvironment().getObjectMapper())).build().…`.

I hope this helps someone who faces a similar issue after upgrading.

PR #9471, opened 2024-10-20
Fix object identity of ScheduledHealthCheck

rhowe:
The previouslyRecovered and type fields were not included in `equals`, `hashCode` or `toString` of ScheduledHealthCheck

Take the opportunity to align the `toString` methods with other Dropwizard code, using string concatenation instead of `StringBuilder`.


joschi (approved):
@rhowe Thanks!

rhowe:
@joschi  sorry I rebased it and clashed with your button clicking

PR #9598, opened 2024-11-06
Make ProtocolUpgradeEnabled configurable

### LINKED ISSUE #9597: httpclient5 5.4 Adds tls upgrade headers by default, breaking istio integrations
Upgraded a service to dropwizard 4.0.10.
It's deployed in k8s and istio. We saw issues, istio rejected traffic with `response: "upgrade_failed"`

My dear colleague found a number of reported issues:
- https://github.com/istio/istio/issues/53239
> Someone upgraded their [httpclient5](https://mvnrepository.com/artifact/org.apache.httpcomponents.client5/httpclient5) from 5.3.1 to 5.4 and in doing so, all outbound requests fail when the application is running behind Istio.
>Digging into it more, it looks like the client is adding the following upgrade headers:
> `curl -v -H "connection: Upgrade" -H "upgrade: TLS/1.2" http://app.search-one-read.svc.cluster.local/`
- https://github.com/envoyproxy/envoy/issues/36305

From the Dropwizard standpoint, I propose that the `io.dropwizard.client.HttpClientConfiguration` is expanded with a  property `boolean protocolUpgradeEnabled` which can be used in the `io.dropwizard.client.HttpClientBuilder` to set the `RequestConfig::setProtocolUpgradeEnabled`, as this would unblock us.

georgespalding:
###### Problem:

Closes #9597
The httpclient5.4 sends request headers `"connection: Upgrade` and `upgrade: TLS/1.2` by default.
When istio/envoy sees these headers the request is aborted with `response: "upgrade_failed"`
The httpclient5 can be configured to not do this, but there is no practical way to access that setting from dropwizard plumbing today.

This PR attempts to remedy that issue.

###### Solution:

Make  `protocolUpgradeEnabled` configurable on the HttpClientConfiguration.

###### Result:

When `protocolUpgradeEnabled` is set to `false`, it will allow the httpclient5 client to successfully interoperate with istio/envoy.

### DIFF
--- a/docs/source/manual/configuration.rst
+++ b/docs/source/manual/configuration.rst
@@ -1481,6 +1481,7 @@ See HttpClientConfiguration_  for more options.
       keepAlive: 0ms
       retries: 0
       userAgent: <application name> (<client name>)
+      protocolUpgradeEnabled: true
 
 
 =============================  ======================================  =============================================================================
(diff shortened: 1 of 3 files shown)


joschi (approved):
@georgespalding Thanks a lot for your contribution! ❤️

PR #9647, opened 2024-11-26
Use mockito mock#strictness instead of mock#lenient

rhowe:
lenient was deprecated in mockito 4.6.1


rhowe:
We can use strict mocking here, closing

PR #9653, opened 2024-11-27
Clean up assert-j usage

rhowe:
###### Problem:
There was a mix of junit and assert-j assertions as well as some overly-verbose usage of assert-j

###### Solution:
Make some subjective improvements

###### Result:
Tests will be more consistently expressed and there will be overall less test code.


joschi (approved):

Issue #9773, opened 2025-01-14
Access request logging fails on bad path in Dropwizard 5


Mahoney:
When running with access request logging (via logback-access), a request such as this:
`GET /..;/manager/html`
results in no access request log entry and this stacktrace in the logs:
```
java.lang.IllegalStateException: Expecting request to be an instance of ServletContextRequest
	at io.dropwizard.request.logging.LogbackAccessRequestLog.log(LogbackAccessRequestLog.java:34)
	at org.eclipse.jetty.server.internal.HttpChannelState$HandlerInvoker.completeStream(HttpChannelState.java:751)
	at org.eclipse.jetty.server.internal.HttpChannelState$HandlerInvoker.failed(HttpChannelState.java:738)
	at org.eclipse.jetty.server.internal.HttpChannelState.failed(HttpChannelState.java:1960)
	at org.eclipse.jetty.server.internal.HttpChannelState$ErrorCallback.succeeded(HttpChannelState.java:1750)
	at org.eclipse.jetty.util.ExceptionUtil.callAndThen(ExceptionUtil.java:442)
	at org.eclipse.jetty.server.handler.ErrorHandler$WriteErrorCallback.succeeded(ErrorHandler.java:630)
	at org.eclipse.jetty.util.thread.SerializedInvoker$Link.run(SerializedInvoker.java:268)
	at org.eclipse.jetty.util.thread.SerializedInvoker.run(SerializedInvoker.java:168)
	at org.eclipse.jetty.server.internal.HttpChannelState$ChannelResponse.succeeded(HttpChannelState.java:1325)
	at org.eclipse.jetty.server.internal.HttpConnection$SendCallback.onCompleteSuccess(HttpConnection.java:916)
	at org.eclipse.jetty.util.IteratingCallback.processing(IteratingCallback.java:354)
	at org.eclipse.jetty.util.IteratingCallback.iterate(IteratingCallback.java:243)
	at org.eclipse.jetty.server.internal.HttpConnection$HttpStreamOverHTTP1.send(HttpConnection.java:1427)
	at org.eclipse.jetty.server.internal.HttpChannelState$ChannelResponse.write(HttpChannelState.java:1300)
	at org.eclipse.jetty.server.handler.ErrorHandler.generateAcceptableResponse(ErrorHandler.java:257)
	at org.eclipse.jetty.server.handler.ErrorHandler.generateResponse(ErrorHandler.java:163)
[...]

PR #9788, opened 2025-01-16
fix: Allow access log to log paths jetty rejects

Mahoney:
Tentative fix for #9773

###### Problem:

When running with access request logging (via logback-access), a request such as this:
`GET /..;/manager/html`
results in no access request log entry and a stacktrace in the logs.

 See #9773

###### Solution:

Use `ch.qos.logback.access:logback-access-jetty12`'s provided ways of adapting Jetty classes to the Servlet API to allow using logbakc access.


Mahoney:
To state the obvious I have failed to add a test that actually demonstrates this works... just putting this here as a tentative direction of travel rather than asking for it to be merged. If someone can point me at an appropriate way to write an integration test proving this works that would be awesome.

Mahoney:
Actually I see I've effectively reverted #8008 / #7965 here.

I do think that would have been better solved upstream in logback-access.

Mahoney:
I think if we can get https://github.com/qos-ch/logback-access/pull/23 merged and released then this PR would become relevant again.

PR #9806, opened 2025-01-23
Remove `@Nullable` from `CacheLoader` return type.

cpovirk:
Our `CacheLoader` never returns `null`. When we upgrade to Caffeine
3.2.0 (https://github.com/dropwizard/dropwizard/pull/9792), we'll have
to either:

- Declare that we're implementing a `CacheLoader<..., @Nullable
  Class<?>>`, in which case we can keep `@Nullable` in the return type.
- Continue to declare that we're implementing a `CacheLoader<...,
  Class<?>>`, in which case we'll need to remove `@Nullable` from the
  return type.

See https://github.com/ben-manes/caffeine/issues/1826.


joschi (approved):
@cpovirk Thanks a lot for looking into this so quickly!

The explanation makes sense and I think the `@Nullable` annotation wasn't needed in that case.

PR #9827, opened 2025-01-29
Use JsonNode for serialisation tests, not the object under test

andrewaylett:
###### Problem:
The example test given for testing serialisation uses the object under test to construct the expected result.

This means we are relying on serialisation being effective in order to test that serialisation -- if a class has issues with its serialisation then the old example would apply the problematic serialisation to the expected value as well as to the actual value.

###### Solution:
We use JsonNode as a mechanism to make sure we retain the exact structure of the input, while still ensuring we don't fail due to formatting differences.

###### Result:
An updated code example in the documentation that provides a more robust test.

### DIFF
--- a/docs/source/manual/testing.rst
+++ b/docs/source/manual/testing.rst
@@ -89,6 +89,7 @@ Next, write a test for serializing a ``Person`` instance to JSON:
 
 .. code-block:: java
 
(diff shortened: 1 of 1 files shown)


joschi (approved):
@andrewaylett Thanks a lot for your contribution!

PR #10042, opened 2025-04-10
Use Jetty 12 VirtualThreadPool

dennyac:
###### Problem:
The current implementation of enableVirtualThreads is unbounded and doesn't respect the maxThreads option in the server configuration which could potentially cause resource/memory exhaustion.

###### Solution:
They've added [VirtualThreadPool](https://github.com/jetty/jetty.project/blob/jetty-12.0.x/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/thread/VirtualThreadPool.java) in Jetty 12 and it seems like it's the [preferred way](https://jetty.org/docs/jetty/12/programming-guide/arch/threads.html#thread-pool-virtual-threads-queued)

Since we still are in the release candidate phase with 5.x, this shouldn't be a big change. Thoughts?


joschi (approved):

hbzxl:
CAN you merge this change to dropwizard 4.0.x?

joschi:
@hbzxl No. Dropwizard 4.x is based on Jetty 11.

PR #10112, opened 2025-05-22
build: disable pgpverify-maven-plugin

joschi:
https://www.simplify4u.org/pgpverify-maven-plugin/check-mojo.html#keysMapLocation
https://www.simplify4u.org/pgpverify-maven-plugin/keysmap-format.html
https://github.com/s4u/pgp-keys-map


pstackle:
@joschi Can you share some more context for this change?
It looks like you started out adding a keysmap file, but ended up disabling the pgp signature checking altogether. What motivates that change?

joschi:
@pstackle Verifying the dependency signatures has been broken in the `release/5.0.x` branch for quite a while now.

Unreliable or rate-limiting key servers, invalid/missing/revoked keys, etc.

Feel free to pick up before I disabled the plugin. I'll be happy to enable it again.

The last error was about a revoked key.
https://github.com/dropwizard/dropwizard/actions/runs/15198209193/job/42747164304#step:5:10303

PR #10140, opened 2025-05-29
Update pgpverify-maven-plugin configuration.

pstackle:
Improves the configuration of the pgpverify-maven-plugin to allow it to be re-enabled.

Defines plugin configuration to use both the file from the
`org.simplify4u:pgp-keys-map` dependency and a custom keys-map file to allow for
customizations when the `pgp-keys-map` project (https://github.com/s4u/pgp-keys-map/blob/master/resources/pgp-keys-map.list) is missing something.

Related: #10112

### DIFF
--- a/.mvn/pgpverify/custom-keys-map.list
+++ b/.mvn/pgpverify/custom-keys-map.list
@@ -2,7 +2,7 @@
 # This fills in the gaps for https://github.com/s4u/pgp-keys-map
 
 # io.smallrye:jandex:jar:3.2.0 PGP key has been revoked and public key is not available
-io.smallrye:jandex:3.2.0    = !0x94976E17E18DD3201447286954963C3E875A56AE
+io.smallrye:jandex:jar:3.2.0    = !0x94976E17E18DD3201447286954963C3E875A56AE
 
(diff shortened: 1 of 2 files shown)


.mvn/pgpverify/custom-keys-map.list
```diff
@@ -0,0 +1,10 @@
+# Defines a custom keys map file for use by pgpverify-maven-plugin
+# This fills in the gaps for https://github.com/s4u/pgp-keys-map
+
+# io.smallrye:jandex:jar:3.2.0 PGP key has been revoked and public key is not available
+io.smallrye:jandex:3.2.0    = !0x94976E17E18DD3201447286954963C3E875A56AE
```
joschi:
```suggestion
io.smallrye:jandex:jar:3.2.0    = !0x94976E17E18DD3201447286954963C3E875A56AE
```

joschi:
Addressed via https://github.com/dropwizard/dropwizard/pull/10998, https://github.com/dropwizard/dropwizard/pull/10999, and https://github.com/dropwizard/dropwizard/pull/11000.

PR #10253, opened 2025-07-07
CONTRIBUTING.md: remove Google Group instructions

kevinburkesegment:
It looks like the Google Group previously linked in the README has had one post in the last year, which received no replies. By contrast, Github Discussions is a much more active forum for discussion.


joschi (approved):
@kevinburkesegment Thanks! 😄

PR #10284, opened 2025-07-22
Skip Http2WithConscryptTest on ARM MacOS

rnorth:
###### Problem:

When run on an ARM MacOS machine, I'm always getting:
```
[ERROR]   Http2WithConscryptTest.testHttp1WithCustomCipher » UnsatisfiedLink no conscrypt_openjdk_jni-osx-aarch_64 in java.library.path: /Users/richardnorth/Library/Java/Extensions:/Library/Java/Extensions:/Network/Library/Java/Extensions:/System/Library/Java/Extensions:/usr/lib/java:.
[ERROR]   Http2WithConscryptTest.testHttp2WithCustomCipher » NoClassDefFound Could not initialize class io.dropwizard.http2.Http2WithConscryptTest
```

I believe this is due to https://github.com/google/conscrypt/issues/1034 - lack of an ARM native build for MacOS for the Conscrypt library.

###### Solution:
Skip the affected tests when running on ARM MacOS (`aarch64`), so that build does not fail.

It's never a particularly satisfying thing to have to skip tests, but I think it's probably the simplest/clearest way to unblock contributions from developers on ARM Macs. Please do say if there's a better approach that you'd prefer, though.

### DIFF
--- a/dropwizard-http2/src/test/java/io/dropwizard/http2/Http2WithConscryptTest.java
+++ b/dropwizard-http2/src/test/java/io/dropwizard/http2/Http2WithConscryptTest.java
@@ -7,13 +7,13 @@ import io.dropwizard.testing.junit5.DropwizardExtensionsSupport;
 import org.conscrypt.OpenSSLProvider;
 import org.eclipse.jetty.http.HttpVersion;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


joschi (approved):
@rnorth Thanks a lot for your contribution! ❤️

PR #10287, opened 2025-07-24
Make it impossible to instantiate JettyAuthenticationUtil

rhowe:
This is a helper class of static methods - add a private constructor to prevent instantiation.

(appeasing a Sonarqube grumble).

Technically a breaking change - should I raise this against 5.x instead?


joschi (approved):
@rhowe Thanks!

PR #10288, opened 2025-07-24
Add more temporary file patterns to .gitignore

rhowe:


.gitignore
```diff
@@ -17,6 +20,9 @@ release.properties
 .classpath
 .project
 
+# vim
+*.swp
+
```
joschi: This should be handled in a global git-ignore file.


```suggestion
```

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Git
+*.orig
+
```
joschi: Not sure when this can happen. 🤔 


```suggestion
```

.gitignore
```diff
@@ -28,3 +34,6 @@ logs
 .DS_Store
 
 out/jreleaser/
+
+# Test files
+/dropwizard-migrations/src/test/resources/test-db.mv.db
```
joschi: Could you please move this into a `.gitignore` file inside `dropwizard-migrations`?

joschi (approved):
@rhowe Thanks!

.gitignore
rhowe: It's `git mergetool` detritus but global ignore works

PR #10289, opened 2025-07-24
Suppress warnings about Slf4j initialisation

rhowe:
These cases all look intentional and the warnings are just adding noise to the build output.


joschi (approved):
@rhowe Thanks!

PR #10290, opened 2025-07-24
Simplify DropwizardSlf4jRequestLogWriterTest

rhowe:
There were some unused mocks in this test, so remove them.

Also, we can replace `ArgumentCaptor` with mockito's new `assertArg` helper


joschi (commented):

PR #10291, opened 2025-07-24
Replace some ArgumentCaptor use with assertArg

rhowe:
Newer releases of mockito added support for embedding test assertions in verify statements, which means we don't always need to use `ArgumentCaptor`.


joschi (commented):

rhowe:
> @rhowe Could you please use explicit imports instead of wildcard imports?
> 
> Otherwise this looks good to me.

Sure, will amend. I blame IntelliJ here!

joschi (approved):
@rhowe Thanks!

PR #10292, opened 2025-07-24
Avoid string conversions in tests which can test for zero size instead

rhowe:


joschi (commented):

PR #10386, opened 2025-08-20
Simplify GraphiteReporterFactoryTest

rhowe:
We can replace use of ArgumentCaptor with assertArg here, as well as extracting all the fields of interest in one go with Assert-J.


joschi (approved):
@rhowe Thanks

PR #10446, opened 2025-08-31
Replace call to deprecated logback PatternLayout#getDefaultConverterMap

rhowe:
This was deprecated in logback 1.5.13 and replaced by the more typesafe getDefaultConverterSupplierMap


joschi (approved):
@rhowe Thanks!

PR #10453, opened 2025-08-31
Use non-deprecated httpclient methods for managing client timeouts

rhowe:
Revalidation timeouts and connection timeouts have moved to a ConnectionConfig class which allows them to be adjusted per-route.


joschi (approved):
@rhowe Thanks a lot!

PR #10472, opened 2025-09-04
Suppress warning about deprecated logback getDefaultConverterMap overrides

rhowe:


dropwizard-logging/src/test/java/io/dropwizard/logging/common/TestPatternLayoutFactory.java
```diff
@@ -38,7 +38,11 @@ protected Map<String,Supplier<DynamicConverter>> getDefaultConverterSupplierMap(
             return Collections.emptyMap();
         }
 
+        /**
+         * @deprecated Use {@link #getDefaultConverterSupplierMap()} instead
+         */
         @Override
+        @Deprecated
```
jplock, line 42: should we add a "since" in here? (same comment below)

rhowe: Actually on second thoughts I'm not sure the javadoc is adding anything - this a test class and the other one is a private inner class. I've taken it out.

jplock: good point, thanks

PR #10532, opened 2025-09-17
docs: list jackson 2.20.x in Upgrade Notes for Dropwizard 5.0.x

InnokentiyStruchkov:
###### Problem:
The upgrade notes page (https://www.dropwizard.io/en/stable/manual/upgrade-notes/upgrade-notes-5_0_x.html) lists Jackson 2.19.x as a new dependency, but 5.0.0 uses com.fasterxml.jackson:jackson-bom v2.20.0

###### Solution:
Fix the mentioned version.

###### Result:
Clarification of docs.

### DIFF
--- a/docs/source/manual/upgrade-notes/upgrade-notes-5_0_x.rst
+++ b/docs/source/manual/upgrade-notes/upgrade-notes-5_0_x.rst
@@ -24,7 +24,7 @@ Caffeine            2.8.x                 3.2.x
 Guava               31.x                  33.4.x-jre
 Hibernate           6.1.x                 6.6.x
 Hibernate Validator 7.x                   8.0.x
-Jackson             2.15.x                2.19.x
+Jackson             2.15.x                2.20.x
 Jersey              3.0.x                 3.1.x
 Jetty               11.x                  12.1.x
 JUnit               5.9.x                 5.13.x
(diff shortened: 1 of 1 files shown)


docs/source/manual/upgrade-notes/upgrade-notes-5_0_x.rst
```diff
@@ -24,7 +24,7 @@ Caffeine            2.8.x                 3.2.x
 Guava               31.x                  33.4.x-jre
 Hibernate           6.1.x                 6.6.x
 Hibernate Validator 7.x                   8.0.x
-Jackson             2.15.x                2.19.x
+Jackson             2.15.x                2.20.x
```
InnokentiyStruchkov, line 27: I'd like to clarify one more thing. The latest 4.0.x series version--4.0.16--uses Jackson 2.19.2. This comparison table just lists initial Jackson versions for first major Dropwizard versions (4.0.0 and 5.0.0)?

joschi: Yes, it originally only referred to the versions used early in the 4.0.x branch.

Jackson got better with breaking changes, so we updated the minor versions as well.

joschi (approved):
@InnokentiyStruchkov Thanks for your contribution!

PR #10692, opened 2025-10-13
Move format strings into helper methods in dropwizard-auth

rhowe:
errorprone isn't a fan of extracting format strings, accusing it of making code harder to read.

Extract the generation of the response header value into a helper methods instead.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/DefaultUnauthorizedHandler.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/DefaultUnauthorizedHandler.java
@@ -5,14 +5,16 @@ import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 
 public class DefaultUnauthorizedHandler implements UnauthorizedHandler {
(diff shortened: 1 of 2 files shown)


pstackle (approved):

dropwizard-auth/src/main/java/io/dropwizard/auth/DefaultUnauthorizedHandler.java
```diff
@@ -5,14 +5,16 @@
...
     public Response buildResponse(String prefix, String realm) {
         return Response.status(Response.Status.UNAUTHORIZED)
-                .header(HttpHeaders.WWW_AUTHENTICATE, String.format(CHALLENGE_FORMAT, prefix, realm))
+                .header(HttpHeaders.WWW_AUTHENTICATE, getChallengeHeader(prefix, realm))
                 .type(MediaType.TEXT_PLAIN_TYPE)
                 .entity("Credentials are required to access this resource.")
                 .build();
     }
+
+    private String getChallengeHeader(String prefix, String realm) {
```
rhowe, line 17: I guess really this ought to be `static`?

PR #10713, opened 2025-10-22
fix: Remove Logback dependency when external logging is requested

### LINKED ISSUE #10711: Dropwizard Log4j setup fails to build with Dropwizard 5
I am using Dropwizard 4.x with Log4j as described in the Note at this link https://www.dropwizard.io/en/stable/manual/core.html#logging

Configuration:

```yaml
logging:
  type: external

server:
  applicationConnectors:
    - type: http
      port: ${SERVICE_PORT:-8080}
  adminConnectors:
    - type: http
      port: ${ADMIN_PORT:-8081}
  requestLog:
    type: external
```

Gradle build:

```groovy
configurations.configureEach {
    // No Logback
    exclude group: 'ch.qos.logback', module: 'logback-access'
    exclude group: 'ch.qos.logback', module: 'logback-classic'
    exclude group: 'ch.qos.logback', module: 'logback-core'
    exclude group: 'io.dropwizard.logback', module: 'logback-throttling-appender'
    exclude group: 'io.dropwizard.metrics', module: 'metrics-logback'
    exclude group: 'org.slf4j', module: 'log4j-over-slf4j'
}
```

This worked fine for a very long time, but now it fails to build with Dropwizard 5:

```
    java.lang.NoClassDefFoundError: ch/qos/logback/core/spi/AppenderAttachable
        at io.dropwizard.core.server.AbstractServerFactory.addRequestLog(AbstractServerFactory.java:702)
        at io.dropwizard.core.server.DefaultServerFactory.build(DefaultServerFactory.java:197)
        at io.dropwizard.core.cli.ServerCommand.run(ServerCommand.java:48)
        at io.dropwizard.core.cli.EnvironmentCommand.run(EnvironmentCommand.java:67)
        at io.dropwizard.core.cli.ConfiguredCommand.run(ConfiguredCommand.java:98)
```

The trace points to this method where Dropwizard seems to have a hardcoded dependency on the `LogbackAccessRequestLog` class from `logback-core` that is excluded:

https://github.com/dropwizard/dropwizard/blob/be820242653c15d32d730ff5197c7121eeb8e8e7/dropwizard-core/src/main/java/io/dropwizard/core/server/AbstractServerFactory.java#L699-L707

This looks like a breaking change since 4.x that implemented request log handling like this:

https://github.com/dropwizard/dropwizard/blob/08ca9138ee8708d65a9f1798f24abce11e948487/dropwizard-core/src/main/java/io/dropwizard/core/server/AbstractServerFactory.java#L756-L768

My goal is to be able to continue using Dropwizard with Log4j without pulling in any Logback dependencies.

mihalyr:
Dropwizard 5 initially hardcoded a Logback class reference that prevented the use of external logging without Logback dependencies on the classpath.  This commit changes the instanceof check to a class name check that does not require the Logback class to be present at runtime.

###### Problem:

https://github.com/dropwizard/dropwizard/issues/10711

###### Solution:

Replace `instanceof` with a dynamic class name comparison.

###### Result:

No runtime dependency on Logback classes and existing code that worked on Dropwizard 4 keeps working on Dropwizard 5 with external logging without extra Logback dependencies.

### DIFF
--- a/dropwizard-core/src/main/java/io/dropwizard/core/server/AbstractServerFactory.java
+++ b/dropwizard-core/src/main/java/io/dropwizard/core/server/AbstractServerFactory.java
@@ -699,7 +699,8 @@ public abstract class AbstractServerFactory implements ServerFactory {
     protected void addRequestLog(Server server, String name, MutableServletContextHandler servletContextHandler) {
         if (getRequestLogFactory().isEnabled()) {
             RequestLog log = getRequestLogFactory().build(name);
-            if (log instanceof LogbackAccessRequestLog) {
+            // Do not hard-code a dependency to logback when external logging such as Log4j is used
+            if ("io.dropwizard.request.logging.LogbackAccessRequestLog".equals(log.getClass().getName())) {
                 servletContextHandler.insertHandler(new LogbackAccessRequestLogAwareHandler());
             }
             server.setRequestLog(log);
--- a/dropwizard-core/src/test/java/io/dropwizard/core/server/AbstractServerFactoryTest.java
+++ b/dropwizard-core/src/test/java/io/dropwizard/core/server/AbstractServerFactoryTest.java
@@ -8,6 +8,12 @@ import io.dropwizard.jersey.DropwizardResourceConfig;
 import io.dropwizard.jersey.setup.JerseyContainerHolder;
 import io.dropwizard.jersey.setup.JerseyEnvironment;
 import io.dropwizard.jetty.MutableServletContextHandler;
+import io.dropwizard.logging.common.ConsoleAppenderFactory;
+import io.dropwizard.request.logging.ExternalRequestLogFactory;
+import io.dropwizard.request.logging.LogbackAccessRequestLog;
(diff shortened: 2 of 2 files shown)


mihalyr:
I didn't want to go too far here so just sent the first draft with the simplest thing I come up with. 

1. I tested this with my build that was failing after upgrading to DW 5 (#10711) and it was working with the new snapshot build - "works on my machine for my use case"
2. I replaced `instanceof` with a direct class name check, which are not equivalent. The `instanceof` check will work also with class hierarchies, while the class name comparison only works with the exact class. I am not a Logback user, so I couldn't test this with a Logback setup and I am not sure if this is a concern - is overriding the `LogbackAccessRequestLog` class that should be considered? If yes, I can change the impl to compare the names of all classes in the hierarchy instead of just the concrete class.
3. There weren't any tests, I've added some, but they do not cover the dynamic nature of the dependency check. That would need perhaps a separate tests module that excludes Logback and ensures that it still works at runtime. Since this wasn't covered earlier, I thought it was overkill also for this change.

joschi (approved):
@mihalyr Thank you for your contribution! ❤️

PR #11227, opened 2026-06-27
feat: Add nonce-based Content Security Policy (CSP) support

insaf021:
Problem:

Dropwizard applications rendering HTML views lack native support for nonce-based Content Security Policy (CSP), making it difficult to securely authorize inline scripts and styles and increasing the risk of XSS attacks.

Solution:

Implemented nonce-based CSP support by adding a request-scoped filter that generates secure nonces, configurable CSP policies, and a thread-safe mechanism to expose the nonce during view rendering. The nonce is available as `${cspNonce}` in FreeMarker templates and `{{{cspNonce}}}` in Mustache templates, with accompanying unit and integration tests.

Result:

Dropwizard applications can now enable nonce-based CSP protection through configuration, allowing templates to safely use dynamic inline scripts and styles while maintaining thread safety.


insaf021:
Gentle Ping

PR #11230, opened 2026-06-29
fix: Harden ByteRange parsing in AssetServlet against invalid index bounds     

insaf021:
Problem:
Invalid HTTP byte range requests were not fully validated, allowing out-of-bounds or reversed ranges to propagate and potentially cause incorrect response handling.

Solution:
Added range validation in `ByteRange.parse` for resource length, bounds, and range ordering. Updated `AssetServlet` to handle `IllegalArgumentException` and return HTTP `416 Range Not Satisfiable`. Added tests covering invalid range scenarios.

Result:
Invalid HTTP range requests are now safely rejected with HTTP `416 Range Not Satisfiable`, preventing invalid range processing and improving robustness.


dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/ByteRange.java
```diff
@@ -11,50 +11,80 @@
...
+    private final long end;
 
-    public ByteRange(final int start, final int end) {
+    public ByteRange(final long start, final long end) {
         this.start = start;
         this.end = end;
     }
 
-    public int getStart() {
+    public long getStart() {
```
accwebs, line 22: Given that at present ByteRange is only used to index into byte[] values fully cached in memory, the effective use of ByteRange is only for buffers constrained to int lengths.  

I agree the concern of arithmetic overflow could result in bad behavior in extreme length cases.  Thus, changing the math to long like you have done in `parse()` seems sensible.

But what I'm less sold on is changing the start/end interfaces of ByteRange to long.  Instead, shouldn't we just:

1.  Keep the members of type int
2. Add validation at the end of parse that checks 
`long_value >= 0 && long_value <= Integer.MAX_VALUE` and if violated, throws IllegalArgumentException like you're already doing for similar cases?

This would mean that AssetServlet could then - with confidence - simply index into the byte[] array using ByteRange.getStart() and getEnd() without further validation?

What do you think about that approach?

accwebs (commented):
Hi, trying to help review a bit of the built-up PR backlog :-)

The feedback here is speaking only as myself.

It's hard to fully review this PR because of its overlap with various other PRs you opened.  However, I think the value this PR mainly brings is a concern about ByteRange and its handling.  So I will focus on that here.

See my individual comment for more details

PR #11233, opened 2026-06-30
Security hardening: path-traversal guard, file-descriptor leak fix, and timing-oracle fixes

insaf021:
Problem:
Several security hardening opportunities were identified in AssetServlet, ResourceURL, and BasicCredentials, including path traversal validation, a file descriptor leak, password usage in hashCode(), and ETag comparison.

Solution:
Added path traversal validation in AssetServlet, fixed the file descriptor leak in ResourceURL by properly closing the opened InputStream, updated BasicCredentials.hashCode() to exclude the password, and replaced the ETag comparison with a constant-time comparison in AssetServlet.

Result:
Improves security hardening and resource handling without changing the public API or the behavior of valid requests. New tests were added, and all existing tests continue to pass.

### DIFF
--- a/dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentials.java
+++ b/dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentials.java
@@ -1,5 +1,7 @@
 package io.dropwizard.auth.basic;
 
+import java.security.MessageDigest;
+import java.nio.charset.StandardCharsets;
 import java.util.Objects;
 
 import static java.util.Objects.requireNonNull;
@@ -43,9 +45,32 @@ public class BasicCredentials {
 
     @Override
     public int hashCode() {
-        return Objects.hash(username, password);
+        // Intentionally exclude the password from the hash.
+        //
+        // BasicCredentials is used as the cache key of CachingAuthenticator, which stores
+        // credentials in a hash-map. Including the password in hashCode() would mean that
+        // two credentials for the same user with *different* passwords land in different
+        // buckets, so the constant-time equals() comparison (which uses MessageDigest.isEqual)
+        // is never reached. An attacker who can measure cache-hit vs. bucket-miss timing would
+        // gain a coarser timing oracle on the password.
+        //
+        // Excluding the password ensures that same-username lookups always enter the same
+        // bucket, forcing the constant-time equals() path for every authentication attempt.
+        // The Java hashCode/equals contract still holds: equal objects share the same username
+        // and therefore the same hashCode.
+        return username.hashCode();
     }
 
+    /**
+     * Returns {@code true} if both the username and password match.
+     *
+     * <p>The password comparison uses {@link MessageDigest#isEqual} — a constant-time
+     * byte comparison — to eliminate a timing side-channel. Because {@link
+     * io.dropwizard.auth.CachingAuthenticator} uses {@code BasicCredentials} as a
+     * cache key, {@code equals} is on the critical authentication path; a short-circuit
+     * {@link String#equals} comparison would leak prefix information about the stored
+     * password to a remote attacker who can measure response latency.
+     */
     @Override
     public boolean equals(Object obj) {
         if (this == obj) {
@@ -55,7 +80,10 @@ public class BasicCredentials {
             return false;
         }
         final BasicCredentials other = (BasicCredentials) obj;
-        return Objects.equals(this.username, other.username) && Objects.equals(this.password, other.password);
+        return Objects.equals(this.username, other.username)
+                && MessageDigest.isEqual(
+                       this.password.getBytes(StandardCharsets.UTF_8),
+                       other.password.getBytes(StandardCharsets.UTF_8));
     }
 
 
--- a/dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/AssetServlet.java
+++ b/dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/AssetServlet.java
@@ -13,14 +13,17 @@ import java.io.InputStream;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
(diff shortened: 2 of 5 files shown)


dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentials.java
```diff
@@ -43,9 +45,32 @@ public String getPassword() {
 
     @Override
     public int hashCode() {
-        return Objects.hash(username, password);
+        // Intentionally exclude the password from the hash.
```
accwebs, line 48: Overall, my opinion of this is that yes - BasicCredentials should be adjusted in this way; a timing attack right now could conceivably be used to extract a secret value from memory.

Nice catch @insaf021!

dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/AssetServlet.java
```diff
@@ -47,9 +50,17 @@ private CachedAsset(byte[] resource, long lastModifiedTime) {
         }
 
         private static String hash(byte[] resource) {
-            final CRC32 crc32 = new CRC32();
-            crc32.update(resource);
-            return Long.toHexString(crc32.getValue());
+            try {
```
accwebs, line 53: I'm confused about the 'security' labeling of this change.  Certainly CRC32 is not super collision resistant.  But at the same time, Etag values are never secret, right?  Is there a situation that I'm not thinking of where they could conceivably be secret or a collision results in a security issue?

Basically - are you trying to fix a security problem here or just fix an annoyance/bug where you occasionally hit etag collisions in practice?

dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/AssetServlet.java
```diff
@@ -308,7 +331,13 @@ private boolean isCachedClientSide(HttpServletRequest req, CachedAsset cachedAss
         // Indicates that with the presense of If-None-Match If-Modified-Since should be ignored.
         String ifNoneMatchHeader = req.getHeader(IF_NONE_MATCH);
         if (ifNoneMatchHeader != null) {
-            return cachedAsset.getETag().equals(ifNoneMatchHeader);
+            // Use constant-time comparison to prevent a timing side-channel.
+            // A variable-time String.equals() would allow an attacker to recover the
+            // ETag value character-by-character via network latency measurements and
+            // then forge If-None-Match headers to suppress delivery of updated assets.
+            return MessageDigest.isEqual(
```
accwebs, line 338: I'm confused why we're worried about timing attacks when Etag values are non-secret.  Can you explain the rationale?

dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/ResourceURL.java
```diff
@@ -102,14 +102,21 @@ public static long getLastModified(URL resourceURL) {
...
+                    // descriptor on every call.
+                    stream = connection.getInputStream();
                     return connection.getLastModified();
                 } catch (IOException ignored) {
                 } finally {
-                    if (connection != null) {
+                    if (stream != null) {
                         try {
-                            connection.getInputStream().close();
+                            stream.close();
```
accwebs, line 119: I'm not at all following this change; the original code appears correct (albeit counter-intuitive).  I might be missing something though.  Can you explain?

Best I can tell, for `FileURLConnection` instances, `resourceURL.openConnection()` fundamentally refers to the same `Closeable` thing that connection.getInputStream() returns.  Calling connection.getInputStream() in finally does not "connect" anything because it's already connected via `openConnection` to get the last modified date. The original comments here (prior to this PR) are correct - namely - that due to Java's poor design (likely due to these classes coming before resource closing was 'solved' in Java), the `URLConnection` class is not Closeable/AutoCloseable, despite in the case of being a file having closeable-like properties.  Thus what must be done to close it is to get the input stream and close that.

Are you seeing something else on your specific Java version?  It might be that this behavior has changed across versions.  I tested with Java 17, up to date with latest patches.

accwebs (commented):
Hi, trying to help review a bit of the built-up PR backlog :-) 

The feedback here is speaking only as myself.

Overall, you are attempting to solve 4 issues in this PR where perhaps it would be better to split out each issue as a separate PR?

1.  `BasicCredentials` equality testing timing attack concern
    - overall, nice catch, I agree we should fix this.
    - your solution seems reasonable

2.  Etag collision concerns due to CRC32
    - I'm confused about the security framing. See my individual comment on this.  Is this perhaps simply a non-security improvement to try to make collisions less frequent?  Or do I misunderstand?

3.  Etag equality testing timing attack concern
    - I'm confused why we are worried about this given the non-sensitive nature of etags. Am I missing something?  See my individual comment on this.

4.  Resource leak and extra file handle creation in ResourceURL
    - The PR change here appears to be a no-op; the original behavior was correct, albeit counter-intuitive.  I might have missed something though.

I know it's been a while since you raised this @insaf021 but I would suggest breaking each of these items (assuming we keep them after this initial discussion) out to separate PRs?  That could be reviewed/discussed separately?

dropwizard-servlets/src/main/java/io/dropwizard/servlets/assets/AssetServlet.java
```diff
@@ -201,7 +212,7 @@ protected void doGet(HttpServletRequest req,
 
             final String rangeHeader = req.getHeader(RANGE);
 
-            final int resourceLength = cachedAsset.getResource().length;
+            final long resourceLength = cachedAsset.getResource().length;
```
accwebs, line 215: This change and the one below seems to have 'snuck in' from another PR.

dropwizard-auth/src/main/java/io/dropwizard/auth/basic/BasicCredentials.java
accwebs: I've done a more in-depth analysis of this specific issue/concern here:
https://github.com/dropwizard/dropwizard/issues/11334

Would be interested in your thoughts.

PR #11265, opened 2026-07-27
Fix SafeRequestParameterConverter registration

unkish:
###### Problem:
`LogbackAccessRequestLayout` is an Access-log layout, but it registers `SafeRequestParameterConverter` in Logback Classic's converter map, therefore leading to usage of (default) `RequestParameterConverter` instead of `SafeRequestParameterConverter`.

###### Solution:
* Register `SafeRequestParameterConverter` with Logback Access's PatternLayout rather than Logback Classic's unrelated PatternLayout
* Replace deprecated `OptionHelper.isEmpty` with `OptionHelper.isNullOrEmptyOrAllSpaces`

### DIFF
--- a/dropwizard-request-logging/src/test/java/io/dropwizard/request/logging/layout/LogbackAccessRequestLayoutTest.java
+++ b/dropwizard-request-logging/src/test/java/io/dropwizard/request/logging/layout/LogbackAccessRequestLayoutTest.java
@@ -1,18 +1,26 @@
 package io.dropwizard.request.logging.layout;
 
+import ch.qos.logback.access.common.spi.AccessEvent;
+import ch.qos.logback.access.common.spi.ServerAdapter;
 import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.core.Context;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
 
 import java.util.TimeZone;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.AdditionalAnswers.delegatesTo;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 
 class LogbackAccessRequestLayoutTest {
-    final Context context = mock(LoggerContext.class);
+    private final Context context = mock(LoggerContext.class);
     private final TimeZone timeZone = TimeZone.getTimeZone("UTC");
-    final LogbackAccessRequestLayout layout = new LogbackAccessRequestLayout(context, timeZone);
+    private final LogbackAccessRequestLayout layout = new LogbackAccessRequestLayout(context, timeZone);
 
     @Test
     void outputPatternAsHeaderIsFalse() {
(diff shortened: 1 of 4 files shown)


josephlbarnett (commented):
DEFAULT_CONVERTER_MAP is also gone from logback-classic 1.6.x, I think this should fix that too?

mattnelson:
> DEFAULT_CONVERTER_MAP is also gone from logback-classic 1.6.x, I think this should fix that too?

Looks like it.

Some more context > https://github.com/dropwizard/dropwizard/pull/11262#issuecomment-5106363220

josephlbarnett (approved):
effectively running with this and haven't seen any problems fwiw

dropwizard-request-logging/src/test/java/io/dropwizard/request/logging/layout/LogbackAccessRequestLayoutTest.java
```diff
@@ -30,4 +38,22 @@ void hasAPatternWithATimeZone() {
...
+        ServerAdapter serverAdapter = mock(ServerAdapter.class);
+        AccessEvent accessEvent = mock(
+                AccessEvent.class,
+                delegatesTo(new AccessEvent(context, httpServletRequest, httpServletResponse, serverAdapter)));
+
+        layout.setPattern("%" + patternKey + "{parameterName}");
+        layout.start();
+        layout.doLayout(accessEvent);
+
+        verify(accessEvent).getRequestParameterMap();
```
accwebs, line 59: I honestly was (and possibly still am) struggling to understand the logic of this test.  I think it is predicated on:

1. Right now, Logback's stock behavior doesn't call `AccessEvent.getRequestParameterMap()`
2. The `SafeRequestPArameterConverter` implementation DOES call `getRequestParameterMap()`

Is that the thinking?

If so, I follow. I've certainly done far worse in test code so I am not trying to be critical here.  But I wonder if this is overly-fragile since if a future Logback-access starts calling getRequestParameterMap somehow, and this breaks again, this test would not catch it.

I locally had build a tentative code change with this test:
```
// Verifies that the Dropwizard override installed in LogbackAccessRequestLayout's static initializer:
    // reqParameter / requestParameter must resolve to SafeRequestParameterConverter, not upstream's
    // RequestParameterConverter (which reads live request state and is unsafe under async appenders).
    @Test
    @SuppressWarnings("NullAway")
    void reqParameterUsesSafeConverter() {
        // when
        final Supplier<DynamicConverter> reqParamSupplier = layout.getEffectiveConverterMap().get("reqParameter");
        final Supplier<DynamicConverter> requestParamSupplier
            = layout.getEffectiveConverterMap().get("requestParameter");

        // then
        //    Invoke each supplier to materialize the converter instance and check its class.
        assertThat(reqParamSupplier)
            .isNotNull();
        assertThat(requestParamSupplier)
            .isNotNull();
        assertThat(reqParamSupplier.get())
            .isInstanceOf(SafeRequestParameterConverter.class);
        assertThat(requestParamSupplier.get())
            .isInstanceOf(SafeRequestParameterConverter.class);
    }
```

Now, is this better?  IDK.  For sure at least checking the type of the supplied thing might be less fragile, yet, I'm also making a bold assumption that what I get from the `layout` is actually something representative of what AccessEvent will use in reality.

Thoughts?  Would it be smart to include both of these test variants to hopefully be a bit more resilient in case future changes invalidate the assumptions of one of them?

accwebs: Nice job finding and fixing the bug ofc <3

unkish: Hi

> Is that the thinking?

Correct.
I agree that it is fragile/susceptible to potential change of `RequestParameterConverter` starting calling `getRequestParameterMap`.

> Thoughts? Would it be smart to include both of these test variants to hopefully be a bit more resilient in case future changes invalidate the assumptions of one of them?

Personally I see no harm in adding proposed test, having both tests or adding `isInstanceOf` assertion to existing one ie.
```java
        layout.setPattern("%" + patternKey + "{parameterName}");
        layout.start();
        layout.doLayout(accessEvent);

        verify(accessEvent).getRequestParameterMap();

        final Supplier<DynamicConverter> converterSupplier = layout.getEffectiveConverterMap().get(patternKey);
        assertThat(converterSupplier).isNotNull().extracting(Supplier::get).isInstanceOf(SafeRequestParameterConverter.class);
    }
```

Would gladly hear opinions as to which approach to take.

P.S. Thank you for the feedback!

unkish:
To bump logback-classic to 1.6.x for dropwizard build - `LogbackClassicRequestLogFactory` would also have to be modified by removing following override:
```java
        @Override
        @Deprecated
        @SuppressWarnings("InlineMeSuggester")
        public Map<String, String> getDefaultConverterMap() {
            return Collections.emptyMap();
        }
```

dropwizard-request-logging/src/test/java/io/dropwizard/request/logging/layout/LogbackAccessRequestLayoutTest.java
pstackle: @unkish First, thank you for taking the time to make this contribution. 
Also, thank you @accwebs for the feedback. 

I agree that just testing that the method gets called is a bit fragile, so I'd vote to also include the assertion that verifies the expected class is returned in the one test.

accwebs (approved):

pstackle (approved):
Thanks @unkish for your contribution and your patience!

PR #11297, opened 2026-08-24
build(deps): update actions/setup-java to v6

brunoborges:
###### Problem:

Active workflows use an older actions/setup-java release.

###### Solution:

Update every pinned reference to actions/setup-java v6.0.0.

###### Result:

All active workflows use the current setup-java major release.


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #11302, opened 2026-08-30
Allow Liquibase 4.33.0 revoked signing key

skrcode:
## Summary

- allow the existing Liquibase 4.33.0 artifact's revoked signing key in the repository PGP map
- scope the exception to that exact group, artifact, and version
- preserve signature verification and every existing key rule for all other artifacts and versions

This is a narrow companion fix for Renovate PR #11300 and targets its branch so the FreeMarker upgrade remains attributable to Renovate.

## Why

The FreeMarker 2.3.35 upgrade compiles and tests successfully, but every Java CI matrix entry now stops in `dropwizard-migrations` because the previously accepted Liquibase 4.33.0 signer has been revoked and its public key is unavailable:

> org.liquibase:liquibase-core:jar:4.33.0 PGP key 0x4F474D657AAB9B387B812DA2915C4D2EC1FE0D33 has been revoked and public key is not available

The repository already uses exact dependency/version exceptions for this PGP failure class. This applies the same narrow policy shape and does not allow the key for another Liquibase release or artifact.

## Verification

- `mvn -B --no-transfer-progress clean install -DskipTests` on JDK 17: all 42 modules passed, including PGP verification throughout the reactor
- `mvn -B --no-transfer-progress -pl dropwizard-migrations,dropwizard-views-freemarker verify` on JDK 17
- 64 focused tests run, 0 failures, 0 errors, 1 skipped
- Maven Enforcer dependency convergence and banned-dependency rules passed
- PGP verification validated 102 artifacts for migrations and 94 for FreeMarker views
- complete diff reviewed: one exact-version PGP map entry and its diagnostic comment

Built and verified with [JAIPilot](https://github.com/JAIPilot/jaipilot):
- Skills: `jaipilot-maintainer-intent`, `jaipilot-fast-execution`, and `jaipilot-review-diff`
- Execution profile: `gpt-5.6-sol · xhigh · fast`

### DIFF
--- a/pgp-keys-map.list
+++ b/pgp-keys-map.list
@@ -2,6 +2,8 @@
 io.reactivex.rxjava2:rxjava = !0x1D9AA7F9E1E2824728B8CD1794B291AEF984A085
 # io.smallrye:jandex:jar:3.2.0 PGP key 0x94976E17E18DD3201447286954963C3E875A56AE has been revoked and public key is not available
 io.smallrye:jandex:3.2.0 = !0x94976E17E18DD3201447286954963C3E875A56AE
+# org.liquibase:liquibase-core:jar:4.33.0 PGP key 0x4F474D657AAB9B387B812DA2915C4D2EC1FE0D33 has been revoked and public key is not available
+org.liquibase:liquibase-core:4.33.0 = !0x4F474D657AAB9B387B812DA2915C4D2EC1FE0D33
 # Not allowed artifact org.hibernate.orm:hibernate-core:jar:6.6.41.Final and keyID
 org.hibernate.orm:hibernate-core = 0x1452F35849B50750F6A3BBB4B54011358B352F85
 # Not allowed artifact org.jboss.logging:jboss-logging:jar:3.6.3.Final and keyID


accwebs:
@pstackle I dug into this (was also looking at it since the broken build was annoying me .... thanks @skrcode for working on this!):

My review asks two questions:
1. Is the fingerprint in this PR the correct key for Liquibase's prior artifact signing?
2. Why was the key revoked (does this represent a concern for allow-listing this key)?


# Question 1

This key is confirmed to be still listed when you download https://repo.liquibase.com/liquibase.asc and import those keys.  The current bundle contains:

```
pub   rsa3072/0x915C4D2EC1FE0D33 2024-02-16 [SC] [revoked: 2026-08-18]
      reason for revocation: Key has been compromised
      Key fingerprint = 4F47 4D65 7AAB 9B38 7B81  2DA2 915C 4D2E C1FE 0D33
uid                   [ revoked] Liquibase <support@liquibase.org>

pub   rsa4096/0x9CB34520F4F52C15 2025-08-12 [SC] [revoked: 2026-08-18]
      reason for revocation: Key has been compromised
      Key fingerprint = A1DD 405B E11C 10B6 8642  177F 9CB3 4520 F4F5 2C15
uid                   [ revoked] Liquibase <devops@liquibase.com>

pub   rsa4096/0x08263CBA8BCD9EE8 2026-08-18 [SC] [expires: 2028-08-17]
      Key fingerprint = A33B A2DD 58AB DFC5 441F  425F 0826 3CBA 8BCD 9EE8
uid                   [ unknown] Liquibase <devops@liquibase.com>
sub   rsa4096/0x24423F6796EEFC73 2026-08-18 [E] [expires: 2028-08-17]
      Key fingerprint = D1CD C4C1 1B2A 541C B010  46C9 2442 3F67 96EE FC73
```

The first key of the three matches this PR.  Thus, this PR lists a valid key fingerprint for Liquibase.  **But, since I have no built up web-of-trust for these keys, it would be smart if a few others could confirm these key fingerprints are the liquibase keys they also have.**

# Question 2

Liquibase seems to have revoked this keys due to 'key compromise':
```
pub   rsa3072/0x915C4D2EC1FE0D33 2024-02-16 [SC] [revoked: 2026-08-18]
      reason for revocation: Key has been compromised
      Key fingerprint = 4F47 4D65 7AAB 9B38 7B81  2DA2 915C 4D2E C1FE 0D33
uid                   [ revoked] Liquibase <support@liquibase.org>
```

This might indicate that allow-listing it for further builds might not be advisable.  However, if you read the release notes for 5.0.4: https://github.com/liquibase/liquibase/releases#release-v5.0.4

> Release signing key rotated: The Liquibase release signing key was rotated following the revocation of the old key. GPG verification requires Liquibase Community 5.0.4 or later. Past releases are safe and untampered.

The 8-18 revocation date for both key `0x915C4D2EC1FE0D33` and key `0x9CB34520F4F52C15` lines up very much with 5.0.4's release date where this statement was made.  There is no explicit listing of what the old/new keys are, that I could find.  So I think the best we can do is the date correlation.  

Their statement in the release notes is of course is a contradictory statement to a 'key compromised' revocation reason.  That being said, like with many PKI certificate revocations, I find that people usually just blindly select a revocation reason without thinking about it and making a point to be accurate.  So it wouldn't surprise me if they blindly picked reason code 1 in error, rather than a more appropriate code 2.

----

In sum, I think this PR (with key fingerprint `0x4F474D657AAB9B387B812DA2915C4D2EC1FE0D33`) is good to merge once we get maybe one more confirmation on the Liquibase key fingerprint.

pstackle:
Thank you @skrcode and @accwebs! 

I can see that the referenced key got added to the `s4u/pgp-keys-map` repo back in 2024: https://github.com/s4u/pgp-keys-map/pull/2159

@accwebs I acknowledge the truth behind your analysis of question 2, but I wish a more accurate reason code was chosen to reduce the ambiguity with the more problematic scenario of an actual key compromise.

Issue #11303, opened 2026-09-01
Request logging of user ID and thread name is broken


accwebs:
At present, with Logback Access's request logging implementation, Dropwizard request logs fail to log a few fields.  After some extensive testing, the affected fields are:
 * user ID
 * thread name

Other bugs:
 * 'response header' logging logs "" (empty) instead of - when the header is not present. 
     * (Where the 'request header' logging DOES log a dash.)
 * request/response content logging - when enabled - does not honor character encodings
 * the IAccessEvent.getRequest/getResponse's decision to return a fake HttpServletRequest/HttpServletResponse breaks esoteric features in logback-access like `AccessEventDiscriminator`. This can't be wired up in Dropwizard today best I can tell, but it's sloppy.

The source of the user ID problem is again the `RequestWrapper` class in the `logback-access-jetty12` library:

```
    @Override
    public String getRemoteUser() {
        return null;
    }

    @Override
    public boolean isUserInRole(String role) {
        return false;
    }

    @Override
    public Principal getUserPrincipal() {
        return null;
    }
```

Separately, the lack of logging "thread name" come down to simply not calling `AccessEvent.setThreadName()` in the `RequestLogImpl`/`LogbackAccessRequestLog` log(...) implementation (neither implementation does this).  Obviously this is not nearly as much of a concern as the lack of user logging.

accwebs:
# Analysis

Obviously, we could keep fixing up `RequestWrapper` and `ResponseWrapper` like we are already doing:

```
        // TODO: remove build*HeaderMap overrides once https://github.com/qos-ch/logback-access/pull/23
        //       is merged and released in logback-access
        HttpServletRequest httpServletRequest = new RequestWrapper(jettyRequest) {
            @Override
            public Map<String, String> buildRequestHeaderMap() {
                return buildHeaderMap(jettyRequest.getHeaders());
            }
        };
        HttpServletResponse httpServletResponse = new ResponseWrapper(jettyResponse);
        ServerAdapter adapter = new JettyServerAdapter(jettyRequest, jettyResponse) {
            @Override
            public Map<String, String> buildResponseHeaderMap() {
                return buildHeaderMap(jettyResponse.getHeaders());
            }
        };
```

This feels fragile.  We have been 'bitten' a few times now by similar bugs and it's time to do something smarter IMO.

---

[...]

PR #11307, opened 2026-09-02
<argLine> cleanups (submodules are inadvertently overriding parent argLine)

accwebs:
Spotted while I was working on #11303: argLine right now is a bit of a jumbled mess.

This tries to clean it up:
1. Remove `--illegal-access=deny` since Java 17 is required (thus it's a no-op just generating warnings)
2. Remove unnecessary dropwizard-jackson argLine customization
3. Remove dropwizard-example and dropwizard-http2 <argLine> which is overriding the parent argLine which is likely unintentional

Further thoughts:  if we anticipate needing custom argLines per submodule we could later add a `<argLine.parent>` property in the parent POM and then define `<argLine>${argline.parent}</argLine>` in the parent.  This would give submodules a simple way to include the parent value and add to it.

But TBH I was having trouble anticipating when we'd really need this.  IMO the whole project should be tested with the same argLine which should (ideally) be the one documented as required for use of Dropwizard + perhaps a few affects-tests-only changes.  So for now I did not do that.

### DIFF
--- a/dropwizard-example/pom.xml
+++ b/dropwizard-example/pom.xml
@@ -11,7 +11,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <maven.compiler.release>11</maven.compiler.release>
-        <argLine>-Duser.language=en -Duser.region=US -Duser.timezone=UTC</argLine>
 
         <!-- You'll probably want to remove this for your project. We're just using it here
              so that dropwizard-example doesn't get deployed as a library. -->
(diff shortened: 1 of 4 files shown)


dropwizard-example/pom.xml
```diff
@@ -11,7 +11,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <maven.compiler.release>11</maven.compiler.release>
-        <argLine>-Duser.language=en -Duser.region=US -Duser.timezone=UTC</argLine>
```
accwebs, line 14: So, I could see wanting to leave this here since it's ?maybe? part of an example that is intended to be copiable.  But since this is a member of the project I removed it.  But I could see either way.

accwebs: Notably examples-parent in the docs dir has an `<argLine>` but that seems to be intended to be a separate Maven module in that case, so I left it alone.

Issue #11304, opened 2026-09-01
DropwizardTestSupport should own the Jersey Client object (not DropwizardAppExtension)


accwebs:
Best I can tell, the concept of `DropwizardTestSupport` is to give a common integration test 'hook' point regardless of test framework in use.  Then, classes like `DropwizardAppExtension` add a Junit-specific registration pattern on top.  But fundamentally `DropwizardAppExtension` and `DropwizardExtensionsSupport` simply expose the underlying `DropwizardTestSupport` functionality.

Today this is mostly how things work.  However, `DropwizardAppExtension` has a Jersey `Client` object which is there as a convenience for making API calls to the running Dropwizard application under test.  This `Client` object was implemented directly in `DropwizardAppExtension` rather than putting it in `DropwizardTestSupport` and 'relaying' from `DropwizardAppExtension` to it.

While I don't really know for sure, I **have** to think this was accidental.  Probably when the `Client` was added to `DropwizardAppExtension` the author forgot or didn't know about the intention to offer feature-parity in `DropwizardTestSupport`?

Am planning to raise a PR to move the Client implementation to DropwizardTestSupport and continue to expose it transitively in `DropwizardAppExtension`.

accwebs:
Fixed by #11308

PR #11308, opened 2026-09-02
Move the Client instance out of DropwizardAppExtension into DropwizardTestSupport

accwebs:
Issue #11304: Move the Client instance out of `DropwizardAppExtension` into `DropwizardTestSupport` so it can be used via either class.

###### Problem:
See #11304.

###### Solution:
1. I've directly moved `Client` to `DropwizardTestSupport`.
2. I've retained 'relay' methods in `DropwizardAppExtension`.
3. Doing a "direct" relay of `clientBuilder()` is difficult unless `DropwizardTestSupport` is converted from protected to public scope.  I therefore have made it public in both cases and added some Javadocs.  It can be used via the old pattern (subclassing and overriding), or via a new pattern of direct invocation.

Tweak along the way:
* Added `getClient()` method which is an alias for `client()`.  My IDE (IntelliJ) by default prints warnings for client logic invoking methods that return `AutoCloseable`/`Closeable`, if the calling logic doesn't close them.  In this case the caller should NOT close the result of client() since it is shared.  Switching to a `getClient()` name causes the IDE warning to be suppressed without a special rule.  So this feels helpful as a linting/IDE hint if callers prefer that.

### DIFF
--- a/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
+++ b/dropwizard-testing/src/main/java/io/dropwizard/testing/DropwizardTestSupport.java
@@ -11,14 +11,23 @@ import io.dropwizard.core.cli.EnvironmentCommand;
 import io.dropwizard.core.cli.ServerCommand;
 import io.dropwizard.core.setup.Bootstrap;
 import io.dropwizard.core.setup.Environment;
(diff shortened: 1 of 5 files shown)


pstackle (approved):
Thanks @accwebs for your contribution!

PR #9970, opened 2025-03-20
fix: Allow access log to log paths jetty rejects

josephlbarnett:
Tentative fix for #9773

Extends #9788 to maintain duplicate header handling while we wait for https://github.com/qos-ch/logback-access/pull/23

PR #11309, opened 2026-09-02
Resurrect @zUniQueX's approach of changing Logback request log implementation (drop logback-access-jetty12 and AccessEvent) - fixes user logging & other logging bugs

accwebs:
Issue #11303: Resurrect @zUniQueX's approach of changing Logback request log implementation

This is intended to replace PR #10210.

**Now ready for review.**

###### Problem:
See #11303. Prior issues #9970, #8008.

###### Solution:
1. Relocate existing "auth logging test" (not actually doing auth logging testing) to a new `AuthIntegrationTest` class.
2. Add extensive integration tests for classic and Logback Access request log implementations:
   * `CustomRequestLogPatternIntegrationTest` is now merged with `RequestLogPatternIntegrationTest` to form a combined (and largely rewritten) `LogbackAccessRequestLogIntegrationTest`.  Similar rename for `AbstractRequestLogPatternIntegrationTest` -> `AbstractRequestLogIntegrationTest`.
   * Don't rely on nondeterministic/timing behaviors in the request log tests.
   * Consolidate `CustomRequestLogPatternIntegrationTest` into `LogbackAccessRequestLogIntegrationTest`.
   * Remove `@DisabledOnOs(OS.WINDOWS)` for logging tests
        * Ran the build on Windows and don't experience any `TempDir` problems so I doubt this is still a concern.
    * All the new tests definitely cross the line somewhat into testing Logback Access itself (vs Dropwizard).
    * BUT, we've repeatedly seen Logback bugs causing major issues over the years, so having extensive tests seems warranted.
    * PLUS I needed a way to evaluate what might break when we switch to an alternative logging approach.
3. Entirely replace the `IAccessEvent` and `RequestWrapper` aspects of the existing Logback Access implementation:
    * Resurrect @zUniQueX's approach of creating a correctly-implemented `IAccessEvent` class (`JettyAccessEvent`) instead of the ill-advised `AccessEvent` implementation in Logback Access.  This drops the `logback-access-jetty12` library entirely for now.  Logback Access is itself retained however a reduced set is used (as mentioned `AccessEvent` is not salvageable for now, thus is no longer used).
    * Tweak zUniQueX's approach to fix various bugs:
       * (New bug) Don't remove `LogbackAccessRequestLogAwareHandler` from the new design. This allows the decorated `HttpServletRequest` to flow back to the `log()` method.  Without this, things like session ID retrieval and esoteric `getRequest()` `getResponse()` getters do not work right.
       * (New bug) Use case-insensitive `Map` for HTTP headers.
       * (Cleanup of introduced) Fix sequenceNumber to be final.
       * (New bug) Fix `getRequestContent()` and `getResponseContent()` to be usable in the Logback Access pattern (with TeeFilter registration) rather than throwing (which is a behavior change and also hard breaks the logging).
       * (Cleanup of introduced) Improve `getServerAdapter()` to proactively build & re-use the returned `DropwizardJettyServerAdapter` instance instead of creating each time.
       * (Existing bug) Fix `getResponseHeader()` to return "-" when header is not present; for logging consistency with `getRequestHeader()`.  This was a bug/inconsistency in the prior implementation but IMO we might as well fix it now.
       * (Existing bug) Fix `getRequest()` and `getResponse()` getters to work. Esoteric use case at best, but might as well.
       * (Existing bug) Fix thread name request log logging to work.  I opted to log the request thread name, not the background logging thread name, which I think is reasonable.
    * Hopefully we can upstream this at some point.  Not optimistic given that https://github.com/qos-ch/logback-access/pull/23 is still pending which was opened back in the #9970 days.  Long time ago.

Fix along the way:
4. Fix `dropwizard-request-logging` module's incorrect scope of `jetty-security` and `jetty-session` (transitively relied upon by `jetty-ee10-servlet`).

###### Result:
1. Request logging hopefully works reliably/correctly now regardless of field. Hopefully.
   * user ID logging is fixed
   * thread name logging is fixed
   * request/response content logging - when enabled - now honors character encodings
   * Ability to extract values synchronously from the AccessEvent via getRequest/getResponse is fixed per the intended Logback Access design of some esoteric features.
2. Anyone logging HTTP response headers will see a "-" instead of "" (blank) if not set.
3. Anyone using Logback Access's spaghetti/ill-advised request body [form-url-encoding custom logging feature](https://github.com/qos-ch/logback-access/blob/d6fd5fe394a355d2bfad47bc175b014f3506138e/logback-access-common/src/main/java/ch/qos/logback/access/common/spi/AccessEvent.java#L530) will see that stop working for them.
4. Anyone using Logback Access's spaghetti/ill-advised response body [logging suppression for certain types](https://github.com/qos-ch/logback-access/blob/d6fd5fe394a355d2bfad47bc175b014f3506138e/logback-access-common/src/main/java/ch/qos/logback/access/common/spi/AccessEvent.java#L585) will see that stop working for them.
5. Anyone directly depending on `dropwizard-request-log` module won't be missing the required `jetty-security` and `jetty-session` transitives.  Not that `jackson-databind` remains incorrectly scoped.  Tracked in #11319 

###### Further work:
1. Integration tests for custom logging.
2. Integration tests for JSON access logging.
3. Adding substantial tests to the new `AuthIntegrationTest` class would seem warranted.  All I did here was port over the existing test to the right place based on what it was doing.

----

BEFORE MERGING I will squash my changes built on @zUniQueX 's resurrected commit into one commit.  I'd propose a final merge of 4 commits:

1. Move RequestLogPatternIntegrationTest.testRemoteUserIsSetCorrectly() to a new AuthIntegrationTest class. It's not testing the request log. 
2. Rename files without contents changes (to aid with history tracking):
3. Redesign/rewrite request logging integration tests
4. Combined commit of @zUniQueX's changes + my changes made to those + threadname fixes.

The idea is to make different aspects of this changeset more revert-able/reviewable in the future.


dropwizard-e2e/src/test/java/com/example/request_log/LogbackAccessRequestLogIntegrationTest.java
```diff
@@ -0,0 +1,2325 @@
...
+            assertThat(resultingLogLine.get("userAgent"))
+                .isEqualTo("kitchen-sink-agent");
+        }
+
+        @Test
+        void userAgent_isDashWhenMissing() throws Exception {
+            // when
+            try (Socket clientSocket = new Socket("localhost", testSupport.getLocalPort())) {
+                // bypass HTTP client restrictions where they force a user agent to be supplied
+                PrintWriter socketPrinter = new PrintWriter(clientSocket.getOutputStream());
```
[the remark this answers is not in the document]
accwebs: PrintWriter is flushed and Socket object being closed closes the output stream.  So I think this is a non-issue unless we prefer to close() it just because?  I'm personally not a fan of closing decorators to Input/OutputStreams (they usually close the underlying resource but it's not always clear, and in this case I just wanted to close the Socket instead of shutting down output stream).

dropwizard-e2e/src/test/java/com/example/request_log/AbstractRequestLogIntegrationTest.java
```diff
@@ -0,0 +1,424 @@
...
+        @Path("/authenticated")
+        public String getAuthenticatedUser(@Auth PrincipalImpl principal) {
+            return String.format("Hello, %s!", principal.getName());
+        }
+
+        @POST
+        @Path("/submit")
+        @Produces(MediaType.APPLICATION_JSON)
+        @Consumes(MediaType.APPLICATION_JSON)
+        public String submitSomething(@NotNull @Valid PostParameter parameter) {
```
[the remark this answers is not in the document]
accwebs: Largely exists so we can send a POST request for the logging tests

accwebs:
@zUniQueX if you have a few minutes, would appreciate a sanity-check from you on this. Specifically, maybe eyeball the changes I made to "your commit" (which I tried to resurrect faithfully and then update via additional commits so people can track it during the review).  Am I off base in any of the changes I made?  Anything that feels suspicious?

zUniQueX:
@accwebs I haven't looked at the tests, but the changes to the code look reasonable. Exposing the servlet contexts however could cause strange behavior as you've already denoted in the comments.

Generally, I do like the idea of staying compatible with logback access. But I think the maintenance overhead is too high for us on that. My PR was only a POC and I think it's not a bad way to overcome logback's problems. But in this situation we have to review this every time Jetty or Logback bring updates on their related components. (The Logback changes should be negligible, but Jetty will move forward).

If you have time to dive into this, further work on the topic would be appreciated. Else we probably have to re-evaluate dropping support for logback-access.

accwebs:
Hi @zUniQueX !

> @accwebs I haven't looked at the tests, 

That's fine.  I mainly was interested in what you gave me - an opinion of 'should we do this, and if so, how far should we go'.  Thank you.

> Generally, I do like the idea of staying compatible with logback access. But I think the maintenance overhead is too high for us on that. 

This is exactly the type of feedback/review I was looking for <3 .... basically, I'm pretty sure my code is correct (certainly it's more correct than logback's impl), but **correct** code does not mean **wise**.  In your opinion what's the best way to proceed?  I see a few options:

1. 'move out' with this PR as-is
    * Of course with any implementation-level feedback and history cleaned up 
2. move out with it generally speaking, but back out/drop certain compatibility items I added (to hopefully gain some future reduced maintenance effort)?
    * Which items would we want to change/deviate from Logback on, in your opinion?
3. Hold on this PR for now; instead, try to spot-patch some of the most egregious Logback access bugs like we're doing with the HTTP header stuff.  
    * Later we could separately explore a follow up to eliminate logback-access (or possibly Logback) entirely?
4. Hold on this PR for now; instead, try to more systematically replace Logback Access right now.

My thinking is we do some variant of 1 or 2 now, and then we explore 4 as a follow up item.  But I am plagued with doubt.  So any thoughts are most welcome :-) 

> My PR was only a POC 

I know!  But honestly, it was a good POC and quite close to being done already!  Having attempted a few other options myself (and hitting serious roadblocks), your approach seemed most solid hence why I used it as a foundation and basically just added tons of tests and then made minor tweaks to pass the tests.

> If you have time to dive into this, further work on the topic would be appreciated. Else we probably have to re-evaluate dropping support for logback-access.

I have concerns about the future sustainability of Logback, particularly Logback Access.  While the Logback and SLF4J maintainers have contributed enormous value to the Java ecosystem, my observation over the years has been that maintaining compatibility and resolving deep-seated concurrency/lifecycle issues within Logback has been increasingly difficult.  SLF4J of course remains amazing.

I hope the state of the world improves; however, as things stand, I too believe we need to seriously consider dropping Logback Access.  Possibly dropping Logback entirely if we can find something better.  

Logging is a frustrating and annoying problem space; I'm not sure I could do better but at the same time I must be honest about what I see.  Maybe we could try to participate in the project in a more serious way.  I certainly have been remiss and should have helped years ago (easy to complain, but then if I'm not part of the solution, the fault is mine).  But of course in parallel, Dropwizard should also be considering alternatives that could work for it.  Maybe a switch could be done in a future major release. I'm torn.

zUniQueX:
@accwebs Thanks for the detailed feedback.

I don't feel like doing this change in a patch or minor update is a good option. Nevertheless, leaving it with the current state isn't a good option too.

My concern is whether other users of logback and jetty have the same problems like we have - although they're not using dropwizard - or if that's just a problem for us. Do you think moving out the logback access log to a separate module/repo like the throttling appender would be a good idea?

I personally don't know, if we'd want to ship this ASAP or wait for the next major release. If anybody extends the classes, this would break the API, which we certainly would't want until we push a new major release.

If we hold this change back for now, we could directly discuss dropping logback access support with the next major release. But this would need to revise some other components like JSON logging.

accwebs:
I mean, we need to fix the logging now somehow, ASAP.  So we have I think 3 general options:
1. Use the approach here.  Namely, we still use logback-access but none of it's Jetty binding logic.  Instead we inject our own custom classes for capturing the request attributes.  Yes, there is some risk of breakage but best I can tell that's pretty minimal. 
    * Subsequent major versions might scrap this work and do something entirely different.
2. Say approach 1 is too heavy-handed and keep using overriding/subclassing tricks to fix up the RequestWrapper and AccessEvent impls.  To me, this is less sustainable than 1 (given the plethora of bugs) hence why I rejected it.  But it is a realistic option, probably.
3. Immediately begin planning toward a 6.x release (and EOL'ing 5.x) where we drop logback-access entirely, possibly more of Logback.  If that's the intent, we'd might validly argue that doing 1 is pointless (wasted) effort.  
   * Or at that point, maybe we contribute to Logback and fix the problems there.

Of course, no matter what we pick now, we still can and I expect we would change course later.

dropwizard-e2e/src/test/java/com/example/request_log/AbstractRequestLogIntegrationTest.java
```diff
@@ -0,0 +1,428 @@
...
+
+// Conventions for derived classes:
+//
+// 1. The abstract base is @TestInstance(PER_CLASS) and @Execution(SAME_THREAD). Concrete subclasses inherit both.
+//
+// 2. In each nested class that owns an app:
+//      @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+//      @Nested
+//      class MyNestedApp {
+//          private final DropwizardTestSupport<Configuration> testSupport = createAppTestSupport(...);
```
accwebs, line 72: Something I'd like to call out here -- so with #11318 done now, it would be theoretically possible to pivot these tests all away from `DropwizardTestSupport` back to `DropwizardExtensionsSupport` / `DropwizardAppExtension`.  I did make a local attempt at this, but I found annoying ordering problems remained related to the temp dir management and similar.  It probably WOULD be possible to work around those issues with enough effort, but I figured just keep the use of `DropwizardTestSupport` here since we support both, and the whole reason `DropwizardTestSupport` exists is to give you more flexibility of JUnit's behavior doesn't work for you.

pstackle:
@zUniQueX 
> But in this situation we have to review this every time Jetty or Logback bring updates on their related components. (The Logback changes should be negligible, but Jetty will move forward).

I'm not sure I follow what you mean here. If `logback-access` makes edits to the `IAccessEvent` interface, we'd get a compile-time failure until we address the gaps. If Jetty makes structural API changes, the same thing would happen, and if Jetty makes more subtle changes, the tests will hopefully detect if a behavior we care about changes. 

This (merging this PR) seems like a more preferable state to be in than before with how easy it was to let a regression in to DW v5 by just letting `logback-access`'s  jetty 12 integration silently break things.


> I don't feel like doing this change in a patch or minor update is a good option. Nevertheless, leaving it with the current state isn't a good option too.

The loss of support to log the username of the request in Dropwizard v5 is too severe of a regression to not fix. I'm open to different approaches, but I think this is definitely in the category of a bug that needs to be fixed and waiting for a major version update isn't tenable. 


> Do you think moving out the logback access log to a separate module/repo like the throttling appender would be a good idea?

I think we could incubate this within the `dropwizard-request-logging` module as a starting place, and if there is a compelling need for it to exist somewhere else later, we could migrate it. I don't really think it would be a good idea to advocate to have non-Dropwizard projects depend on it, because I think that could sign us up for more support work than it'd be worth. In an ideal world, the `logback-access` project would address the bugs and we go back to letting that project handle request logging functionality, but I'm not sure how quickly we could expect that to happen.

> If anybody extends the classes, this would break the API, which we certainly would't want until we push a new major release.

Which classes are you talking about? The existing ones that extend/build upon `logback-access-jetty12`? Or the new ones that @accwebs is adding (by resurrecting your PoC)? 

If the former, and if they were extending them to fix issues that we are now addressing, maybe those extended classes were only needed to fix the implementation we shipped. If they were extending them to add other/different functionality, I don't think it would be a huge lift for them to either port their customizations to the new approach or to bring more of the old functionality over and continuing to build on `logback-access-jetty12` themselves.


@accwebs and @zUniQueX: both of you make reference to dropping `logback-access` entirely. Unless we were able to identify a higher quality access/request logging library, which I'm not aware of one, I think it would be a tall ask to need to completely roll our own request logging library. Going back to the pre logback-access days, with very minimal logging configuration support is not really an option.

accwebs:
>@accwebs and @zUniQueX: both of you make reference to dropping logback-access entirely. Unless we were able to identify a higher quality access/request logging library, which I'm not aware of one, I think it would be a tall ask to need to completely roll our own request logging library. Going back to the pre logback-access days, with very minimal logging configuration support is not really an option.

Indeed.  I am not sure what to suggest but we could start looking.  But yes, this would be a much longer-running action and a bit of research project.  I also am not 100% sure I want to give up on Logback (in the long term) vs just trying to become an active maintainer and fix it.

Thanks for all your feedback @pstackle

zUniQueX:
@pstackle

> I'm not sure I follow what you mean here. If logback-access makes edits to the IAccessEvent interface, we'd get a compile-time failure until we address the gaps. If Jetty makes structural API changes, the same thing would happen, and if Jetty makes more subtle changes, the tests will hopefully detect if a behavior we care about changes.

> This (merging this PR) seems like a more preferable state to be in than before with how easy it was to let a regression in to DW v5 by just letting logback-access's jetty 12 integration silently break things.

Indeed. This would be a benefit for the users. But then we'd have to remove `logback-access-jetty12` entirely and write our own request logging component (which we do with this PR). We won't get any updates from the logback community to our implementation and would have to look at what they're doing to decide, if some changes of logback should be ported to our codebase.

> The loss of support to log the username of the request in Dropwizard v5 is too severe of a regression to not fix. I'm open to different approaches, but I think this is definitely in the category of a bug that needs to be fixed and waiting for a major version update isn't tenable.

Agreed. But then we should reduce the breaking changes to the minimum. The current state of this PR removes a managed dependency and this could break something downstream. So we should at least keep the managed dependency until we do a major release.

> In an ideal world, the logback-access project would address the bugs and we go back to letting that project handle request logging functionality, but I'm not sure how quickly we could expect that to happen.

I don't know, if this will happen at all. Over the past releases the major servlet versions were all integrated very slowly. If the next EE release will bring Servlet 7.0, logback will probably take a longer time to integrate the API changes.

> Which classes are you talking about? The existing ones that extend/build upon logback-access-jetty12?

I was talking about the classes currently shipped with the framework. If anyone makes own extensions or fixes (or both), this would break the implementation. However, if the current implementation is broken, this could make the changes of users obsolete. So the migration path would be easy by deleting the extended classes.

> I think it would be a tall ask to need to completely roll our own request logging library.

This would probably be very high maintenance cost for us and I don't think we currently would want to do this. If anyone from the community like @accwebs was interested in helping to maintain this, it could be feasible however.

accwebs:
Thanks.  Struggling to understand the proposed way ahead overall.  Merge this PR for now?  Do something else?

---

One notable point I captured from your latest comment is that maybe we should keep `logback-access-jetty12` on the dependency tree just to avoid breaking anyone with downstream code using it (assuming it will stay on the dependency tree in the patch release).

Assuming I understand that right, I am happy to adjust the PR to do that.

accwebs:
I have now rebased this commit.  Notable changes:

1.  Since #11325 is now merged, eliminated dependency scope changes that were now handled as part of that PR.  
2. `ch.qos.logback.access:logback-access-jetty12` dependency is re-added to the dependency tree to try to avoid the small possibility of downstream breakage. Added a `maven-dependency-plugin` ignore rule for that lib with a comment attached.
3. I realized my approach to call setThreadName() to capture it was logically inconsistent with the general approach of `JettyAccessEvent` for everything else.  Admittedly that attribute is a tad different due to `IAccessEvent` declaring `setThreadName()` but I have now changed the logic to capture thread name during `prepareForDeferredProcessing()`, which I think is more sane than before.

dropwizard-request-logging/pom.xml
```diff
@@ -53,6 +53,10 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <dependency>
```
accwebs, line 56: I'm at a loss why `maven-dependency-plugin` complains *now* about this, given that jspecify is already referenced by `SafeRequestParameterConverter.java`.  My guess is the way the annotation is used in that class is not detectable by maven-dependency-plugin, whereas `JettyAccessEvent` does now use it in a way where it gets picked up.  Anyway, this solves the error that now gets generated.

dropwizard-request-logging/pom.xml
```diff
@@ -121,4 +125,28 @@
...
+                    <execution>
+                        <id>analyze</id>
+                        <configuration>
+                            <ignoredDependencies>
+                                <!-- Issue #11303: We've replaced all use of logback-access-jetty12 for now to work -->
+                                <!--               around various request log bugs in Logback Access. But we have   -->
+                                <!--               deliberately kept this unused library on the dependency tree for -->
+                                <!--               DW 5.x to avoid a shift in the dependency tree that could break  -->
+                                <!--               existing 5.x users.                                              -->
+                                <ignoredDependency>ch.qos.logback.access:logback-access-jetty12</ignoredDependency>
```
accwebs, line 144: I can still be 'talked out' of this.  I'm torn.  It HAS occurred to me of late that maybe if I'm a downstream user of DW 5.x and I somehow rely on this on my dependency tree due to some request log customization, I might **prefer** to have my app hard break vs silently change its operation.

PR #11316, opened 2026-09-09
Fix dropwizard-request-logging module's incorrect scope of jetty-security and jetty-session

accwebs:
Caught while I was working on #11303 / #11309: Fix `dropwizard-request-logging` module's incorrect scope of `jetty-security` and `jetty-session` (transitively relied upon by `jetty-ee10-servlet`).

###### Problem:
At present, `dropwizard-request-logging`'s dependency scope for `jetty-security` and `jetty-session` are being manually overridden to `test` when `compile` scope is required (since `jetty-ee10-servlet` is compile scope, and depends on these modules).

###### Solution:
1. Fix the scope.
2. Override the erroneous output of `maven-dependency-plugin` for these modules.

###### Result:
Anyone directly depending on `dropwizard-request-log` module won't be missing the required `jetty-security` and `jetty-session` transitives.

### DIFF
--- a/dropwizard-request-logging/pom.xml
+++ b/dropwizard-request-logging/pom.xml
@@ -73,6 +73,14 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-security</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-session</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


accwebs:
In discussion with @pstackle we prefer to try to fix the problem 'across the codebase' rather than just a partial fix (which is still even leaving the similar `jackson-databind` scope issue in `dropwizard-request-logging` unfixed here in this PR).

I have captured the overall problem and more context in #11319.

Because #11309 depends on this particular change I will simply add the changes here as part of the same #11309 PR.

Issue #11310, opened 2026-09-02
DropwizardAppExtension works incorrectly when used with PER_CLASS @TestInstance


accwebs:
Something hit when I was working on integration tests in #11309.

I believe `DropwizardExtensionsSupport`/`DropwizardAppExtension` works incorrectly/inapprorpriately when used with Junit Jupiter's `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` option.

## How it behaves right now

Case 1 (CORRECT BEHAVIOR): `@BeforeEach`/`@AfterEach` start/stop when `DropwizardAppExtension` member is non-static:

```
@ExtendWith(DropwizardExtensionsSupport.class)
class TestMyAppIT {
    // starts and stops once per test method (matching JUnit's instantiation of TestMyAppIT, one per test method)
    private final DropwizardAppExtension<Configuration> app = new DropwizardAppExtension<>(TestApplication.class);

    @Test
    public void test1() {

    }

    @Test
    public void test2() {

    }
}
```

This makes sense and lines up with JUnit's per-test instantiation of `TestMyAppIT`.  However, when there are lots of tests, this will be quite slow since starting/stopping the application per test method is time consuming.

---

Case 2 (CORRECT BEHAVIOR): `@BeforeAll`/`@AfterAll` start/stop when `DropwizardAppExtension` member is static:

```
@ExtendWith(DropwizardExtensionsSupport.class)
class TestMyAppIT {
    // starts and stops once for the whole test class 
    // (JUnit is still instantiating TestMyAppIT once per test method, but this is static so this does not matter)
    private static final DropwizardAppExtension<Configuration> app = new DropwizardAppExtension<>(TestApplication.class);

    @Test
    public void test1() {

    }

    @Test
    public void test2() {

    }
}
```

This still makes sense.  This will save much execution time provided all tests in the class can use the same configuration of tested application.  

However, the problem with requiring `static` is it prevent the use of any OOP in one's test classes since static effectively breaks dynamic binding / V-tables thus goodbye the benefits of OOP.  Still, this is often enough for most test cases.

---

Case 3 (INCORRECT BEHAVIOR, IMO):  Use of PER_CLASS `@TestInstance`

What is PER_CLASS?  Well, to solve the problems with using `static` to avoid extra work in integration tests, a third option now offered as of JUnit Jupiter (wasn't supported in 4.x) is an alternative test class lifecycle.  This can also be configured (or not) in `@Nested` classes.

```
[...]

PR #11318, opened 2026-09-09
Rewrite DropwizardExtensionsSupport to behave properly in various respects (PER_CLASS, more reliable after() behavior, etc)

accwebs:
###### Problem:
See #11310.  Note that there are now 5 separate bugs identified on that ticket's subsequent discussion:

1. Works incorrectly/inappropriately when used with JUnit Jupiter's `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`
2. Multi-field iteration order doesn't follow JUnit conventions; declaration order and `@Order` annotation are ignored
3. When one field's `before()` throws, subsequent fields' `after()` methods are still called
4. When one field's `after()` throws, subsequent fields' `after()` methods are NOT called
5. `beforeAll` / `afterAll` walk enclosing classes, causing double-firing of outer statics

###### Solution:
1. Add extensive tests for `DropwizardExtensionSupport` (which helped detect issues 2-5).
2. Rewrite `DropwizardExtensionSupport` till the tests all pass :-)

###### Result:
Behavior hopefully is now correct / sync'd up with JUnit expectations.

###### Risks:
Various "breaking" behavior changes from the programmer's perspective. 

However I would argue the new behavior is correct and what people would intuitively expect/want.


accwebs:
Appveyor build failed due to a transient build runner error best I can tell.  I think this PR remains ready to merge.

pstackle (approved):
Thanks @accwebs for the contribution!

PR #11325, opened 2026-09-14
Issue #11319: Dependency tree cleanups (fixing scopes, removing pointless exclusions, etc)

### LINKED ISSUE #11319: Various dependency scope mistakes (incorrect "test" scope) across the codebase (driven by maven-dependency-plugin failures/errors)
I'm filing this issue to track the larger "dependency scope" problem I stumbled on and attempted to do a partial fix of one specific case in #11316.  After talking with @pstackle we prefer to try to solve the problem across the board rather than playing whack-a-mole.

In a number of cases, I have identified various `<dependency>` blocks where `<scope>test</scope>` is declared, overriding a compile-time transitive dependency (in non-Maven lingo, a "runtime" dependency) to a `test`-scope dependency simply to satisfy the build failure error generated by `maven-dependency-plugin`.  The result is that the tests pass, but anyone relying on Maven's dependency resolution for direct references to an affected module is likely going to get `NoClassDefFoundError`s at runtime.

Current known examples:
* dropwizard-request-logging
   * `jetty-security` and `jetty-session` (transitively relied upon by `jetty-ee10-servlet`, which is in compile scope).
   * `jackson-databind` (transitively relied upon by `dropwizard-jackson`, which is in compile scope)

There will likely be many across the codebase.

The fundamental problem with `maven-dependency-plugin` is tracked here:
* https://issues.apache.org/jira/browse/MDEP-791
* https://github.com/apache/maven-dependency-plugin/issues/1297

Both issues are closed out as basically WON'T FIX.  They claim the problem is in Maven itself and point at https://github.com/apache/maven/issues/7682.  I tend to disagree (IMO, the plugin simply needs to walk the dependency tree to filter out false 'test scope' claims, but I admit I'm no expert about Maven internals.  Maybe there is some implementation issue that prevents that.)

Anyway, there will not be any fix for this anytime soon.  At present, the only workaround recommended is to set:
```
<ignoredNonTestScopedDependencies>
   <ignoredNonTestScopedDependency>*</ignoredNonTestScopedDependency>
</ignoredNonTestScopedDependencies>
```

If I get consensus from everyone, I'm happy to take an action to:
1. Set that workaround across the project.
2. Audit/fix all the existing 'incorrect scope' issues in the repo right now.

accwebs:
## Problem:

As reported/discussed on issue #11319, this PR's primary goal is to correct a class of accumulated dependency-scope mistakes across Dropwizard's Maven pom.xml's.  We also work around a known limitation with `maven-dependency-plugin` which fails the build when a transitive dependency demands `compile` scope, but only the local `test`-scope code directly imports it (see https://issues.apache.org/jira/browse/MDEP-791, apache/maven-dependency-plugin#1297, and apache/maven#7682).


## Solution:

### Quick Summary of Changes

I fixed the scope of various transitive dependencies back from `<scope>test` to `<scope>compile`.

To do this work, it was necessary to look at various aspects of the dependency tree, especially to prevent a situation where I might have re-introduced dependencies that were previously `<exclude>`'d transitively.  To do that, I had to make sense of each **existing** exclusion rule and why it was there.

In "making sense of the exclusion rules", I realized many are unnecessary.  Thus, I opportunistically made various related cleanups/improvements:
 * Auditing/removing obsolete `<exclusion>` rules.
 * Adding a few items that could remain concerning (but no longer needed exclusions) back as `bannedDependency` enforcer rules
 * Dropping two `<ignoredDependency>` entries (`org.jspecify:jspecify`, `io.dropwizard.logback:logback-throttling-appender`) which were empirically no-ops with the current plugin version.

As part of this work, `<ignoredNonTestScopedDependencies>*</...>` is now set at the parent POM level to avoid false `maven-dependency-plugin` scope errors failing the build.

In addition to adding the `ignoredNonTestScopedDependencies` config, I corrected some configuration issues with `maven-dependency-plugin` (config blocks being globally set vs being set on a per-execution level).

### Detailed List of Changes

#### 1. Scope corrections (54 promotions across 19 modules)

Promoted `<scope>test</scope>` declarations to `<scope>compile</scope>` where the coord was also reachable via a compile-scoped transitive. This makes the declared scope match the effective downstream scope.

**Modules affected**: `dropwizard-auth`, `dropwizard-client`, `dropwizard-core`, `dropwizard-e2e`, `dropwizard-forms`, `dropwizard-health`, `dropwizard-hibernate`, `dropwizard-http2`, `dropwizard-jdbi3`, `dropwizard-jersey`, `dropwizard-json-logging`, `dropwizard-metrics`, `dropwizard-migrations`, `dropwizard-request-logging`, `dropwizard-testing`, `dropwizard-validation`, `dropwizard-views`, `dropwizard-views-freemarker`, `dropwizard-views-mustache`.

#### 2. maven-dependency-plugin config changes / analyzer accommodation for (1)

##### 2.a. ignoredNonTestScopedDependencies

Added `<ignoredNonTestScopedDependencies>*</...>` to the root pom's `analyze-only` config to silence this class of warning (which does not work correctly, today) globally.

##### 2.b. Removal of now-pointless ignoredDependency entries

Dropped two `<ignoredDependency>` entries (`org.jspecify:jspecify`, `io.dropwizard.logback:logback-throttling-appender`) which were empirically no-ops with the current plugin version.

##### 2.c. Promoted maven-dependency-plugin "global" configs to their relevant per-execution config blocks

Fixed that parent pom's `maven-dependency-plugin` configs resolving `byte-buddy-agent` were defined globally for all executions rather than the `download-byte-buddy-agent-jar` execution.

Fixed that `dropwizard-migrations` defined a module-specific `ignoredUnusedDeclaredDependency` config but this was done at the global level rather than the `analyze` execution.

#### 3. Obsolete `<exclusion>` removals (63 total)

##### 3.a. Where did these (removed) rules all come from?

Three root causes of the 'build-up' of pointless rules (caveat: my best guess OFC):

**Upstream upgrades** - coords that current versions of Jetty 12, Jersey 3.x, HK2 3.x no longer produce or transitively pull:
- `org.glassfish.hk2.external:jakarta.inject` (34 removals) - HK2 now uses `jakarta.inject:jakarta.inject-api` directly.
- `org.eclipse.jetty.toolchain:jetty-jakarta-servlet-api` (15 removals) - Jetty 12 replaced it with `jakarta.servlet:jakarta.servlet-api`.
- `org.eclipse.jetty.toolchain:jetty-servlet-api` (2 removals) - older Jetty toolchain name.
- `junit:junit` (4 removals) - Jersey 3.x / `dropwizard-testing` no longer transitively pull JUnit 4.

**`javax` → `jakarta` migration leftovers** - exclusions that predate the project's migration and no longer apply:
- `javax.servlet:javax.servlet-api` (2 removals)
- `javax.activation:javax.activation-api` (1 removal)

**Exclusions now redundant with BOM management (`slf4j-bom` import) or Maven scope rules (`provided`-scope deps don't propagate)** - most of the dropwizard-dependencies/pom.xml changes
- `logback-classic` → `slf4j-api`
- `liquibase-slf4j` → `slf4j-api`
- `liquibase-slf4j` → `liquibase-core`
- `jetty-setuid-jna` → `jetty-server`
- `guava` → `com.google.code.findbugs:jsr305` (Guava 33.4.6+ no longer declares it)

##### 3.b. Coupled cleanup in `dropwizard-http2`

Alongside removing the `jetty-jakarta-servlet-api` exclusions, dropped an explicit `<dependency><scope>runtime</scope>` for `jakarta.servlet-api`. That declaration existed only to compensate for the exclusion cutting off Jetty's transitive path to a servlet API. With the exclusion gone, Jetty 12 naturally provides `jakarta.servlet-api:6.0.0:compile` transitively via `jetty-server`. Both pieces of the obsolete pattern removed together.

This is the source of the one scope widening in the classpath diff: `jakarta.servlet-api` moves from `:runtime` to `:compile` for `dropwizard-http2` consumers. Strictly more permissive - cannot break anything.

##### 3.c. Exclusions retained (table shows AI-generated summary when I was beating it into a corner on each dependency)

I'm not 100% sure about the `commons-collections4` one to be honest, but at the same time the origin of that rule was quite deliberate and until I understand the rationale of the author on it, I don't want to touch it.  Especially when the way it is right now is clearly "fine" albeit requires a custom exception rule for maven-dependency-plugin.

| # | Pom                     | Parent                                                       | Excluded                                      | Verdict             | Rationale                                                                                                                                                                                                                                                                                                                                                                         |
|---|-------------------------|--------------------------------------------------------------|-----------------------------------------------|---------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 1 | dropwizard-dependencies | `org.liquibase:liquibase-core`                               | `org.yaml:snakeyaml`                          | 🔴 **LOAD-BEARING** | liquibase 4.33.0 brings `snakeyaml:2.4`; `jackson-dataformat-yaml:2.22.2` brings `snakeyaml:2.5`. Removing exclusion → `dependencyConvergence` fails                                                                                                                                                                                                                              |
| 2 | dropwizard-hibernate    | `com.fasterxml.jackson.datatype:jackson-datatype-hibernate6` | `jakarta.transaction:jakarta.transaction-api` | 🟡 **DEFENSIVE**    | Coord arrives via `hibernate-core` in current tree. `jakarta.transaction-api` is NOT in `<dependencyManagement>` anywhere, so version control depends entirely on nearest-wins mediation. Exclusion prevents `jackson-datatype-hibernate6` from being an alternate source if the two libs diverge in future - ensures `hibernate-core` remains the sole authority for the version |
| 3 | dropwizard-migrations   | `org.liquibase:liquibase-core`                               | `javax.xml.bind:jaxb-api`                     | 🔴 **LOAD-BEARING** | `javax.xml.bind:jaxb-api` is in the `bannedDependencies` list in `dropwizard-parent/pom.xml:166` (replaced by `jakarta.xml.bind:jakarta.xml.bind-api`). Removing exclusion → banned-dep failure                                                                                                                                                                                   |
| 4 | dropwizard-migrations   | `org.liquibase:liquibase-core`                               | `org.apache.commons:commons-collections4`     | 🟡 **COSMETIC**     | Redundant with the direct top-level declaration + BOM management (the "version ownership" pattern). Effective classpath unchanged either way. Kept for now; whether to drop the exclude+redeclare pattern project-wide is a separate discussion                                                                                                                                                                                                                     |

#### 4. maven-enforcer-plugin `bannedDependencies` additions

Two entries added to `dropwizard-parent/pom.xml`'s enforcer config, replacing per-module exclusion patterns with centralized project-wide protection:

- `org.eclipse.jetty.toolchain:jetty-jakarta-servlet-api` - replaced by `jakarta.servlet:jakarta.servlet-api`; Jetty 12 dropped it.
- `com.google.code.findbugs:jsr305` - use `org.jspecify:jspecify` instead. Also blocks a future transitive reintroducing an abandoned coord.

## Verification

Beyond the standard `mvn verify`, I asked Claude to perform various checks of the before/after dependency tree, giving it a bunch of guidance.  It performed a programmatic runtime-classpath diff across all 42 modules using `mvn dependency:list -DincludeScope=runtime` (which produces the exact set of coords on the runtime classpath - compile + runtime scopes - that downstream consumers receive).

**Full aggregate results:**

- 19 modules had additive changes to their runtime classpath (168 total additions).
- Every added artifact traces cleanly to an intended scope promotion or its expected transitive cascade.
- **Zero artifacts removed** from any module's runtime classpath.
- **Zero version changes** anywhere in the tree.
- **Zero scope narrowings**.
- **One scope widening** (dropwizard-http2 `jakarta.servlet-api: runtime → compile`, explained above).

## Result:

1. Anyone importing one or more specific dropwizard modules without the dropwizard-dependencies BOM should get a functional dependency tree now.
2. Future builds won't complain that a `compile`-scope dep needs to be `test`-scope when that's not the case.  We do however lose detection for cases where an unnecessary compile-scope dependency is accidentally declared.


pstackle (approved):
LGTM. Thanks @accwebs !

